PR #10, opened 2014-10-28
Jena OSGi bundle

stain:
Embedding core Jena modules in a single OSGi bundle.

This works around classloader issues such as Jena's use of `Class.forName()`, but does not yet support other OSGi bundles to easily plug in 3rd party Jena implementations of say readers and writers.

Modules included: 
* jena-core
* jena-arq
* jena-tdb
* jena-sdb
* jena-iri

This OSGi bundle includes a dependency on the `httpclient-osgi` and `httpcore-osgi` version. All other packages used are embeded within the OSGi bundle, making them appear on a single classloader in OSGi.


stain:
I've added `jena-osgi-test` which tests use of jena-osgi as an OSGi bundle from within both Equinox and Apache Felix.

It has a couple of [simple tests](https://github.com/stain/jena/blob/f3535fc391df8a7b91bd988d8f4920196956bf7a/jena-osgi-test/src/main/java/org/apache/jena/osgi/test/JenaOSGITestImpl.java) that try to cover the areas of jena-iri, jena-core, jena-arq and jena-tdb. All tests pass on both frameworks.

As a side effect, in `jena-osgi-test/target/eosgi-dist` you get a pre-made set of bundles for using Jena within OSGi for the two frameworks. 
```
stain@biggie-utopic:~/src/jena/jena-osgi-test/target/eosgi-dist$ ls felix/lib

felix/lib:
commons-csv-1.0.jar	       jcl-over-slf4j-1.7.6.jar			 org.everit.osgi.dev.testrunner.junit4-3.0.4.jar
commons-lang3-3.3.2.jar        jena-osgi-2.12.2-SNAPSHOT.jar		 org.ops4j.pax.tipi.hamcrest.core-1.3.0.1.jar
httpclient-osgi-4.2.6.jar      jena-osgi-test-2.12.2-SNAPSHOT.jar	 org.ops4j.pax.tipi.junit-4.11.0.1.jar
httpcore-osgi-4.2.5.jar        jsonld-java-0.5.1.jar			 slf4j-api-1.7.6.jar
jackson-annotations-2.3.0.jar  libthrift-0.9.2.jar			 slf4j-log4j12-1.7.6.jar
jackson-core-2.3.3.jar	       log4j-1.2.17.jar
jackson-databind-2.3.3.jar     org.everit.osgi.dev.testrunner-4.0.3.jar
```

afs:
After seeing some feedback, is this now ready to put in (some comments on namign elsewhere)? I saw one comment for a different packaging but that seemed to be for a specific use.  I don't think we wil be providing the one-and-only way to package Jena libraries for OSGi; we can provide a "frequently useful" form.

If so, as a new module, I'd like to see a proposal to dev@ that stands out.  This new functionality in delivery is a high ceremony point, and the naming is very user visible (we hope!)

@stain - That make sense?

stain:
Agreed!

I went for just jena-osgi as that is the naming style used by other Apache
projects that do separate OSGi bundles, like httpclient-osgi.


There's more to be done - ideally the jena-* should be OSGi bundles
themselves - we might be able to adapt the headers used by Clerezza and see
how it works. I am however currently more comfortable about my jena-osgi
proposal, just to get something out there, and to avoid messing about with
the dependencies of the core Jena packages at this point.


On 26 January 2015 at 16:30, Andy Seaborne <notifications@github.com> wrote:

> After seeing some feedback, is this now ready to put in (some comments on
> namign elsewhere)? I saw one comment for a different packaging but that
> seemed to be for a specific use. I don't think we wil be providing the
> one-and-only way to package Jena libraries for OSGi; we can provide a
> "frequently useful" form.
>
> If so, as a new module, I'd like to see a proposal to dev@ that stands
> out. This new functionality in delivery is a high ceremony point, and the
> naming is very user visible (we hope!)
>
> @stain <https://github.com/stain> - That make sense?
>
> —
> <https://github.com/apache/jena/pull/10#issuecomment-71489219>.
>



-- 
Stian Soiland-Reyes
Apache Taverna (incubating)
http://orcid.org/0000-0001-9842-9718

afs:
Please could you send email to dev@ so the dev community is fully aware.  A new module is a signifcant event in the projects life!  I'll take it as lazy consensus and merge after a 24h delay (or when I have time).

PR #11, opened 2014-11-24
Added yasqe as query editor

LaurensRietveld:
I've added YASQE as query editor, and YASR as resultset visualizer.

Some of the changes I made:
- upped the version of datatables
- removed the datatables css (already bundled and slightly modified in YASQE css)
- moved the query execution logic to YASQE (so removed quite a bit from the qonsole)
- removed the codemirror addon dependencies: these are all bundled in YASQE
- removed the different output selectors in the dataset htm (already part of YASR)
- move the endpoint selection to the top (looked a bit strange without the output selectors there)
- upped the require.js version
- perhaps some more things I forgot to mention

As this interface is still a work in progress, I wasnt able to test everything (e.g. adding prefixes).
I'm not sure whether the current prefix system (with the buttons) is still needed with the prefix autocompletion in YASQE (using prefix.cc as well). If you do decide to still use them, quite some code can be cleaned up (some regex expressions) by using the YASQE prefix functions (see http://yasqe.yasgui.org/doc)

Finally, an unrelated comment: I -really- like the edit function you've added. It took me a while to understand what it was though. You could consider checking whether Eetu Mäkelä from the snapper turtle editor (see http://jiemakel.github.io/snapper/) would be interested in creating a pull request as well


LaurensRietveld:
Ah, looks like the other issues are quite small. I'll submit a new PR to fix these in a bit

PR #17, opened 2015-01-05
Fixed firefox UI issues of query editor

LaurensRietveld:
- fixed alignment of endpoint selector in firefox
- updated YASQE to newest version. This one contains a firefox fix as well, as the query button was not visible in some situations.


LaurensRietveld:
related to the original PR https://github.com/apache/jena/pull/11 , where we added the YASQE and YASR libraries

afs:
Thanks.  

Just to let you know:

The Fuseki2 code has just been reorganised into multiple maven modules.  The webapp is now in jena-fuseki2/jena-fuseki-core but the structure is unchanged.  It's easy to fix up the PR from its patch.

PR #21, opened 2015-02-02
Updates to jena-osgi according to email thread

stain:
From email thread [Jena OSGi pull request](http://mail-archives.apache.org/mod_mbox/jena-dev/201502.mbox/%3C54CE2655.20006%40apache.org%3E)

jsonld comments in pom, submodules under apache-jena-osgi, millioseconds, and only build jena-osgi-test on -Papache-release.

In addition a rat-plugin configuratoin in parent was modified as it didn't exclude all in target/

### DIFF
--- a/jena-osgi-test/pom.xml
+++ b/jena-osgi-test/pom.xml
@@ -98,12 +98,19 @@
             <environment>
               <id>felix</id>
               <framework>felix</framework>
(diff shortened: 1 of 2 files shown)


stain:
Also included the NOTICE to include the historical Jena NOTICEs and Xerces.

afs:
Please separate thing like comments and POM fixes from setting NOTICE.  They need separate discussions.

PR #28, opened 2015-02-10
Quieter jena-osgi-test logging

stain:
.. by setting system property
now the only noise is because of duplicate slf4j binding, as slf4j-log4j is inherited from Jena parent.

Also fixes the missing `LATEST` version for jena-osgi


afs:
A few lines for the duplicate slf4j binding is practical.  Theer are various strange warnings elsewhere in Jena, some to do with Java8 and JavaFX (which we don't use!).

PR #29, opened 2015-02-10
Jena osgi integration test

stain:
Only compile+run jena-osgi-test in `mvn install` - ignore in `mvn test`.

Note that with #28 this will also work with `mvn verify`.


stain:
There will be a conflict with #28 - I'll update if/when #28 is in.

afs:
#28 now pull/pushed into Jena.

stain:
Rebased to master.

PR #30, opened 2015-02-12
JENA-878: Explicit imports of org.apache.xerces.impl.dv

stain:
.. by doing Organize Imports in Eclipse.

not that it solves JENA-878, but at least we would know
more explicitly which org.apache.xerces.impl classes
we depend on.


afs:
Part of #31

PR #31, opened 2015-02-13
JENA-878 Avoid exposing xerces.impl.dv

stain:
Not quite a fix to [JENA-878](https://issues.apache.org/jira/browse/JENA-878) - but avoid the dangerous OSGi implications of requiring an export of `org.apache.xerces.impl.dv`.

This patch makes any method signature that involves org.apache.xerces.impl.dv package-private, e.g. those sending around `ValidationInfo`.

This meant moving `XSDGenericType` to become an inner class of XSDDatatype.  It could in theory be a separate class in the same package - but it shouldn't "really" be exposed or promoted - so I thought it was better to hide it in plain sight inside XSDDatatype which is the only class using it directly.

I suggested now for `XSDDatatype.XSDGenericType` remains public as the other `impl.XSD*` classes are public - I guess they could be used mainly for `instanceof` checks? If not the class itself can also be made package-private or even private.

This removes these particular warnings from jena-osgi about `org.apache.xerces.impl.dv`:

```
[WARNING] Bundle org.apache.jena:jena-osgi:bundle:2.13.0-SNAPSHOT : 
  Export com.hp.hpl.jena.datatypes.xsd,  has 1,  private references [org.apache.xerces.impl.dv], 
[WARNING] Bundle org.apache.jena:jena-osgi:bundle:2.13.0-SNAPSHOT : 
  Export com.hp.hpl.jena.datatypes.xsd.impl,  has 1,  private references [org.apache.xerces.impl.dv],
```


jena-core/src/main/java/com/hp/hpl/jena/datatypes/xsd/impl/XSDGenericType.java
```diff
@@ -1,66 +0,0 @@
-/*
```
afs: Why is this moved to XSDDatatype? Seems to be just stylistic.

jena-core/src/main/java/com/hp/hpl/jena/datatypes/xsd/XSDDatatype.java
```diff
@@ -18,34 +18,52 @@
...
+import com.hp.hpl.jena.datatypes.xsd.impl.XSDDouble;
+import com.hp.hpl.jena.datatypes.xsd.impl.XSDDurationType;
+import com.hp.hpl.jena.datatypes.xsd.impl.XSDFloat;
+import com.hp.hpl.jena.datatypes.xsd.impl.XSDMonthDayType;
+import com.hp.hpl.jena.datatypes.xsd.impl.XSDMonthType;
+import com.hp.hpl.jena.datatypes.xsd.impl.XSDPlainType;
+import com.hp.hpl.jena.datatypes.xsd.impl.XSDTimeType;
+import com.hp.hpl.jena.datatypes.xsd.impl.XSDYearMonthType;
+import com.hp.hpl.jena.datatypes.xsd.impl.XSDYearType;
+import com.hp.hpl.jena.graph.impl.LiteralLabel;
```
afs: Gnerally, we havn't enforced style of imports.  These will get lost is anyone uses Eclipse to reformat and has * collapse on (which it is by default).

jena-core/src/main/java/com/hp/hpl/jena/datatypes/xsd/XSDDatatype.java
```diff
@@ -251,7 +269,7 @@ public XSDDatatype(String typeName, Class<?> javaClass) {
      * @param xstype the XSSimpleType definition to be wrapped
      * @param namespace the namespace for the type (used because the grammar loading doesn't seem to keep that)
      */
-    public XSDDatatype(XSSimpleType xstype, String namespace) {
+    XSDDatatype(XSSimpleType xstype, String namespace) {
```
afs: Jena provides all the meaningful XSD types (just some XML specific types are missing) so it is unlikely that external code is calling this public constructor.  Howver, it is public.  If the aim is to hide Xerces XSSimpleType, then hop 
about:

1. Make this protected.
2. Add `public XSDDatatype(Object xstype, String namespace)` and cast to `XSSimpleType`

I thhnk that will keep OSGi happy(ier).

stain:
No, the imports can't be enforced.. (without applying the checkstyle plugin) but the above is what the default Eclipse Import Reorganize gave me :). It would hopefully give more reasonable compile errors if at a later xerces upgrade the classes disappeared.

`XSDGenericType` moved up from `xsd.impl` so that it is able to access the now private constructor of `xsd.XSDDatatype`. I didn't want to make it a top-level class at `xsd` as that would 'promote' it a bit too much. The inner class `XSDGenericType` could (should?) be made private as well, but then you can't do `instanceof` checks against it.

PR #35, opened 2015-02-25
JENA-869 Disable Remove button until it works again

stain:
Disabled in UI - but bug JENA-869 remains open as Remove doesn't work properly.


afs:
The "remove" button not the "download" button.

stain:
Right - too many buttons. Correct bug number - but wrong commit message. Updated.

With regards to JENA-871 I  had a quick look, but I was unable to remove the download button of the SPARQL result as it is part of the table handling library yasr.

PR #37, opened 2015-02-26
JENA-732 jena-maven-tools outputs to target/generated-sources/jena

stain:
fixes JENA-731 and JENA-732

This reverts the reverted commit by Benson Margulies so that generated sources are put in `target/generated-sources/jena` according to Maven plugin conventions, instead of straight into `target/generated-sources`.  (JENA-732)

This also solves JENA-731 so that the source files are also compiled by Maven. I've therefore added tests to `src/it` to test for presence of the `*.class` files 

.. except for where `<sources>` configuration has been overridden as in [schemagen-integration-2](https://github.com/stain/jena/blob/JENA-732-reborn/jena-maven-tools/src/it/schemagen-integration-2/verify.bsh#L31) - but should it perhaps also add those sources to the compile path? Other plugins like jaxb-maven-plugin usually do..

### DIFF
--- a/jena-maven-tools/README
+++ b/jena-maven-tools/README
@@ -15,7 +15,7 @@ Changes
 
 Version 0.8
   * Output folder is now by default target/generated-sources/jena
-    Use <output> to override.
+    Use <outputDirectory> to override.
   * Generated Java source files are also compiled by Maven
 
 Version 0.4
--- a/jena-maven-tools/README
+++ b/jena-maven-tools/README
@@ -13,6 +13,11 @@ Available plugins:
 
 Changes
 
+Version 0.8
+  * Output folder is now by default target/generated-sources/jena
+    Use <output> to override.
+  * Generated Java source files are also compiled by Maven
+
 Version 0.4
 
   * Moved source code from org.openjena package tree to org.apache.jena
--- a/jena-maven-tools/src/it/schemagen-integration-0/verify.bsh
+++ b/jena-maven-tools/src/it/schemagen-integration-0/verify.bsh
@@ -25,6 +25,6 @@ if (!testDir.exists()) {
 
 Verifier verifier = new Verifier( testDir.getAbsolutePath() );
 
-String test1 = verifier.getBasedir() + "/target/generated-sources/Test1.java";
+String test1 = verifier.getBasedir() + "/target/generated-sources/jena/Test1.java";
 // note that ?s modifier makes . pattern match newlines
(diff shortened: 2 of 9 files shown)


stain:
*edit*: I got confused  by `<output>` vs `<outputDirectory>` - so I updated the release notes in README and added a test that verifies that if specifying a custom outputDirectory it is used AND compiled.

ajs6f:
Given that #38 is merged, is this still live?

ajs6f:
@stain Just pinging again-- is this PR still in flight, or does it still make sense?

kinow:
>Given that #38 is merged, is this still live?

From reading the comments in #38, I think these PR's were related, but before #38 was merged it was simplified (see @stain comments in #38).

I am working on `schemagen`, and coincidentally related to this issue. So had a try with `master`, and successfully reproduce JENA-732. See undert he `generated-sources`.

screenshot from 2018-12-30 17-55-25

Then `rebased` this PR onto `master` (result plus another change here: https://github.com/kinow/jena/commits/pr-37-2). And I can confirm that the changes in this PR are still good, except for a package name that had openjena.

Using the change in this PR, my sources were generated correctly under `generated-sources/jena`. See below.

screenshot from 2018-12-30 18-17-37

Even `Skos.java` which had no package. Sources used for testing available here: [https://github.com/kinow/jena-tests](https://github.com/kinow/jena-tests).

Having the generated sources in `generated-sources` is not ideal, and may break things depending on how people are packaging/building. Jenkins plug-ins put translation classes under `generated-sources/localizer`, `maven-jaxb-plugin` puts their generated sources under `generated-sources/xjc`, and so it goes.

A tool that outputs sources directly to `generated-sources` could be considered wrong, I reckon, as normally each tool/plugin creates its own sub directory under `generated-sources`. At least that's my rationale for supporting the change in JENA-732, disregarding the argument that changing this could break builds/workflows (I think we should in this case, for correctness).

So I'm +1 to fix conflicts & merge it, then fix JENA-732. 

Cheers
Bruno

afs:
I don't use schemagen via maven so +1 to fix and merge.

Related:
At the moment (3.10.0) the maven module is commented out of the main build.
Last release was 3.6.0. A change in the Apache parent broke this module.

Not directly related:
Shall we move the source from the top of Jena to under jena-extras?

afs:
Given [JENA-1760 "Retire Apache Jena maven-tools module"](https://issues.apache.org/jira/browse/JENA-1760), we can close this PR. 

The PR is not lost - it can be revived at some later date.

PR #38, opened 2015-02-26
JENA-849 maven tools demo versions

stain:
Fixes JENA-849 in updated `jena-maven-tools/demo`

.. tries to not list the version numbers explicitly as they would go out of date.


afs:
Relationship to #37?  To first approximation, this includes all of #37 so should #37 by closed?

stain:
Uhm.. well, it accidentally got based on #37 as that was the plugin version I updated it for, but in the final version the demo/ also works straight of master - so I'll rebase this.

stain:
Now independent of #37

afs:
Thanks.

stain:
Anything blocking this from going in? Sorry I forgot to check if it was applied for the last release..

PR #39, opened 2015-02-27
Updated text indexing

ehedgehog:
This change addresses JENA-686, support for cross field conjunctive queries in jena-text, by
allowing TextDocProducers to be specified in a TextDatasetAssembler assembly and
existing index entities to be updated as well as added. 

DatasetTextGraph - commit's monitor finish() moved to the top
	of the method. This is because a batching TextDocProducer
	(as we have in our external app) may have quads buffered up
	awaiting end-of-batch and they must be flushed by finish();
	if they are not, they  are auto-flushed after the commit has
	been run and an exception is thrown.

TextDatasetFactory - needed methods to create datasets with
	doc producers and close-index-on-close flags

TextIndex - added new operation updateEntity to allow update
	of (possibly) existing entities

TextIndexLucene - added implementation of updateEntity. Added
	deleteDocuments(Term) for deletion of documents (as is used
	in ppd-text-index text doc producer batch).

TextDatasetAssembler - updated to allow specification of doc
	producer (since done by mainline Jena) and close-index-on
	close.

AbstractTestDatasetWithLuceneIndex, DummyDocProducer,
TestTextDatasetAssembler - fix up some issues in the test 
	framework (use of statics vs use of instance variables,
	remembering to close datasets @After done, etc).


ehedgehog:
Closed; to be replaced by smaller/simpler pull requests.

PR #42, opened 2015-03-13
jena-text updates for constructing documents suitable for conjunctive queries

ehedgehog:
This change to jena-text allows TextDocumentProducers
to access the dataset they are monitoring, and to
create indexes where there is a single (Lucene) document
for a given subject and all its defined properties
rather than a separate document per triple that the
subject appears in.

This allows indexes to be used for conjunctive query
eg a request such as

	city: Plymouth AND street: Station

See https://github.com/epimorphics/ppd-text-index for an
an example project that uses conjunctive queries and 
provides a bulk index creation utility.

The changes to jena-text are spread over six files as follows:

# jena-text/src/main/java/org/apache/jena/query/text/DatasetGraphText.java 

The two-phase commit protocol is modified to ensure that a
DatasetChanges monitor finishes() before the commit protocol
starts. It is possible for a DatasetChanges to have buffered-up
changes which have not yet been applied; applying these
changes mid-commit can cause errors.

[This problem was detected in https://github.com/epimorphics/ppd-text-index
where TextDocProducerBatch does have buffered state and closing
the DatasetGraphText threw an exception when the state was flushed.]

A test for this behaviour is not currently available.

# jena-text/src/main/java/org/apache/jena/query/text/TextIndex.java 

Added the abstract method updateEntity, contracted with addEntity().
Updating an index with updateEntity is intended to discard any
existing Document with the entity key and create a new one with
all and only the fields specified by the given Entity, as opposed
to addEntity which creates a new Document from the Entity even if
one with that key already exists.

This allows Documents suitable for conjunctive query to
be created. The project 

# jena-text/src/main/java/org/apache/jena/query/text/TextIndexLucene.java 

Implement updateEntity for a TextIndexLucene.

# jena-text/src/main/java/org/apache/jena/query/text/TextIndexSolr.java 

Placeholder for implementation of TextIndexLucene.updateEntity; at the
moment we do not support it.

# jena-text/src/main/java/org/apache/jena/query/text/assembler/TextDatasetAssembler.java 

The DatasetAssembler may construct a non-default TextDocProducer to
feed to the TextDatasetFactory, passing in to the constructor the
TextIndex that the TextDocProducer uses. This change additionally
allows for a two-argument constructor taking the DatasetGraph as
well as the TextIndex as arguments.

The TextDocProducer can use this to query the dataset for triples.
(EG other triples with the same subject as one it has received, so
as to build all the properties into a single Document.)

#  jena-text/src/test/java/org/apache/jena/query/text/assembler/TestTextDatasetAssembler.java 

Changes to have a test that the two-argument constructor is called
when appropriate.

### DIFF
--- a/jena-text/src/main/java/org/apache/jena/query/text/DatasetGraphText.java
+++ b/jena-text/src/main/java/org/apache/jena/query/text/DatasetGraphText.java
@@ -142,6 +142,7 @@ public class DatasetGraphText extends DatasetGraphMonitor implements Transaction
      */
     @Override
     public void commit() {
+    	super.getMonitor().finish() ;
         // Phase 1
         if (readWriteMode.get() == ReadWrite.WRITE) {
             try {
@@ -167,7 +168,6 @@ public class DatasetGraphText extends DatasetGraphMonitor implements Transaction
             throw new TextIndexException(t);
         }
         readWriteMode.set(null);
-        super.getMonitor().finish() ;
     }
 
     @Override
--- a/jena-text/src/main/java/org/apache/jena/query/text/TextIndex.java
+++ b/jena-text/src/main/java/org/apache/jena/query/text/TextIndex.java
@@ -36,6 +36,7 @@ public interface TextIndex extends Closeable //, Transactional
     
     // Update operations
     void addEntity(Entity entity) ;
+    void updateEntity(Entity entity) ;
     
     
     // read operations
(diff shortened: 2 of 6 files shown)


jena-text/src/main/java/org/apache/jena/query/text/DatasetGraphText.java
```diff
@@ -142,6 +142,7 @@ public void abort() {
      */
     @Override
     public void commit() {
+    	super.getMonitor().finish() ;
```
Stephen-Allen: > The two-phase commit protocol is modified to ensure that a DatasetChanges monitor finishes() before the commit protocol starts. It is possible for a DatasetChanges to have buffered-up changes which have not yet been applied; applying these changes mid-commit can cause errors.

This makes more sense.

Stephen-Allen:
Changes look good!

ehedgehog:
Fixes #686.

afs:
Fixes JENA-686

PR #44, opened 2015-03-20
JENA-901 Upper bound on LPDRuleEngine cache

stain:
Fixes [JENA-901](https://issues.apache.org/jira/browse/JENA-901).

A better solution might be to use guava caches - but I didn't want to add another dependency, so this fix can go into a patch release.

### DIFF
--- a/jena-core/pom.xml
+++ b/jena-core/pom.xml
@@ -45,6 +45,13 @@
       <scope>test</scope>
     </dependency>
 
+	<dependency>
+		<groupId>com.carrotsearch</groupId>
+		<artifactId>java-sizeof</artifactId>
+		<version>0.0.4</version>
+		<scope>test</scope>
+	</dependency>
+
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
--- a/jena-core/pom.xml
+++ b/jena-core/pom.xml
@@ -45,13 +45,6 @@
       <scope>test</scope>
     </dependency>
 
-	<dependency>
-		<groupId>com.carrotsearch</groupId>
-		<artifactId>java-sizeof</artifactId>
-		<version>0.0.4</version>
-		<scope>test</scope>
-	</dependency>
-
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
--- a/jena-core/src/main/java/com/hp/hpl/jena/util/BoundedLRUMap.java
+++ b/jena-core/src/main/java/com/hp/hpl/jena/util/BoundedLRUMap.java
@@ -33,17 +33,17 @@ import java.util.Map;
  * @param <V>
  *            Type of values
  */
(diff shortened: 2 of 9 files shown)


jena-core/pom.xml
```diff
@@ -45,6 +45,13 @@
...
     </dependency>
 
+	<dependency>
+		<groupId>com.carrotsearch</groupId>
+		<artifactId>java-sizeof</artifactId>
+		<version>0.0.4</version>
+		<scope>test</scope>
+	</dependency>
+
     <dependency>
```
afs: What is this? What is the license?

jena-core/src/main/java/com/hp/hpl/jena/util/BoundedLRUMap.java
```diff
@@ -0,0 +1,58 @@
...
+			throw new IllegalArgumentException("maxEntries <= 0");
+		}
+		this.maxEntries = maxEntries;
+	}
+
+	@Override
+	protected boolean removeEldestEntry(java.util.Map.Entry<K, V> eldest) {
+		return size() > maxEntries;
+	}
+}
```
afs: Seems to duplicate existing oaj.atlas.lib.cache.CacheLRU.  At a minimum move and use that once so it is then findable for any later upgrade to Guava rather than leave yet another separate cache implementation around. Better still the cache discussion on the dev@ list. Putting in ad-hoc fixes drives up the long term maintence overhead.

stain:
Thanks for pointing out [CacheLRU](https://github.com/apache/jena/blob/master/jena-arq/src/main/java/org/apache/jena/atlas/lib/cache/CacheLRU.java) which I missed.

I have now looked into this, and moving CacheLRU would mean also moving its interface [Cache](https://github.com/apache/jena/blob/master/jena-arq/src/main/java/org/apache/jena/atlas/lib/Cache.java),  `ActionKeyValue` and `CacheImpl`.

If these go into a new package, e.g. `org.apache.jena.lib.cache`, then for theit is easy to make `@deprecated` subclasses/subinterfaces for `Cache`, `CacheLRU` and even `CacheImpl`. It is trickier for `ActionKeyValue` which is more wide-spread in use within Atlas - unless I do some kind of proxy.

So that hints to me that I would have to rather let them stay in the same package, but live in `jena-core`. This would work -- but it would either make a nasty split of that package across two Maven modules -- or the whole of `atlas.lib.*` moves over.

Moving/fixing/etc CacheLRU is sounding more and more like something to reorganize in Jena 3 - Can we have a Fix for Version 3.0 in Jira that we can put something like this against?

Now simply adding the Guava dependency sounds like a much easier path!

stain:
Pull Request now no longer includes the java-sizeof test dependency - I would suggest go ahead to merge it until we don't get in the guava patch. (I'll try that afterwards - but can't commit too much time)

([license is Apache license btw](https://github.com/dweiss/java-sizeof/blob/master/pom.xml))

stain:
I'll close this if #45 / #47  goes in.

stain:
Closing in preference for #47 as #45 went in.

PR #45, opened 2015-03-20
Guava as shaded dependency

stain:
(Pull request mainly for discussion)

Add Guava as a shaded dependency in a new module `jena-shadowed-ext` 
that can then be used by `jena-core` for [JENA-901](https://issues.apache.org/jira/browse/JENA-901).

As this means `jena-core` gets a new dependencies (even our own!), and we have been shot down before for adding those in patch releases, this means a new minor version on anything from `jena-core` and downstream.

I did not use this shaded jar within Elephas as it does not itself use Guava - Elephas does however have conflicting version dependencies on Guava 11.0.2 (from provided hadoop-common 2.6.0), 16.0.1 (from curator-client) and 18.0 (from airline dependency in jena-elephas-stats). I therefore changed `jena-elephas`  outdated dependencyManagement on 11.0.2 to [18.0,)  -- although this could be reduced to 16 as jena-elephas-stats is a demo-app and it can upgrade to 18 locally in its own pom (ideally airline etc. should themselves use ranges)


stain:
Also.. what semantic versioning should the jena-shadowed-ext  itself use?? :)

stain:
... or - we could just shadow Guava directly inside jena-core - the effect would be the same, but no need for an additional JAR, and thus no need to bump minor version number. Anything in Jena (except jena-iri) would then still be able to see/use the shaded Guava.

jena-shadowed-ext/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+          <relocations>
+            <relocation>
+              <pattern>com.google</pattern>
+              <shadedPattern>org.apache.jena.ext.com.google</shadedPattern>
+            </relocation>
+          </relocations>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
```
afs: I don't understand why this needed.

jena-shadowed-ext/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+        <configuration>
+          <artifactSet>
+            <includes>
+              <artifact>com.google.guava:guava</artifact>
+            </includes>
+          </artifactSet>
+          <relocations>
+            <relocation>
+              <pattern>com.google</pattern>
+              <shadedPattern>org.apache.jena.ext.com.google</shadedPattern>
```
afs: More focused is to shade com.google.common and com.google.thirdparty

rvesse:
-1 because of the changes to Elephas

See https://issues.apache.org/jira/browse/JENA-842

Elephas use of an outdated version of Guava is intentional if unfortunate.  Sadly Hadoop dependencies are something of a mess - they even have versions of Commons libraries that conflict with each other - but not something we are empowered to fix

rvesse:
Also -1 for bumping the minor version just for dependency changes and a strong -1 to dependency ranges

Dependency ranges are a PITA, unfortunately the NuGet folks chose to default to dependency ranges instead of fixed versions as the default and this causes all manner of grief and nastiness in the .Net world.  Let's not bring that over to Maven and Java please!

jena-shadowed-ext/pom.xml
stain: The relocation is needed because jena-shadowed-ext would effectively become
a second Guava implementation, which might be older or newer from what
might also be on the class path (as in Elephas / Hadoop). Maven uses only
groupId and artifactId to decide in versions.

You do not want those to mix, as you could end up with say
com.google.common.FooBuilder from Guava 16 and com.google.common.FooImpl
from 10.

There is only com.google.common now. If a later update of guava adds
additional packages, are you saying those should NOT be relocated, but left
in place? (This can be mitigated by modifying the file include patterns,
but that could lead to linkage errors )

stain:
I'll remove the elephas change, it is not affected as long as there is a a
class relocation of the shadowed one.

jena-shadowed-ext/pom.xml
afs: I was wodnering what if there is a com.google, not related to guava unrelated that is not to be shaded.
I have been trying out:
<relocation><pattern>com.google.common</pattern>
then
<relocation><pattern>com.google.thirdparty</pattern>

stain: `guava:18` is the only artifact that is selected to be shaded by the specified `<includes>` - so if that JAR contained other things under `com.google` (e.g. someone updates this to Guava 22 later) those would also be shaded. Is that not what we would want?

Anyway, for consistency's sake (to match the bundle export), I've modified it to only include files under `com.google.common` and `com.google.thirdparty`.

Note, transitives are not followed with the use of `<includes>`, so if the guava.pom started depending on other things, they would not get shaded. Alternatively I would need to use a different `<parent>` to avoid the <dependencies>` sneaking in from `jena-parent`, or try to keep matching `<excludes>` for those up to date here.

The current dependency of guava includes only:
```xml
    <dependency>
      <groupId>com.google.code.findbugs</groupId>
      <artifactId>jsr305</artifactId>
      <optional>true</optional><!-- needed only for annotations -->
    </dependency>
```
I have updated the pull request to use explicit packages for `com.google.common` and `com.google.thirdparty`.

stain:
@rvesse, I have now modified the PR to not touch elephas poms. I've also added a comment to this pom that `jena-shadowed-ext` should not be used outside Jena.

afs:
Why is felix involved?

> + <groupId>org.apache.felix</groupId>

stain:
I thought for any new modules, if you can, make it an OSGi bundle. That does not cause any problems. Thus making the job easier when we get rid of jena-osgi.

afs:
That is a discussion and decision for the project to make.  Putting extra, hidden consequences makes it harder to accept a pull request.

Making everything an OSGi bundle is not free - it needs to be maintained and it adds to build complexity. Let's keep things modular and not making everyone have to have some understanding of OSGi to work with and test other modules.

stain:
I am not sure what you mean by "is not free". I am not sure what understanding of OSGi is required here.

The OSGi metadata is simply additional entries in `META-INF/MANFEST.MF` that, with the Maven settings I put in, simply replicating what is already in the upstream guava JAR (but with the transcribed package names). 

I didn't feel a need to move this into `jena-osgi` as it would be harder to update there.

afs:
Then I don't understand the OSGi setup: the stated contract is "jena internal use only" and I assumed that because jena-osgi has `<Export-Package>...org.apache.jena.*` it ended up in that as with all dependences in the bundle that weren't excluded by `Import-Package ... !...`.

stain:
Good point, I shall rewrite the pull request so it is a Private-Package of
jena-osgi and not a OSGi manifest itself.

It should not be leaked out through signatures as long as I make the field
of the cache private instead of protected.

stain:
Would the package name org.apache.jena.impl.ext.com.google.common (... I
know) perhaps hammer in this point also to non-OSGi folks? I guess we want
to be able to replace guava at any point..

I will update the cache pull request #47 to make cache private, not
protected (that will leak Guava in Import-Package).

PR #46, opened 2015-03-21
JENA-904 Avoid activeInterpreters leak

stain:
Fixes [JENA-904](https://issues.apache.org/jira/browse/JENA-904) by closing the interpreter of the generator of /ConsumerChoicePointFrame - and making sure CCPF.close() is called through super.close() calls of its owners.


stain:
You might notice the test is very similar to #44 - but I kept it in a separate file here now to avoid merge conflicts. Feel free to manually merge the two tests!

stain:
Anything blocking this fix..? It does not depend on #44, although both must go in to solve JENA-904.

PR #47, opened 2015-03-21
JENA-901 LPDRuleEngine cache Guava from jena-shadowed-ext

stain:
This patch builds on the discussion pull request #45 - so if #45 goes through, this is how to use Guava within Jena - importing `org.apache.jena.ext.com.google.common.cache.Cache`.

The considerations of #45 apply here - so because `jena-core` gets a new dependency `jena-shadowed-ext` - even though that is our own brand new JAR - that should not go out as a patch-level upgrade, but forces a new minor version on pretty much everything jena-core and downstream. (We've been told of for adding dependencies in a patch before)

See also a second pull request #48 .

### DIFF
--- a/jena-core/pom.xml
+++ b/jena-core/pom.xml
@@ -59,7 +59,7 @@
 
     <dependency>
     	<groupId>org.apache.jena</groupId>
-    	<artifactId>jena-shadowed-guava</artifactId>
+    	<artifactId>jena-shaded-guava</artifactId>
     	<version>3.1.1-SNAPSHOT</version>
     </dependency>
 
--- a/jena-core/pom.xml
+++ b/jena-core/pom.xml
@@ -58,27 +58,33 @@
     </dependency>
 
     <dependency>
+    	<groupId>org.apache.jena</groupId>
+    	<artifactId>jena-shadowed-guava</artifactId>
+    	<version>3.1.1-SNAPSHOT</version>
+    </dependency>
+
+    <dependency>
       <groupId>xerces</groupId>
       <artifactId>xercesImpl</artifactId>
     </dependency>
-    
-     <dependency> 
-       <artifactId>commons-cli</artifactId> 
-       <groupId>commons-cli</groupId> 
-     </dependency> 
+
+     <dependency>
+       <artifactId>commons-cli</artifactId>
+       <groupId>commons-cli</groupId>
+     </dependency>
 
     <dependency>
       <groupId>org.xenei</groupId>
       <artifactId>junit-contracts</artifactId>
       <scope>test</scope>
     </dependency>
-    
+
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-all</artifactId>
       <scope>test</scope>
     </dependency>
-    
+
     <dependency>
       <groupId>com.jayway.awaitility</groupId>
       <artifactId>awaitility</artifactId>
@@ -90,7 +96,7 @@
       <artifactId>jena-base</artifactId>
       <version>3.1.1-SNAPSHOT</version>
     </dependency>
-    
+
     <dependency>
       <groupId>org.apache.jena</groupId>
       <artifactId>jena-base</artifactId>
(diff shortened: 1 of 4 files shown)


stain:
Updated patch - now based on #45 (so review that first)

stain:
Updated to not expose guava cache through a protected field. (An exclusion was also needed in jena-osgi).

jena-core/src/main/java/com/hp/hpl/jena/reasoner/rulesys/impl/LPBRuleEngine.java
```diff
@@ -52,16 +66,22 @@
     protected boolean recordDerivations;
         
     /** List of engine instances which are still processing queries */
-    protected List<LPInterpreter> activeInterpreters = new ArrayList<>();
-    
+    protected List<LPInterpreter> activeInterpreters = new LinkedList<>();
+
+    protected final int MAX_CACHED_TABLED_GOALS = 
+			Integer.getInteger("jena.rulesys.lp.max_cached_tabled_goals", 512*1024);
+
```
afs: getting System properties is problematic in some environments.  See JenaRuntime.getSystemProperty for some mitigation of this.

jena-core/src/main/java/com/hp/hpl/jena/reasoner/rulesys/impl/LPBRuleEngine.java
```diff
@@ -252,19 +272,31 @@ public synchronized void tablePredicate(Node predicate) {
...
-            LPInterpreter interpreter = new LPInterpreter(this, goal, clauses, false);
-            activeInterpreters.add(interpreter);
-            generator = new Generator(interpreter, goal);
-            schedule(generator);
-            tabledGoals.put(goal, generator);
-        }
-        return generator;
+    public synchronized Generator generatorFor(final TriplePattern goal, final List<RuleClauseCode> clauses) {
+		return getCachedTabledGoal(goal, new Callable<Generator>() {
+		 	@Override
```
afs: This can be Java8'ed!

jena-core/src/main/java/com/hp/hpl/jena/reasoner/rulesys/impl/LPBRuleEngine.java
```diff
@@ -252,19 +272,31 @@ public synchronized void tablePredicate(Node predicate) {
...
+		    public Generator call() {
+		 		/** FIXME: Unify with #generatorFor(TriplePattern) - but investigate what about
+		 		 * the edge case that this method might have been called with the of goal == null
+		 		 * or goal.size()==0 -- which gives different behaviour in 
+		 		 * LPInterpreter constructor than through the route of
+		 		 * generatorFor(TriplePattern) which calls a different LPInterpreter constructor
+		 		 * which would fill in from RuleStore. 
+		 		 */
```
afs: I don't know enough about the rules engine to comment.  We need another reviewer here.

stain:
OK, I've rebased for 3.1.1-SNAPSHOT, updated for Java 8 and fixed the System properties. 

As for the `FIXME` comment let's just leave it here, unless someone understands or would be willing to investigate what is the semantic difference between the `LPInterpreter` constructor [without clauses](https://github.com/apache/jena/blob/master/jena-core/src/main/java/org/apache/jena/reasoner/rulesys/impl/LPInterpreter.java#L99) and [with clauses](https://github.com/apache/jena/blob/master/jena-core/src/main/java/org/apache/jena/reasoner/rulesys/impl/LPInterpreter.java#L111) (which could be [empty](https://github.com/apache/jena/blob/master/jena-core/src/main/java/org/apache/jena/reasoner/rulesys/impl/LPInterpreter.java#L130))  but then would not use `engine.getRuleStore().codeFor(goal)`.

(Perhaps raise that as a new Jira issue?)

afs:
+1 : Looks good to me.

I'll merge the java changes if there are no other comments. 

The POM change isn't needed.  jena-shaded-guava comes via jena-base.

PR #48, opened 2015-03-21
JENA-901 Use Guava for cache, shadowed in jena-core

stain:
In comparison with #47, this shadows Guava directly in `jena-core` without introducing `jena-shadowed-ext`. 

This is thus acceptable as a patch-version update - but at the downside that `jena-core` have to do

    import com.google.common.cache.Cache

directly - while all other Jena modules would have to use 

    import org.apache.jena.impl.ext.com.google.common.cache.Cache

as it is shadowed with the JAR. This will cause problems in the other modules in an Eclipse-like environment, as the IDE will re-expose com.google version if `jena-core` is open - and they would then seemingly not be able to find the `oaj.impl.ext.*` version as that only exist in the Maven-built JAR.

In this scenario, it would probably be cleaner for the other modules to also shadow guava - in addition to the space-wastage, this probably becomes a bit awkward - it also means you can't pass any Guava structures between Jena modules. 

One advantage however is that if they all shadow individually, then they can shadow with the [minimizeJar](https://maven.apache.org/plugins/maven-shade-plugin/shade-mojo.html#minimizeJar) option to only get the few Guava classes used (e.g. in jena-core only the Cache).

### DIFF
--- a/jena-core/pom.xml
+++ b/jena-core/pom.xml
@@ -61,6 +61,22 @@
       <artifactId>xercesImpl</artifactId>
     </dependency>
 
+    <dependency>
+
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <!-- Version number specified explicitly and not in jena-parent, 
+           as it should only be used here, and further shadowing/bundle/NOTICE
+           consideration might be needed on upgrade. -->
+      <version>18.0</version>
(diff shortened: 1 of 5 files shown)


stain:
I am closing this as the feedback on guava shadowed in jena-core was negative (and I agree!)

PR #50, opened 2015-04-02
JENA-909 - Docker image for Fuseki 2

stain:
JENA-909 - Docker image for Fuseki

This is based on Maven, which is run either inside the Docker image, or if target/dist already exist in the host, use that (thus it should also work for local snapshots, or download from Maven on demand).

The Maven build does not make the docker image - and the readme now refers to stain/jena-fuseki (this should be changed to jena/jena-fuseki probably - presumably someone in PMC should make the organization in http://registry.hub.docker.com/ ?)

### DIFF
--- a/jena-fuseki2/jena-fuseki-docker/Dockerfile
+++ b/jena-fuseki2/jena-fuseki-docker/Dockerfile
@@ -13,7 +13,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-FROM maven:3.3-jdk-7
+FROM maven:3.3-jdk-8
 
 # Config and data
 VOLUME /fuseki
--- a/jena-fuseki2/jena-fuseki-docker/Dockerfile
+++ b/jena-fuseki2/jena-fuseki-docker/Dockerfile
@@ -1,0 +1,62 @@
+#   Licensed to the Apache Software Foundation (ASF) under one or more
+#   contributor license agreements.  See the NOTICE file distributed with
+#   this work for additional information regarding copyright ownership.
+#   The ASF licenses this file to You under the Apache License, Version 2.0
+#   (the "License"); you may not use this file except in compliance with
+#   the License.  You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+FROM maven:3.3-jdk-7
+
+# Config and data
+VOLUME /fuseki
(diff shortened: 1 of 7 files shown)


afs:
Having all the commits used in development is useful for the developer and short term history.  

For the long term record, a single commit (or, in general, a small number of independent units) at a specific time point, would be better as part of the handover.

stain:
Now a single squash commit - also handle creation of databases/ if not present.

afs:
It would be great if this could be tested by someone.

The README describes a registry.hub.docker.com image "stain/jena-fuseki" and the document says "stain/jena-fuseki" in quite a few places. 

What is being delivered by this module or is it more an example that can be pointed to?

kinow:
I can test it. I need to use Docker for another project, so that would be a chance to practice and learn more.

>What is being delivered by this module or is it more an example that can be pointed to?

@stain, you can answer that in the meantime, and/or I can give my point of view too.

ajs6f:
@stain @kinow Were we able to get testing of some kind done? I can see real utility for a Docker image, because it would let a wider audience try out Fuseki more easily, and for some people, would give them a headstart on constructing their own Docker setups. (Docker is becoming an increasingly popular deployment mechanism.)

stain:
I have updated https://hub.docker.com/r/stain/jena/ and https://hub.docker.com/r/stain/jena-fuseki/ to the latest releases (3.2.0, 2.5.0).

It's now based on `java:8-jre-alpine` base image so it should automatically updated with a newer java arrives.

I am a bit hampered trying to contribute to this as I am not a Jena committer and don't want to step on toes; after all, releasing Jena is the job of the PMC.

I think this pull request is moot, it is very difficult to test it with the Docker build as part of the source build, as basically you will require a new compilation of the whole thing inside `Dockerfile`, or you would have to use a Maven plugin to build the Docker image (which would then require Docker installed). 

Jena already have a massive git repository which can take considerable time & disk to build, and I don't think it's right to add more to it.

I think Jena and Fuseki should become two official images under the `_` namespace in Docker H
hub - and then there should be a mini-git repository "jena-docker" - similar to [my jena-docker](https://github.com/stain/jena-docker) - but hosted under `apache/`, which could then be referenced from https://github.com/docker-library/official-images. 

That image will then simply pull the latest release from `dist.apache.org` and one only need to update version number and sha1 on any new release. As the binary is the same (just unpacked), no additional PMC release vote should be needed.

Alternatively there could be a `jena` repository under https://github.com/docker-library/ managed by some PMC members.

Obviously as such an image should be "official" I am unable to try to work on it in that way, as I am neither a committer nor on the PMC.

ajs6f:
@stain I'm not that familiar with Docker-- is `apache/` in Docker Hub an official Apache account for some kind of central Docker repo?

stain:
It looks like it - although they seem to violate the ASF release processes as non-voted over commits are being released  straight to the hub from their master branch.

ajs6f:
Are commits showing up there a release, or a mirroring of the source code repository like Github?

PR #51, opened 2015-04-02
jena-text multilingual indexing

amiara514:
Hi,
Instead of having a default Lucene index, this version of jena-text allows to associate localized indexes with jena datasets (english, french and spanish for the moment).

See README.md for use cases.


afs:
Hi there - thank you very much for the pull request.  It looks very interesting.

Can I ask a couple of things:

1. Who owns the copyright on the code?  if you work for a company or institution, often the company or institution owns the copyright.
1. This is based on jena 2.12.1?  I tried to apply it to the current codebase but there have been 2 significant contributions since then and the pull request does not align with the codebase nowadays.

amiara514:
Hi,
Thanks for the comment,
 
1. I'm working for a university research center. 
There is no explicit copyright for this code, changes are done inside a GNU GPL open source project (https://github.com/LICEF/comete).
Is it a standard way to proceed ?
 
2. Yes, more precisely on tag jena-1.12.1-rc2 (rev 050c298ada38749a1ff166a77851b963991e4785)
 
 
PS: In the proposed version of code, both transactions in Lucene and Jena dataset are synchronized.
 
 
 
Regards
Alexis
 
 
 
 

 
Date: Tue, 7 Apr 2015 10:11:50 -0700

afs:
Hi Alexis,

I see that Comète is GPL3-licensed. That means that the project can incorporate source code licenced under the Apache License and still make the overall license GPL3, this is what required by GPL3.  The Free Software Foundation and the Apache Software Foundation both recognize this.

The other way isn't possible - if software, or changes, are licenced under GPL3, then an Apache License system would need to become GPL3 to fulfil the requirements of GPL3.  

If the changes can be separated from the rest of the GPL3-codebase and if you are able to license them as Apache License, then at least the licensing issues are sorted out.  I acknowledge that they might not be possible because of the way Comète is setup.

amiara514:
Hi Andy,

I think that my previous message was not clear.

Changes made for jena-text are independant from Comète. They are already under the Apache License. Comète uses this fork as an external dependency.
So I think there's no licensing issue.

osma:
This looks very interesting - especially the synchronization of transactions, which is very useful if your data changes often (mine does). Did I understand correctly that if you remove a triple that was indexed with jena-text, the corresponding entry gets deleted from the jena-text index as well? That would be a great improvement!

I think the README.md needs some work. Now this is called a "Fork of Apache Jena" but if this gets merged, it's no longer a fork.

Andy already mentioned that this doesn't play well with the current codebase, it cannot be automatically merged.

afs:
Hi Alexis, 

I'm glad we have got that cleared up.

@osma has touched an important point - the integration with full ACID transactions.  That's probably a major part of the fact the merge doesn't work.

amiara514:
Hi Osma, 

yes it is, for both of quad actions (add or delete).
I've changed the commit behavior that was done automatically on each action on the index.
For the moment, it's the responsability of the client code to commit or rollback the index (finishIndexing() or abortIndexing() methods) in phase with the associated dataset transaction.
I think that synchro should be manage directly from jena (jena-arq?) 

Andy, I have not yet checked why the merge doesn't work. Probably that ACID requirements are not totally respected. There's also maybe a conflict with the contributions you have mentionned last time.


However, I realize that we have here two distinct issues, the multilingual-index and the dataset/index transactions synchronization.
Do you think it should be a better way to close this issue and re post 2 distinct pull requests from 2 related branches ?

osma:
Hi Alexis!

> For the moment, it's the responsability of the client code to commit or rollback the index (finishIndexing() or abortIndexing() methods) in phase with the associated dataset transaction.
I think that synchro should be manage directly from jena (jena-arq?) 

Yes, that sounds sensible. For example I use jena-text mainly via Fuseki, so there is no opportunity for me to tell Fuseki to call any jena-text methods.

> However, I realize that we have here two distinct issues, the multilingual-index and the dataset/index transactions synchronization.
> Do you think it should be a better way to close this issue and re post 2 distinct pull requests from 2 related branches ?

Indeed it seems that this pull request addresses two separate issues. If you can easily split that into two separate pull requests, it would help merging them. It's easier to address new features one at a time.

Also rebasing your code on the current upstream jena-text would of course help. A pull request that doesn't apply to the current code is not very easy to merge...

-Osma

amiara514:
Hi Osma
If you want to use jena-text with Fuseki, you need to attach an assembler description. Read the configuration section from [Text searches with SPARQL](https://jena.apache.org/documentation/query/text-query.html). In my opinion, using jena-text by code is more flexible. It let's you combine usage of 'default' and 'indexed' datasets.

amiara514:
Hi Andy, 
Finally, I will try to prepare soon 3 new separate proposals (based on current codebase) :
- multilingual index
- update of text index on triple deletion
- synchro of dataset/index transactions
Are you agree with that ?

osma:
> If you want to use jena-text with Fuseki, you need to attach an assembler description. Read the configuration section from Text searches with SPARQL. 

I know, I've done this for a long time. I just wanted to point out that if your current code requires extra synchronization calls in the calling code, that won't work for the Fuseki use case.

> In my opinion, using jena-text by code is more flexible. It let's you combine usage of 'default' and 'indexed' datasets.

I'm sure it's more flexible, but this is really an apples vs oranges comparison since there are different use cases with different requirements. First and foremost I need a SPARQL endpoint on top of my data, the text index is a nice extra. I could use Virtuoso or some other triple store instead, but Fuseki+jena-text works pretty well for me.

> Finally, I will try to prepare soon 3 new separate proposals (based on current codebase) :

Sounds very good to me!

amiara514:
> I just wanted to point out that if your current code requires extra synchronization calls in the calling code, that won't work for the Fuseki use case. 

Indeed, so it's definitely the dataset transaction mechanism to manage the index transaction.

amiara514:
> I just wanted to point out that if your current code requires extra synchronization calls in the calling code, that won't work for the Fuseki use case. 

Indeed, so it's definitely the dataset transaction mechanism to manage the index transaction.

PR #53, opened 2015-04-29
Lucene index synchro on triple deletion (jena-text)

amiara514:
Hi,
This code synchronize the Lucene index (from jena-text) when a triple is removed from the associated graph. It is based on a trick  for exact match on Lucene. See it [here](http://blogs.perl.org/users/mark_leighton_fisher/2012/01/stupid-lucene-tricks-exact-match-starts-with-ends-with.html).


osma:
Hi Alexis! Thanks for the code!

This just screams for unit tests that show that it actually works in all cases...

I'm a bit worried about the BORDER_DELIMITER trick used here. It seems very fragile. For example, what happens if I have these two triples indexed using jena-text:

ex:paris rdfs:label "Paris"@fr .
ex:paris rdfs:label "Paris"@en .

Then the second triple is deleted. Will both entries be dropped from the Lucene index?

Similar things may happen for variants that differ only in letter case (which will be folded to lowercase by most analyzers), or singular/plural forms that get stemmed into the same base form. These may all cause false matches when entries are about to be deleted.

If I had to implement something like this, I'd instead try to store the original literal in the Lucene index, as intact as possible. In practice, putting the string value and language tag in separate fields which store the values without tokenizing (like the uri and graph fields currently). Then at deletion time it would be easy to query for the exact same entry and remove it, and only it, from the Lucene index.

amiara514:
Hi Osma,
> ex:paris rdfs:label "Paris"@fr .
ex:paris rdfs:label "Paris"@en .

Indeed, both of them will be dropped (however, can be solved with multi language proposal)

>Similar things may happen for variants that differ only in letter case (which will be folded to lowercase by most analyzers), or singular/plural forms that get stemmed into the same base form. These may all cause false matches when entries are about to be deleted.

Yes... it works for my needs but it's true that it's not a global solution.

I don't think that storing the original literal in non-debug mode is a good way to follow. It will result in a huge index and will affect performance.
Instead of that, storing a smaller identifier generated from a hash of a key (like uri+literal+language) will surely be a better solution.

Agree with that ?

osma:
> Indeed, both of them will be dropped (however, can be solved with multi language proposal)

I'm curious, how would the multi language proposal help with this problem?

> I don't think that storing the original literal in non-debug mode is a good way to follow. It will result in a huge index and will affect performance.
> Instead of that, storing a smaller identifier generated from a hash of a key (like uri+literal+language) will surely be a better solution.

I agree. A hash with sufficient uniqueness guarantees could work here. I think it should incorporate all parts of the quad (subject, predicate, literal value + language, named graph). This way, the entries in the Lucene index would correspond exactly to quads. When a quad gets deleted, the corresponding entry can be uniquely identified via the hash and deleted from the Lucene index as well.

But I'm not sure how this would interact with the recent work done by Chris Dollin (@ehedgehog) on cross field conjunctive queries (pull request #42 and [JENA-686](https://issues.apache.org/jira/browse/JENA-686)). In that work, the Lucene index is modified so that entries ("documents") in the Lucene index do not correspond to triples/quads, but instead there is one entry per subject resource (which may incorporate information from several triples/quads), allowing more efficient conjunctive queries on the Lucene index level. How would you deal with deletions in such a scenario?

amiara514:
>I'm curious, how would the multi language proposal help with this problem?

Multilingual index manages dynamically one index per language. Hence two same literals with different languages are not stored in the same index.

For the "hash solution", it works fine with a sha1. So we have one more field by doc, but I don't think it's embarrassing for the final index size. Should I commit it ? 

I don't know either if it can disturbs the conjonctive stuff.
However, the addEntity interacts with the updateEntity, and entries are still corrspond to triples/quads isn't it ?

osma:
> Multilingual index manages dynamically one index per language. Hence for the case where we have two same literals with different languages, they will be not stored into the same index. 

Ah, I see. But this still doesn't help for cases where there are small differences between literals within the same language, for example singular/plural forms that get stemmed by the analyzer, or variations in capitalization.

> For the "hash solution", it works fine with a sha1. So we have one more field by doc, but I don't think it's embarrassing for the final index size. Should I commit it ? 

For me this looks like a sensible solution. But I would love to hear comments from others, in particular on the next issue:

> I don't know either if it can disturbs the conjonctive stuff.
> However, the addEntity interacts with the updateEntity, and entries already correspond to triples/quads isn't it ?

With the current default configuration, yes, jena-text entries correspond to triples/quads. But with the conjunctive query support, that is no longer the case. There is a wider issue here - is jena-text primarily an alternative triple/quad index, or is it actually an entity index that just happens to work on triples in the default configuration? The latter case makes deletion much more difficult, as there is no longer a 1:1 mapping between quads and Lucene documents.

osma:
There was a discussion thread around this topic on the Jena dev mailing list in late February, starting with this message: http://mail-archives.apache.org/mod_mbox/jena-dev/201502.mbox/%3CCAPTxtVNzCAEKNQepeQ3hZsBmd7a1%2BNhJo0My1GOJF0GorMmYKA%40mail.gmail.com%3E

amiara514:
> Ah, I see. But this still doesn't help for cases where there are small differences between literals within the same language, for example singular/plural forms that get stemmed by the analyzer, or variations in capitalization.

It's exactly that!
So, I push the hash solution which cover all previous cases.

For the other issue (with conjonctive query), maybe deletion have to be managed with an updateDocument ?

osma:
[copying from the email I sent to the dev list, apparently the bot didn't notice my message]

Thanks a lot Chris, this was indeed very useful! So you're already doing synchronization in your ppd-index code and handling deletions in an appropriate, if somewhat brutal, way. That should make it easier to implement synchronization in the regular case, where quads correspond exactly to Lucene documents.

afs:
What's now the relation of this #53 to #64?

osma:
My impression is that this is on hold, Alex has been pushing for #64 to be merged first. @amiara514 should confirm.

Content-wise, these are not very closely related, but they touch the same jena-text code.

amiara514:
Like @osma said, there's no direct relation between those pulls (except jena-text)
Both were mixed in the previous proposal. I had splitted in 2 distinct pulls for a better understanding.

[#53](https://github.com/apache/jena/pull/53) is a small patch for cleaning Lucene index on SPARQL triple deletion (with literal objects).

[#64](https://github.com/apache/jena/pull/64) is a larger one for enabling linguistic index.
About that, a resume of #64 have been posted in another thread of dev mailing list (20-05-2015).
See details [here](http://mail-archives.apache.org/mod_mbox/jena-dev/201505.mbox/raw/%3CBLU181-W56D79F6BA6AC7103317BC6ECC20%40phx.gbl%3E/1/2).

afs:
Thanks for the explanation.  (I was just wondering about cleaning up after the earlier discussion on another out of date PR.)

osma:
@amiara514 What's the status of this? Obviously there are now some merge conflicts that need to be resolved, but other than that, do you think it's ready for merging? I for one would like to see (a version of) this merged, because it would help a lot in making efficient text index queries if you could rely on the index not containing stale entries.

Looking at the diffs, I see there are some changes to jena-arq that don't seem related to the main purpose of this code. Some of that is just cosmetic changes like adjusting whitespace. Then there is some QueryVisitor code, and some changes to ARQ tests. I doubt these patches are related to jena-text?

A couple of notes on the jena-text part:

1. You seem to always use the "uid" field to store the SHA-1 checksum. However, all other field names in jena-text, including entityField to store the URI, have so far been configurable using the assembler or EntityDefinition. Should this field name be made configurable as well? Not having it set could then mean "backwards compatible mode" where checksums are not stored and old entries are thus not deleted.

2. You are using DigestUtils.shaHex() to calculate the checksum, but this seems to be a deprecated method nowadays. I suggest using sha1Hex() or one of the other more explicit variants instead.

3. I'm not sure TextQueryFuncs.getLiteralLanguage is a very useful method (at least in its current form, with no error handling for null values etc), as one could just do node.getLiteral().getLanguage() directly. But this is a minor style issue.

amiara514:
@osma The status is still pending. Ok, I will fix minor changes to provide a re-mergeable version.

1. yes maybe the unique ID could be configurable. Hence the feature would be a kind of: "Enabling deletion mode". Agree with that ?

2. no problem for change it.

3. I will take a look

osma:
> yes maybe the unique ID could be configurable. Hence the feature would be a kind of: "Enabling deletion mode". But insertions without this option enabled are unremovable.
Agree with that ?

Sure, that's how it would work then. Without stored id's deletion is impossible.

Unit tests are missing too.

amiara514:
Hi, PR is mergeable again after conflict fixing of #72.

osma:
Thanks for fixing, and sorry for causing the conflict with #72.

It's good that you've added unit tests, however I think there could be more of them.

The current test adds and removes a resource, and only then checks that it's gone. I think it should check that it got into the index in the first place, otherwise it could be that text indexing is completely broken (no hits ever) and the test would still pass.

Would it be possible/easy the structure the unit tests so that all regular tests get executed also with the uid field enabled? After all, it shouldn't affect the current functionality if you enable deletion support (if it does it's a bug, either in implementation or the tests). You could get a lot of free tests this way and there would perhaps be no need for further tests of uid/deletion functionality.

A similar trick done with the graph-specific indexing, i.e. there are general tests in AbstractTestDatasetWithTextIndex, then a couple of extra tests for graph-aware functionality in AbstractTestDatasetWithGraphTextIndex, and finally TestDatasetWithLuceneGraphTextIndex pulls it together with the right (graph-aware) configuration. You could similarly try to reuse all the tests in AbstractTestDatasetWithTextIndex for the uid case. I admit the class hierarchy and naming is a bit complicated...

amiara514:
Ok I see, I will add a similar case of "graph-specific" for deletion support.

One question about graph indexing. In jena-text documentation you mention: "This allows for more efficient text queries when the query targets only a single named graph." But there's no example of using this (even in the tests).

amiara514:
@osma oups, forget my message.

amiara514:
I reorganized tests part

osma:
Excellent! Looks good to me now, I'm in favor of merging this.

Only minor thing I noticed: there are now two places where the SHA-256 checksum is calculated. I suggest moving this functionality to the Entity class (e.g. Entity.getUid()) so that the checksum logic is in only one place.

amiara514:
Done. 
I'll do changes of the "Deletion of Indexed Entities" part of jena-text doc once PR will be merged

osma:
@afs would you consider this ready for merging? I'd like to see this merged, so that I could start relying on it and perhaps implement some more jena-text features on top ;)

amiara514:
Yes absolutly, you can merge it... before another conflict ;-)

afs:
Ack!

afs:
PR merged. Many thanks for working on this.

Please could we have some documentation for the feature?

amiara514:
@afs like for the other doc I put it in a gist : https://gist.github.com/amiara514/ef839adb7c5cb9dd697d
It covers only the "Deletion of Indexed Entities" section of text-query.mdtext

afs:
Great - I'll integrate it as soon as I have a moment free to do so.

PR #54, opened 2015-04-30
Deprecating Map1 in favor of Java8 Function

ajs6f:
Deprecating Map1 in favor of Java8 Function.

I believe I have managed to avoid mangling any public contracts presented by Jena.

### DIFF
--- a/jena-core/src/main/java/jena/schemagen.java
+++ b/jena-core/src/main/java/jena/schemagen.java
@@ -30,6 +30,7 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.text.SimpleDateFormat;
 import java.util.*;
+import java.util.function.Function;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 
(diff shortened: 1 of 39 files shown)


afs:
Looks good but I'm having a problem with {{OntResourceImpl.AsMapper}}.  This is removed in {{OntResourceImpl}] but still in use in {{OntClassImpl}}, {{EnumeratedClassImpl}}, {{BooleanClassDescriptionImpl}}, {{DataRangeImpl}}. Maybe some other changes didn't get committed/pushed?

ajs6f:
Hm. That does sound weird. I'll take a look and straighten it out.

afs:
Created JENA-929 to help track this long-term. Just mention JENA-929 in comments and the JIRA wil be updated automatically.

afs:
Restoring {{AsMapper}} seemed to fix things - at least , it compiled then, just a bunch of warnings about use of {{Map1}} and unused {{java.util.Function}} imports so it does look like a minor sync issue.

ajs6f:
Yep-- Git weirdness. I think that should square it. I'm not mentioning the Jira issue here because Git mistakes don't seem like part of the long-term record.

ajs6f:
Right-- I just amended the commit to fully excise AsMapper.

afs:
Got it - it's clean now.  I remove unused imports, replaced Statement.Util and some Triple statics by rewriting the few uses in place e.g. Statement::getSubject.

ajs6f:
Thanks!

P.S. Sorry about unused imports and the like. That was sloppy.

PR #55, opened 2015-05-01
Migrating Filter to Java 8 Predicate

ajs6f:
Continued work on JENA-929.

Specifically, we now begin to migrate o.a.j.util.iterator.Filter and associated machinery to Java 8's new Predicate and associated machinery.


ajs6f:
NOT TO BE MERGED YET. This is just a first draft. There is clearly more to do to use Java 8 syntax to clarify filtered iteration and I just wanted to get something out to the committers to avoid any surprises, hence this "early" PR.

afs:
Looks good.  If it would be helpful to you, we can break this up into several steps in anyway so as to make it progressive improvement? e.g. even as specific as as rename `Filter.accept` as `Filter.test` as one step on the overall path.

ajs6f:
Totally up to you. It's all clear to me and I don't require to parcel it out, but that's because I've been knee-deep in it. {grin}
If you think it would make for a clearer evolution in the record, then sure, let's slice it up into logical moves. The method rename would be a little more granular than I would like to get. Perhaps maybe these steps:

- Change method names, extend `Filter` to `Predicate`, massive use of `Predicate` instead of `Filter`.
- Switch from explicit construction of `Predicate`s to lambda-syntax.
- Manipulations to iterators to improve the readability of filtering and make it look more like stream processing.

What do you think?

afs:
FYI: For iterators see `Iter` (in jena-base).  Again, pre-java8 but has a use in applying stream-list functions to iterators, instead of converting iterator to stream to iterator.  (It came about because not all iterators start as Jena-originated iterators; it applied to Iterable to but that's gone with Java8.)

afs:
Looks like a good plan.  The last point (manipulations) is possibly open-ended than the first two.

ajs6f:
Okay-- thanks for pointing out `Iter`! It's a bit strange, though. It doesn't seem to be used all that much. For example, `NiceIterator.toSet()` doesn't use `Iter.toSet(Iterator)`, it reimplements. Is `Iter` where you'd like to center the migration to `Stream`s? Is the intention to use `Iter` as much as possible?

As to "manipulations" being open-ended, are you suggesting that I should break this PR apart or add more to it? (That is, add more to the last point.)

ajs6f:
It looks like `Iter` makes use of some types that also have nice Java 8 replacements. Would a migration from `org.apache.jena.atlas.iterator.Transform<T, R>` to `Function` belong in this PR or another?

ajs6f:
Actually, there's also `org.apache.jena.atlas.iterator.Filter<T>`, which seems to be superseded by Java 8's `Predicate`. Maybe another PR is due here.

afs:
History! It's used in jena-sdb and jena-tdb.  It originated out of work on SDB (pre-iterables) and needed to work on general iterators, not (just) ExtendedIterators.  It has not been accessible to core until the jena3 refactoring.

I think keep as two PRs - from memory, `Iter` and `ExtendedIterator` don't intersect much.

ajs6f:
Ah, history. {grin}

Okay, sounds like we should leave this current PR alone, if you feel that you can accept it as is.

I'll send another which moves through `jena-base` and swaps out any parts that have straightforward Java 8 replacements that I find.

Then I'll move on to the saga of `ExtendedIterator`, and as I do that, I'll try to shift the footprint onto `jena-base` and away from re-implementations.

Does that sound reasonable?

afs:
OK - to confirm this PR is not "draft" any more but proposed final.

`ExtendedIterator` is different because it is in the `Graph` interface from `find`.  That means it is used not just within the codebase of Jena but also by external projects/products that extend Jena.

That does not mean it can't change; it needs wider discussion of the tradeoffs though. Jena3 is not compatible with unchanged Jena2 applications but at the same time, if the gap is too great, the changeover is likely longer.

ajs6f:
Confirmed.

Your remark about `ExtendIterator` completely makes sense to me. It's out there as a public contract. When I suggested on the mailing list changing it to get closer to `Stream`, I meant without breaking the methods that are already there. That's why I believe that up through that change, my proposal shouldn't really create much disruption. Of course, I'll wait for a clear go-ahead to spend any time on that.

To be clear though, the move to pushing `NiceIterator`'s implementation code up into `ExtendedIterator` as default implementation is pretty non-controversial, right? Shall I  send in a nice small PR for that?

PR #56, opened 2015-05-04
Jena base to java8

ajs6f:
Work for https://issues.apache.org/jira/browse/JENA-931

Using Java 8 types and removing code that duplicates functionality found inside the current footprint.


afs:
This PR is quite big so comments will be incremental. In fact, it might be better to tease out different sets of changes.  Could you give a quick overview of the changes?

Overall, this looks like a great set of changes, including using classes available in Google Guava to replace ones in Jena (e.g. MultiMap, MultiSet).

#### `oaj.atlas.lib.Closeable`

Adding `extends java.io.Closeable` causes about 76 more warnings in the code base because `java.io.Closeable` is interpreted by the compiler specially. Some of the classes that use `oaj.atlas.lib.Closeable` aren't suitable for try-resource.

Maybe some of these classes don't really need closeable but that needs checking.  Can we pull this out as a separate strand and not include it in the general Java8 overhaul?

#### `NodeTransform`

This is an important abstraction in the SPARQL optimizer.  Can we keep it, define it as `extends Function<Node, Node>` and keep use of `NodeTransform`? (There are times in Java when it would be nice to say "interface/class A is the same as B" where B is a complex expression, but you can't).  It makes finding `NodeTransform`s easier and makes interfaces clearer as to purpose.  The optimizer code is not general java code.

#### `@Deprecated`

I noticed this on `Lib.equal` and it seems to be a style in the PR to say "Prefer" in the javadoc. It gets highlighted in javadoc if `@deprecated` is used.

ajs6f:
I'm definitely up for breaking this down. :) 

The major moves here are: 

- I replaced any interface inside oaj.atlas for which I could find a functionally-equivalent Java 8 type with that Java 8 type, and resolved the various syntactical changes that resulted (mostly just renaming of things, of which you can see quite a lot).

- Following on that, I used lambda notation to shorten and clarify code that directly involved those types from step 1. (In other words, I didn't go looking for opportunities, I just took them when I happened to be "in the neighborhood" and saw them: there are plenty of more places, I think, to exercise lambda syntax).

- A few places where I could determine clearly that a method or type was not in use anywhere in the codebase and that the functionality provided was either available elsewhere (typically in Guava) or constructable in Java 8 in an obvious "one-liner", I removed it.

- Lastly, for a few types (notably 'MultiMap' and 'MultiSet') I happened to know that excellent well-exercised Guava versions already existed, so I swapped them out and toasted the Jena versions.

To your specific points:

- `oaj.atlas.lib.Closeable`:  I clearly misunderstood the semantics of this type. I thought it was simply a narrowing of ` java.io.Closeable` to disallow checked exceptions, so that the additional warnings would be a _good_ thing. Since it is independent, I will remove the new relationship: would it not also make sense to add some kind of comment explaining the particular semantics on oaj.a.l.Closeable`? I have to admit, I didn't look at it twice before thinking "try-with-resources".

- `NodeTransform` I will put it back in in the optimizer. If I understand you correctly, the type itself should be declared inside oaj.sparql and used only there, yes?

- Ah, I completely forgot about `@deprecated`. Just plain forgetfulness. I'll get those comments tagged correctly.  

Shall I make these changes, and then we can figure out how to parcel this stuff out into reasonable commits?

JENA-931

ajs6f:
Actually, for `NodeTransform`, perhaps it could be even more tightly scoped to just `org.apache.jena.sparql.algebra.optimize`?

afs:
The vast majority of the PR looks good including converting code "in passing". MultiMap?mulitSet are unlikely to have leaked (and can be restored if need be); I've already converted my own code elsewhere not to use them.

Where NodeTransform is located is relatively minor - it's not just in "optimize" as the engine to replace nodes by other nodes is used elsewhere (e.g. update, few other places). Other systems may have used it where it is for their own optimizer operations so its a balance of impedance to adopting Jena3 compared to where might go NodeTransform, NodeTransformLib and NodeTransformOp. While such cases are not going to be numerous, they already have the most work to do to migrate. 

In other words, I'd class it as "minor" and concentrate on other things but that's just my personal take. The sooner we can get to a sufficiently stable version to start basing a release on the better - no "second system" effect, or "third system" effect in this case - and keeping the migration work down encourages users to migrate.

ajs6f:
Okay, I'll plug `NodeTransform` back in at oaj.sparql.graph, since that is where it was: minimal disruption. I've got the JavaDocs square and re-separated the two `Closeable`s in a third commit (which I assume we can squash and reslice as seems meet for the long-term record.

ajs6f:
`NodeTransform` is back-- I think that's all the emendation we wanted, yes? 

Did you have thoughts about dividing out this large PR?

afs:
I'll look again but the NodeTransform and the Closeable issues seem to be the main things and after that the size is due to the number of places that Java8 improves things.  Checking in my own code MultiMap/MulitSet don't seem likely to have leaked too far.

So if we have discussed the different aspects of the PR, it can be applied in the current form as far as I'm concerned. The next step is any other review and discussion then integrate it.

ajs6f:
Sounds good. After this, wherever goes the conversation about Java 8, the Streams API, Commons RDF, etc., I will keep PRs much smaller. I felt comfortable trying to "get away" with this goliath because so much of it is the most brainless renaming.

I'll wait to hear from other commiters/devs and I'll be ready to incorporate additional feedback.

PR #57, opened 2015-05-07
Relying more on Guava impl for caching

ajs6f:
The idea here is to simplify Jena's code by relying a little more on the Guava implementation for caching.


jena-base/src/main/java/org/apache/jena/atlas/lib/cache/Cache1.java
```diff
@@ -1,127 +0,0 @@
...
-import java.util.Objects;
-import java.util.concurrent.Callable ;
-import java.util.function.BiConsumer;
-
-import org.apache.jena.atlas.iterator.SingletonIterator ;
-import org.apache.jena.atlas.lib.Cache ;
-
-/** A one-slot cache.*/
-public class Cache1<K, V> implements Cache<K,V>
-{
```
afs: Proposal: keep this.

jena-base/src/main/java/org/apache/jena/atlas/lib/CacheFactory.java
```diff
@@ -33,7 +33,7 @@
 
     /** Create a null cache */
     public static <Key, Value> Cache<Key, Value> createNullCache() {
-        return new Cache0<>() ;
+        return new CacheGuava<>(0) ;
     }
```
afs: Cache0 is really just a "complete the set" implementation.  If any use as a dummy, the weight of CacheGuava might not be worth it.  Proposal: just remove Cache0 (it does not fit with getOrFill anyway).

jena-base/src/main/java/org/apache/jena/atlas/lib/CacheFactory.java
```diff
@@ -43,7 +43,7 @@
 
     /** One slot cache */
     public static <Key, Value> Cache<Key, Value> createOneSlotCache() {
-        return new Cache1<>() ;
+        return new CacheGuava<>(1) ;
     }
```
afs: CacheGuava is a bit heavy for a one-slot cache.  While Cahce1 may not be used currently, if it were, then keeping it as simple, small implementation is better. Proposal: Keep Cache1 as is.

jena-base/src/main/java/org/apache/jena/atlas/lib/CacheFactory.java
ajs6f: You mean see to it that `Cache0` is not used at all, then instead of reimpling it with Guave, just get rid of it entirely?

jena-base/src/main/java/org/apache/jena/atlas/lib/CacheFactory.java
ajs6f: Shall I check and see whether `Cache1` is really in use? It tastes to me like memoization, which might be better supplied by some other means in Java 8, e.g via statics that use [Supplier](http://docs.oracle.com/javase/8/docs/api/java/util/function/Supplier.html).

jena-base/src/main/java/org/apache/jena/atlas/lib/cache/CacheSimple.java
```diff
@@ -37,120 +37,81 @@
 
 public class CacheSimple<K,V> implements Cache<K,V>
 {
-    private final V[] values ; 
-    private final K[] keys ;
-    private final int size ;
-    private int currentSize = 0 ;
-    private BiConsumer<K,V> dropHandler = null ;
+	private Map<K,V> internalCache;
+
```
afs: A Map grows as more, different, keys are added whereas the existing CacheSimple is a fixed size slot replacement cache with the property that it never resizes. 

CacheSimple is a fast cache, traded against a poorer than LRU replacement/eviction policy of same-hash slot replacement.  Using a Map changes its characteristics.

ajs6f: I did not understand that-- I will rework this PR with attention paid to that point and to whatever answer you give about `Cache0` and `Cache1`. Thanks for the attention!

jena-base/src/main/java/org/apache/jena/atlas/lib/CacheFactory.java
afs: Yes. It would not be hard to bring it back!

ajs6f: Yes, hooray for source control. {grin} I will adjust this PR accordingly.

jena-base/src/main/java/org/apache/jena/atlas/lib/CacheFactory.java
afs: It's not used (if it is used elsewhere, then as it's not API, reasonable change is OK). The cache interface is useful though and `Supplier` does not cover the `getOrFill` pattern.  It also allows a switch from `Cache1` to a bigger, better cache as the interface is the same.

jena-base/src/main/java/org/apache/jena/atlas/lib/cache/CacheSimple.java
afs: If keys K1 and K2 have the same short hash, then one adding K2 overwrites k1 in CacheSimple but in a java Map, you end up with two entries K1 and K2, with internal reorganisation as needed and the Map is bigger. CacheSimple is a sort of fixed size map.

jena-base/src/main/java/org/apache/jena/atlas/lib/CacheFactory.java
ajs6f: Okay, `Cache1` stays with us. I'll adjust this PR to that effect.

jena-base/src/main/java/org/apache/jena/atlas/lib/cache/CacheSimple.java
ajs6f: How about using [`LinkedHashMap`'s ability to keep fixed-size?](http://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashMap.html#removeEldestEntry-java.util.Map.Entry-)?

jena-base/src/main/java/org/apache/jena/atlas/lib/CacheFactory.java
ajs6f: Hm. There is one use of `Cache0`. It's indirectly, via `CacheFactory`, and it occurs in `BlockMgrCache` at [line 62](https://github.com/apache/jena/blob/master/jena-tdb/src/main/java/org/apache/jena/tdb/base/block/BlockMgrCache.java#L62). I'm not sure enough as to `BlockMgrCache`'s responsibilities to feel comfortable rewiring it to not require the existence of a `BlockMgrCache.readCache` without your go-ahead. I would feel better if I better understood the comment "Caches are related so we can't use a Getter for cache management." Does this refer to `readCache` and `BlockMgrCache.writeCache`?

jena-base/src/main/java/org/apache/jena/atlas/lib/cache/CacheSimple.java
ajs6f: As you can see, I've added this change and also a test to demonstrate that `CacheSimple` now has the required fixed-size behavior.

jena-tdb/src/main/java/org/apache/jena/tdb/base/block/BlockMgrCache.java
```diff
@@ -58,10 +58,7 @@ private BlockMgrCache(int readSlots, int writeSlots, final BlockMgr blockMgr)
     {
```
ajs6f: I've made changes to `BlockMgrCache` sufficient to permit the removal of `Cache0` from the codebase, but I'd much appreciate a careful review here. Also, I have to admit I think this new code (with all the `null`-checks) to be pretty ugly. I'd like to rewrite it using `Optional`, but I'd first like to be sure that my basic understanding is correct.

P.S. Actually, I'd like to make `Cache::getIfPresent` return `Optional`. That would really eliminate a lot of `null`-checking.

jena-base/src/main/java/org/apache/jena/atlas/lib/cache/CacheSimple.java
afs: LinkedHashMap implement an LRU policy and is what the old CacheLRU used.  We swapped that to CacheGuava and removed CacheLRU. Using LinkedHashMap again means we have two LRU caches.

CacheSimple has a specific policy of slot replacement should that be wanted.  We can leave it be and decide whether to delete it later.

ajs6f: Okay, I'll back out the changes to `CacheSimple`.

jena-tdb/src/main/java/org/apache/jena/tdb/base/block/BlockMgrCache.java
afs: I hadn't realized that Cache0 was used.  My bad.

The use in BlockMgrCache for a "not a cache really" avoid null checking or some such in BlockMgrCache.

ajs6f: No problem. It was definitely to avoid `null`-checking, because `BlockMgrCache` sprouted all kinds of `null`-checks when I killed `Cache0`. As I commented there, I would like to rewrite `BlockMgrCache` to use `Optional` but I'm not yet confident enough that I understand it well enough, and I'm not yet familiar enough with the test framework to understand how well-covered `BlockMgrCache` is and how careful I need to be around it.

afs: `Cache::getIfPresent` returning Optional means that the calling code ends up checking Optional rather than null.  Looking in TDB, for example, it's no gain, in fact it's a loss because of the additional object creation.  

The reason is that these aren't in streams where Optional skipping is natural.

The node cache is performance sensitive and uses getIfPresent so changes here might have undesirable effects.

What would be good is looking at the unrelated cache code in jena-core and seeing if it can now be consolidated (including using LRU caching which IIRC it does not).

jena-base/src/main/java/org/apache/jena/atlas/lib/cache/CacheSimple.java
ajs6f: Just to be clear, am I correct to be thinking that a `CacheSimple` should never expand beyond the size of its constructor argument? I'm asking because that's exactly what my new test checked, and it passes for my `LinkedHashMap` impl, but not for the original. I build a `CacheSimple<Integer>(5)` and put 10 things in it, mapped from 10 different keys each to the same value, and then its `size()` is 10 when it should only be 5. Or am I misunderstanding the policy here?

jena-tdb/src/main/java/org/apache/jena/tdb/base/block/BlockMgrCache.java
afs: Cache0 is the `Optional` of caches :-)

`BlockMgrCache` is mainly used for 32-bit TDB only - 64-bit uses memory mapped file and does not layer a `BlockMgrCache` on top.

ajs6f: Okay, I did notice that stuff (in `jena-core`) but I wanted to be sure I thoroughly understood this `jena-base` stuff before tackling it. I'm glad I did, because I clearly did _not_ understand these classes well enough to shift things onto them.

ajs6f: If this whole class is only for 32-bit operation, then I reckon maybe I should put `Cache0` back in and move on. It doesn't seem like a good place to invest time, no?

jena-base/src/main/java/org/apache/jena/atlas/lib/cache/CacheSimple.java
afs: CacheSimple has a plain old, fashioned bug in maintaining the size!

jena-tdb/src/main/java/org/apache/jena/tdb/base/block/BlockMgrCache.java
afs: The current clearing up makes a good checkpoint. 

This is ready to integrate now?

ajs6f: Up to the presence of that bug in `CacheSimple`: see my comment there.

ajs6f: Wait-- that comment you made about a bug seems to be gone. I guess the question I'm asking is whether we want to include `TestCacheSimple` and fix the bug, or leave out `TestCacheSimple` and just mark the bug.

afs:
I fixed the bug while I was looking at it :-)  I'll review and try out the current state and merge it if there are no further comments.

ajs6f:
Sounds good to me!

ajs6f:
Because the comment disappeared-- clearly something else was going on, as @afs later indicated he already fixed the bug. What is the problem?

ajs6f:
Oh, cool. Yeah, there's a number of message-emitting sources here, eh? Asynchronous transactions are tricky. {grin}

ajs6f:
One last thing: it seems like a good idea to me to put a note in `CacheSimple` about the slot-replacement and fixed-size qualities that are demanded there. Or maybe instead to add a test to `TestCacheSimple` to indicate the slot-replacement requirement. Would you like me to do that?

afs:
Thank you.  Merged into code base, and reflecting the conversion here: Keep Cache0 (avoid nulls!) and keep the general CacheSet over a Cache due to otherwise changing the eviction policy in TDB missing nodes cache.

ajs6f:
Cool. Maybe there should be a comment on `CacheSetImpl` explaining why it must "inherit" the eviction policy from the wrapped `Cache`. It's not obvious to someone who doesn't know how it's being used downstream in TDB.

afs:
Good suggestion. I'll add that as I do #59.

PR #60, opened 2015-05-11

ajs6f:
Remove some forms of dead code from jena-arq, jena-base, jena-core, jena-csv, jena-iri, an jena-permissions.  Includes removals for:

Unthrown checked exceptions
Needless typecasts
Unused imports


afs:
The files in package "thrift/wire" starting `RDF_` are autogenerated by the Thrift code generator. They'll get uncleaned if that's remade. (Thrift code generation is not part of the build to reduce the dependency; the script and master grammar are in jena-arq/Grammar/).

ajs6f:
Hm. Do you think it's worth taking a look at those Thrift tools to see if I can get them to gen cleaner code? (It has to be said that I've never used Thrift.)

afs:
Thrift is another Apache project. In the Thrift case, all Jena adds the `@SuppressWarnings("all")` at the top.  As the warnings also depend on the local compiler settings, its hard for a general code generator to generate clean code. From my experience, that's typical for code generators.

ajs6f:
Yeah, I know what Thrift is, but I have no direct experience with it. I've worked with a few code generating frameworks (mostly for WSDL-related stuff) that gave relatively good clean code and allowed you to put filters in for some amount of customization.

It sounds like that's just not worth the effort here. Shall I back out the changes in that package for this PR?

afs:
Please, if it's easy to do that; ditto the javacc overwrites.  (sparql_10 are sparql_11 actually code directly from SPARQL spec so anything touching them gets special attention!)

ajs6f:
Okay, I will make that happen, but I'm not sure what you mean by "javacc overwrites". You mean methods `sparql_10` and `sparql_11`?

afs:
Yes.  (Sorry about assuming that was clear - over-familiarity with the code layout)

ajs6f:
Okay, I've added two commits that should revert those bad changes.

afs:
Another code generated batch : org/apache/jena/iri/impl/Lexer* are JFlex generated 

(these files do all say "autogenerated" in them)

afs:
I've picked out as much material from the PR as I can and applied according to the discussions (so not removing interface declarations when the base type includes it, left exceptions on \@overrides in main code, removed unnecessary exceptions more completely in test code).

I have not handled jena-permissions, a module I know less about. One of the reasons split by module is helpful is that different people tend to look after different modules so one person does not have oversight of all code.

PR #61, opened 2015-05-11

ajs6f:
Remove some forms of dead code from jena-elephas and its submodules, jena-querybuilder, and jena-fuseki1 and jena-fuseki2, including:

Unthrown checked exceptions
Needless typecasts
Unused imports

### DIFF
--- a/jena-elephas/jena-elephas-io/src/main/java/org/apache/jena/hadoop/rdf/io/input/QuadsInputFormat.java
+++ b/jena-elephas/jena-elephas-io/src/main/java/org/apache/jena/hadoop/rdf/io/input/QuadsInputFormat.java
@@ -18,8 +18,6 @@
 
 package org.apache.jena.hadoop.rdf.io.input;
 
-import java.io.IOException;
-
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.mapreduce.InputSplit;
 import org.apache.hadoop.mapreduce.RecordReader;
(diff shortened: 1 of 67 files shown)


rvesse:
I not declaring unthrown checked exceptions is a Java 8-ism?

rvesse:
Looking at this I realise that Java 8 potentially presents a problem for Elephas because the entire Jena 3 stack is Java 8 and so Elephas has to be Java 8 yet most Hadoop clusters are still Java 6/7 which is going to create issues for end users.

However other than simply documenting this there is not much that we can do about it since recompiling Jena 3 for Java 7 is no longer an option either because of all the Java 8 code changes that have gone in

kinow:
Umbrella ticket in Hadoop to support Java 8 https://issues.apache.org/jira/browse/HADOOP-11090

Guess it will still take a while.

afs:
Actually, it looks like there is quite an impetus to get Hadoop on Java8 (even if Java6/7 bytecodes). EOL is making it a focus of attention.

afs:
One possibility is to keep Java8-isms out of Elephas (on a practical basis) for the moment so backporting to 2.13.0 is a low barrier.  Obviously, bug fixes that go into 3.0.0 and not backported are a potential issue and hard to spot in all the cleaning.  JIRA should help there. My sense is that most fixes have not been in core/arq which Elephas is reliant on.

afs:
Would it be an easy task to possible to split this into jena-elephas related changes and changas related to other modules?  I'm not so familiar with jena-elephas but I could progress the other bits and pieces.  (If it's not an easy task, I can take a patch and edit it but the history is lost).

ajs6f:
I don't know if my Git-fu is up to splitting it, but I could certainly just regenerate a new PR that is more restricted. Do you want to see one module : one PR? Or can all of the Elephas-related deltas come together?

afs:
jena-elephas, querybuilder, and the rest (Fuseki's) would be great.   (Actually, looking at the querybuilder file, I wonder if that is a style for a purpose ... @Claudenw?)

ajs6f:
Should I wait for comment from @Claudenw or move to make three new PRs:

- `jena-elephas`
- `jena-querybuilder`
- everything else that was included here

?

afs:
I can process the "everything else" PR because I think I understand those code modules well enough.  If it's not much work, 3 PRs is a clear cut split but if it turns into a lot of work, plan B is that I can take a patch file and edit it.

ajs6f:
Closed in favor of #67, #68, #69.

PR #62, opened 2015-05-11
JENA-938: Nonfunctional cleanup in jena-jdbc

ajs6f:
Remove some forms of dead code from jena-jdbc, including:

Unthrown checked exceptions
Needless typecasts
Unused imports

### DIFF
--- a/jena-jdbc/jena-jdbc-core/src/main/java/org/apache/jena/jdbc/JenaDriver.java
+++ b/jena-jdbc/jena-jdbc-core/src/main/java/org/apache/jena/jdbc/JenaDriver.java
@@ -266,7 +266,7 @@ public abstract class JenaDriver implements Driver {
     }
 
     @Override
-    public boolean acceptsURL(String url) throws SQLException {
+    public boolean acceptsURL(String url) {
         if (url.startsWith(DRIVER_PREFIX + this.implPrefix)) {
             return true;
         } else {
@@ -812,7 +812,7 @@ public abstract class JenaDriver implements Driver {
      * @param props
      *            Properties
      */
-    protected void modifyProperties(Properties props) throws SQLException {
+    protected void modifyProperties(Properties props) {
         // Default implementation does nothing
     }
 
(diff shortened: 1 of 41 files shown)


afs:
Would it be an easy task to possible to split this into jena-jdbc related changes and changes related to other modules? (If it's not an easy task, I can take a patch and edit it but the history is lost).

ajs6f:
See my comment [here](https://github.com/apache/jena/pull/61#issuecomment-102834531). Same question applies.

afs:
In this PR, jena-jdbc and the rest would make it easier from my point of view.

ajs6f:
Meaning split it into changes to 
- jena-jdbc and its submodules
- all other modules
? For a total of two new PRs? Sorry to be so precise-- I just want to make sure we're on the same wavelength.

afs:
Good to be precise!

2 PRs would be fantastic - (1) jena-jdbc and everything below it; they all go to make up jena-jdbc and (2) the rest.

ajs6f:
Cool. I'll get on that. Might take a day or two, because I'm going to use it as a chance to improve my Gitcraft and try to understand how to split commits on a condition like "is under a directory / isn't under that directory".

ajs6f:
Okay, I've split this guy between this PR and #70.

afs:
Thanks for the splitting.

ajs6f:
No problemo. Because it was a single commit, it was no harder than a `git reset` to before that commit and then some tap dancing between `git add`ing changes, committing them and making PRs from the commits. 

Does it seem reasonable to you to go forward to making some module-by-module PRs that remove unused `private` methods? Then people familiar with each module in question could decide whether the removed code was really dead or whether it has some potential future use. I would expect that at least some of it will be genuinely removable, right?

afs:
Re: private methods

I turned on Eclipse's "warn on unused private member" which includes private methods, (it does not seem to separate the two).

A quick scan of code I recognized suggests they represent functionally there for completeness, maybe also used at one time, just not currently active.  That represents useful code IMO.  

I didn't find much that seemed removable so, from my limited experiment, it does not look like it would be so useful / much return-on-investment.

"Mostly harmless"

ajs6f:
Okay, that's cool. If you're confident that that stuff is "real", only not currently interesting, then we need it. I would have removed it and retrieved it when wanted from VC, but as you say, mostly harmless.

PR #63, opened 2015-05-11

ajs6f:
Remove some forms of dead code from the src/main resources in jena-core and jena-base, including:

Unnecessary @SuppressWarnings annotations
Unused local variables

### DIFF
--- a/jena-core/src/main/java/jena/rdfcompare.java
+++ b/jena-core/src/main/java/jena/rdfcompare.java
@@ -20,9 +20,6 @@ package jena;
 
 import static jena.cmdline.CmdLineUtils.setLog4jConfiguration;
 
-import java.net.URL;
-import java.io.FileInputStream;
-
 import org.apache.jena.rdf.model.* ;
 
 /** A program which read two RDF models and determines if they are the same.
@@ -108,11 +105,6 @@ public class rdfcompare extends java.lang.Object {
     
     protected static void read(Model model, String in, String lang) 
       throws java.io.FileNotFoundException {
-        try {
-            URL url = new URL(in);
-            model.read(in, lang);
-        } catch (java.net.MalformedURLException e) {
-            model.read(new FileInputStream(in), "", lang);
-        }
+        model.read(in, lang);
     }
 }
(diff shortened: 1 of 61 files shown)


jena-core/src/main/java/jena/rdfcompare.java
```diff
@@ -108,11 +105,6 @@ protected static void usage() {
     
     protected static void read(Model model, String in, String lang) 
       throws java.io.FileNotFoundException {
-        try {
-            URL url = new URL(in);
-            model.read(in, lang);
-        } catch (java.net.MalformedURLException e) {
-            model.read(new FileInputStream(in), "", lang);
-        }
+        model.read(in, lang);
```
afs: An historical note: 

That is a way to handle URLs and plain file names.  It only works as plain `model.read(in, lang);` if RIOT is around otherwise there is no HTTP processing of `model.read(String, .....`

afs:
This PR looks easier to process (it does not mix a lot of modules; sometimes that mix means several people are going to be interested, but in different parts).

How does this fit with the other PRs? Are all the JENA-938 PR independent and so can be processed in any order?

ajs6f:
They are, I hope, okay to merge in arbitrary order.

I made these PRs based on the idea that each covers enough ground to be substantive without being too large to review comfortably. If there is a policy about making PRs that don't address more than one module, I'm happy to work within that, but it would be great to have that published somewhere. I really could not find a lot of documentation at all about how to contribute beyond [this](https://jena.apache.org/getting_involved/index.html), which doesn't address any of the issues of style vs. substance or scope that have come up in this ticket.

afs:
When I applied the changes, I get a lot of "hiding" warnings.  Seems to me the root cause is the hiding in the first place, not the use of  `@SuppressWarnings("hiding")` which, for me at least in Eclipse, is a warning to be suppressed.  Does the use of `@SuppressWarnings("hiding")` cause any problems?

ajs6f:
I didn't see any such expansion of warnings, but that may have to do with my IDE settings.

Generally, my take is that it's better to remove the warnings by fixing the underlying problems, but I realize that that's far from a universal opinion. Maybe we should can this PR until there is a little more consensus about code style/code cleaning? The discussion, it seems to me, is not really done yet.

afs:
Generally yes - fixes are better when possible.  In this case though (on `.factory`) there is a coding pattern than was fine at the time and now isn't.  The hiding effect is due to inheritance of `ResourceImpl` and `OntResourceImpl` .  I'll integrate what I can and see if fix up does not muddy things up in this case.

ajs6f:
Okay, whatever is useful is hopefully there for you without too much trouble. Let me know if I can help-- part of my personal motive for these wild-swinging "clean up" PRs is to familiarize myself with a large code base.

PR #75, opened 2015-06-04
Enhance Javadocs for QueryEngineFactory

ajs6f:
Based on remarks by Claude Warren at http://markmail.org/message/tingkcxzs3sf2xz4


afs:
Thanks - I've used this with clarifications.  Normally, only query execution process would call the factory operations.

PR #76, opened 2015-06-09
Eliminate <jena-core>/jena.cmdline, deprecate rdfcat

ajs6f:
https://issues.apache.org/jira/browse/JENA-962


jena-core/src/main/java/jena/RuleMap.java
```diff
@@ -50,7 +53,7 @@
  * </p>
  */
 public class RuleMap {
-    static { setLog4jConfiguration() ; }
+    static { setCmdLogging("jena-log4j.properties") ; }
```
afs: Can this be plain `setCmdLogging()` which (1) is sensitive to a local log4j.properties if the user that and (2) has a built-in config in case no properties file is found.

jena-core/pom.xml
```diff
@@ -56,12 +56,24 @@
 			<groupId>xerces</groupId>
 			<artifactId>xercesImpl</artifactId>
 		</dependency>
+    
+        <dependency>
+            <artifactId>commons-cli</artifactId>
+            <groupId>commons-cli</groupId>
+            <version>1.3</version>
+        </dependency>
```
afs: Can we have dependency version management in jena-parent please?

jena-core/src/test/java/org/apache/jena/test/TestPackage.java
```diff
@@ -32,7 +33,7 @@
 public class TestPackage extends TestCase {
```
afs: `Test_rdfcat` now has a lot of deprecation warnings. Is it just a case of `@SuppressWarnings("deprecation")` on the class or is there something else going on?

jena-core/src/main/java/jena/RuleMap.java
ajs6f: Sure, I'll make that change.I'm still getting my head (slowly) around the logging facilities.

jena-core/pom.xml
ajs6f: Yep, I'll take care of that.

jena-core/src/test/java/org/apache/jena/test/TestPackage.java
ajs6f: It's because I `@Deprecate`d `rdfcat`. I'll suppress those warnings, since they convey no real information.

afs:
Thanks for clearing it up.  I've just pulled it into the codebase.

PR #77, opened 2015-06-16
Move arq.cmd.* to jena-base's jena.cmd

ajs6f:
https://issues.apache.org/jira/browse/JENA-961

This is a multi-module PR, but I could find no other way to do it: since several Jena modules depend on `jena-arq` for their command-processing, moving machinery from `jena-arq` to `jena-base` caused changes in them. However, those are minor changes (e.g. changes in imports) and the meat of the PR is where it should be, in `jena-arq` and `jena-base`.


afs:
Moving into jena-base is probably the best first step.  It separates it from gathering all the commands which may well not be good idea anyway, or at least, a pointless exercise that risks breaking people's scripts (that may not use the provided ones, or have installed copies).

ajs6f:
Sorry, @afs, when you say "Moving into jena-base is probably the best first step." are you referring to `ArqCmdGeneral`/`CmdGeneral`?

ajs6f:
Okay, I've got `ModVersion` out to `jena-base` and moved `Version` into it as an inner class (since `ModVersion` is the only client of `Version`). We are still doing the the reflection thing because I need to track down all the classes that supply the `VERSION` constant. Then I can try to replace that machinery with something else.

@afs when you suggest handing in a callback to `ModVersion`, are you thinking that there would be an interface `Versioned` or the like that would feature a member `VersionInfo getVersion()` or something like that? And `Jena`, `ARQ` and the other classes that currently offer the `VERSION` constant, once I track them down, will instead impl that type?

afs:
Thank you. Code merged with some changes: 

* "Version" was used elsewhere (jena-elephas) so I put it back.
* Didn't update rdfcat; the `BiConsumer` additions are only use that old code so it seemed better to leave the old code self contained.
* Kept `processArgs`, not `Consumer.accept`. It is only of the same signature style. The operation name to "accept" didn't seem a good name in all places in the code.
* ModVersion left in CmdGeneral.

PR #78, opened 2015-06-17
Fix jena-querybuilder test

ajs6f:
The change to how rdf:type is serialized seems to have zonked a test in `jena-querybuilder`.


ajs6f:
https://issues.apache.org/jira/browse/JENA-969

afs:
It would be better to be robust as to which form is produced.  Both are correct.

PR #80, opened 2015-06-24
Deprecating and removing some iterators from jena-core

ajs6f:
https://issues.apache.org/jira/browse/JENA-966


afs:
Hi - could you try closing this pull request from your end? It seems to have got stuck in some way, whereby the commit controls don't have any effect. It's not the only one; the same happened to one of my PRs and the way I sorted it was from the requester end. Thanks.

afs:
That unstuck the PR - thanks.

PR #81, opened 2015-06-26
jena-text stored literals: initial functionality and tests for Lucene

osma:
This PR implements a feature where it's possible to store the original literal values in the jena-text Lucene index and to access them when querying the index. It works like this:

1) Configure jena-text to store literals (default is off) using the new `text:storeValues` setting. Note that you also need the `text:langField` setting in the entity map for language tags and datatypes to be handled correctly.

```
    #text:directory <file:Lucene> ;
    text:directory "mem" ;
    text:storeValues true ;
    text:entityMap <#entMap> ;
    .

<#entMap> a text:EntityMap ;
    text:entityField "uri" ;
    text:langField "lang" ;
    [...]
```

2) Add some data, say this triple:

```
:myresource rdfs:label "My resource"@en .
```

3) Query like this:

```
SELECT * {
  (?s ?score ?literal) text:query "resource" .
}
```

In the query result, `?literal` will be bound to `"My resource"@en.`

It also works with typed literals (as requested by @ehedgehog). The datatype will be stored in the langField using a special prefix (currently `^^`) which ensures that it cannot be interpreted as a language tag.

There are unit tests for all the basic cases (simple, non-default property, language tags, datatypes).

I had to change the TextIndex API slightly again, to pass the queried property from TextQueryPF to TextIndexLucene/TextIndexSolr so that they know which field to look up values from. Since it was already changed recently, to return TextHit objects instead of Nodes, I wouldn't expect another change to hurt.

I've done a basic implementation for Solr as well which doesn't handle the language tags and datatypes (TextIndexSolr didn't have support for langField...), but it should be able to return at least the lexical value. I haven't been able to test this because of lack of documentation for the jena-text/Solr combination and possibly some bitrot in TextIndexSolr - last time I tried I couldn't get it working at all.

I can do documentation for this after it has been merged. Now that I have committer access I could merge this myself, but I'd like to get a couple of +1's before doing that.

### DIFF
--- a/jena-text/src/main/java/org/apache/jena/query/text/DatasetGraphText.java
+++ b/jena-text/src/main/java/org/apache/jena/query/text/DatasetGraphText.java
@@ -99,7 +99,7 @@ public class DatasetGraphText extends DatasetGraphMonitor implements Transaction
             String f = textIndex.getDocDef().getField(predicate) ;
             queryString = f + ":" + queryString ;
         }
-        List<TextHit> results = textIndex.query(queryString, limit) ;
+        List<TextHit> results = textIndex.query(predicate, queryString, limit) ;
         return results.iterator() ;
     }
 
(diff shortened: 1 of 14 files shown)


osma:
@afs Sure, now there is one: [JENA-978](https://issues.apache.org/jira/browse/JENA-978)

PR #83, opened 2015-06-29
Removing out-of-date comment and empty @Overrides from Dataset

ajs6f:
A comment in `Dataset` indicated that it did not implement `Transactional` when in fact it now does. I removed that comment and the now-unnecessary overriden methods.


afs:
I've removed the comment and left the pulled down redeclarations.  Transactions are central to datasets sop this makes it clearer. And it gets into the javadoc directly, not indirectly.

PR #84, opened 2015-06-30
[JENA-977] tdbloader2 script rewrite

rvesse:
This pull request contains a substantial rewrite of the `tdbloader2` scripts to make them more user friendly, flexible and robust.

## Dev Environment Notes

Previously it was a pain to run the scripts in a dev environment because they assume a class path of `$JENA_HOME/lib/*` which does not exist in a dev environment.  Therefore the POM for the distribution module was updated to use the maven dependencies plugin to generate the `lib/` directory during a `package` phase and clean it up during the `clean` phase which makes it much easier to set `JENA_HOME` to your working copy distribution module directory and enhance the scripts.

## Script Changes

The script changes are fairly extensive covering a number of areas.  The existing two scripts were split into four:

- `tdbloader2` - Main entry point which coordinates running the other scripts
- `tdbloader2data` - Script which runs the data phase of the build
- `tdbloader2index` - Script which runs the index phase of the build
- `tdbloader2common` - Script which provides functions common to all scripts

The now defunct `tdbloader2worker` script was removed, there was also outdated and broken scripts in `jena-tdb/bin/` which were also removed

### Symbolic Link and relative path handling

In rewriting the scripts some bugs with current treatment of `JENA_HOME` were addressed:

- If `JENA_HOME` is not set it tries to locate it from the scripts path but if the script is symbolic linked then it uses `readlink -f` however the `-f` option has completely different meaning on BSD/OS X so could fail in some cases.  The scripts now all contain a `resolveLink` function which handles the OS specific behaviour appropriately.
- If `JENA_HOME` is itself set to a symbolic link then the scripts could fail to invoke the other scripts, if `JENA_HOME` is a symbolic link it is now resolved appropriately

There were also similar bugs that could occur if the database location given or data file paths were themselves relative and/or symbolic links.  At various points the scripts will now resolve symbolic links and make paths absolute which makes the scripts less error prone.

### Option Handling

The scripts now all support a variety of user friendly options and has built-in help for those.  The main script `tdbloader2` accepts all the options and handles passing relevant options through to the appropriate child scripts as necessary.

All options that previously were only exposed via environment variables are now exposed as command line options.  For some the existing environment variables (`JVM_ARGS` and `SORT_ARGS`) are still honoured if these options are not otherwise specified.

Each of the tdbloader2 scripts now provides a `printUsage` function which contains a detailed and user-friendly help summary.  A user can view this by running with the `-h` or `--help` option on each script.

### Incremental Builds

A `--phase` option is now supported on `tdbloader2` which takes a value of `all`, `data` or `index`.  `all` does a full build and is the default behaviour if phase is omitted.

The other two perform the appropriately named phase of the build.  This allows a build to be done in smaller incremental steps and also allows for the index phase of the build to be restarted which is useful because in my experience if you get past the data phase then the index phase has far more scope for error.

### Indexing Improvements

There have been a lot of improvements made to the indexing scripts:

- Warns if it looks like the disk where sort is storing temporary files may be too full
- Aborts if there is insufficient free disk space to sort an input file
- Warns if a given sort is likely to be external, adds additional warnings if the same sort may be short of disk space on the disk where sort is storing temporary files
- Provides progress reporting for sort when running in the foreground provided that the `pv` ([PipeViewer](http://www.ivarch.com/programs/pv.shtml)) tool is available

### Debugging

All scripts now support `--debug` and `--trace` options which add extra output

- `--debug` will add various additional debugging output during a build 
- `--trace` will set `set -x` on the scripts


afs:
The find `JENA_HOME` fragment was from `apache-jena/template.bin` -- should that be updated as well? (and the scripts regenerated).

rvesse:
Good point, I have now applied this fix to `template.bin` and regenerated all the scripts.  I also updated `cmd-maker` slightly since `tdbloader2` is no longer based on the template at all

rvesse:
@afs Any further comments or should I go ahead and merge?

PR #87, opened 2015-07-21
Bump versions of JUnit, Guava for Elephas

ajs6f:
https://issues.apache.org/jira/browse/JENA-994

Possibly off-target or too late for this ticket, but bumping Hadoop to latest release enables bringing Guava to latest release as well, a nice bonus.

### DIFF
--- a/jena-elephas/pom.xml
+++ b/jena-elephas/pom.xml
@@ -35,7 +35,7 @@
   <properties>
     <plugin.compiler.version>2.6.0</plugin.compiler.version>
     <arq.version>3.0.0-SNAPSHOT</arq.version>
-    <junit.version>4.11</junit.version>
+    <junit.version>4.12</junit.version>
     <mrunit.version>1.0.0</mrunit.version>
     <airline.version>0.9.2</airline.version>
   </properties>
@@ -70,7 +70,7 @@
       <dependency>
         <groupId>com.google.guava</groupId>
         <artifactId>guava</artifactId>
-        <version>11.0.2</version>
+        <version>16.0.1</version>
       </dependency>
 
       <!-- Jena Dependencies -->


afs:
Hadoop 2.7.0 was first released in April of this year and marked as not a production release. 2.7.1 was this month (July).  AFAICT neither CDH nor HDP are using 2.7.x at all.

Hadoop installations don't get upgraded very frequently; it's "non-trivial". IMO it is better to wait for users to ask for an update to get the pacing right.

ajs6f:
Have users asked for upgrades in the past?

ajs6f:
The version of Hadoop in use is controlled by a Maven profile. Perhaps we could offer a "reliable" default profile and a "latest release" profile for those looking to explore?

rvesse:
The existing profile is what I would define as a reliable default

We could perhaps move to 2.6.x as that is starting to make it into the latest builds of CDH, Hortonworks etc but as Andy notes people tend to upgrade infrequently in this space

ajs6f:
Okay, I'll back off on Hadoop and just offer a simple bump for Guava up to the last version that is compatible with the current Hadoop.

ajs6f:
I brought this back to just a bump for Guava (to an old, but less old, version) and JUnit.

afs:
Why does Elephas have it's own version management of JUnit and of jena-arq (aligned to the main build)?  Could this be left from original development?

ajs6f:
I know nothing of the story behind the chosen JUnit version, but I can tell you that the current version of Guava used in the main codebase (18.0) will not work with the version of Hadoop in use in Elephas now. (That's actually how I ended up trying to bump Hadoop.) This (16.0.1) is the latest version I could find that would. Guava 18.0 _will_ work with a later version of Hadoop, but I don't know how much later-- I only tried with 2.7.1.

afs:
On 21/07/15 16:00, Rob Vesse wrote:
> The existing profile is what I would define as a reliable default
>
> We could perhaps move to 2.6.x as that is starting to make it into the
> latest builds of CDH, Hortonworks etc but as Andy notes people tend to
> upgrade infrequently in this space
>
> —
> <https://github.com/apache/jena/pull/87#issuecomment-123360536>.
>

Rob - the current dependency is already Hadoop 2.6.0

	Andy

afs:
org.apache.hadoop:hadoop-common:jar:2.6.0 depends on com.google.guava:guava:jar:11.0.2 (provided).  

Just because Jena tests work isn't evidence Guava 16.0.1 works unfortunately.  Jena is a library and may be co-resident with other code.  See HADOOP-10101 including the initial attempt at migration to 15.0 which did not go in. At this very late stage, it needs a strong advantage for users for change.

(and this is why we shade guava for Jena!)

afs:
Junit version is now the same as the rest of Jena.  Version of jena-arq managed in dependencyManagement with an explicit version, not a property, because that keeps the release plugin happy (well, happier).

rvesse:
It may be possible to remove the explicit dependency on guava as a stop-gap

If memory serves the reason behind this was that dependency management in the Jena parents was causing a version conflict.  Since guava is no longer a dependency of other Jena modules (because it is shaded) it may be possible to simply remove the explicit dependency on Guava

Also completely agree with Andy, even if the tests pass the real test is that the demo application runs correctly on an actual Hadoop cluster where various dependencies are provided by the cluster

ajs6f:
Is that something that can be automated as a system test (test-driving the demo app in a simple cluster)?

afs:
```
[INFO] +- com.github.rvesse:airline:jar:0.9.2:compile
[INFO] |  +- javax.inject:javax.inject:jar:1:compile
[INFO] |  +- com.google.guava:guava:jar:11.0.2:compile
[INFO] |  \- org.apache.commons:commons-lang3:jar:3.3.2:compile
```
Presumably the explicit `dependencyManagement` is causing that version to be choosen over com.github.rvesse:airline.

I've fixed the JUnit and jena-arq dependency management.

Jena itself does need to depend on a runtime Guava because of shading (looking like a better decision all the time!).

I don't see this as a change needed for 3.0.0.  It can be sorted out at leisure.

rvesse:
Agreed, that reminds me that I am intending to strip Guava out of Airline anyway

afs:
Makes sense - then the choice of guava direct is whatever Hadoop dictates.

PR #92, opened 2015-09-03
JENA-1019 - Making Model attributes private and final in many vocabulary classes.

ramiromagalhaes:
As requested in https://issues.apache.org/jira/browse/JENA-1019, here is my change proposal.


afs:
The changes to what look like hand-written vocabularies look good.

Some of these files appear to have been autogenerated by schemagen (e.g the ones where the model is `static private m_model` already).

1. As they are private, the compiler will treat them as "effectively final" anyway.
1. The files may get overwritten.

afs:
Changes applied.  Thank you very much.

PR #93, opened 2015-10-01
JENA-1039: Avoid race-condition to OWL constants

stain:
Fixes [JENA-1039](https://issues.apache.org/jira/browse/JENA-1038) race condition.


afs:
See the discussion on JENA-1039.  A more comprehensive set of changes applied which should mean that this specific fix is not needed.  Testing will tell.

PR #94, opened 2015-10-14
JENA-624 

ajs6f:
JENA-624: new in-memory dataset implementation using persistent data structures

There are a lot of commits in this: I can squash them if desired, just let me know.

I will be happy to make any changes, large or small, no matter how numerous, right here in this actual branch and PR so that if/when it gets merged, the history and my authorship is preserved. Thanks!

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/query/DatasetFactory.java
+++ b/jena-arq/src/main/java/org/apache/jena/query/DatasetFactory.java
@@ -16,253 +16,226 @@
  * limitations under the License.
  */
 
-package org.apache.jena.query ;
+package org.apache.jena.query;
 
-import java.util.List ;
+import java.util.List;
 
-import org.apache.jena.assembler.Assembler ;
-import org.apache.jena.rdf.model.Model ;
-import org.apache.jena.rdf.model.Resource ;
-import org.apache.jena.sparql.ARQException ;
-import org.apache.jena.sparql.core.DatasetGraph ;
-import org.apache.jena.sparql.core.DatasetGraphFactory ;
-import org.apache.jena.sparql.core.DatasetImpl ;
-import org.apache.jena.sparql.core.assembler.DatasetAssembler ;
-import org.apache.jena.sparql.util.DatasetUtils ;
-import org.apache.jena.sparql.util.graph.GraphUtils ;
-import org.apache.jena.util.FileManager ;
+import org.apache.jena.assembler.Assembler;
+import org.apache.jena.rdf.model.Model;
+import org.apache.jena.rdf.model.Resource;
+import org.apache.jena.sparql.ARQException;
+import org.apache.jena.sparql.core.DatasetGraph;
+import org.apache.jena.sparql.core.DatasetGraphFactory;
+import org.apache.jena.sparql.core.DatasetImpl;
+import org.apache.jena.sparql.core.assembler.DatasetAssembler;
+import org.apache.jena.sparql.util.DatasetUtils;
+import org.apache.jena.sparql.util.graph.GraphUtils;
+import org.apache.jena.util.FileManager;
 
-/** Make Datasets and DataSources in various ways */
-
+/**
+ * Makes {@link Dataset}s in various ways.
+ *
+ */
 public class DatasetFactory {
-    /** Create an in-memory, modifiable Dataset */
-    public static Dataset createMem() {
-        return create(DatasetGraphFactory.createMem()) ;
-    }
 
-    /**
-     * Create an in-memory, modifiable Dataset. New graphs must be explicitly
-     * added using .addGraph.
-     */
-    public static Dataset createMemFixed() {
-        return create(DatasetGraphFactory.createMemFixed()) ;
-    }
+	/**
+	 * @return an in-memory, modifiable Dataset
+	 * @deprecated Prefer {@link #createGeneral()} or {@link #createTxnMem()}
+	 */
+	@Deprecated
+	public static Dataset createMem() {
+		return create(DatasetGraphFactory.createMem());
+	}
 
-    /**
-     * Create a dataset with the given model as the default graph
-     * 
-     * @param model
-     *            The model for the default graph
-     * @return Dataset
-     */
-    public static Dataset create(Model model) {
-        return new DatasetImpl(model) ;
-    }
+	/**
+	 * @return a transactional, in-memory, modifiable Dataset with MR+SW locking
+	 */
+	public static Dataset createTxnMem() {
+		return create(DatasetGraphFactory.createTxnMem());
+	}
 
-    /**
-     * Create a dataset: clone the dataset structure of named graohs, and share
-     * the graphs themselves.
-     * 
-     * @param dataset
-     *            Dataset to clone structure from.
-     * @return Dataset
-     */
-    public static Dataset create(Dataset dataset) {
-        return new DatasetImpl(dataset) ;
-    }
+	/**
+	 * @return a general-purpose, in-memory, modifiable Dataset
+	 */
+	public static Dataset createGeneral() {
+		return createMem();
+	}
 
-    /**
-     * Wrap a datasetgraph to make a mutable dataset
-     * 
-     * @param dataset
-     *            DatasetGraph
-     * @return Dataset
-     */
-    public static Dataset create(DatasetGraph dataset) {
-        return DatasetImpl.wrap(dataset) ;
-    }
+	/**
+	 * @return an in-memory, modifiable Dataset. New graphs must be explicitly added using .addGraph.
+	 */
+	public static Dataset createMemFixed() {
+		return create(DatasetGraphFactory.createMemFixed());
+	}
 
-    /**
-     * Create a dataset based on a list of URIs : these are merged into the
-     * default graph of the dataset.
-     * 
-     * @param uriList
-     *            URIs merged to form the default dataset
-     * @return Dataset
-     */
+	/**
+	 * @param model The model for the default graph
+	 * @return a dataset with the given model as the default graph
+	 */
+	public static Dataset create(final Model model) {
+		return new DatasetImpl(model);
+	}
 
-    public static Dataset create(List<String> uriList) {
-        return create(uriList, null, null) ;
-    }
+	/**
+	 * @param dataset Dataset to clone structure from.
+	 * @return a dataset: clone the dataset structure of named graohs, and share the graphs themselves.
+	 */
+	public static Dataset create(final Dataset dataset) {
+		return new DatasetImpl(dataset);
+	}
 
-    /**
-     * Create a dataset with a default graph and no named graphs
-     * 
-     * @param uri
-     *            URIs merged to form the default dataset
-     * @return Dataset
-     */
+	/**
+	 * Wrap a {@link DatasetGraph} to make a mutable dataset
+	 *
+	 * @param dataset DatasetGraph
+	 * @return Dataset
+	 */
+	public static Dataset create(final DatasetGraph dataset) {
+		return DatasetImpl.wrap(dataset);
+	}
 
-    public static Dataset create(String uri) {
-        return create(uri, null, null) ;
-    }
+	/**
+	 * @param uriList URIs merged to form the default dataset
+	 * @return a dataset based on a list of URIs : these are merged into the default graph of the dataset.
+	 */
+	public static Dataset create(final List<String> uriList) {
+		return create(uriList, null, null);
+	}
 
-    /**
-     * Create a dataset based on a list of URIs : these are merged into the
-     * default graph of the dataset.
-     * 
-     * @param uriList
-     *            URIs merged to form the default dataset
-     * @param fileManager
-     * @return Dataset
-     */
+	/**
+	 * @param uri URIs merged to form the default dataset
+	 * @return a dataset with a default graph and no named graphs
+	 */
 
-    /**
-     * Create a named graph container of graphs based on a list of URIs.
-     * 
-     * @param namedSourceList
-     * @return Dataset
-     */
+	public static Dataset create(final String uri) {
+		return create(uri, null, null);
+	}
 
-    public static Dataset createNamed(List<String> namedSourceList) {
-        return create((List<String>)null, namedSourceList, null) ;
-    }
+	/**
+	 * @param namedSourceList
+	 * @return a named graph container of graphs based on a list of URIs.
+	 */
 
-    /**
-     * Create a dataset based on two list of URIs. The first lists is used to
-     * create the background (unnamed graph) by merging, the second is used to
-     * create the collection of named graphs.
-     * 
-     * (Jena calls graphs "Models" and triples "Statements")
-     * 
-     * @param uriList
-     *            graphs to be loaded into the unnamed, default graph
-     * @param namedSourceList
-     *            graphs to be atatched as named graphs
-     * @return Dataset
-     */
+	public static Dataset createNamed(final List<String> namedSourceList) {
+		return create((List<String>) null, namedSourceList, null);
+	}
 
-    public static Dataset create(List<String> uriList, List<String> namedSourceList) {
-        return create(uriList, namedSourceList, null) ;
-    }
+	/**
+	 * Create a dataset based on two list of URIs. The first lists is used to create the background (unnamed graph) by
+	 * merging, the second is used to create the collection of named graphs.
+	 *
+	 * (Jena calls graphs "Models" and triples "Statements")
+	 *
+	 * @param uriList graphs to be loaded into the unnamed, default graph
+	 * @param namedSourceList graphs to be atatched as named graphs
+	 * @return Dataset
+	 */
 
-    /**
-     * Create a dataset container based on two list of URIs. The first is used
-     * to create the background (unnamed graph), the second is used to create
-     * the collection of named graphs.
-     * 
-     * (Jena calls graphs "Models" and triples "Statements")
-     * 
-     * @param uri
-     *            graph to be loaded into the unnamed, default graph
-     * @param namedSourceList
-     *            graphs to be attached as named graphs
-     * @return Dataset
-     */
+	public static Dataset create(final List<String> uriList, final List<String> namedSourceList) {
+		return create(uriList, namedSourceList, null);
+	}
 
-    public static Dataset create(String uri, List<String> namedSourceList) {
-        return create(uri, namedSourceList, null) ;
-    }
+	/**
+	 * Create a dataset container based on two list of URIs. The first is used to create the background (unnamed graph),
+	 * the second is used to create the collection of named graphs.
+	 *
+	 * (Jena calls graphs "Models" and triples "Statements")
+	 *
+	 * @param uri graph to be loaded into the unnamed, default graph
+	 * @param namedSourceList graphs to be attached as named graphs
+	 * @return Dataset
+	 */
 
-    /**
-     * Create a named graph container based on two list of URIs. The first is
-     * used to create the background (unnamed graph), the second is used to
-     * create the collection of named graphs.
-     * 
-     * (Jena calls graphs "Models" and triples "Statements")
-     * 
-     * @param uri
-     *            graph to be loaded into the unnamed, default graph
-     * @param namedSourceList
-     *            graphs to be atatched as named graphs
-     * @param baseURI
-     *            baseURI for relative URI expansion
-     * @return Dataset
-     */
+	public static Dataset create(final String uri, final List<String> namedSourceList) {
+		return create(uri, namedSourceList, null);
+	}
 
-    public static Dataset create(String uri, List<String> namedSourceList, String baseURI) {
-        return DatasetUtils.createDataset(uri, namedSourceList, baseURI) ;
-    }
+	/**
+	 * Create a named graph container based on two list of URIs. The first is used to create the background (unnamed
+	 * graph), the second is used to create the collection of named graphs.
+	 *
+	 * (Jena calls graphs "Models" and triples "Statements")
+	 *
+	 * @param uri graph to be loaded into the unnamed, default graph
+	 * @param namedSourceList graphs to be atatched as named graphs
+	 * @param baseURI baseURI for relative URI expansion
+	 * @return Dataset
+	 */
 
-    /**
-     * Create a named graph container based on two list of URIs. The first is
-     * used to create the background (unnamed graph), the second is used to
-     * create the collection of named graphs.
-     * 
-     * (Jena calls graphs "Models" and triples "Statements")
-     * 
-     * @param uriList
-     *            graphs to be loaded into the unnamed, default graph
-     * @param namedSourceList
-     *            graphs to be atatched as named graphs
-     * @param baseURI
-     *            baseURI for relative URI expansion
-     * @return Dataset
-     */
+	public static Dataset create(final String uri, final List<String> namedSourceList, final String baseURI) {
+		return DatasetUtils.createDataset(uri, namedSourceList, baseURI);
+	}
 
-    public static Dataset create(List<String> uriList, List<String> namedSourceList, String baseURI) {
-        return DatasetUtils.createDataset(uriList, namedSourceList, baseURI) ;
-    }
+	/**
+	 * Create a named graph container based on two list of URIs. The first is used to create the background (unnamed
+	 * graph), the second is used to create the collection of named graphs.
+	 *
+	 * (Jena calls graphs "Models" and triples "Statements")
+	 *
+	 * @param uriList graphs to be loaded into the unnamed, default graph
+	 * @param namedSourceList graphs to be atatched as named graphs
+	 * @param baseURI baseURI for relative URI expansion
+	 * @return Dataset
+	 */
 
-    public static Dataset make(Dataset ds, Model defaultModel) {
-        Dataset ds2 = new DatasetImpl(ds) ;
-        ds2.setDefaultModel(defaultModel) ;
-        return ds2 ;
-    }
+	public static Dataset create(final List<String> uriList, final List<String> namedSourceList, final String baseURI) {
+		return DatasetUtils.createDataset(uriList, namedSourceList, baseURI);
+	}
 
-    // Assembler.
-    /**
-     * Assembler a dataset from the model in a file
-     * 
-     * @param filename
-     *            The filename
-     * @return Dataset
-     */
-    public static Dataset assemble(String filename) {
-        Model model = FileManager.get().loadModel(filename) ;
-        return assemble(model) ;
-    }
+	public static Dataset make(final Dataset ds, final Model defaultModel) {
+		final Dataset ds2 = new DatasetImpl(ds);
+		ds2.setDefaultModel(defaultModel);
+		return ds2;
+	}
 
-    /**
-     * Assembler a dataset from the model in a file
-     * 
-     * @param filename
-     *            The filename
-     * @param resourceURI
-     *            URI for the dataset to assembler
-     * @return Dataset
-     */
-    public static Dataset assemble(String filename, String resourceURI) {
-        Model model = FileManager.get().loadModel(filename) ;
-        Resource r = model.createResource(resourceURI) ;
-        return assemble(r) ;
-    }
+	// Assembler-based Dataset creation.
 
-    /**
-     * Assembler a dataset from the model
-     * 
-     * @param model
-     * @return Dataset
-     */
-    public static Dataset assemble(Model model) {
-        Resource r = GraphUtils.findRootByType(model, DatasetAssembler.getType()) ;
-        if ( r == null )
-            throw new ARQException("No root found for type <" + DatasetAssembler.getType() + ">") ;
+	/**
+	 * Assemble a dataset from the model in a file
+	 *
+	 * @param filename The filename
+	 * @return Dataset
+	 */
+	public static Dataset assemble(final String filename) {
+		final Model model = FileManager.get().loadModel(filename);
+		return assemble(model);
+	}
 
-        return assemble(r) ;
-    }
+	/**
+	 * Assemble a dataset from the model in a file
+	 *
+	 * @param filename The filename
+	 * @param resourceURI URI for the dataset to assembler
+	 * @return Dataset
+	 */
+	public static Dataset assemble(final String filename, final String resourceURI) {
+		final Model model = FileManager.get().loadModel(filename);
+		final Resource r = model.createResource(resourceURI);
+		return assemble(r);
+	}
 
-    /**
-     * Assembler a dataset from a resource
-     * 
-     * @param resource
-     *            The resource for the dataset
-     * @return Dataset
-     */
+	/**
+	 * Assemble a dataset from the model
+	 *
+	 * @param model
+	 * @return Dataset
+	 */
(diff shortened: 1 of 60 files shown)


afs:
Really good to see this.  

If you could squash commits that would be great.

It will probably take some while and some discussion.

If it is meaningful to split into parts that can be considered separately, that might also help but I would not be surprised if that is not natural to the PR.

ajs6f:
Okay, @afs, I've squashed everything into two commits. They are basically orthogonal, although there is a tiny leak from the first into the second (in `DatasetGraphWithRecord`), because I found a clearer way of typing mutations later in the summer. The first is the dataset-with-operation-record that I did earlier, that provides a real abort function to any wrapped MRSW dataset. The second is the persistent-data-structure-based MR+SW dataset impl. Let me know if you would prefer these to be broken out into two separate PRs-- that's no problem, I just didn't want to fire multiple PRs at once without clearing it. I can easily remedy the small leak and build two PRs.

You are right in that other than that, there aren't really natural boundaries inside each commit. That's a little unfortunate, because I know it will make review more of a pain in the neck, but at least I can promise that of the extant code into which this is being flung, nothing is touched outside of adding dependencies to POM files. This is all new stuff and shouldn't bother anything else.

Now back to ISWC!

afs:
This comment is mostly reviewing the Memory Dataset design, not the Journalling part.

#### Performance Status

@ajs6f, you [mentioned that using Java8 streams](http://mail-archives.apache.org/mod_mbox/jena-dev/201510.mbox/%3CBC838704-0486-4094-90C6-D8C8DA044351%40email.virginia.edu%3E) was leading to overhead costs.  Is it the streams or the fact that the datastructure has to be traversed and it's less efficient than the array bunches in the memory graph?

#### DatasetFactory

I don't see factory operations and, what will be important, javadoc.  What's the status here?

See discussion on dev@ [(part of this email)](http://mail-archives.apache.org/mod_mbox/jena-dev/201510.mbox/%3C5610F36D.8090602%40apache.org%3E).

#### Assembler
Ditto.

#### Documentation

A couple of paragraphs
* what is it
* how to use it 
* Choice between general and  txnmem datasets

Maybe best to integrate with other documentation - pull out "transactions" from TDB and modify it.

#### Persistent datastructures

These could go in their own package.

Some basic tests needed.

#### Dependency management

The jena-arq POM changes should have version management carried up in jena-parent.

mockito already has some version in jena-parent and is used in jena-core.

Add mockito and awaitility to jena-arq/DEPENDENCIES for accounting.

#### Mocking in tests

(minor) Any reason to not just do the action and test for an effect?

#### Warnings

Remove warning or suppress if necessary.
(unnecessary semicolons, a couple of casts, one javadoc error)

### Journaling

Even though not used, I'm minded to include it in the codebase with the understanding it is "subject to change". I can imagine it could help jena-text for example.

LockMRPlusSW : tests?

ajs6f:
Thanks for the feedback, @afs! I'll take it point-by-point:

- Performance Status: The suspicion to which I was referring is that it's because using the Streams API generates more objects and those objects have more state than using iterators. It should be noted that I have not been able to verify this. I could, by rewriting the PR to avoid the Streams API, but it would require a good bit of work to do that and I think it would be _much_ harder to read. I think the majority of the performance decline is because of the changes in design, not incidental stuff like this.

- DatasetFactory: I just hadn't touched that yet to avoid touching anything outside of my code until we are happy with the basic machinery. I will happily put that in as per your proposal in that message. See a forthcoming commit. {grin}

- Assembler: I'm still reading through the assembler subsystem and making sure I understand it before extending it. I should be able to supply something soon.

- Documentation: Yes, I haven't done anything here. I assume you are talking about documentation in the website, right? Not Javadocs?

- Persistent datastructures: Cool, I'll break these out and give them some tests.

- Dependency management: Okay, I'll square that away. I didn't realize that $module/DEPENDENCIES even existed. What does that support?

- (Not) Mocking in tests: I think that should be doable in most cases. I usually prefer mocking because it makes everything very explicit, but if Jena's habit is otherwise, I'm happy to follow it.

- Warnings: Hm, didn't see those (except for the casts). I will make sure everything is clean. Those casts are really weird, because sometimes I have seen compilation fail without them, but they do seem to be unnecessary and the types are infer-able. I will check into it more thoroughly.

- Journaling: I think including it could be a good way also to get some feedback. Is there some Jena-standard way to mark something as "slightly experimental" a la Google's `@Beta` annotation?

- LockMRPlusSW: I will slap some tests on that.

ajs6f:
@afs I can't find any of the unnecessary semicolon warnings-- can you give me a pointer?

afs:
On 26/10/15 13:51, A. Soroka wrote:
> @afs <https://github.com/afs> I can't find any of the unnecessary
> semicolon warnings-- can you give me a pointer?
>
> —
> <https://github.com/apache/jena/pull/94#issuecomment-151139579>.
>

maybe you fixed them but I get:

Unnecessary semicolon	DatasetGraphInMemory.java 
/jena-arq/src/main/java/org/apache/jena/sparql/core/mem	line 191Problem

Unnecessary semicolon	DatasetGraphInMemory.java 
/jena-arq/src/main/java/org/apache/jena/sparql/core/mem	line 195

Unnecessary semicolon	PMapTupleTable.java 
/jena-arq/src/main/java/org/apache/jena/sparql/core/mem	line 69

Unnecessary semicolon	TestQuadTableForms.java 
/jena-arq/src/test/java/org/apache/jena/sparql/core/mem	line 80

Unnecessary semicolon	TestTripleTableForms.java 
/jena-arq/src/test/java/org/apache/jena/sparql/core/mem	line 59

ajs6f:
Got 'em, thanks. I had forgotten to turn on Eclipse's warnings for "empty statements".

ajs6f:
@afs For "Persistent datastructures: These could go in their own package." are you thinking of a subpackage of the one I'm currently using in `jena-arq`: `org.apache.jena.sparql.core.mem`, or since these structures are really just about arranging `Nodes`, maybe stick 'em in `jena-core`?

afs:
The datastructure code in jena-base:org.apache.jena.atlas.lib.

As things stand today "dataset" related stuff needs to go in jena-arq.  In jena-core+jena-arq is one thing, split for historical and "not too big" reasons. A reorganisation, e.g. DatasetGraph+Quad next to Graph and Triple, would be disruptive to many applications unfortunately.

ajs6f:
Okay, you mean then to add a `jena-base`:`org.apache.jena.atlas.lib.persistent`? Can do.

ajs6f:
Oh, okay, that's different: so you're thinking `jena-base`:`org.apache.jena.atlas.lib.persistent` or the like? Cool.

ajs6f:
@afs On the assembler front, did you want me to contribute a totally new assembler that works for this new dataset impl, or to alter the current `DatasetAssembler` to use this new impl instead of what it now uses?

afs:
The assembler for general purpose datasets can do a lot of things because it can build many kinds of model and include them in the dataset.

The in-memory dataset is somewhat simpler. It can load files, either quads, or triples (to the default graph) or load triples to a specific graph. 

Looks to me like start afresh. Maybe the best way is sketch out the design in email and get discussion going.

ajs6f:
Makes sense, @afs. I'll head down that road.

jena-arq/src/main/java/org/apache/jena/sparql/core/assembler/InMemDatasetAssembler.java
```diff
@@ -0,0 +1,67 @@
...
+				pDefaultGraph) : root.hasProperty(pGraph) ? getResourceValue(root, pGraph) : null;
+		if (defaultGraphDef != null) dataset.setDefaultModel(retrieve(defaultGraphDef, assembler, mode));
+		// or with ja:data
+		final Predicate<RDFNode> isResource = n -> {
+			if (n.isResource()) return true;
+			throw new DatasetAssemblerException(root, "Not a resource: " + stringForRDFNode(n));
+		};
+		multiValue(root, data).parallelStream().filter(isResource)
+				.forEach(defaultGraphNode -> read(dataset, defaultGraphNode.asResource().getURI()));
+
```
afs: .parallelStream() does not make sense here.

I think it needs to be sequential and all in the same transaction. The memory dataset does not support parallel updates. At best, it'll just get queued on the lock. But then error messages are going to be random order.

jena-arq/src/main/java/org/apache/jena/sparql/core/assembler/InMemDatasetAssembler.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * An {@link Assembler} that creates in-memory {@link Dataset}s.
+ *
+ */
+public class InMemDatasetAssembler extends AssemblerBase {
+
+	@Override
+	public Dataset open(final Assembler assembler, final Resource root, final Mode mode) {
+		checkType(root, MemoryDataset);
+		final Dataset dataset = createTxnMem();
```
afs: Transactions for data reading?

jena-arq/src/main/java/org/apache/jena/sparql/core/assembler/InMemDatasetAssembler.java
```diff
@@ -0,0 +1,67 @@
...
+		final Predicate<RDFNode> isResource = n -> {
+			if (n.isResource()) return true;
+			throw new DatasetAssemblerException(root, "Not a resource: " + stringForRDFNode(n));
+		};
+		multiValue(root, data).parallelStream().filter(isResource)
+				.forEach(defaultGraphNode -> read(dataset, defaultGraphNode.asResource().getURI()));
+
+		// Assemble or load named graphs
+		multiValue(root, pNamedGraph).parallelStream().filter(isResource).forEach(namedGraphNode -> {
+			final Resource namedGraphResource = namedGraphNode.asResource();
```
afs: can be just  'multiValueResource(root, pNamedGraph)'

jena-arq/src/main/java/org/apache/jena/sparql/core/assembler/InMemDatasetAssembler.java
afs: `defaultGraphNode` is a misleading name.   More like `file`?

ajs6f: The hope from using `parallelStream` was indeed to queue on the lock, but get _reading_ going in parallel. IOW, reading could require network access or filesystem access that could very well be parallelized. But I take your point about the order of error messages. That could be really terribly confusing. I'll pull out the parallelization.

ajs6f: `defaultGraphNode` could be a file, could be a network resource: how about `defaultGraphdocument`?

jena-arq/src/main/java/org/apache/jena/sparql/core/assembler/InMemDatasetAssembler.java
ajs6f: Yes, that's a bad oversight. I'll fix that.

jena-arq/src/main/java/org/apache/jena/sparql/core/assembler/InMemDatasetAssembler.java
ajs6f: Yeah, I wondered about that, but the semantics of what I've got here are actually different. If you look at `isResource`, it throws an exception when the thing is a literal (which would make no sense in an assembler construction). That's what the other assemblers seem to do. I'm happy to just ignore any "weird" constructions like that, but I thought I'd better do what the others did and throw an exception. I'll happily simplify this to ignore instead.

jena-arq/src/main/java/org/apache/jena/sparql/core/assembler/InMemDatasetAssembler.java
```diff
@@ -0,0 +1,67 @@
...
+public class InMemDatasetAssembler extends AssemblerBase {
+
+	@Override
+	public Dataset open(final Assembler assembler, final Resource root, final Mode mode) {
+		checkType(root, MemoryDataset);
+		final Dataset dataset = createTxnMem();
+		setContext(root, dataset.getContext());
+		// Default graph can be defined with ja:graph or ja:defaultGraph
+		final Resource defaultGraphDef = root.hasProperty(pDefaultGraph) ? getResourceValue(root,
+				pDefaultGraph) : root.hasProperty(pGraph) ? getResourceValue(root, pGraph) : null;
```
afs: This makes pGraph is just an alternative for pDefaultGraph. We don't need two names for the same thing. My suggestion was ja:graph was an inline description ofgraph name (optional) and data file.  See email for example.

(The code in more lines would be clearer for the next person to look at it!)

ajs6f: I'm happy to make them do whatever we think is going to be most useful but I'm not yet understanding why we would want them to be different here when in [DatasetAssembler](https://github.com/apache/jena/blob/master/jena-arq/src/main/java/org/apache/jena/sparql/core/assembler/DatasetAssembler.java#L58) they really do seem to be the same?

afs: They are different. `ja:defaultGraph` leads to a full model description. Assembling it builds the model in memory and then takes a copy.

`ja:graph` reads a file into a graph. Not general, not a copy.  It's needed too put a triples file into a named graph.

ajs6f: With all respect, that really doesn't seem to be what is happening in DatasetAssembler. A value from either predicate is taken and used at [this](https://github.com/apache/jena/blob/master/jena-arq/src/main/java/org/apache/jena/sparql/core/assembler/DatasetAssembler.java#L64) line to build from a model description. A value on `ja:graph` will be used there, not to load from a file. 

Is "load-from-a-file" the correct meaning for `ja:graph`? If so, what was the point of introducing `ja:data`?

ajs6f: I will add some `if-then` statements to unpack the code a bit.

afs: My suggested design was to enable files to be read directly into an in-memory assembler. That to me is the major use case for getting data into dataset.

Building a model and copying it in (ja:defaultGraph, ja:namedGraph) allows precomputation (e.g. rules) - I think that's a specialist use case (if it's confusing, lets leave that for the general dataset).

ajs6f: Okay, I thought it would be cool, but I certainly admit that it's gotten confusing already, and no users have even gotten involved yet! {grin}

I will strip out this jazz entirely and take it back to the plain case. To do that, I will use only one predicate, `ja:data`, with the meaning of "load tuples from this URI".

afs:
meta observation:  Github says "added some commits on 20 Jul".  This is not the whole story because 037fd1a, for example, is today (5 Nov).

ajs6f:
Weird. I don't know what that means, but it certainly doesn't seem right. Several of the intermediate commits in that last also came well after 19 July (which is what it says on my page!).

jena-arq/src/main/java/org/apache/jena/query/DatasetFactory.java
```diff
@@ -16,253 +16,226 @@
  * limitations under the License.
  */
 
-package org.apache.jena.query ;
```
afs: Huge amount of change for adding/changing a couple of operations.

jena-arq/src/main/java/org/apache/jena/query/DatasetFactory.java
```diff
@@ -16,253 +16,226 @@
...
+		return create(DatasetGraphFactory.createTxMem());
+	}
+
+	/**
+	 * @return a general-purpose, in-memory, modifiable Dataset
+	 */
+	public static Dataset createGeneral() {
+		return createTxnMem();
+	}
+
```
afs: This should go to the used to be `createMem()` i.e. `create(DatasetGraphFactory.createMem()) ;`.

jena-arq/src/main/java/org/apache/jena/query/DatasetFactory.java
```diff
@@ -16,253 +16,226 @@
...
+	public static Dataset createMemFixed() {
+		return create(DatasetGraphFactory.createMemFixed());
+	}
+
+	/**
+	 * @param model The model for the default graph
+	 * @return a dataset with the given model as the default graph
+	 */
+	public static Dataset create(final Model model) {
+		return new DatasetImpl(model);
```
afs: Why adding al the `final`?  The compiler, with "effectively final", does this anyway.

Adding `final` to arguments just increases the syntax line noise. A lot of this file has changed for nu functionality change.

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
```diff
@@ -0,0 +1,310 @@
...
+	public Iterator<Node> listGraphNodes() {
+		return access(() -> quadsIndex().listGraphNodes().iterator());
+	}
+
+	private Iterator<Quad> quadsFinder(final Node g, final Node s, final Node p, final Node o) {
+		if (isUnionGraph(g)) { // union graph is the merge of named graphs
+			final Set<Triple> seen = new HashSet<>();
+			return quadsIndex().find(ANY, s, p, o).filter(q -> !q.isDefaultGraph() && seen.add(q.asTriple()))
+					.iterator();
+		}
```
afs: Is the q.isDefaultGraph needed?  Aren't default graph triples held separately from the quads?

afs: Is there anyway to stream this? If the quads come out of the quadsIndex from SPOG then duplicates come out adjacent don't they?  Hence only need to do an adjacency test.

afs: I don't see any tests for the union graph.

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/QuadTable.java
```diff
@@ -0,0 +1,59 @@
...
+	Stream<Quad> find(Node g, Node s, Node p, Node o);
+
+	/**
+	 * Discover the graphs named in the table
+	 *
+	 * @return an {@link Stream} of graph names used in this table
+	 */
+	default Stream<Node> listGraphNodes() {
+		return find(ANY, ANY, ANY, ANY).map(Quad::getGraph).distinct();
+	}
```
afs: If the index traversed is SPOG, then duplicate graph names are adjacent which is much cheaper than distinct.

afs: Given the structure, isn't the top level of any index starting with a G enough to get all the graph names? The key set of the FourTupleMap for GSPO is enough, much faster, no retention of state needed for `distinct()`.

jena-arq/src/main/java/org/apache/jena/query/DatasetFactory.java
ajs6f: Ah, right. Will fix.

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
ajs6f: No, the `q.isDefaultGraph` is an oversight of mine, left over from when the default graph was in the same structures as named graphs. I'll fix that.

ajs6f: 1. I'm not sure what you mean by "stream this". It _is_ streaming. Unless you mean to avoid the buildup of state in the `Set`, in exchange for your suggestion about adjacency? In that case, see below.
2. This is the `DatasetGraph`. It doesn't know anything about SPOG or any other particular kind of index. It only knows about `QuadTable`s and `TripleTable`s, which may happen (do, in the default case) to be implemented with maps, but could just as well be implemented with some other technique that wouldn't produce adjacency as an artifact. (E.g. @Claudenw 's idea about Bloom filters, or your suggestion about "mixed structures".)

ajs6f: I'll get some tests in for the union graph case.

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/QuadTable.java
ajs6f: See my comment on `DatasetGraphInMemory` as to union graph functionality. It applies here too. This is an interface which has no reason or right to assume anything about adjacency.

ajs6f: As to starting with G, that's why this is a `default` method. Look at `HexTable` and you will see that very advantage being taken. But this is a `default` method and must not rely on facts about an implementation.

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
afs: `.seen` retains state so while it streams, it grows.  If there could be a way to process only a small windows of results (e.g current and previous), the there is no state accumulation.

jena-arq/src/main/java/org/apache/jena/query/DatasetFactory.java
```diff
@@ -16,253 +16,226 @@
  * limitations under the License.
  */
 
-package org.apache.jena.query ;
-
```
ajs6f: See your comment [here](https://mail-archives.apache.org/mod_mbox/jena-dev/201510.mbox/%3C5610F36D.8090602%40apache.org%3E) about "other clearing up of DatasetFactory ...", although I'm not sure why Git created the diff in this confusing way.

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
ajs6f: Right, see my comment above about assumptions that are available here. If we can't rely on adjacency (and we can't) then I _think_ this is the best we can do. But I would be very happy to be wrong.

Another alternative would be to require a special kind of returned order in `QuadTable::find`, but that seem s _way_ too restrictive to me.

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/QuadTable.java
afs: Agreed about default methods.  I am confused that HexTable and QuadTableForm both provide QuadTabl.  Good that the real code path exploits the efficiency. `GRAPH ?g {}` which even has a special operator in SPARQL algebra, is something that users want to be efficient.

ajs6f: Cool. I agree that the "forked impl" is a bit odd, but I did it for what I think are real gains in concision and clarity _within_ those types. The important point is that `QuadTableForm` is an `enum` and it's the six values _within_ `QuadTableForm` that actually impl `QuadTable`. Then `HexTable` binds up all those forms into a structure that acts as _one_ `QuadTable` by selecting the most efficient form(s) for any given operation. I'll add some comments to explain that relationship more fully and clearly.

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
afs: Union graph is important, in fact, it's one of the good things that this new dataset brings to users.

If necessary, we can add `DatasetGraphBaseFind.findInUnionGraph`. Actually that looks like a good idea but let's make that a completely separate issue for now.

`QuadTable` could have an operation so as to push it down to a place (is that HexTable?) where it can be done efficiently.

Maybe, an operation like: `QuadTable.findUnique(....)`. (Adding a flag to `find()` feels yukky to me - a bias against powerful methods with different characteristics based on additional arguments).

ajs6f: Okay, that makes a lot of sense to me. It's not dissimilar to the way I'm currently handling `listGraphNodes`, as we discuss elsewhere in this PR. I entirely agree about the preferrable signature. I'll get another commit in on this point.

ajs6f:
@afs , I've factored a `findInUnionGraph` through the code to provide access to implementation efficiencies. Let me know what you think!

jena-arq/src/main/java/org/apache/jena/sparql/core/DatasetGraphFactory.java
```diff
@@ -53,45 +54,38 @@ private static void copyOver(DatasetGraph dsgDest, DatasetGraph dsgSrc)
...
-    
+    public static GraphMaker graphMakerNull = () -> null ;
+
+    private static GraphMaker memGraphMaker = () -> GraphFactory.createDefaultGraph() ;
+
+    /**
+     * @return a DatasetGraph which features transactional in-memory operation
+     */
+    public static DatasetGraph createTxMem() { return new DatasetGraphInMemory(); }
+
```
afs: createTxMem => createTxnMem

ajs6f: Changed.

afs:
Looks good. Time to think about merging. This is a significant contribution so I'll send a message to dev@ and formally let people know it is happening with lazy consensus.

ajs6f:
Ah, that is exciting! I will begin learning the documentation CMS so as to offer a good set of pointers and explanations to help people try this out.

ajs6f:
@afs Shall I squash again in preparation?

afs:
That would be nice - the merged commits become the "public record" so to speak so the audience is changing.  The journalling separate for the mem dataset for example.

ajs6f:
Right, I'll do what I did before and squash to two more-or-less orthogonal commits, one for "Introducing a journaling dataset!" and one for "Introducing a MR+SW in-mem dataset!". I'll add more useful and full commit comments for each.

ajs6f:
@afs Squashed 'em pretty flat. The first commit is journaling operation, the second is the new in-memory design.

I will wait to hear from you, otherwise moving on the docs and trying to understand the ARQ test framework in `/testing`. I think I'm going to have to come to the list on that, because while I think I can correctly read much of the RDF that is setting up tests, I can't figure out for the life of me how it gets executed or how to hook in a section so that my new designs are used.

afs:
Great. 

You don't need `/testing`. The material under there is for testing query execution, not the dataset storage.

For functional tests, `AbstractDatasetGraphTests` and others in that package should be enough.  It's all programmatic, not script driven.

There are other sets around that might be worth at least checking they work, and if not, put more tests in the datasetgraph programmatic tests ("testing the tests"). e.g. We can try to feed it through the SPARQL test suite as a one-off but they aren't parameterized to implementation.  They test ARQ, not the dataset storage and if theer are failures, they should be captured in dataset tests.

I have some more tests (coverage of code paths in `DatasetGraphBaseFind`) which I'll put in but after this is merged so we have a stable base to get the merge done.

ajs6f:
Ah, cool. Well, I've thrown every abstract `DatasetGraph` test I could find in the source at this stuff (and you showed me some about which I didn't know) and added a bunch of my own to test out qualities specific to these impls, so I think we're doing pretty good there.

afs:
and `AbstractTestTransaction`. Think that's it.

ajs6f:
Yep, got that one via `TestDatasetGraphInMemory.TestDatasetGraphInMemoryTransactions`.

afs:
All this can be refined post merge. I've been checking that the existing test hierarchies run as well.

`TestDatasetGraphInMemory` and `TestInMemory` seem to both call `TestDatasetGraphInMemory`.

To make it clear what's a suite and what are the tests, ARQ uses a convention that "TS_" are suites and `Test*` are the `@Test`s themselves (and occasionally `TC_` are collections of suites, e.g. RIOT vs ARQ).  It's just a style thing and in a large set of tests, a bit of control over order (building blocks before combinations helps in navigating failures. Putting each nested class in `TestDatasetGraphInMemory` in a separate files and having TS_ makes navigation a bit easier in an IDE. [All "IMO".]

ajs6f:
I'm not quite sure what you mean by "`TestDatasetGraphInMemory` and `TestInMemory` seem to both call `TestDatasetGraphInMemory`" It doesn't seem to be calling itself when I read the code... `TestDatasetGraphInMemory` itself is just a container for the things inside it, which all together form the tests for the one class `DatasetGraphInMemory`. It runs using the suite runner to operate all the tests inside it, and it definitely gets called by `TestInMemory`, which itself gets linked up into `TS_Core`. 

I think I need to rename `TestInMemory` to `TS_InMemory` to come into line with the naming convention, right? It is responsible for testing all the machinery that together is self to implement the new in-memory dataset design, several different classes, not just `DatasetGraphInMemory` itself.

I can definitely break the various things in `TestDatasetGraphInMemory` out, but given that it is currently only about 250 lines, most of which is Java formatting and boilerplate, and that most of those nested guys are just trivial subtypes of abstract dataset test classes from elsewhere, I thought it might actually be _harder_ to read broken out.  In a different language, I might use multiple concrete inheritance to clarify/simplify this, letting `TestDatasetGraphInMemory` inherit from all these various abstract test classes directly. But I'm fine either way.

Shall I make another commit renaming `TestInMemory` to `TS_InMemory` and breaking apart `TestDatasetGraphInMemory` to a `TS_DatasetGraphInMemory` and multiple classes like `TestDatasetGraphInMemoryThreading`, `TestDatasetGraphInMemoryLock`, etc.?

ajs6f:
It occurs to me that a better way to explain myself is: `TestDatasetGraphInMemory` looks the way it does (lots of nested classes) because I wanted to keep the usual correspondence of one-class-to-one-test-class and I needed to use the test behavior from many abstract dataset test classes.

afs:
It is not the code - it's the way tests are found.

This is what seems to be happening - when running the JUnit tests on the package (at least in Eclipse and I believe in surefire if we didn't use "TS_"), `TestDatasetGraphInMemory`, `TestInMemory` and the inner class of `TestDatasetGraphInMemory` all get found and run.  But `TestInMemory` runs `TestDatasetGraphInMemory`, and` TestDatasetGraphInMemory` runs it's inner classes.  So the tests run x3, one for each level. To check, I put a break point on and it got hit three times.

(Also - running the same test more than on confuses Eclipse in minor ways - some don't get marked as "run".)

No need to do anything now - lets get the code in.

ajs6f:
Ah, I didn't realize that we were using the automatic "execute-by-naming-convention" test running as well as the explicit style of gathering things up into suites.
I'll keep my grubby hands off now and stay out of the way. {grin}

PR #96, opened 2015-10-29
JENA-1058: ASCIIFoldingLowerCaseKeywordAnalyzer for jena-text

osma:
See https://issues.apache.org/jira/browse/JENA-1058

### DIFF
--- a/apache-jena/bin/tdbloader2index
+++ b/apache-jena/bin/tdbloader2index
@@ -266,23 +266,6 @@ debug "Jena Classpath is $JENA_CP"
 # All files are written S P O / G S P O columns per row but in different sort orders.
 info "Index Building Phase"
 
-# Check whether Pipe Viewer is available
-# Needs to temporarily disable exit on error as which produces an error
-# if the given command is not found
-#set +e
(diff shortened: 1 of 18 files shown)


osma:
The class names are getting ridiculously long, but I just followed the same pattern as before.

afs:
+1 : Looks good

By the way - is there documentation (i.e. a list) of all the analyzers we have?

osma:
@afs Yes there is, in the jena-text documentation, though calling it a list is a bit of a stretch since there are only four currently (StandardAnalyzer,  SimpleAnalyzer, KeywordAnalyzer and LowerCaseKeywordAnalyzer) and only LowerCaseKeywordAnalyzer is specific to jena-text, the other three come directly from Lucene.

See https://jena.apache.org/documentation/query/text-query.html#configuring-an-analyzer 

About this PR: I think I will try to implement a more generic, configurable analyzer, as suggested by Claude Warren in a comment on JIRA-1058. If it's not too difficult, it would obviously be better than this very specific analyzer.

osma:
I drop this in favor of JENA-1062 / PR #97

PR #97, opened 2015-11-04
JENA-1062: configurable Lucene analyzer for jena-text

osma:
This is a configurable Analyzer implementation for jena-text / Lucene. It is similar to what can be achieved in [Solr configuration](https://wiki.apache.org/solr/AnalyzersTokenizersTokenFilters). The current implementation only supports a few basic tokenizers and filters included in Lucene. More can be added later if necessary, though some tokenizers and filters require extra configuration parameters and currently there is no mechanism for specifying these.

Tokenizers:
* StandardTokenizer
* KeywordTokenizer
* WhitespaceTokenizer
* LetterTokenizer

Filters:
* StandardFilter
* LowerCaseFilter
* ASCIIFoldingFilter

Configuration can be done in the assembler like this:
```
text:analyzer [
  a text:ConfigurableAnalyzer ;
  text:tokenizer text:KeywordTokenizer ;
  text:filters (text:ASCIIFoldingFilter, text:LowerCaseFilter)
]
```

When used directly from Java code, the ConfigurableAnalyzer accepts one String parameter specifying the Tokenizer name and another parameter which is a List<String> specifying (optional) filters in the order they should be applied.

### DIFF
--- a/jena-text/src/main/java/org/apache/jena/query/text/analyzer/ConfigurableAnalyzer.java
+++ b/jena-text/src/main/java/org/apache/jena/query/text/analyzer/ConfigurableAnalyzer.java
@@ -1,0 +1,93 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.jena.query.text.analyzer ;
+
+import java.io.Reader ;
+import java.util.List ;
+
+import org.apache.jena.query.text.TextIndexException;
+import org.apache.lucene.analysis.Analyzer ;
+import org.apache.lucene.analysis.TokenFilter ;
+import org.apache.lucene.analysis.Tokenizer ;
+import org.apache.lucene.analysis.TokenStream ;
+import org.apache.lucene.analysis.core.KeywordTokenizer ;
+import org.apache.lucene.analysis.core.LetterTokenizer;
+import org.apache.lucene.analysis.core.LowerCaseFilter ;
+import org.apache.lucene.analysis.core.WhitespaceTokenizer ;
+import org.apache.lucene.analysis.miscellaneous.ASCIIFoldingFilter;
+import org.apache.lucene.analysis.standard.StandardFilter;
+import org.apache.lucene.analysis.standard.StandardTokenizer;
+import org.apache.lucene.util.Version ;
+
+
+/** 
+ * Lucene Analyzer implementation that can be configured with different
+ * Tokenizer and (optionally) TokenFilter implementations.
+ */
+
+public class ConfigurableAnalyzer extends Analyzer {
+        private final Version version;
+        private final String tokenizer;
+        private final List<String> filters;
+        
+        private Tokenizer getTokenizer(String tokenizerName, Reader reader) {
+                switch(tokenizerName) {
+                        case "KeywordTokenizer":
+                                return new KeywordTokenizer(reader);
+                        case "LetterTokenizer":
+                                return new LetterTokenizer(version, reader);
+                        case "StandardTokenizer":
+                                return new StandardTokenizer(version, reader);
+                        case "WhitespaceTokenizer":
+                                return new WhitespaceTokenizer(version, reader);
+                        default:
+                                throw new TextIndexException("Unknown tokenizer : " + tokenizerName);
+                }
+        }
+        
+        private TokenFilter getTokenFilter(String filterName, TokenStream source) {
+                switch(filterName) {
+                        case "ASCIIFoldingFilter":
+                                return new ASCIIFoldingFilter(source);
+                        case "LowerCaseFilter":
+                                return new LowerCaseFilter(version, source);
+                        case "StandardFilter":
+                                return new StandardFilter(version, source);
+                        default:
+                                throw new TextIndexException("Unknown filter : " + filterName);
+                }
+        }
+        
+        public ConfigurableAnalyzer(Version ver, String tokenizer, List<String> filters) {
+                this.version = ver;
+                this.tokenizer = tokenizer;
+                this.filters = filters;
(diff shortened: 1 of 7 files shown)


rvesse:
Looks good to me

One open question, how does this interact with past work for language specific indexing and multi-lingual indexing in general?

It's been a while since I poked around Lucene but I seem to remember that it was often necessary to use alternative analysers particularly when you get into languages with compound words, non-Latin alphabets, symbolic alphabets etc.  This shouldn't be a requirement for merging this work but just wanted to check that the current design won't preclude support for this in the future?

Looking over the code it looks like it should be relatively easy to add new analysers and filters as needed but just wanted to make sure I had understood the code correctly

osma:
Good questions @rvesse !

Right now (before this PR) one can either use a few generic, non-language-specific Analyzers: StandardAnalyzer, SimpleAnalyzer, KeywordAnalyzer and LowerCaseKeywordAnalyzer. 

Then there is MultilingualAnalyzer, which looks at the language tag of literals and picks a language-specific Analyzer based on the language tag (falling back on StandardAnalyzer in case there's no suitable Analyzer implementation for the language). The list of language-specific Analyzers is hardwired in the implementation though.

What this adds is a non-language-specific Analyzer that can be configured in a little bit more detail: it is possible to select a Tokenizer and zero or more TokenFilters. However, it does not look at the language tags at all and it is also limited to a few recognized Tokenizers and Filters, none of which require any special parameters.

Things that were possible before:
* use StandardAnalyzer/SimpleAnalyzer/KeywordAnalyzer for everything
* use EnglishAnalyzer for "en" and FrenchAnalyzer for "fr" literals (MultilingualAnalyzer does this)

Things that become possible with this PR:
* use KeywordTokenizer (i.e. don't split into tokens), but drop accents with ASCIIFoldingFilter and make everything lowercase with LowerCaseFilter (my original use case for JENA-1058)
* use WhitespaceTokenizer without filters (perhaps good for handling e.g. a whitespace-separated list of product codes or URIs)
* dozens of other combinations of the non-language-specific Tokenizers and TokenFilters, though probably only some combinations make any sense

Things that are still not possible:
* use EnglishAnalyzer for "en" language but StandardAnalyzer for everything else (in MultilingualAnalyzer the analyzers are hardwired)
* use language-specific analyzers when available but fall back on SimpleAnalyzer (ditto, the fallback to StandardAnalyzers in hardwired in MultilingualAnalyzer)
* use StandardAnalyzer with LengthFilter to remove excessively short or long words (LengthFilter requires `min` and `max` parameters and there is no way to pass those parameters to ConfigurableAnalyzer, so it doesn't support LenghtFilter)

In short, the universe of Analyzers (Tokenizer + TokenFilter combinations, with or without special treatment for language tags) is potentially huge and this PR tackles only one rather small part of it, but it expands the options in a way that I think is useful.

jena-text/src/main/java/org/apache/jena/query/text/analyzer/ConfigurableAnalyzer.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.lucene.util.Version ;
+
+
+/** 
+ * Lucene Analyzer implementation that can be configured with different
+ * Tokenizer and (optionally) TokenFilter implementations.
+ */
+
+public class ConfigurableAnalyzer extends Analyzer {
+        private Version version;
```
ajs6f: Couldn't these files be `final`?

osma:
@ajs6f You're right, I did that now.

jena-text/src/main/java/org/apache/jena/query/text/assembler/ConfigurableAnalyzerAssembler.java
```diff
@@ -0,0 +1,101 @@
...
+           ]
+         ]
+        .
+    */
+
+
+    @Override
+    public Analyzer open(Assembler a, Resource root, Mode mode) {
+        if (root.hasProperty(TextVocab.pTokenizer)) {
+            Resource tokenizerResource = (Resource) root.getProperty(TextVocab.pTokenizer).getObject();
```
ajs6f: Would `Resource::getPropertyResourceValue` be more readable here? (No cast.)

jena-text/src/main/java/org/apache/jena/query/text/assembler/ConfigurableAnalyzerAssembler.java
```diff
@@ -0,0 +1,101 @@
...
+            Statement stmt = current.getProperty(RDF.first);
+            if (stmt == null) {
+                throw new TextIndexException("filter list not well formed");
+            }
+            RDFNode node = stmt.getObject();
+            if (! node.isResource()) {
+                throw new TextIndexException("filter is not a resource : " + node);
+            }
+            
+            Resource res = (Resource)node;
```
ajs6f: Isn't `Node::asResource` a little better here and below?

jena-text/src/test/java/org/apache/jena/query/text/TestDatasetWithConfigurableAnalyzer.java
```diff
@@ -0,0 +1,63 @@
...
+                "."
+                );
+        String queryString = StrUtils.strjoinNL(
+                QUERY_PROLOG,
+                "SELECT ?s",
+                "WHERE {",
+                "    ?s text:query ( rdfs:label '\"feeling ä déja\"*' 10 ) .",
+                "}"
+                );
+        Set<String> expectedURIs = new HashSet<>() ;
```
ajs6f: Our shaded Guava offers various `Sets::newHashSet` functions for this kind of use, which read a pinch better.

osma:
Thanks @ajs6f for your detailed notes! I fixed the things you pointed out but I think especially the Guava Sets.newHashSet pattern could be applied in many other places in the jena-text code as well. My test case code was just adapted from similar other jena-text tests and they all use the same pattern. Fixing those as well isn't really in the scope for this PR but could be done separately.

ajs6f:
@osma I am a nitpicker of the worst kind! {grin}

If you would like me to make a PR against `jena-text` looking for this kind of thing (using Guava or Java 8 idioms to shorten things up) I'm happy to do that sometime in the next week or three. I'm a born "code janitor".

afs:
The test should be "is it clear?". `new HashSet<>()` is common usage.

ajs6f:
Sure, there's nothing wrong with what's there. All I meant is that `newHashSet` seems as clear as `new HashSet<>()` and allows elements, so is shorter.

osma:
@ajs6f :
>If you would like me to make a PR against jena-text looking for this kind of thing (using Guava or Java 8 idioms to shorten things up) I'm happy to do that sometime in the next week or three. I'm a born "code janitor".

Sure, that would be great!

About this PR: good for merging? I can do the merging myself, but just wanted to check first with others.

afs:
Looks good.  With a documentation modifications to follow? The text index seems to becoming quite popular.

osma:
Yes, of course I will also update the jena-text documentation.

osma:
Rebased on current master and squashed my commits into one, preparing to merge to Apache git

PR #100, opened 2015-11-18
Rename in-memory dataset test classes.

afs:
To help navigate tests, this is a reorganisation of the tests created for JENA-624 into the same naming style and organization as the other tests in ARQ.  The main reason for reorganizing is so that there is one style, not that one is better than another.

* Files with tests are `Test*` (Junit3 legacy naming!)
* Test frameworks using abstract classes are `AbstractTest*`
* Files `TS_` for test suites, usually `@RunWith(Suite.class)`
* Each `TS_*` is in a test collection file - `TC_General` in this case.


ajs6f:
I'm totally cool with this renaming, of course. I did want to ask about tests for the Assembler system-- I didn't see any, when I came to write the ones for my little addition to that system. Did I miss them or are they not there? Also, the same goes for the `LockMRPlusSW` I wrote-- I didn't see any tests for the other locks impls. Should I file a ticket to remind us to write those with all the copious spare time we have :) , or did I just miss them?

afs:
IIRC `LockMRSW` gets tested with `TestConcurrency` and when used with `TestDatasetGraphWithLock`.

ajs6f:
And the various `Assembler` subtypes?

afs:
Most Assemblers are tested in jena-core.  Probably weak-to-nonexistent for dataset assemblers directly - Fuseki uses them of course.

I have now tested all combinations of quad patterns (see last nights commit). Getting the tests organised and made useable is proving messy/tricky and taking longer than I hoped. I'll commit what I have in and then they can be improved on.

Sort of want open parametrization on suites of tests all at the same time - usual wanting two dimensions when each framework (in this case "Runner") is one at a time. Or multiple inheritance so maybe interface mixins will work - the data is static - then one method can satisfy the get empty dataset" operation from several mixins.  Dynamic generation requires writing a Runner in Junit4 (?? - it was easier in Junit3).

PR #101, opened 2015-11-20
A few suggestions for @osma

ajs6f:
These are a few small suggestions for @osma for using Guava or Java 8 features to slightly tighen code in `jena-text`. They are far from exhaustive, just meant to be examples for features like `Optional` or the Streams API.


osma:
Thanks @ajs6f - there's quite a lot of fixes in here!

In general, I feel positive about cleanups like this. And it's nice to have a patch which actually *reduces* the amount of code quite significantly!

I'm a bit worried about the readability aspect here though. There are some techniques in here that are unfamiliar to me - having learnt Java more than a decade ago, around the time of JDK 1.4, and mostly avoiding Java in favor of other programming languages ever since (though I have picked up some techniques on the way at least up to Java 6). Using modern, clean and efficient techniques is good, but how modern can you be and still have people understand the code?

For example, the way you replaced the TextHitConverter class with a [forEach expression](https://github.com/apache/jena/pull/101/files#diff-48122cd8c64b1e120b37581f50fc55ceR252) - I didn't even know such a syntax existed! I do like getting rid of the converter class, but is this a style we want to encourage? It looks a bit like typical jQuery code to me, with all the (nested) callbacks...

What do others feel? @afs?

Also in terms of code review, it would be easier to take this one piece (commit) at a time, with descriptive commit messages, instead of everything in one lump.

ajs6f:
Maybe there's confusion about what this PR is for-- I wasn't intending that it would be merged, but that you would use it as a source of examples, like I wrote. The intention was to show off the various techniques (lambdas, etc.) in context. You would want to make a PR using whatever you like here. For example, I show a few uses of `Optional` as the return type of methods that return a value-or-`null`, but there are many such places.

osma:
I see, thanks for the clarification. But I think my question about what kind of code we want is still valid - after all, this isn't my project, but maintained collectively. I can certainly apply some of this in the places where I think it's useful, but I'd like to know what others feel about these techniques before I do that.

ajs6f:
Sure, I was just responding to the question about commits. I agree that code must be written in a style that supports maintenance over the long term. I'd make two points about that: code that is written in archaic style is just as hard to maintain as code that is written in ultra-modern style, and it would be really wonderful if Jena could create some formatting standards (e.g. for indentation, etc.) Common formatting can help a lot with readability. I tried to begin that conversation some months ago, but I didn't seem to get much enthusiasm.

ajs6f:
Just to be clear, I don't mean to argue that Jena's code is particularly archaic, just that we should aim for balance between taking advantage of new capabilities and avoiding the bleeding edge.

afs:
@osma makes a valid point - Java8 is still quite new and many (most?) people are not going to see some of the new idioms as natural or easier, and they can be slower.

PRs that just hang about clutter seeing the state of the project.  It would be nice for [jena/pulls](https://github.com/apache/jena/pulls) not to have PRs that aren;t to go into the codebase. As the header says "ajs6f wants to merge 1 commit" though I don't know the best way to have discussion on code in GH. 

General policy discussions ought to happen on dev@ and factor in that change to something that works is creating work, people don't always have much time, and that there is not "one style" that everyone will prefer.

And spaces, not tabs. That's has been the style for a long time.

ajs6f:
Sure, @afs . It might be good to get some of this written down in the docs. I'll happily close this PR and leave the one directly to @osma's clone in place for him to do just as he likes with.

ajs6f:
As far as the formatting questions, whitespace maybe isn't the best example. For a moe "pungent" case, there are many, many different ways to combine braces, newlines and whitespace through the codebase (maybe an effect of its long history)? Some code uses braces for every control statement, some only when required, some mixed. That sort of thing makes it difficult to read code. I will try to restart this conversation on @dev and try to spur some interest in normalizing on some of these things. There would have to be some change (non-functional formatting commits) but at low cost and I think it worth it.

PR #102, opened 2015-11-24
Fixing namespaceTTL (was recursively calling itself)

escowles:


rvesse:
Thanks for the fix, applied to trunk

I notice the method definition looks a bit odd in general - @afs should `namespaceTTL()` have a similar definition to `localnameTTL()` where it first calls `namespace()` and then applies escaping?

escowles:
I think that the escaping is only needed for the localname -- the namespace URI will usually contain characters (`.-/#`) that would be improperly escaped by escape_PN_LOCAL_ESC().

afs:
@escowles is right - escaping is only allowed in the local part.  Also, the return from `namespaceTTL` is a IRI, to go in `<...>`, not the prefix name that is written in output as a bare word.

PR #103, opened 2015-11-30
JENA-624: Correction to transaction begin for DatasetGraphInMemory

ajs6f:
Correction to transaction begin for DatasetGraphInMemory to include default graph as well as quad table.


afs:
Are there some tests for this?

I noticed that `DatasetGraphInMemory.abort` does not call anything. Note the contract for transactions is that `end()` is optional for write transactions (see javadoc).

afs:
This is important to get in and i don't want it to miss the release so applied as-is.  Also applied a test in `end()` for being called multiple times.

PR #105, opened 2015-12-08
JENA-1078: Logging warning when transaction is ended without commiting

ajs6f:


jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
```diff
@@ -161,17 +161,17 @@ public void abort() {
 	@Override
 	public void close() {
 		if (isInTransaction()) abort();
-
 	}
 
 	@Override
 	public void end() {
         if (isInTransaction()) {
+            log.warn("Ending transaction without commit!");
```
afs: begin(READ)-end() is a normal pattern.  The code needs to know if it is considering a READ or WRITE transaction.

ajs6f: Fixed with latest commit.

afs:
This causes about 12 warnings from the test suite (Eclipse and maven).

Please do add tests and check the test suite runs.  I will check them and fix or silence each one but it would be nice to be at least notified that it would happen.

ajs6f:
I had no idea that throwing WARNINGs in the test suite was any kind of issue. ERRORs, yes, but not WARNINGs. That's a policy that might be recorded somewhere, perhaps [here](https://jena.apache.org/getting_involved/reviewing_contributions.html)?

ajs6f:
In any event, I can review them and eliminate them. No need for you to spend time doing that-- this is my code and I take responsibility for it. Would you reopen the ticket for me?

afs:
We have had the discussion about warnings again on the dev@ just recently.

I've already started on a review and some fixes.  There are some other problems with abort/end. I will submit a PR here for detailed code review. The issues look harmless but that's because of the way the implementation of indexes happen to work.  Different plugged in indexes would not be so safe.

PR #106, opened 2015-12-10
JENA-1083: Refactoring in TupleTable and subtypes for clarity and concision

ajs6f:


ajs6f:
There are a lot of whitespace changes in this PR, for which I apologize. They are present because I have switched my IDE to 4-space indents to match the more commons usage in the Jena codebase. If this is too confusing and it would be preferably to go back to the form that I had been using, please say so and I will be happy to so do.

afs:
The whitespace changes make it tricky to follow the real changes so I have converted all the file in the package in master from tabs to spaces to give a common reference point.

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/QuadTableForm.java
```diff
@@ -124,31 +96,21 @@ public PMapQuadTable get() {
 			return new PMapQuadTable(name()) {
 
 				@Override
-				protected Quad quad(final Node s, final Node p, final Node o, final Node g) {
-					return Quad.create(g, s, p, o);
-				}
+                public Quad tuple(final Quad q) {
+                    return create(q.getSubject(), q.getPredicate(), q.getObject(), q.getGraph());
```
afs: It'll be clear when the white space is standardised but am I right in seeing:

* Using Quad and putting strange things in the slots which is very confusing. `getGraph` returns the `getSubject`.
* There seems to be some creating objects just to take them apart again. The strength of the design is that quads/triples are not stored, instead nodes are placed in indexes. To reflect that, operations based on g/s/p/o in some order are clearer to me at least as well as reducing object churn.

ajs6f: These criticisms are sadly accurate. I seem to have improved in one way by creating other defects. I'm going to close this PR and see if I can solve the problems I identified (repeated code, unclear responsibility) in a better, tighter way, perhaps involving `ColumnMap`.

ajs6f: To be clear, I will send that next PR in the next few days.

afs: We have time. The release build for the vote has already been done.

PR #107, opened 2015-12-11
JENA-1085 : Common pattern for passing down transaction lifecycle operations.

afs:
* Abort called `end` directly causing the warning to be emitted
* Abort did not call quadIndex.abort/defaultGraph.abort
* Pair up quadIndex.end/defaultGraph.end for read transactions
* Add locks around all pairs of quadIndex, defaultGraph operation calls

Discussion points:

* `writeLock` is used even for read transactions (it's a no-op but it's called). Should we rename it as e.g. `readerWriterLock`?
* `TupleTable.abort` defaults to end.  Given the importance of the transaction cycle, might it be better to require each implementation to do that. It puts all the begin/commit/abort/end code for one component in one place for the next person to look at the code.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
@@ -26,7 +26,7 @@ import static org.apache.jena.sparql.core.Quad.isUnionGraph;
 import static org.slf4j.LoggerFactory.getLogger;
 
 import java.util.Iterator;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.concurrent.locks.ReentrantLock ;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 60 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
```diff
@@ -166,16 +179,31 @@ public void close() {
...
     public void end() {
         if (isInTransaction()) {
-            if (transactionType().equals(WRITE))
-                log.warn("end() called for WRITE transaction without commit or abort having been called");
+            if (transactionType().equals(WRITE)) {
+                log.warn("end() called for WRITE transaction without commit or abort having been called causing a forced abort");
+                _abort() ;
+            }
+            finishTransaction();
+            return ;
```
ajs6f: Why an "naked" `return` here?

ajs6f:
Yes, I think that renaming `writeLock` makes sense. The alternative would be to actually check the type of transaction and avoid fiddling with that lock if it is a READ. I think either is okay and either is clearer than what is there now.

ajs6f:
I'm fine with pulling out the default impl of `TupleTable::abort`. I myself don't find it any less clear, but if you do, that's evidence enough.

afs:
Revised pull request (entangled with changes at Apache).

The revision

* Puts the transaction finalization code inside the commit/abort code that takes the system lock. This mean the lock is taken once at the finishing of a transaction.
* Reorganises the code to a rigid pattern for operations.
* Renames `commitLock` as `systemLock`
* Uses `ReentrantLock`, not `ReentrantReadWriteLock` (only the write lock was used; the implement of a `ReentrantReadWriteLock.WriteLock` and `ReentrantLock` is much the same for our usage, just calls to the `sync` object).

As well as code review, design review would be appreciated.  Compared to the original design, this makes `abort` take a lock and call `abort` on indexes whereas previously it simply did nothing, knowing the index implementations would cope.  Because `abort` is a not a performance critical operation (usually!), the symmetric and explicit calls to index operations makes things clearer in my view.

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
afs: Will remove.

afs:
This PR has become mixed in with Jena master evolution.  I'll start again.

PR #108, opened 2015-12-12
JENA-1085: Common pattern for transaction lifecycle. Always call down to index operations.

afs:
(Cleaned version of PR #107)

This revision

* Puts the transaction finalization code inside the commit/abort code that takes the system lock. This mean the lock is taken once at the finishing of a transaction.
* Reorganises the code to a rigid pattern for operations.
* Renames commitLock as systemLock
* Uses ReentrantLock, not ReentrantReadWriteLock (only the write lock was used; the implement of a ReentrantReadWriteLock.WriteLock and ReentrantLock is much the same for our usage, just calls to the sync object).

As well as code review, design review would be appreciated. 

Compared to the original design, this makes abort take a lock and call abort on indexes whereas previously it simply did nothing, knowing the index implementations would cope. Because abort is a not a performance critical operation (usually!), the symmetric and explicit calls to index operations makes things clearer in my view.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
@@ -26,7 +26,7 @@ import static org.apache.jena.sparql.core.Quad.isUnionGraph;
 import static org.slf4j.LoggerFactory.getLogger;
 
 import java.util.Iterator;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.concurrent.locks.ReentrantLock ;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
@@ -37,11 +37,7 @@ import org.apache.jena.query.ReadWrite;
 import org.apache.jena.shared.Lock;
 import org.apache.jena.shared.LockMRPlusSW;
 import org.apache.jena.sparql.JenaTransactionException;
-import org.apache.jena.sparql.core.DatasetGraph;
-import org.apache.jena.sparql.core.DatasetGraphTriplesQuads;
-import org.apache.jena.sparql.core.DatasetPrefixStorage;
-import org.apache.jena.sparql.core.Quad;
-import org.apache.jena.sparql.core.Transactional;
+import org.apache.jena.sparql.core.* ;
 import org.slf4j.Logger;
 
 /**
@@ -55,21 +51,15 @@ public class DatasetGraphInMemory extends DatasetGraphTriplesQuads implements Tr
 
     private final DatasetPrefixStorage prefixes = new DatasetPrefixStorageInMemory();
 
+    /** This lock imposes the multiple-reader and single-writer policy of transactions */
     private final Lock writeLock = new LockMRPlusSW();
 
-    private Lock writeLock() {
-        return writeLock;
-    }
-
-    private final ReentrantReadWriteLock commitLock = new ReentrantReadWriteLock(true);
-
     /**
-     * Commits must be atomic, and because a thread that is committing alters the various indexes one after another, we
-     * lock out {@link #begin(ReadWrite)} while {@link #commit()} is executing.
+     * Transaction lifecycle operations must be atomi, especially begin and commit where
+     * the global state of the indexes is read or written one after another.
+     *  and because a thread that is committing alters the various indexes.
      */
-    private ReentrantReadWriteLock commitLock() {
-        return commitLock;
-    }
+    private final ReentrantLock systemLock = new ReentrantLock(true);
 
     private final ThreadLocal<Boolean> isInTransaction = withInitial(() -> false);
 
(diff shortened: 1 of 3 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
```diff
@@ -55,21 +51,15 @@
...
-    private Lock writeLock() {
-        return writeLock;
-    }
-
-    private final ReentrantReadWriteLock commitLock = new ReentrantReadWriteLock(true);
-
     /**
-     * Commits must be atomic, and because a thread that is committing alters the various indexes one after another, we
-     * lock out {@link #begin(ReadWrite)} while {@link #commit()} is executing.
+     * Transaction lifecycle operations must be atomi, especially begin and commit where
```
ajs6f: This comment is a bit confusing. How about:
```
We have a lock for committing because transaction lifecycle operations must be atomic, especially {@link Transactional#begin} and {@link Transactional#commit}, for which the global state of the (possibly multiple) indexes may be read or written one after another.
```

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
```diff
@@ -125,57 +115,104 @@ public DatasetGraphInMemory(final QuadTable i, final TripleTable t) {
...
-        commitLock().readLock().lock(); // if a commit is proceeding, wait so that we see a coherent index state
-        try {
+        if (isInTransaction()) 
+            throw new JenaTransactionException("Transactions cannot be nested!");
+        startTransaction(readWrite) ;
+        _begin(readWrite) ;
+    }
+
+    private void _begin(ReadWrite readWrite) {
+        withLock(systemLock, () ->{
```
ajs6f: Nice idiom!

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
```diff
@@ -125,57 +115,104 @@ public DatasetGraphInMemory(final QuadTable i, final TripleTable t) {
...
+            abort();
     }
 
     @Override
     public void end() {
         if (isInTransaction()) {
-            if (transactionType().equals(WRITE))
-                log.warn("end() called for WRITE transaction without commit or abort having been called");
+            if (transactionType().equals(WRITE)) {
+                log.warn("end() called for WRITE transaction without commit or abort having been called causing a forced abort");
```
ajs6f: This message is a tiny bit ambiguous if you don't already understand that `end` calls `_abort`. Maybe just add a sentence break: `"end() called for WRITE transaction without commit or abort having been called. This causes a forced abort!"`

ajs6f:
I agree about`ReentrantLock`. Lighter for the same value.

ajs6f:
I'm not sure why you renamed `commitLock`. I would think that `systemLock` would be a better name for the lock supporting `getLock`, which is not the same and is in fact a separate MRSW lock.

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
afs: OK, I'll simplify but the lock isn't just for commit. It is taken once at the start and once at the finish of each transaction. The code used to have an assumption about how quad/triple index.end() worked. Plug in other impls and it may not be true.  The lock now makes any index action over the indexes thread-safe.  The cost is the lock being taken on abort and in read-end. For abort, we can assume does not matter too much (frequent aborts are a sign of something in the application being wrong.  Taking the lock in read-end is safe; we'll be able to assess whether it is a significant cost by comparing with the experiment on a non-transaction implementation.

afs:
`systemLock` because it is used (internally) by the system, not the application.

rvesse:
LGTM

ajs6f:
But the `writeLock` isn't used directly by applications either. Why aren't they both "system locks"?

ajs6f:
If it's a problem with the fact that the former `commitLock` is now being used for other transaction moments as well, I get that. But `systemLock` is pretty generic and I would assume that it was meant to support `getLock` (`system` being "the rest of the system outside this class" and not "the system and not the application"). How about `lifecycleLock` or `transactionLock`?

afs:
I hope the comment explains usage especially as the variable is private.

Other names are equally problematic. This is still technically a "lock free implementation" because "lock", means a lock on data items.  Usually this is made clear by the term alternative name "latch".

ajs6f:
I guess I thought of this "lock" as a semaphore. I've not heard the term "latch" used in this context. In any event, it is a `private` variable, as you say. I'm a little bit allergic to calling things "systemX" because "system" is such a multivalent word.

afs:
`latch` is a very database name and also it is separately used for locks that have two states e.g. java's `CountDownLatch`.

Shall we call the `writeLock` the `transactionLock`? It makes more sense to me as it's held for the lifetime of the whole transaction.  The name reflects the implementation as MR+SW. If this is a framework, maybe that would be a different policy.

ajs6f:
That might be a good idea. If I understand correctly where you're taking this, we have a lock that must be acquired to change transactional state (for example, to commit) but can be released as soon as that is accomplished (`systemLock`) and a lock that must be held as long as a transaction is in progress (`writeLock`). `transactionLock` does sound like a better name for the latter, especially since it's held by readers, too.

afs:
Ok - I will rename `writeLock` as `transactionLock`.

PR #109, opened 2015-12-14
Introduce interface TransactionalComponent.

afs:
Mantis has a more general concept of transaction.  As part of that, it differentiates between `Transactional` (the view from users of a transaction system) and `TransactionalComponent` (units of code that participate in providing transactionality).

See [the code for TransactionalComponent](https://github.com/afs/mantis/blob/master/dboe-transaction/src/main/java/org/seaborne/dboe/transaction/txn/TransactionalComponent.java) -- for example, that has a explicit `Transaction` object and only the upper layers manage the `Transaction` by thread.

For the in-memory dataset, the difference is currently small but introducing this now prepares for more complex setups, without disturbing `Transactional`.  It also removes the need for components to know whether they are in a transaction or not; that is a concept of the coordinator function currently in `DatasetGraphInMemory`.  During times of state change, some components may be in different states to others, making the idea of a component being "in a transaction" imprecise. It also removes 11 ThreadLocals that were not used except in testing. The testing is adapted.

To use individual components on their own (if ever needed) then a simple wrapper that provides `Transactional` over a single component is a better design as it separates the roles of collection and component. Included is `TransactionalOfOne` but that is just illustrative and not for inclusion.

### DIFF
--- a/jena-arq/src/test/java/org/apache/jena/sparql/core/mem/AbstractTestTupleTable.java
+++ b/jena-arq/src/test/java/org/apache/jena/sparql/core/mem/AbstractTestTupleTable.java
@@ -22,15 +22,16 @@ import static java.util.stream.Collectors.toSet;
 import static org.apache.jena.ext.com.google.common.collect.ImmutableSet.of;
 import static org.apache.jena.query.ReadWrite.READ;
 import static org.apache.jena.query.ReadWrite.WRITE;
+import static org.junit.Assert.assertEquals ;
+import static org.junit.Assert.assertTrue ;
 
 import java.util.Set;
 import java.util.stream.Stream;
 
 import org.apache.jena.ext.com.google.common.collect.ImmutableSet;
-import org.junit.Assert;
 import org.junit.Test;
 
-public abstract class AbstractTestTupleTable<TupleType, TupleTableType extends TupleTable<TupleType>> extends Assert {
+public abstract class AbstractTestTupleTable<TupleType, TupleTableType extends TupleTable<TupleType>> {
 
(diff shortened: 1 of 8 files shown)


jena-arq/src/test/java/org/apache/jena/sparql/core/mem/AbstractTestTupleTable.java
```diff
@@ -42,71 +43,68 @@
 
 	protected static final Set<TupleSlot> allWildcardQuery = of();
 
+	protected long transactionalCount() {
+	    table().begin(READ);
+	    long x = rawCount() ;
```
ajs6f: Could this not be
```
table().begin(READ);
try { return rawCount(); }
finally { table().end(); }
```
That seems more readable to me, but maybe that's just my taste.

ajs6f:
I think this is basically awesome. Separating different views of the things-that-are-transacting is good, and it gets us a little further down the road to having a proper `Transaction` object, which would be great.

afs:
I would not be confident yet that all the machinery in Mantis is right - it has not been used enough to know and may be too complicated.  One test is to use it with transactions involving Lucene (weak transactions) and TDB2. But components was one thing I had trouble with until introducing the difference between a component and the overall transaction.

ajs6f:
I suppose it's early days in that sense. But if (as seems at least possible) Jena acquires more and more kinds of indexes (text, geospatial, etc.) and more and more "pluggability" (SPIs) in the index layer, this could really be an important point. Also, won't multi-thread transactions pretty much demand something like a proper `Transaction` object?

PR #110, opened 2015-12-14
JENA-1083: Using ColumnMap in transactional in-memory dataset implementation

ajs6f:
Okay, @afs : Second try. This time I avoid unneeded object creation or packing and unpacking, and I don't do weird things to the semantics of `Quad` and `Triple` (I hope!).

I did add some convenience methods to `ColumnMap` and also let the two tuple types subtype `Tuple<Node>` in order to work more closely with `ColumnMap`. I don't think this should be any problem, since they truly are tuples, but I'm sure you will let me know if that is not the case :).

The gains here are much more clarity for the forms of `PMapTupleTable`-based in-memory indexes and it should also be possible much more easily to create other index orderings, if users find that useful for special purposes. (I.e. saving space under conditions where the possible queries are known and very restricted.)

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/core/Quad.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/core/Quad.java
@@ -181,17 +181,7 @@ public class Quad extends Tuple<Node>
     @Override
     public boolean equals(Object other) 
     { 
-        if ( this == other ) return true ;
-
-        if ( ! ( other instanceof Quad) )
-            return false ;
-        Quad quad = (Quad)other ;
-        
-        if ( ! Objects.equals(tuple[0], quad.tuple[0]) ) return false ;
-        if ( ! tuple[1].equals(quad.tuple[1]) ) return false ;
-        if ( ! tuple[2].equals(quad.tuple[2]) ) return false ;
-        if ( ! tuple[3].equals(quad.tuple[3]) ) return false ;
-        return true ;
+        { return other instanceof Quad && super.equals(other); }
     }
     
     public boolean matches(Node g, Node s, Node p, Node o)
--- a/jena-arq/src/main/java/org/apache/jena/sparql/core/Quad.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/core/Quad.java
@@ -20,11 +20,12 @@ package org.apache.jena.sparql.core;
 
 import java.util.Objects;
 
+import org.apache.jena.atlas.lib.Tuple;
 import org.apache.jena.graph.Node ;
 import org.apache.jena.graph.NodeFactory ;
 import org.apache.jena.graph.Triple ;
 
-public class Quad
+public class Quad extends Tuple<Node>
 {
     // Create QuadNames? GraphNames?
     
@@ -48,8 +49,6 @@ public class Quad
      */
     public static final Node tripleInQuad           =  null ;
     
-    private final Node graph, subject, predicate, object ;
-    
     public Quad(Node graph, Triple triple)
     {
         this(graph, triple.getSubject(), triple.getPredicate(), triple.getObject()) ;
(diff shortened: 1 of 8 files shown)


afs:
These are some fairly drastic to get `ColumnMap` to work here.

`Triple` and `Quad` are central classes used everywhere in Jena so the implications are far ranging.

Are there cost implications?  I'm not sure the compiler implications of using arrays rather than object slots. For example, there are checks on array access - sometime they can be removed, some times not.  The objects are larger by several slots.

And it all adds many new methods to Triple and Quad. They are tuple-like but not full tuples (`Tuple` was not designed for this).

But Tuple of 3 is not a Triple.  Different `.equals`.

I'm not clear this PR meets the objective of avoiding packing and unpacking.  There is implicit creation which I don't understand the implication of via varargs parameters. e.g. `QuadOrdering.map(quad)` unpacks and calls the varargs method.

What is more, the naming spills over into other parts it's less clear than before: `first` is clearer than `nodes[0]` for example. While the original is more verbose in places, it's repetitive (which is tested extensively now) to be replaced by creating Tuples so it is not free. [I'm not sure that I actually understand _find anyway.]

ajs6f:
1. I totally understand that changing `Triple` and `Quad` is significant, so that caution is the order of the day.
2. I'm not sure of what you mean by the "compiler implications". I think that what's in this PR is just as typesafe as what is in `master` now without any more overhead, but I'm happy to correct any examples that pop up. We could get rid of the extra slots by offering a fixed-size tuple constructor `Tuple(byte size, T... elements)`, which might be nice anyway.
3. I don't understand how the notions of equality are different here, but I haven't thought as much about triples as have you. Only two `Triple`s with the same subject, predicate and object are equal now, and only two such triples are equal in this PR. (More or less the same for `Quad`s.) I've added a commit to show that. Am I missing something about how two `Triple`s could be `equal` when they aren't equal as `Tuple`s? I was careful to fix subject, predicate and object to _specific_ components of the `Tuple` (`0`, `1`, and `2`). I don't think this PR can generate problems with triples with different orders of components.
4. I actually find `nodes[1]`, etc. much clearer than `first`, etc., because the numbers help me understand what is going one, but I think that's a matter of personalities, so it's no argument in favor of the PR.

Basically, since `ColumnMap` is designed to work with `Tuple`s only, I did this in order to use it without incurring lots of extra object creation. If we don't go that road, cool, I'll look at creating something like `ColumnMap` but targeted at `Triple` and `Quad`.

ajs6f:
Do you want me to close this in light of #115 ?

afs:
Putting an extra indirection in Triple to get to the array makes it less cache friendly.  As a data structure that can be right at the heart of CPU bound operations, this may make a noticeable difference to Triple. At a bare minimum, experimentation would be necessary to verify the consequences of the change.

Every time the code has `(Node ... arg)` there is an array created.  So if I read it right, in `QuadOrdering.unmapAndCreate` there are 2 temporary arrays and several method calls including inside `unmap`, yet the current code does this directly (`QuadTableForm.quad`) by reordering use of arguments.  Actually, the important case is the reverse direction: `PMapQuadTable._find` (more results than patterns so more comes out that goes it).

Have you run a profiler on the code?

I think there is a lot to be said for the reordering version - write out the combinations once, test extensively. TDB is not a good example here. (1) it is disk backed and the costs are different (2) seeing the current TxnMem code, it could be better though there are more important gains elsewhere.

This PR seems like a lot of change to tidy up relatively few lines.

The concept of reordering is specific to indexing.  At some point, somewhere there is going to be some fiddly bit of code to do reordering. It's the mismatch between a hierarchical index where order is important and uniform accessible of sequences - it's analogously SQL tables and e.g. ISAM, VSAM style access which can be used to build SQL tables.

One of the things I tried to improve with `TupleMap` is the ability to use it for the index mapping, not just `Tuple` to `Tuple`.  I found get the naming right hard; it depends on how you think about mapping - whether it getting from the unmapped tuple to putting in the mapped tuple.

An illustration of one possibility: [TripleOps](https://github.com/afs/jena-workspace/blob/master/src/main/java/tupledev/TripleOps.java) shows two things: adding an external function to give triples an SPO order and using that operation to work between triples and tuples.

ajs6f:
Well, this branch is clearly not worth rebasing post #115, and I haven't run a profiler on it anyway. I'll take a look at `TripleOps` and think a little more about the problem.

I completely agree with your assessment, especially  "At some point, somewhere there is going to be some fiddly bit of code to do reordering." My aim is for there to be only _one_ fiddly bit per reordering, for more clarity and maintainability and less chance of error. Right now code like 'TripleTableForm` and `QuadTableForm` have many places where exactly the same reordering logic is repeated, with only a method name difference. It's hard for me to believe that there isn't a better way to do this. I just need to find one that doesn't incur extra expenses while working with Java's single-valued functions. I've tried by packing values into objects and then into arrays, so I've got to find some other way.

ajs6f:
Closing in favor of finding a different approach.

PR #111, opened 2015-12-15
Fix for JENA-1093: revert JENA-999 and add unit test ensuring that all matching literals are returned by jena-text

osma:
See https://issues.apache.org/jira/browse/JENA-1093

### DIFF
--- a/jena-text/src/main/java/org/apache/jena/query/text/TextQueryPF.java
+++ b/jena-text/src/main/java/org/apache/jena/query/text/TextQueryPF.java
@@ -1,4 +1,4 @@
-/*
+/**
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


afs:
I looked at the history of `TextQueryPF` and looks like we have struck lucky here - no post JENA-999 changes. The changes in `PropertyFunctionBase` look like tidying up except for just one line.

jena-arq/src/main/java/org/apache/jena/sparql/pfunction/PropertyFunctionBase.java
```diff
@@ -88,24 +88,25 @@ public QueryIterator exec(QueryIterator input, PropFuncArg argSubject, Node pred
...
 
         @Override
         protected QueryIterator nextStage(Binding binding)
         {
-            QueryIterator iter = exec(binding, argSubject, predicate, argObject, getExecContext()) ;
+            QueryIterator iter = exec(binding, argSubject, predicate, argObject, super.getExecContext()) ;
             if ( iter == null ) 
-                iter = IterLib.noResults(getExecContext()) ;
+                iter = IterLib.noResults(execCxt) ;
             return iter ;
```
afs: This change is potentially significant.  The execution context can change during execution where a new context derived from the current one is produced so please keep L108/L109.

osma:
All the (rather small) changes to PropertyFunctionBase seemed to be related to each other, but not directly to the changes in TextQueryPF, so I just reverted the whole file back to how it is currently (i.e. post JENA-999 state).

osma:
I squashed the two previous comments to avoid back-and-forth changes to PropertyFunctionBase. So in effect the new commit now only reverts the changes made to TextQueryPF but keeps PropertyFunctionBase intact.

afs:
+1 : we have git revision control to keep the JENA 999 version.

PR #112, opened 2015-12-22
JENA-1093: return multiple literals from text query with bound subject

osma:
Now both variableSubject and boundSubject can return multiple bindings with different literals.

I factored out the conversion from TextHits to bindings (the core part of which was previously in TextHitConverter and nowadays in a lambda expression) into a separate resultsToQueryIterator method as it is used both in the bound and variable subject cases.

### DIFF
--- a/jena-text/src/test/java/org/apache/jena/query/text/TestDatasetWithLuceneStoredLiterals.java
+++ b/jena-text/src/test/java/org/apache/jena/query/text/TestDatasetWithLuceneStoredLiterals.java
@@ -311,6 +311,9 @@ public class TestDatasetWithLuceneStoredLiterals extends AbstractTestDatasetWith
                 TURTLE_PROLOG,
                 "<" + RESOURCE_BASE + testName + ">",
                 "  rdfs:comment 'a nontext comment', 'another nontext comment'",
+                ".",
+                "<" + RESOURCE_BASE + "irrelevant>",
+                "  rdfs:comment 'an irrelevant nontext comment'",
                 "."
                 );
         String queryString = StrUtils.strjoinNL(


jena-text/src/main/java/org/apache/jena/query/text/TextQueryPF.java
```diff
@@ -210,44 +212,36 @@ private QueryIterator variableSubject(Binding binding, Node s, Node score, Node
...
 
-        Var scoreVar = (score==null) ? null : Var.alloc(score) ;
-        Var literalVar = (literal==null) ? null : Var.alloc(literal) ;
         String qs = match.getQueryString() ;
         List<TextHit> x = query(match.getProperty(), match.getQueryString(), -1, execCxt) ;
         
         if ( x == null ) // null return value - empty result
             return IterLib.noResults(execCxt) ;
         
+        List<TextHit> r = new LinkedList();
```
ajs6f: You might or might not like this style more, but you could do this with a filter:
```
List<TextHit> r = x.stream().filter(hit -> hit.getNode().equals(s)).collect(Collectors.toList());
```

osma: Thanks, I sort of knew there was a shorter way of doing this but didn't know the correct idioms (if this were Python, I'd used a list comprehension). Though I'm not sure I like that style either. At least the for loop is clear and makes it blatantly obvious that the whole list gets traversed, which is a potential performance problem in itself...

I plan to replace this part of the code very soon with a cache layer, see the discussion in JENA-999.

afs: Even if you plan to replace this, it would be good to get the fix in so the functionality is correct in master.  It gives work on the cache it's own space to happen.

osma: @afs That was exactly my intent - to fix the functionality in master first (and verify with unit tests), then continue with the cache. I just wanted to point out that the lines for which @ajs6f suggested another formulation are likely to be short-lived.

I will merge this soon if there are no other comments.

osma:
I used the Streams expression as suggested by @ajs6f and also amended the unit test so that it will check for proper filtering. Unless anyone objects I'm going to merge this soon.

PR #114, opened 2015-12-27
JENA-632: Generate JSON from SPARQL directly

kinow:
This pull request contains code for [JENA-632](https://issues.apache.org/jira/browse/JENA-632). The original work is still in a branch in [my fork of Jena](https://github.com/apache/jena/compare/master...kinow:JENA-632). It has been updated after the work on Jena 3 (mainly package renaming). And the web layer has been implemented in fuseki 2, but not backported to fuseki 1.

Besides reviewing the code, the follow steps can be used to quickly test the code.
- Start Fuseki (debug in Eclipse after checking out this branch, for example)
- Load the books.ttl from fuseki1/Data directory
- Query with something as

```
PREFIX purl: <http://purl.org/dc/elements/1.1/>
PREFIX w3: <http://www.w3.org/2001/vcard-rdf/3.0#> 
PREFIX : <http://example.org/book/> 

JSON {
"author": ?author, 
"title": ?title 
}
WHERE 
{
?book purl:creator ?author .
?book purl:title ?title . 
FILTER (?author = 'J.K. Rowling')
}
```

Which follows the syntax proposed in the issue in JIRA.

I am still reviewing the code after porting to the new code base, but an extra pair of eyes reviewing it is always welcome! :grin: 

ps: the SPARQL editor may need some tweaking to support the new syntax

ps2: tried to change the key name in the JSON query but it didn't work. Will try to update the PR if that's really a bug in the code in the next days


jena-arq/Grammar/master.jj
```diff
@@ -100,6 +100,38 @@ import org.apache.jena.sparql.core.Quad ;
 public class CLASS extends PARSERBASE
 {
     boolean allowAggregatesInExpressions = false ;
+
```
afs: This looks like debug code - could it go elsewhere?

jena-arq/Grammar/master.jj
```diff
@@ -327,6 +359,35 @@ void AskQuery() : {}
...
+  WhereClause()
+  SolutionModifier()
+}
+
+void JsonClause() : { Object o ; String s ; }
+{
+  <JSON> { getQuery().setQueryJsonType() ; }
+  <LBRACE>
+  s = String() < PNAME_NS >
+  (
```
afs: Shouldn't PNAME_NS be COLON?  PNAME_NS matches  `foo:` so  "s = String() < PNAME_NS >" matches `"abcd" xyz:`

It may be hard to parse JSON using the SPARQL tokens,in which case using some javacc ability to switch details may be needed.

jena-arq/Grammar/master.jj
```diff
@@ -2166,6 +2227,17 @@ String String() : { Token t ; String lex ; }
     }
 }
 
+Number Number() : { Token t ; Number number ; }
+{
```
afs: Number -> JSONNumber

But will rule NumericLiteral work here? Then parse the lexical form to a number.

However, generally, working in nodes is probably going to be easier because bindings map variables to Nodes.

jena-arq/Grammar/master.jj
```diff
@@ -100,6 +100,38 @@ import org.apache.jena.sparql.core.Quad ;
 public class CLASS extends PARSERBASE
```
afs: Please update tokens.txt as well - it is used to generate HTML.

jena-arq/src/main/java/org/apache/jena/sparql/lang/sparql_11/SPARQLParser11Constants.java
```diff
@@ -431,6 +433,7 @@
     "\"select\"",
     "\"distinct\"",
     "\"reduced\"",
+    "\"json\"",
```
afs: This the new JSON construct sneaking into pure SPARQL 1.1

jena-arq/Grammar/master.jj
```diff
@@ -100,6 +100,38 @@ import org.apache.jena.sparql.core.Quad ;
...
 {
     boolean allowAggregatesInExpressions = false ;
+
+    public static void main(String args[]) {
+        while (true) {
```
afs: This looks like debug code - could it go elsewhere?

afs:
Main comment : the grammar changes haven't been put inside `#ifdef ARQ` so they happen in strict SPARQL 1.1 as well as the extended ARQ language.

Keeping the SPARQL 1.1 parser exactly as the SPARQL 1.1 spec is important. There should just be some noise changes and no more.  (see for example the construct-quad changes).

jena-arq/Grammar/master.jj
kinow: Not sure where else it could go. I copied (shamelessly) from an example I found online while re-reading about JavaCC. With this main method, you can run the grammar in Eclipse, and in the Eclipse Console it will be waiting for a String+LFLF (two break lines IIRC).

Then it will use the grammar to parse the string and will output the QueryUnit. I found it useful for reviewing the changes without running some extra class with a main method, or Fuseki.

What do you think? I'm OK with removing it, or moving it somewhere else. Just don't know where else it could go :-)

jena-arq/Grammar/master.jj
kinow: Looking at the Notes file, there is a comment to run "jj2tokens sparql_11.jj > tokens.txt" to create an initial tokens.txt file, and then manually tidy it up.

Should I re-run jj2tokens, or just manually add the missing entries?

Not sure if that hasn't been used in a while and could generate strange behaviour in other grammars later.

jena-arq/Grammar/master.jj
kinow: I will need some time to check in Eclipse why I used PNAME_NS. If I remember correctly, using COLON, JavaCC would find it ambiguous with some other character or rule?

jena-arq/src/main/java/org/apache/jena/sparql/lang/sparql_11/SPARQLParser11Constants.java
kinow: Oh, when updating the branch (it was written when Jena was 2.x) I simply applied the same changes to master.jj and sparql_11.jj, and looked at other classes that I had patched, and looked for similar classes that looked like had to be updated as well.

But maybe I added Should I revert this one line?

jena-arq/Grammar/master.jj
afs: New `CLASS` don't appear very often so this could be in a java source file for each language

In fact, it only needs to work for language ARQ which is a superset of SPARQL 1.1 

Also - have you seen the command `arq.qparse`? It reads in a query and prints it out (and performs internal checks on `.equals`, `.hashcode`, output sameas input and the algebra generated.

jena-arq/Grammar/master.jj
afs: tokens.txt is used to produce HTML - hence the additional syntax to indicate inline vs a token rule for each definition.

It is safer to hand edit for minor changes because you risk loosing the additional edits alreayd made.

(If running jj2tokens, send to a temporary file and pick out the new bits and edit in manually)

jena-arq/Grammar/master.jj
afs: Yes - javacc will tokenize to PNAME_NS.  These is no COLON and if there were, theer would be other problems.

I have a devious idea!

Use the  PNAME_NS and follow with a java fragment that does additional checking:
```
   t = <PNAME_NS>
   { 
       if ( t.image is not exactly a ":" )
          throwParseException("message",  t.beginLine, t.beginColumn)
   }
```
This then restricts the legal token and means you won't to have to play complicated games with javacc to switch to a different token set (if that is even possible due to lookahead of characters).

jena-arq/src/main/java/org/apache/jena/sparql/lang/sparql_11/SPARQLParser11Constants.java
afs: This is a generated file - Put inside a `#ifdef ARQ ... #endif` in master.jj.

ajs6f:
@kinow is this still live?

kinow:
@ajs6f sorry for the delay. Checked out the branch, just need to refresh my memory. Will need a few more days before pinging you guys to check out this PR again. Next month finally will have some time to play with Jena again, then plan to have this PR completed, and check if there's any Fuseki or easy tasks in JIRA :-)

Cheers
Bruno

ajs6f:
No prob-- good to see you back, @kinow !

defret5y6u7yhfgtdrfe:
Relevant to this thread: [Return nested JSON results])https://mail-archives.apache.org/mod_mbox/jena-users/201705.mbox/%3Ctrinity-2c35f052-b0a8-4b53-aa95-9963ff11a7a6-1494243601102%403capp-mailcom-lxa06%3E)

tommals:
Relevant to this thread [Return nested JSON results](https://mail-archives.apache.org/mod_mbox/jena-users/201705.mbox/%3Ctrinity-2c35f052-b0a8-4b53-aa95-9963ff11a7a6-1494243601102%403capp-mailcom-lxa06%3E)

ajs6f:
Hey, @kinow , is this ready for rebase and review? I can try rebasing it into a separate PR, or you can if you have time... then (hopefully) we can get it in!

kinow:
Hey @ajs6f , I think so. My weak memory seems to tell me that the last time I did that... shortly after, there was some work on JSON (JSON-LD? Or maybe pure JSON) somewhere else in Jena; and when I saw that happening, I wondered if that would perhaps be a replacement for the functionality in this ticket.

I wonder if you or @afs would be able to confirm that there isn't already another way to get JSON responses in Jena, and that this feature is still a nice to have feature, please?  :)

afs:
The JSON ecosystem moves fast! GraphQL is now significant and how that might be used is also very interesting. There are different ways to use GraphQL and maybe it is not the solution to all needs anyway - this ticket is more about result sets in easy-to-consume JSON than other work in the GraphQL-RDF area and more about JSON services.

To move this ticket forward, how about putting the feature in with a big label saying "Experimental - subject to change" and _really_ mean it?  By getting it out, it generates feedback, requirements and insight. 

You can ask for result sets in JSON but the JSON (http://www.w3.org/TR/sparql11-results-json/) or JSON-LD graphs but that isn't what general data consumers would consider to be "easy-to-consume".

Is the RDF literal to JSON aligned with:
[RDF-JS mapping](http://jena.staging.apache.org/documentation/query/javascript-functions.html#arguments-and-function-results)? (which only came along recently).

kinow:
Rebased using GitHub's interface :tada:  (never used it before, hope didn't make any silly mistakes :-) ).

Good points @afs ! I noticed the existing SPARQL to JSON result writers, so adjusted a bit the code to use that.

Now when I try to rebase against the master branch I get a bunch of conflicts, but GitHub web interface seems to indicate it's all good. I was trying to find a simple way to rebase it, but if I can't fix the conflicts and compile & test it, I'll just revert to what was before and then just manually rebase :)

kinow:
Rebasing again, now manually, testing in Eclipse. Previous branch before the GitHub rebase archived at https://github.com/kinow/jena/tree/JENA-632-3

kinow:
Rebased!

Tested the `ARQParser` main method (which is added here, but I find very handy for quickly testing it, but happy to remove if others prefer). And got the following:

```
Enter input: JSON { "name": ?name } WHERE { ?name ?a ?b } LIMIT 3


log4j:WARN No appenders could be found for logger (Jena).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
Parsed query successfully!
---

Enter input: 
```

Then did a `mvn install -DskipTests -Pdev`, and copied `apache-jena-fuseki-3.7.0-SNAPSHOT.tar.gz` elsewhere. Unpacked it, and started `fuseki-server`.

Used an in-memory dataset in Fuseki2, loading the `books.ttl` example shipped with Jena code.

Finally, used the query found in the JIRA ticket to retrieve the JSON result:

```
PREFIX purl: <http://purl.org/dc/elements/1.1/>
PREFIX w3: <http://www.w3.org/2001/vcard-rdf/3.0#> 
PREFIX : <http://example.org/book/> 

JSON {
"author": ?author, 
"title": ?title 
}
WHERE 
{
?book purl:creator ?author .
?book purl:title ?title . 
FILTER (?author = 'J.K. Rowling')
}
```
And here's the output:

```
[ { 
  "author" : "J.K. Rowling" ,
  "title" : "Harry Potter and the Deathly Hallows"
}
{ 
  "author" : "J.K. Rowling" ,
  "title" : "Harry Potter and the Philosopher's Stone"
}
{ 
  "author" : "J.K. Rowling" ,
  "title" : "Harry Potter and the Order of the Phoenix"
}
{ 
  "author" : "J.K. Rowling" ,
  "title" : "Harry Potter and the Half-Blood Prince"
}
 ]
```

Going to push-force onto this branch now, so the pull request should be rebased onto the latest changes in a couple minutes.

kinow:
Done, and ready for review again.

afs:
I applied 114.patch locally and it integrated just fine.

I tried `JSON { "s": ?s , "p": ?p , "o" : ?o } WHERE { ?s ?p ?o }` on data
```
:s :p 123 .
:s :p "abc" .
:s :p "def"@en .
```
and got:
```
[ { "o" : "123^^http://www.w3.org/2001/XMLSchema#integer" } ,
  { "o" : "abc" } ,
  { "o" : "def@en" }
]
```
I think we should use a more aggressive version of the conversion from RDF term to JS used in the upcoming [JS custom Functions](http://jena.staging.apache.org/documentation/query/javascript-functions.html#arguments-and-function-results), code https://github.com/apache/jena/blob/master/jena-arq/src/main/java/org/apache/jena/sparql/function/js/NV.java. That works on `NodeValue`s to make functions more natural to write and is two-way.

For JSON-SELECT we only need a one way conversion and getting info out even if it is some string is better IMO.

| RDF | Javascript |
| ---- | ---- |
| String | String |
| XSD numeric | JSON number |
| XSD boolean | boolean |
| String@lang | String, no lang |
| URI | String |
| "lex"^^SomeDatatype | String, no datatype |
| Unbound | JSON null |

If the app wants RDF term details, it can use `application/result-set+json`. This feature provides provides a simpler, more JSONy view of the data for consuming RDF-unaware applications.

afs:
Tests?

kinow:
Going to check how to get the correct values with JSON. Great catch, as I was using just Strings. But shouldn't be too hard.

Will also include the initial tests with the next changes, covering both previous features, and the conversion from the table you provided :-)

Thanks!

afs:
Conversion code (draft): 2dd4418a22 (on a branch in my GH at the moment).

The code is in `RDFTerm2Json.fromNode`.

kinow:
+1 for the current draft

I would use it [here](https://github.com/kinow/jena/blob/a948f300fb40701e49e83db9d03e5d83da8145f6/jena-arq/src/main/java/org/apache/jena/sparql/engine/QueryExecutionBase.java#L358 and [here](https://github.com/kinow/jena/blob/a948f300fb40701e49e83db9d03e5d83da8145f6/jena-arq/src/main/java/org/apache/jena/sparql/engine/JsonIterator.java#L71).

This way instead of looking only at the literals, we'd have the output from the table above.

So far I have just two unit tests in my working copy

```java
    @Test public void testExecJson() {
        // JENA-632
        Query query = QueryFactory.create("JSON { \"s\": ?s , \"p\": ?p , \"o\" : ?o } "
                + "WHERE { ?s ?p ?o }", Syntax.syntaxARQ);
        
        try ( QueryExecution qexec = QueryExecutionFactory.create(query, m) ) {
            JsonArray jsonArray = qexec.execJson();
            assertNotNull( jsonArray );
            assertEquals(3, jsonArray.size());
        }
    }

    @Test public void testExecJsonItems() {
        // JENA-632
        {
            Resource r = model.createResource(AnonId.create("s"));
            Property p = model.getProperty("");
            RDFNode node = ResourceFactory.createTypedLiteral("123", XSDDatatype.XSDdecimal);
            model.add(r, p, node);
            r = model.createResource(AnonId.create("s"));
            p = model.getProperty("");
            node = ResourceFactory.createTypedLiteral("abc", XSDDatatype.XSDstring);
            model.add(r, p, node);
            r = model.createResource(AnonId.create("s"));
            p = model.getProperty("");
            node = ResourceFactory.createLangLiteral("def", "en");
            model.add(r, p, node);
        }
        Query query = QueryFactory.create("JSON { \"s\": ?s , \"p\": ?p , \"o\" : ?o } "
                + "WHERE { ?s ?p ?o }", Syntax.syntaxARQ);
        
        try ( QueryExecution qexec = QueryExecutionFactory.create(query, model) ) {
            Iterator<JsonObject> execJsonItems = qexec.execJsonItems();
            while(execJsonItems.hasNext()) {
                JsonObject next = execJsonItems.next();
                System.out.println(next);
            }
        }
```

Where the second test is doing - more or less - the same example you provided in your previous comment. Once we have the conversion in place, I will update it to confirm I get what's expected.

Then after that I was thinking about quickly running with jacoco to see how much of this PR is being tested.

kinow:
FWIW, the output of the second test right now is

```
{ 
  "s" : "s" ,
  "p" : "" ,
  "o" : "def@en"
}
{ 
  "s" : "s" ,
  "p" : "" ,
  "o" : "abc"
}
{ 
  "s" : "s" ,
  "p" : "" ,
  "o" : "123^^http://www.w3.org/2001/XMLSchema#decimal"
}

```

@afs are you planning to send another PR when the draft is ready (it looks good to me as-is now :-) ) or would you prefer to include in this PR and Jena-632? It does seem to be related with/necessary for JENA-632. Happy to get updates to this branch :-)

afs:
PR #383 - we can still change it based on experience.

BTW The grammar accepts the JSON field names in single quotes to make embedding in strings easier. It's a STRING1, either one `"` or one `'` delimiting (not mixed!).

I doubt it will need many tests.

kinow:
Most of the tests are done, just need to add at least basic coverage for the SPARQL changes. Then will spend some minutes playing with the queries in Fuseki, before updating the PR. Aiming to push the changes Friday night NZ time, or this weekend.

afs:
Great timing! but as ever, it it makes it, it makes it, and if it doesn't, it doesn't.

(I will not be able to build 3.7.0 until next week anyway.)

PR #383 merged.

jena-arq/src/main/java/org/apache/jena/query/Query.java
```diff
@@ -707,6 +719,8 @@ public void visit(QueryVisitor visitor)
             visitor.visitDescribeResultForm(this) ;
         if ( this.isAskType() )
             visitor.visitAskResultForm(this) ;
+        if ( this.isJsonType() )
```
kinow: @afs added this one in order to be able to use the `Query` object in some tests...

jena-arq/src/main/java/org/apache/jena/sparql/engine/JsonIterator.java
```diff
@@ -0,0 +1,100 @@
...
+        if (queryIterator == null)
+            throw new NoSuchElementException(this.getClass() + ".next") ;
+        try
+        {
+            Binding binding = queryIterator.next() ;
+            JsonObject jsonObject = new JsonObject() ;
+            for (String resultVar : resultVars)
+            {
+                Node n = binding.get(Var.alloc(resultVar)) ;
+                JsonValue value = RDFTerm2Json.fromNode(n) ;
```
kinow: @afs updated the code where I mentioned before...

jena-arq/src/main/java/org/apache/jena/sparql/engine/QueryExecutionBase.java
```diff
@@ -335,7 +340,45 @@ public boolean execAsk() {
...
+        JsonArray jsonArray = new JsonArray() ;
+        List<String> resultVars = query.getResultVars() ;
+
+        while (queryIterator.hasNext())
+        {
+            Binding binding = queryIterator.next() ;
+            JsonObject jsonObject = new JsonObject() ; 
+            for (String resultVar : resultVars) {
+                Node n = binding.get(Var.alloc(resultVar)) ;
+                JsonValue value = RDFTerm2Json.fromNode(n) ;
```
kinow: @afs and here

jena-arq/src/main/java/org/apache/jena/sparql/serializer/QuerySerializer.java
```diff
@@ -142,7 +144,19 @@ public void visitAskResultForm(Query query)
         out.print("ASK") ;
         out.newline() ;
     }
-    
+
+    @Override
+    public void visitJsonResultForm(Query query) {
+        out.print("JSON {");
```
kinow: This is used only when we create the `Query` manually. Only developers/advanced users might be interested in using it. The results via HTTP are not serialized this way.

jena-arq/src/main/java/org/apache/jena/query/ResultSetFormatter.java
```diff
@@ -516,6 +520,41 @@ static public void outputAsJSON(boolean booleanResult)
...
+       jWriter.startOutput() ;
+       while (jsonItems.hasNext()) 
+       {
+           jWriter.startObject() ;
+           JsonObject jsonItem = jsonItems.next() ;
+           for (Entry<String, JsonValue> entry: jsonItem.entrySet()) 
+           {
+               JsonValue value = entry.getValue() ;
+               String val = "";
+               if (value.isString()) {
```
kinow: @afs but here I didn't have a `Node`, only the `JsonValue`, and had to convert from `JsonValue` to `String`

kinow:
screenshot_2018-03-17_00-18-10
screenshot_2018-03-17_00-17-44

Screenshots of Fuseki running some JSON queries. Everything seems to be working fine :tada: ready for review, take 2 !

jena-arq/src/main/java/org/apache/jena/query/ResultSetFormatter.java
afs: "outputAsJSON" is a bit confusing because the other functions of the same name produce SPARQL results in JSON.  How about "output"?

In the code, numbers still come out as strings.

JSWriter seems to be only good for its current usage. The code below seems to work: it manages the outer array itself and uses `JsonValue.output` to write each item.

```
 public static void outputAsJSON(OutputStream outStream, Iterator<JsonObject> jsonItems)
    {
        IndentedWriter out = new IndentedWriter(outStream);
        out.println("[");
        out.incIndent();

        while (jsonItems.hasNext()) {
            JsonObject jsonItem = jsonItems.next() ;
            jsonItem.output(out);
            if ( jsonItems.hasNext() )
                out.println(" ,");
            else
                out.println();
        }
        out.decIndent();
        out.println("]");
        out.flush();
    }
```

kinow: >"outputAsJSON" is a bit confusing because the other functions of the same name produce SPARQL results in JSON. How about "output"?

Sounds good to me. Done!

>In the code, numbers still come out as strings.

Oh, d'oh me! You raised the issue, I updated the code but not the part that writes it back to the user. And even added a test that was testing it was getting "123" and "true" in the output. Fixed the test, and thanks heaps for the code snippet! Otherwise I'd be scratching my head and thinking how to use JSWriter.

Updated, and uploaded the data below

```
@prefix :          <http://example.org/> .

:s :p 123 .
:s :p "abc" .
:s :p "def"@en .
```

Then queried `JSON { "s": ?s , "p": ?p , "o" : ?o } WHERE { ?s ?p ?o }`. Screen shot below with the `123` value - hopefully - correctly displayed! :man_dancing: :tada: 

screenshot_2018-03-20_21-02-30

kinow:
Replied @afs `s last comment, and also rebased the code. Even though it's rebased, there's quite a few commits... let me know if I should squash the commits or if it's preferable to leave the commit history :-) ready to review, take 3?

jena-arq/Grammar/master.jj
```diff
@@ -99,7 +99,42 @@ import org.apache.jena.sparql.core.Quad ;
 #endif
 
 public class CLASS extends PARSERBASE
-{}
+{
+    boolean allowAggregatesInExpressions = false ;
+#ifdef ARQ
```
afs: Please remove "main".

jena-arq/Grammar/master.jj
```diff
@@ -326,6 +361,46 @@ void AskQuery() : {}
...
+    <COMMA> s = String()
+    t = < PNAME_NS > {
+      if ( ! t.image.equals(":") )
+        throwParseException("Prefix name expression not legal at this point : "+t.image, t.beginLine, t.beginColumn) ;
+    }
+    (
+      o = Var() { getQuery().addResultVar((Var) o) ; getQuery().addJsonMapping(s, o) ; }
+    | o = String() { getQuery().addJsonMapping(s, o) ; }
+    | o = Number() { getQuery().addJsonMapping(s, o) ; }
+    )
```
afs: Does not cover URIs, booleans or JSON null.

I tried `JSON { "F": "string" } WHERE {}` but got `[ {} ]` (one element, no object fields) - which does not look right to me.

A bunch of tests calling a common function like `test(String query, String jsonResult)` would help. e.g.

    @Test public void json_query_01() { test("JSON { 'F': 'string' } WHERE {}", "{ [ 'F' : 'string ] }" ; }

`JsonValue` supports structural `.equals`.

jena-arq/Grammar/master.jj
```diff
@@ -2176,6 +2251,19 @@ String String() : { Token t ; String lex ; }
     }
 }
 
+#ifdef ARQ
+Number Number() : { Token t ; Number number ; }
+{
+  (
+    t = < INTEGER > { number = integerValue(t.image) ; }
+  | t = < DECIMAL > { number = doubleValue(t.image) ; }
```
afs: Any reason to exclude doubles?

jena-arq/Grammar/master.jj
```diff
@@ -99,7 +99,42 @@ import org.apache.jena.sparql.core.Quad ;
 #endif
 
 public class CLASS extends PARSERBASE
-{}
+{
+    boolean allowAggregatesInExpressions = false ;
```
afs: This is not in master.jj nowadays. Is this from the early forking?

jena-arq/src/main/java/org/apache/jena/query/Query.java
```diff
@@ -515,6 +519,14 @@ public void addHavingCondition(Expr expr)
         havingExprs.add(expr) ;
     }
 
+    // SELECT JSON
+
+    public Map<String, Object> jsonMapping = new LinkedHashMap<>();
```
afs: Preference - don't expose data structures which can be modified. Either return a copy (if it is not called repeatedly) or with a `Collections.unmodifiableMap` wrapper.

afs: I can't find any use of `jsonMapping` except serialization which would explain the lack of fields noted above.

jena-arq/src/main/java/org/apache/jena/sparql/engine/JsonIterator.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.jena.graph.Node;
+import org.apache.jena.sparql.core.Var;
+import org.apache.jena.sparql.engine.binding.Binding;
+import org.apache.jena.sparql.lib.RDFTerm2Json;
+
+/** A JSON iterator for JsonObject's, that wraps a QueryIterator, and a list
+ * of result variables.
+ */
+public class JsonIterator implements Iterator<JsonObject>
+{
```
afs: Feel free to reformat int same like `{` and no space before `;`.

(yes - a lot of code is not formatted but we seem to be heading that way).

Or leave for now and we can clean up new classes after merging.

jena-arq/src/main/java/org/apache/jena/sparql/engine/http/QueryEngineHTTP.java
```diff
@@ -532,6 +535,18 @@ else if ( actualContentType.equals(WebContent.contentTypeJSON))
         }
     }
 
+    @Override
+    public JsonArray execJson()
+    {
+        throw new NotImplementedException("JSON queries not implemented for remote calls") ;
```
afs: Minor: for better or worse, we have `org.apache.jena.atlas.lib.NotImplemented`.

jena-arq/src/main/java/org/apache/jena/sparql/serializer/QuerySerializer.java
afs: The `qparse` tool uses serialization (and it parses the result back again).

afs: More test suggestions: `testParse("JSON { "s": "FOO" , "p" : ?p , "o" : ?o } WHERE { }")` to parse to Q, serialize, reparse to Q2 and check Q.equals(Q2).

e.g.a new `org.apache.jena.sparql.syntax.TestJSONParsing`

`JSON { "s": "FOO" }` becomes `JSON { "s": FOO }`

jena-arq/src/main/java/org/apache/jena/sparql/syntax/syntaxtransform/QueryTransformOps.java
```diff
@@ -217,6 +217,11 @@ public void visitAskResultForm(Query query) {
             newQuery.setQueryAskType();
```
afs: Transforming `JSON { ... }` can be left for another ticket.

jena-arq/src/main/java/org/apache/jena/sparql/engine/JsonIterator.java
ajs6f: I've been planning to offer a low-interruption plan for some large-scale reformatting later this spring, so no harm in leaving this.

jena-arq/Grammar/master.jj
kinow: Most likely. Will remove tonight.

jena-arq/src/main/java/org/apache/jena/sparql/engine/http/QueryEngineHTTP.java
kinow: Coincidentally, I noticed this [lang] class here too yesterday, and was going to replace by that Atlas class... but first I searched other occurrences, and I think somewhere in a sparql module it was being used too. So I thought it was OK. Happy to replace this one by the atlas module class. Consider it done in some hours.

jena-arq/src/main/java/org/apache/jena/sparql/engine/JsonIterator.java
kinow: Okeydokey @afs I noticed new classes are getting the new formatting applied, but wasn't sure about changing it in this PR or leaving as it was. This change started somewhere in 2015, so lots of old code formatting here.

jena-arq/Grammar/master.jj
kinow: Consider it done. I'll copy it to a GitHub gist in case I need to test grammars again (not only in Jena... I believe I actually got it from some StackOverflow answer).

afs:
Syntax tests: https://gist.github.com/afs/e069478ec6839267256b9eacd30aa2c4

kinow:
Will apply some of the changes tonight. But I'm afraid I may not be able to fix the serialization issues before this weekend when there might be a new release. This weekend I'll be in Japan for some weeks, so may have to complete working on the other comments by mid-April 😢

kinow:
And only started working again on this because you bumped this PR @ajs6f :-) learning a lot working on this ticket, and with JENA-1488, hope to start contributing back to Jena more now.

kinow:
@afs had a quick look at your gist and it looks good! I won't be able to have a proper look at it and to commit for the next ~8 hours. But feel free to commit to this branch and update the pull request too if you prefer :-)

afs:
Eval tests: https://gist.github.com/afs/9e57b705a0d53889606346f6723238ce

Framework : needs more tests!

afs:
If it does not make the release then - whatever! - there'll be another along soon!

I may be able to do some work on it after the release. I think I now understand it better and it makes a useful contribution to developing applications that want JSON simply.

Have fun in Japan!

jena-arq/Grammar/master.jj
kinow: Oh, no reason at all. Just forgot about it I think. Re-reading that method, I almost feel like removing the number parameter, and just replacing by `return integetValue(t.image);` etc. But will just add the `< DOUBLE >` for now :-)

afs: Would this be simpler? 

Generate `Node` from the parser, 
The data  structure is then `Map<String, Node> jsonMapping`.
To use, lookup variables in the binding to get a `Node` and also use the `Node`s from `jsonMapping`.

In the grammar, use `NumericLiteral() | BooleanLiteral()`, which return `Node`, then the sign of numbers is also handled.  as well as booleans. When needing the item for making the JSON, apply `RDFTerm2Json`, as is done when a variable is used and turned into a Json value.

There isn't a standard alone string rule returning a `Node`. The `String()` returns a string (it is used to make a RDF Literal, optional ^^ and @ parts) - to unify handling turn the string into a `Node` with `NodeFactory.createLiteral(t.image)`.

jena-arq/Grammar/master.jj
kinow: Good idea! Using your TestJsonEval.java as base, and doing a quick change in the master.jj and grammars, to support it.

jena-arq/src/main/java/org/apache/jena/sparql/serializer/QuerySerializer.java
kinow: >JSON { "s": "FOO" } becomes JSON { "s": FOO }

Shouldn't it be

```
JSON { "s": "FOO" } becomes JSON { "s": "FOO" } // last FOO still with double quotes / a string?
```

jena-arq/Grammar/master.jj
kinow: Oh! It was by accident that it got fixed I think?

I copied parts of the `SelectClause()` to `JsonClause()`. More especifically, the Select had:

```
      // Expressions without ()
      ( LOOKAHEAD(2)
        expr = BuiltInCall()        { getQuery().addResultVar((Var)null, expr) ; }
      | expr = FunctionCall()       { getQuery().addResultVar((Var)null, expr) ; }
      | n = RDFLiteral()            { getQuery().addResultVar((Var)null, NodeValue.makeNode(n)) ; }
      | n = NumericLiteral()        { getQuery().addResultVar((Var)null, NodeValue.makeNode(n)) ; }
      | n = BooleanLiteral()        { getQuery().addResultVar((Var)null, NodeValue.makeNode(n)) ; }
```

Which became the following for the JSON counterpart:

```
    o = Var() { getQuery().addResultVar((Var)o) ; getQuery().addJsonMapping(s, o) ; }
  | o = RDFLiteral() { getQuery().addResultVar(s, NodeValue.makeNode(o)) ; getQuery().addJsonMapping(s, o) ; }
  | o = NumericLiteral() { getQuery().addResultVar(s, NodeValue.makeNode(o)) ; getQuery().addJsonMapping(s, o) ; }
  | o = BooleanLiteral() { getQuery().addResultVar(s, NodeValue.makeNode(o)) ; getQuery().addJsonMapping(s, o) ; }
```

More or less the same. But we have the Var (i.e. for cases like `"title" : ?title`), and then I did not include the expressions (hence the lookahead removed).

Does it make sense @afs ? If so, we can actually remove the `Number()` and this whole #if ARQ block. And I believe the code is doing more or less what you suggested?

Thanks!
Bruno

afs: Observa
tion: The JSON clause can be DRY-ed with `JsonObjectMember() ( <COMMA> JsonObjectMember()) *` and put lines 342-355 in `JsonObjectMember`.

afs: Couldn't reply to the number point - I think the rebase has moved things around - it now looks right.  Tests will tell!

kinow: @afs nice catch in that observation. Reduced code duplication and avoids now possible "forgot to apply to the other part of the json clause" issues in the future :-)

Change applied. Rebased, ran all tests, tried three queries mentioned in this pull request (we have a few already! :smiley: ), got no issues.

Pushing the latest code, rebased onto the maste branch.

kinow:
@afs also removed the `Number()` clause as that was not being used any more. Ready for review again :+1: 

And big thanks for all the feedback (and patience!) so far.

jena-arq/src/main/java/org/apache/jena/query/Query.java
```diff
@@ -515,6 +519,18 @@ public void addHavingCondition(Expr expr)
         havingExprs.add(expr) ;
     }
 
+    // SELECT JSON
+
+    private Map<String, Object> jsonMapping = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
kinow: Done!

jena-arq/src/main/java/org/apache/jena/sparql/serializer/QuerySerializer.java
```diff
@@ -142,7 +144,19 @@ public void visitAskResultForm(Query query)
...
+        List<String> terms = new ArrayList<>();
+        for (Map.Entry<String, Object> entry : query.getJsonMapping().entrySet()) {
```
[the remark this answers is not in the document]
kinow: Patch applied! Thanks Andy! TIL learned about that Prologue object.

Regarding the indentation, in `QuerySerializer.java`, there are other 2 occurrences of `incIndent`/`decIndent`, but instead of using `4`, they use the class constant

`static final int BLOCK_INDENT = 2 ;`

Should we use that for JSON as well?

Thanks!!!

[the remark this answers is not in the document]
afs: Sorry about the patch - I couldn't figure how to clone your repo because GH did not like the fact I already had a repo called "jena". Yes, a constant is better.  I thought that 4 looked nicer than 2 and it does not get indented further.  Just use `incIndent` twice!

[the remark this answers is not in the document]
afs: Is it ready to merge?  Any area to look at specially? I've used `qparse` on JSON queries and run a few test queries with all the cases I can think of.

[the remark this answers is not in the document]
kinow: > Just use incIndent twice!

Haha, I like it. Done!

>Is it ready to merge? Any area to look at specially? I've used qparse on JSON queries and run a few test queries with all the cases I can think of.

I think so ! Only minor thing on my mind now are some expressions. We have not implemented support for queries with expressions like

`JSON { 'F' : sum(1+2)} WHERE { } `

Or other functions. Perhaps that could - if necessary - be implemented later?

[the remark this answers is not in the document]
afs: For now, Expressions can be done with `BIND` at the end of the `WHERE` pattern.

(`sum` is a group-aggregator -- whole different ball game!)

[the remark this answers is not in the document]
kinow: Oh, got it! Nice! So no more comments from me :smiley:

kinow:
:tada: should I proceed with the merging? If so, I will squash the 18 commits into a single one before, as said before, but with a backup branch just-in-case. And then start working on the documentation in order to update the ticket in JIRA.

afs:
That works for me.

kinow:
Branch created: [JENA-632-before-merge](https://github.com/kinow/jena/tree/JENA-632-before-merge). Won't delete it this year, in case we need to revisit.

Followed that by:

```
$ git rebase -i HEAD~18
$   # fixed-up all but the first commit
$ git diff JENA-632-before-merge 
$   # empty response, code is identical even though different commit tree
$ git checkout master
$ git merge JENA-632-2
$ git commit --amend
$   # added This closes #114 message to close this pull request
$ mvn clean test install -Pdev
$ git push git-at-apache master
```

PR #115, opened 2015-12-28
JENA-1107 : Improve Tuple.

afs:
Improvements:
* Tuple as Interface
* Implementation focused on being "Tuples", not only for TDB (it's historical origin).
* Less space for small tuples.
* Avoid array overheads for small tuples.
* TupleMap uses more consistent naming than ColumnMap.
* TupleMap can be used for index mapping, array mapping, not use Tuple mapping.


jena-base/src/main/java/org/apache/jena/atlas/lib/tuple/Tuple.java
```diff
@@ -0,0 +1,73 @@
...
+    /** length : elements are 0 to len()-1 */ 
+    public int len() ;
+
+    /** Convert to a List */
+    public default List<X> asList() {
+        return new TupleList<>(this) ;
+    }
+
+    /** stream */
+    public default Stream<X> stream() {
```
ajs6f: `Iterable` has a `default` impl for `::spliterator` that could be used to make a `Stream`. I think that could avoid the extra object creation in `Tuple::asList`. That's how the `default` works for `Collection::stream`.

jena-base/src/main/java/org/apache/jena/atlas/lib/tuple/Tuple.java
```diff
@@ -0,0 +1,73 @@
...
+        copyInto(array, 0, len());
+    }
+
+    /** Copy the Tuple into the array */ 
+    public default void copyInto(X[] array, int start) {
+        copyInto(array, start, len());
+    }
+
+    /** Copy the Tuple into the array */ 
+    public void copyInto(X[] array, int start, int length) ;
```
ajs6f: Isn't there a really obvious `default` here? Maybe it would be good to have a comment here explaining why impls need to do their own thing here...?

jena-base/src/main/java/org/apache/jena/atlas/lib/tuple/Tuple.java
```diff
@@ -0,0 +1,73 @@
...
+
+package org.apache.jena.atlas.lib.tuple;
+
+import java.util.List ;
+import java.util.function.Consumer ;
+import java.util.stream.Stream ;
+
+import org.apache.jena.atlas.lib.ArrayUtils ;
+
+/** A Tuple is the same class of item */
```
ajs6f: This comment is a bit opaque. Does it mean that all items in the tuple are expected to have the same class?

jena-base/src/main/java/org/apache/jena/atlas/lib/tuple/Tuple.java
afs: I can move that (and iterator()) into the interface.  

(Now TupleBase is needed (because you can't default methods from Object in an interface) some of all can go there if it keeps the source clean.)

jena-base/src/main/java/org/apache/jena/atlas/lib/tuple/Tuple.java
afs: Good point.

PR #117, opened 2015-12-31
JENA-583: Using Maven optional tag for log4j dependencies

ajs6f:


afs:
The codebase should compile with a clean maven repo.
Jena does not compile without this dependency.

This is a discussion needed -- see JENA-1005.

ajs6f:
Having the [optional tag](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html) doesn't prevent Jena from using this dependency. It's just as much present as ever it was. It prevents the dependency from being _transitive_. I successfully compiled this PR before sending it and tested the CLI tools.

ajs6f:
This PR builds with `mvn clean install` or, to be extra-sure:
```
mvn clean dependency:purge-local-repository -DreResolve=false
mvn install
```

PR #118, opened 2016-01-04
A maven module for Jena cmds

afs:
This new module puts all the commands in one place.  They retain their original java package names.

Commands covered are those from jena-core, jena-arq, and jena-tdb - the ones in the binary distribution.

Fuseki2 also puts these commands in its server jar - people have found executing from just the server jar a convenient feature.  This can be reconsidered; this PR does not change that aspect.

The binary distribution (zip, tar.gz) now depends on module jena-cmds.

Maven artifact `apache-jena-libs` does not so using that artifact will now miss the commands.

Maybe this module should be called apache-jena-cmds as a delivery unit.  Or maybe we should keep as is, and add apache-jena-cmds as a combined single jar for the main commands.


ajs6f:
+1 to the general plan here, definitely. Is the future plan to migrate the namespaces as well?

rvesse:
Looks like a good first step

afs:
@ajs6f Some of the commands already have two names, one from jena.XYZ. It used a reflection call to call forward in the build chain; now the commands included have direct calls.  There is the same code available if some other command wants to have a consolidated name but the user will need the additional module anyway.

ajs6f:
Okay. I found it odd that so many package names re-appear in the new artifact. I'm used to OSGi habits, but this is nothing to cavil at, just my predilections.

afs:
No package names have been invented.  They are the same as before, just now in one module.

afs:
(wrong button)

ajs6f:
Right, that's why I said "re-appear". They appear in the original modules, and now in the new one. As I wrote, it's not anything to worry about and mostly just about my expectations.

PR #121, opened 2016-01-11
JENA-149: Duration equality by value.

afs:
This PR provides canonicalization of the date/time duration in comparison, equality and hashcode operations based on year-month and day-time parts. (e.g. if it is 110seconds, it become 1minute 50seconds).

Only year-month and day-time durations are fully comparable but applications may, indeed probably do, use the xsd:duration datatype, not the appropriate subtype.

This does not affect SPARQL.


der:
Looks good to me, many thanks. Sorry, I had completely forgotten about that ancient Jira!

Happy for you to merge or I can if that's better practice.

afs:
I can merge it as I'm all setup for that at the moment.

The PR only canonicalizes at the point the test operations are called, not when its created.  This is just caution.  The `XSDDuration.toString` method also works on the `data[]` array so that is unchanged, though I think very unlikely to be relevant as it is not the lexical form.

PR #122, opened 2016-01-15
JENA-1118: bringing back the SDB distribution

osma:
See https://issues.apache.org/jira/browse/JENA-1118

### DIFF
--- a/jena-sdb/assembly.xml
+++ b/jena-sdb/assembly.xml
@@ -41,11 +41,11 @@
 
   <files>
     <file>
-      <source>LICENSE</source>
+      <source>dist/LICENSE</source>
       <destName>LICENSE</destName>
     </file>
     <file>
-      <source>NOTICE</source>
+      <source>dist/NOTICE</source>
       <destName>NOTICE</destName>
     </file>
   </files>
--- a/jena-sdb/assembly.xml
+++ b/jena-sdb/assembly.xml
@@ -41,11 +41,11 @@
 
   <files>
     <file>
-      <source>dist/LICENSE</source>
+      <source>LICENSE</source>
       <destName>LICENSE</destName>
     </file>
     <file>
-      <source>dist/NOTICE</source>
+      <source>NOTICE</source>
       <destName>NOTICE</destName>
     </file>
   </files>
--- a/jena-sdb/dist/LICENSE
+++ b/jena-sdb/dist/LICENSE
@@ -1,0 +1,246 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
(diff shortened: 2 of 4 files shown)


jena-sdb/assembly.xml
```diff
@@ -41,11 +41,11 @@
 
   <files>
     <file>
-      <source>dist/LICENSE</source>
+      <source>LICENSE</source>
```
afs: They are probably different.  The `dist/` version cover everything in the assembly, including shipped dependencies. The ones at the top of the module just cover the code in the module.

Can we just pull `dist/` out of git and check?

afs:
From memory, there were problems making an assembly and the jar in the same module.  Some sort of name clash but maybe it wasn't this module and maybe that is why the assembly is "-distribution".

jena-sdb/assembly.xml
osma: I changed this because the Maven assembler complained that "dist/LICENSE" didn't exist. I understood these paths to be relative to the jena-sdb directory.

Maybe change to "../dist/LICENSE"?

osma:
It works for me. Also, jena-fuseki1 seems to do the same.

afs:
This makes "jena-sdb....distribution" becomes a binary product of the release. No problem with that but we should do it formerly. When this is running and distributions in the snapshots maven repo, could you email dev@ with a message asking for a review of the binary and specifically the N&L. We have to do that at release - it seems to me to be easier to do the task now, rather than at release when it's busy.

jena-sdb/assembly.xml
osma: Sorry, I'm confused now. Which `dist/LICENSE` should this be? There is no `jena-sdb/dist` directory nowadays. There is `jena-fuseki1/dist` and `apache-jena-fuseki/dist` currently.

osma:
Ah, hm. I didn't realize that. I was thinking that it should simply be possible to build this distribution, just like it is currently possible to build the SDB jar.

jena-sdb/assembly.xml
afs: The deleted `jena-sdb/dist` -- it must be in git somewhere.

osma: Ah, now I see. I'll hunt for it...

afs:
Yes, it should "just work" and it does push it to maven.

See http://central.maven.org/maven2/org/apache/jena/jena-sdb/1.4.1/ which is the last one produced.

jena-sdb/assembly.xml
osma: Removed in ec123a3637c8bf115896d5fb2babc41c8eddbf6a

osma:
Ah, okay. So these were actually produced for a while after the 1.3.6 release, but not put on the download site, just available via Maven Central. Okay, I think getting back to that state of affairs would be fine. I'm unsure about "official" downloads, knowing the state of SDB support available from the project.

afs:
Official downloads is definitely another step.  I think that is definitely a project-wide decision.

osma:
I found and updated (year) LICENSE and NOTICE from git history. Good now?

afs:
I checked the original dist/{N,L} in git and it looks like no additions from the main binary distribution. Did you spot anything? I've had an all-to-brief look of assembly.xml and it looks like it excludes all JDBC so no issues there.  As the main N&L have been updated since so copy apache-jena/{N,L} to jena-sdb/dist as a starting point.

When you email dev@, I'll check the snapshot build. (I'm not in a position to do a local build myself to make a "distribution" and check just at the moment.)

osma:
Hmm, there are several LICENSE files in the Jena source tree which mention "code contributed by Plugged In Software". It appears that some of these have old pathnames referring to `com.hp.hpl.jena` packages which have since been renamed:

```
$ grep ExtendedHandler `find . -name LICENSE`
./LICENSE:src/main/java/org/apache/jena/rdf/arp/ExtendedHandler.java
./jena-sdb/dist/LICENSE:src/main/java/com/hp/hpl/jena/rdf/arp/ExtendedHandler.java
./jena-core/LICENSE:src/main/java/com/hp/hpl/jena/rdf/arp/ExtendedHandler.java
./jena-core/src/main/resources/META-INF/LICENSE:src/main/java/com/hp/hpl/jena/rdf/arp/ExtendedHandler.java
./jena-core/target/classes/META-INF/LICENSE:src/main/java/com/hp/hpl/jena/rdf/arp/ExtendedHandler.java
./apache-jena/LICENSE:src/main/java/org/apache/jena/rdf/arp/ExtendedHandler.java
```

As you can see above, this affects not only the jena-sdb/dist/LICENSE file but also jena-core and apache-jena.

osma:
Okay I copied the changes from apache-jena. It seems that the pathnames were already fixed in the `apache-jena/LICENSE` file so they are now fixed in `jena-sdb/dist/LICENSE` too, but there are still stale copies around elsewhere, as evidenced by the above grep results.

osma:
I will merge this soon to master unless someone opposes...

PR #124, opened 2016-01-25
Add Javadoc for Assembler indicating that Model values are advisory

ajs6f:
Attempting to capture @afs's comment [here](https://github.com/apache/jena/pull/123#issuecomment-174306928) as advice in the Javadocs.


afs:
This is useful - I don't know the true semantics of "mode" - I suspect it reflects part of the assembler system where creation and sharing are in a cluster of items that are related in some way so they know the overall sharing intended.  It needs someone with closer knowledge to say (or a lot of retrospective archaeology).

PR #126, opened 2016-02-10
JENA-1139 workaround

stain:
This workaround for JENA-1139 concatenates
the content of the other Jena modules
`src/main/resources/META-INF/services/org.apache.jena.system.JenaSubsystemLifecycle`
from jena-osgi's inlined modules `jena-core`, `jena-arq` and `jena-tdb`


afs:
I tried to pull this and got changes to tdb/base/file/LocationLock (possibly no actual change). What's happening?

afs:
Applied as a diff to avoid confusion over tdb/base/file/LocationLock.

PR #127, opened 2016-03-05
JENA-1147 : Introduce FactoryRDF

afs:
FactoryRDF is an interface to abstract the creation of nodes, triples and quads from constituent elements, e.g. to create a URI Node, the operation takes a string for the URI. This allows different policies to be slotted in.  This PR covers the use of FactoryRDF by ParserProfile. A PR to follow will use a caching implementation of FactoryRDF as the default mechanism when parsing.


afs:
This has got messy.  I'll resubmit a cleaned up version.

PR #128, opened 2016-03-06
JENA-1147 : Introduce FactoryRDF 

afs:
FactoryRDF is an interface to abstract the creation of nodes, triples and quads from constituent elements, e.g. to create a URI Node, the operation takes a string for the URI. This allows different policies to be slotted in. This PR covers the use of FactoryRDF by ParserProfile.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/system/FactoryRDF.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/system/FactoryRDF.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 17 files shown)


jena-arq/src/main/java/org/apache/jena/riot/system/FactoryRDF.java
```diff
@@ -0,0 +1,56 @@
...
+    // ?? Are these too varied?
+    public DatasetGraph createDatasetGraph() ;
+    public Triple createTriple(Node subject, Node predicate, Node object) ;
+    public Quad createQuad(Node graph, Node subject, Node predicate, Node object) ;
+    public Node createURI(String uriStr) ;
+    public Node createTypedLiteral(String lexical, RDFDatatype datatype) ;
+    public Node createLangLiteral(String lexical, String langTag) ;
+    public Node createStringLiteral(String lexical) ;
+    /** Create a blank node */
+    public Node createBlankNode() ;
```
ajs6f: Do we want something for `BlankNodeId` here?

jena-arq/src/main/java/org/apache/jena/riot/system/FactoryRDFCaching.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public FactoryRDFCaching(int cacheSize, LabelToNode labelMapping) {
+        super(labelMapping) ;
+        cache = setCache(cacheSize) ;
+    }
+
+    @Override
+    public Node createURI(String uriStr) {
+        try {
+            return cache.get(uriStr, ()->RiotLib.createIRIorBNode(uriStr)) ;
```
ajs6f: This may be a dumb question, but why `RiotLib` here, why not `NodeFactory`?

afs: `NodeFactory` is fundamental and provide exactly the creation operations without opinion (maybe you want to create a real URI that looks like `<_:label>` e.g. for output.

`RiotLib.createIRIorBNode` understands the `<_:label>` form.

ajs6f: Ah, cool.

PR #131, opened 2016-03-30
JENA-1134: support AnalyzingQueryParser in jena-text

osma:
This PR makes it possible to select either the standard Lucene QueryParser or the AnalyzingQueryParser using jena-text configuration like this:

```
    text:directory <file:Lucene> ;
    text:queryParser text:AnalyzingQueryParser ;
    text:queryAnalyzer [
        a text:ConfigurableAnalyzer ;
        text:tokenizer text:KeywordTokenizer ;
        text:filters (text:ASCIIFoldingFilter text:LowerCaseFilter)
    ] 
    text:entityMap <#entMap> ;
```

The main difference between these query parsers is that AnalyzingQueryParser performs analysis also for wildcard queries. For example, if you use ASCIIFoldingFilter as above, if you want a search for `édu*` to match `éducation` you need AnalyzingQueryParser.

One problem I had with the implementation is that the query parser needs to be constructed dynamically for every query, so I need to store the information about which query parser to use instead of just storing the QueryParser/AnalyzingQueryParser instance directly. I solved this by simply storing the type of query parser as a string, i.e. either `"QueryParser"` or `"AnalyzingQueryParser"`, and then dynamically construct the correct type of parser based on this information. I'm sure there are more elegant ways of doing this, e.g. creating Factories for each parser type and saving the correct kind of Factory, but I don't want to overengineer. Opinions?

This could rather easily be extended to other query parser types supported by Lucene, though I'm unsure how useful that would be in practice. ComplexPhraseQueryParser and/or PrecedenceQueryParser could perhaps be useful to somebody.

### DIFF
--- a/jena-text/src/test/java/org/apache/jena/query/text/TestDatasetWithAnalyzingQueryParser.java
+++ b/jena-text/src/test/java/org/apache/jena/query/text/TestDatasetWithAnalyzingQueryParser.java
@@ -1,0 +1,64 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.jena.query.text;
+
+import java.util.Set ;
+
+import org.apache.jena.atlas.lib.StrUtils ;
+import org.apache.jena.ext.com.google.common.collect.Sets ;
+import org.junit.Before ;
+import org.junit.Test ;
+
+/**
+ * This class defines a setup configuration for a dataset that uses an ASCII folding lowercase keyword analyzer with a Lucene index.
+ */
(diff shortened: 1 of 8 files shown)


jena-text/src/test/java/org/apache/jena/query/text/TestDatasetWithAnalyzingQueryParser.java
```diff
@@ -0,0 +1,64 @@
...
+                "."
+                );
+        String queryString = StrUtils.strjoinNL(
+                QUERY_PROLOG,
+                "SELECT ?s",
+                "WHERE {",
+                "    ?s text:query ( rdfs:label 'édu*' 10 ) .",
+                "}"
+                );
+        Set<String> expectedURIs = Sets.newHashSet(RESOURCE_BASE + testName);
```
ajs6f: For a single element like this, I'd prefer `Collectons::singleton`, but that's just me.

osma: This was copied from another, similar unit test (TestDatasetWithConfigurableAnalyzer.java). I could change both, of course, but I doubt this makes a big difference for readability.

ajs6f: Yes, in that case, definitely more important to keep things consistent.

osma:
Any other comments? I will merge this some time next week unless someone objects.

afs:
Looks good to me.

osma:
Merged after rebasing.

joelkuiper:
👍 I kinda need ComplexPhraseQueryParser for some search, would it be good to add that too?

osma:
@joelkuiper Please create a new JIRA ticket for ComplexPhraseQueryParser support on issues.apache.org, as this issue/PR is already closed.

It shouldn't be very hard to add support for ComplexPhraseQueryParser because the framework is already done - it's basically adding an extra case to `TextIndexLucene.getQueryParser()` along with a unit test that can be pretty much copied and pasted from `TestDatasetWithAnalyzingQueryParser`, except changing the actual test to something which only ComplexPhraseQueryParser can do. Pull requests also very welcome ;)

PR #132, opened 2016-04-01
JENA-967: afn:Sprintf implementation

ales004:
Hi,
I implemented the afn:sprintf function. In order to implement it, I had to:

- create a sprintf.java class in the sparql function part.

- implement in the XSDFuncOp.java a function dealing with making sprintf

- create a series of tests checking that the function is working correctly.

While implementing, I realized that the example described in the bug is just one of the possibilities (printing numbers) as by the Java specification, String.Format works with all types. I thus implemented the function to work with all the type that made sense in my opinion (we can discuss about this).

There is still one problem that I am not sure how to solve:
Test 04 and 07 and 08 depend on the user locale (it will be printed as 1.2.. or 1,2.. or with date the day or month depend on the locale). In order to solve this, we will need to show to String.Format the current locale but I could not find a way to do this from the XSDFuncOp class.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/expr/nodevalue/XSDFuncOp.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/expr/nodevalue/XSDFuncOp.java
@@ -352,6 +352,60 @@ public class XSDFuncOp
         }
     }
 
+    // expecting nvString = format | nvStart = value (int,float, string,....)
+    public static NodeValue javaSprintf(NodeValue nvFormat, NodeValue nvValue) {
+        try {
+            String formatForOutput = nvFormat.getString() ;
+            ValueSpaceClassification vlSpClass = nvValue.getValueSpace();
+            switch(vlSpClass){
+                case VSPACE_NUM:
+                    NumericType type = classifyNumeric("javaSprintf",nvValue);
+                    switch(type) {
+                        case OP_DECIMAL:
+                            BigDecimal decimalValue = nvValue.getDecimal();
+                            return NodeValue.makeString(String.format(formatForOutput,decimalValue)) ;
+                        case OP_INTEGER:
+                            BigInteger integerValue = nvValue.getInteger();
+                            return NodeValue.makeString(String.format(formatForOutput,integerValue)) ;
+                        case OP_DOUBLE:
+                            Double doubValue = nvValue.getDouble();
+                            return NodeValue.makeString(String.format(formatForOutput,doubValue)) ;
+                        case OP_FLOAT:
+                            Float floatValue = nvValue.getFloat();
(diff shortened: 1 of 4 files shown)


afs:
Hi - looks good 

One suggestion : Java printf can take a variable number of arguments from zero upwards.

It would be nice if `afn:sprintf("%d != %d", 1, 2)` worked.  Could the arguments (args 1 on) go into a list, and javaSprintf proces them into a a Object[] for String.format?

afs:
>  These cases for the moment are not supported. Maybe we could treat them all like strings.

Yes - `NodeFunction.str(NodeValue)` can be used to get the lexical form.

ales004:
Yes I think that the idea about the list is very good. I will try to implement it like this.

I am still wondering what to do with the String.Format for Double, Float and Datetime that depend on the Locale. What do you think about this?

jena-arq/src/main/java/org/apache/jena/sparql/expr/nodevalue/XSDFuncOp.java
```diff
@@ -352,6 +352,60 @@ public static NodeValue javaSubstring(NodeValue nvString, NodeValue nvStart, Nod
...
+                            return NodeValue.makeString(String.format(formatForOutput,decimalValue)) ;
+                        case OP_INTEGER:
+                            BigInteger integerValue = nvValue.getInteger();
+                            return NodeValue.makeString(String.format(formatForOutput,integerValue)) ;
+                        case OP_DOUBLE:
+                            Double doubValue = nvValue.getDouble();
+                            return NodeValue.makeString(String.format(formatForOutput,doubValue)) ;
+                        case OP_FLOAT:
+                            Float floatValue = nvValue.getFloat();
+                            return NodeValue.makeString(String.format(formatForOutput,floatValue)) ;
```
ales004: Depending on the Locale: 1.23 can be printed as 1.23 or 1,23

afs:
For now, I think that "local locale" is the best single choice because if the user wants the XSD standard form "1.23" they can workaround it with a call `str()` and print a string. Not perfect but gives them a way.

XSD F&O v3.1 has "picture strings" : https://www.w3.org/TR/xpath-functions-3/#dt-picture-string : that looks like Java DecimalFormat, not printf style.

I think most people are familiar with printf and things like padding strings and alignment are covered. We can refine the capabilities after the machinery is in.

ales004:
I think that commit bf14c982b8ae3eb8f040f55c28f8262efb5dad01 should do what you asked:

- now also tests are Locale dependent.
- sprintf accepts any number of arguments.
- unkwnown classes are translated to string using the function you pointed out.

What do you think?

afs:
Looks good and I've merged it into master. Thanks!

(I've added t the documentation for the function library.)

PR #133, opened 2016-04-02
Support for try-with-resources Statements in ClosableIterator and Models

CesarPantoja:
Implementation of the interface AutoCloseable in ClosableIterator and Model. This would allow to use Java's [try-with resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) Statements, i.e.:

``` Java
        // some definitions
        String personURI = "http://somewhere/JohnSmith";
        String fullName = "John Smith";

        // create an empty Model
        try (Model model = ModelFactory.createDefaultModel()) {

            // create the resource
            Resource johnSmith = model.createResource(personURI);

            // add the property
            johnSmith.addProperty(VCARD.FN, fullName);
        }
```

or

``` Java
            // list the statements in the Model
            try (StmtIterator iter = model.listStatements()) {

                // print out the predicate, subject and object of each statement
                while (iter.hasNext()) {
                    Statement stmt = iter.nextStatement();  // get next statement
                    Resource subject = stmt.getSubject();     // get the subject
                    Property predicate = stmt.getPredicate();   // get the predicate
                    RDFNode object = stmt.getObject();      // get the object

                    System.out.print(subject.toString());
                    System.out.print(" " + predicate.toString() + " ");
                    if (object instanceof Resource) {
                        System.out.print(object.toString());
                    } else {
                        // object is a literal
                        System.out.print(" \"" + object.toString() + "\"");
                    }

                    System.out.println(" .");
                }
            }
```


ajs6f:
The semantics of `AutoCloseable#close` are "Closes this resource, relinquishing any underlying resources." The semantics of `Model#close` are "…_free_ up resources held" (my emphasis). I'm not sure if "relinquish" and "free" are the same here. You can "relinquish" a resource without everything in the system relinquishing it, which is what "free" seems to me to mean. And then, a `Model` can be a view over some underlying resource, like a `Dataset` or even some kind of remote resource. That might make things a little more complicated. But I'm not very subtle with Jena's core abstractions, so my qualms here may be way off base.

afs:
For `Model`: it would be unusual (mainly in small examples!) for a model to be created and closed in the way that scoping for try-with resources tends to emphasis.  Now that model may may be stateless views of datasets, the notion of "close" is not so clear cut. It certainly does not close the dataset. That is what I think is being captured in "Not all implementations of Model require this method to be called."

For `CloseableIterator` is the root of lots of iterator types in Jena, not just StmtIterator and the otehr API-related iterators. Looks to me if the javadoc is expressing situations that are not happening in practice anymore.  There was, a long time ago, a BDB backed storage (C-BDB) and that needed iterator closed so DB cursors were released.

Iterators like `ExtendedIterator<Triple>` get combined, concatenated, filtered, and mapped and don't fit the `try` block style so well.  I added `AutoCloseable` and got about 700 warnings from the codebase.

So the question must be what impact on everyone's existing applications will there be?  A whole bunch of warnings turning up has to be considered in deciding here.

Jena supports different APIs over the same core. Maybe creating a more modern API for model/statement/resource is needed.

afs:
This seems like a high-impact change (even though it will result in warnings not errors) for users and the code.

I think we should not make this change.

The option of new API over the same core is worthwhile pursuing.

ajs6f:
Would that (new API) look like a set of wrappers? Or perhaps a method on `Model` that reflects the instance into an `AutoCloseable` subtype of `Model`?

afs:
I have no idea! Maybe it's a revised current API, maybe it's influenced by different design patterns. 
Any API is more likely to work over Graph/Triple/Node than add another layer on Model.

rwm:
I would strongly vote for adding the AutoCloseable interface, at least for StmtIterator. The sole purpose of AutoCloseable is to allow developers to use the `try-with-resources` statement and it does not cause any side effects. When it was introduced to Java, the interface was added to all objects with a `close()` method.

The javadoc for AutoCloseable says:

> An object that may hold resources (such as file or socket handles) until it is closed. The close() method of an AutoCloseable object is called automatically when exiting a try-with-resources block for which the object has been declared in the resource specification header. This construction ensures prompt release, avoiding resource exhaustion exceptions and errors that may otherwise occur.
>API Note:
>    It is possible, and in fact common, for a base class to implement AutoCloseable even though not all of its subclasses or instances will hold releasable resources. For code that must operate in complete generality, or when it is known that the AutoCloseable instance requires resource release, it is recommended to use try-with-resources constructions. However, when using facilities such as Stream that support both I/O-based and non-I/O-based forms, try-with-resources blocks are in general unnecessary when using non-I/O-based forms.


@afs do you have a list of the generated warnings and maybe could attach them in a file so we can review them?

In my experience, the warnings are mostly useful hints about missing close statements. You can disable them via `@SuppressWarnings("resource")` or by compiling with `-source 7`.

See the following example:
```
	private class C implements AutoCloseable{
		@Override
		public void close() {
		}
	}
	public void a(){
		C c = new C();
		// warning! never closed
		Objects.requireNonNull(c);
	}
	public void b(){
		C c = new C();
		Objects.requireNonNull(c);
		c.close();
		// no warning, the instance is closed
	}
	public C c(){
		C c = new C();
		Objects.requireNonNull(c);
		// no warning, the instance is returned
		return c;
	}
	public void d(){
		C c = new C();
		if( new Random().nextFloat() < 0.2 ){
			// warning! not closed at this location
			return;
		}
		c.close();
	}
```

The AutoCloseable interface should be used on any class with a `close()` method, which may need to be called. Otherwise developers are forced to use Java 6 code with manual exception handling for resource.

ajs6f:
I think we're all probably familiar with he semantics of `AutoCloseable`. Examine the API note: for "... facilities... that support ... I/O-based and non-I/O-based forms, try-with-resources blocks are in general unnecessary when using non-I/O-based forms." That would include any iterator drawing from in-memory constructs, a huge proportion of Jena usage.

Running through a list of warnings together is not a good use of anyone's time. Let me suggest an alternative approach. Fork Jena's codebase, make a branch with the change you desire to see, and correct the compiler warnings in your branch. Then send Jena a PR and we can work from there.

afs:
@rwm : The impact is not solely on Jena codebase but also on all the application code written to the Jena API.

ajs6f:
We've not yet received a PR along these lines that takes into account all the fresh warnings that would be generated, and in the meantime efforts to create new APIs have proceeded apace. @CesarPantoja, would you like to work more on this or would you like to close it?

CesarPantoja:
Hi! Sorry, I had been busy. I'll take a look at the warnings and possibly create a PR in the coming days.

afs:
The key point is that warnings are in user code.

If the consequences are large numbers of warnings, we will have created a support problem when in fact the user code works perfectly well.

ajs6f:
One problem is that "try-with-resources blocks are in general unnecessary when using non-I/O-based forms", and there is no distinction in Jena's type system between `Model`s that are disk-based and those that aren't (and I'm not saying there should be).

ajs6f:
@CesarPantoja Have you had a chance to take the warnings mentioned above into account?

CesarPantoja:
Sorry, I've been quite busy lately :disappointed:  Just finished my PhD a couple of weeks ago. (And Jena had a big role in it :smile:.
I'll try to get around it this weekend.

ajs6f:
NP, @CesarPantoja. If your work concerned Jena, please feel free to let the list know about it. It's good to share good news.

afs:
I am not in favour of this change even if the internal Jena warnings are fixed.

The impact is not solely on Jena codebase but also on all the application code written to the Jena API and the knock-on effect on the support list.

try-with-resource is a good way to handle external resources, like files, and also where a mandatory final operation is necessary.

"Model" does not fit that description.

"ExtendedIterator" is widespread in the public API and Jena has not been enforcing closing for a very long time.

A lot of application code will not be written to the try-with-resource style nor be structured to easily migrate. It only takes returning an iterator from a method to mean the code can not be easily updated.

ajs6f:
I'm certainly not particularly in favor of it as much as I don't like to have PRs hanging. :)

@CesarPantoja, couldn't you just use a wrapper class like
```
class AutoCloseableModel extends Model implements AutoCloseable {}
```
if you like to use `try-with-resource` in this way?

CesarPantoja:
@ajs6f Yes, I think that's a good solution. Closing this PR.

PR #134, opened 2016-04-04
Fix Jena initialization in OSGi environments

jpullmann:
This is a suggestion to fix errors in static initialization of Jena 3 in OSGi environments folowing the OSGi ServiceLoader Mediator spec. The appropriate manifest headers and a bundle activator were added: 

 http://mail-archives.apache.org/mod_mbox/jena-users/201603.mbox/%3C2804-56ee5500-b-5266a380%40181641092%3E

The original integration test (jena-osgi-test) was ineffective. Missing a runtime configuration the test methods did not run/never failed on my side. It works for Apache Karaf now. 

Please install the uploaded, OSGi-fied dependency com.github.andrewoma.dexx:collection: 

mvn install:install-file -Dfile=collection-0.4.0.jar -DgroupId=com.github.andrewoma.dexx -DartifactId=collection -Dversion=0.4.0 -Dpackaging=jar

and run the integration test: mvn -Drat.ignoreErrors=true clean integration-test

If needed, I'll extend the test to cover other OSGi environments (Equinox etc.)

[collection-0.4.0.zip](https://github.com/apache/jena/files/202677/collection-0.4.0.zip)

### DIFF
--- a/apache-jena-osgi/jena-osgi/src/main/java/org/apache/jena/osgi/Activator.java
+++ b/apache-jena-osgi/jena-osgi/src/main/java/org/apache/jena/osgi/Activator.java
@@ -1,3 +1,21 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.apache.jena.osgi;
 
 import org.apache.jena.system.JenaSubsystemRegistry;
@@ -6,28 +24,11 @@ import org.apache.jena.system.JenaSystem;
 import org.osgi.framework.BundleActivator;
 import org.osgi.framework.BundleContext;
 
+// Sample bundle activator explicitly calling the static Jena initialization.
 public class Activator implements BundleActivator {
-	
-	/* 
-	 * Based on http://svn.apache.org/repos/asf/aries/trunk/spi-fly/spi-fly-examples/spi-fly-example-provider-consumer-bundle/src/main/java/org/apache/aries/spifly/pc/bundle/Activator.java
-	 * the Activator#start() waits for bundle extension by Aries SPI Fly, configures JenaSystem logging and requests for initialization.     
-	 */
+
 	public void start(BundleContext context) throws Exception {
 
-		Thread t = new Thread(new Runnable() {
-			@Override
-			public void run() {
-				try {
-					Thread.sleep(500);
-				} catch (InterruptedException e) {
-				}				
-				setUpJena();
-			}
-		});
-		t.start();
-	}
-
-	private void setUpJena() {
 		JenaSubsystemRegistry r = new JenaSubsystemRegistryBasic() {
 			@Override
 			public void load() {
@@ -38,7 +39,6 @@ public class Activator implements BundleActivator {
 		JenaSystem.DEBUG_INIT = true;
 		JenaSystem.init();
 		System.out.println("Jena-OSGi bundle configuration done!");
-		
 	}
 
 	public void stop(BundleContext context) throws Exception {
--- a/apache-jena-osgi/jena-osgi/src/main/java/org/apache/jena/osgi/Activator.java
+++ b/apache-jena-osgi/jena-osgi/src/main/java/org/apache/jena/osgi/Activator.java
@@ -1,0 +1,47 @@
+package org.apache.jena.osgi;
+
+import org.apache.jena.system.JenaSubsystemRegistry;
+import org.apache.jena.system.JenaSubsystemRegistryBasic;
+import org.apache.jena.system.JenaSystem;
+import org.osgi.framework.BundleActivator;
+import org.osgi.framework.BundleContext;
+
+public class Activator implements BundleActivator {
+	
+	/* 
+	 * Based on http://svn.apache.org/repos/asf/aries/trunk/spi-fly/spi-fly-examples/spi-fly-example-provider-consumer-bundle/src/main/java/org/apache/aries/spifly/pc/bundle/Activator.java
+	 * the Activator#start() waits for bundle extension by Aries SPI Fly, configures JenaSystem logging and requests for initialization.     
+	 */
+	public void start(BundleContext context) throws Exception {
+
+		Thread t = new Thread(new Runnable() {
+			@Override
+			public void run() {
+				try {
+					Thread.sleep(500);
+				} catch (InterruptedException e) {
+				}				
+				setUpJena();
+			}
+		});
+		t.start();
+	}
+
+	private void setUpJena() {
+		JenaSubsystemRegistry r = new JenaSubsystemRegistryBasic() {
+			@Override
+			public void load() {
+				super.load();
+			}
+		};
+		JenaSystem.setSubsystemRegistry(r);
+		JenaSystem.DEBUG_INIT = true;
+		JenaSystem.init();
+		System.out.println("Jena-OSGi bundle configuration done!");
+		
+	}
+
+	public void stop(BundleContext context) throws Exception {
+
+	}
+}
--- a/apache-jena-osgi/jena-osgi-features/pom.xml
+++ b/apache-jena-osgi/jena-osgi-features/pom.xml
@@ -1,0 +1,102 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+   Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed with
+   this work for additional information regarding copyright ownership.
+   The ASF licenses this file to You under the Apache License, Version 2.0
+   (the "License"); you may not use this file except in compliance with
+   the License.  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.apache.jena</groupId>
+		<artifactId>apache-jena-osgi</artifactId>
+		<version>3.1.0-SNAPSHOT</version>
+	</parent>
+	<artifactId>jena-osgi-features</artifactId>  
+	<packaging>bundle</packaging>
+	<name>Apache Jena - OSGi Karaf features</name>	
+	<description>Apache Karaf features (deplyoment configurations) for easy installation and integration testing.</description>
+	<build>
+		<pluginManagement>
+            <plugins>
+            	<plugin>
+                	<groupId>org.codehaus.mojo</groupId>
+                	<artifactId>build-helper-maven-plugin</artifactId>
+                	<version>1.8</version>
+            	</plugin>
+            </plugins>
+        </pluginManagement>
+        <resources>
+            <resource>
+                <directory>src/main/resources</directory>
+                <filtering>true</filtering>
+            </resource>
+        </resources>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-resources-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>copy-resources</id>
+                        <phase>validate</phase>
+                        <goals>
+                            <goal>copy-resources</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>${basedir}/target</outputDirectory>
+                            <resources>
+                                <resource>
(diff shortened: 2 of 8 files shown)


ajs6f:
I will try and contact the maintainers of Dexx Collections to see if we can get a properly OSGI-ified release there. Also, they appear to have released a new version 0.5, so I think we will want at some point to move along with them.

ajs6f:
Track the Dexx release at: https://github.com/andrewoma/dexx/issues/5 / https://issues.apache.org/jira/browse/JENA-1160

ajs6f:
To create the OSGi bundle for Dexx, did you use any tooling, or just insert a manifest manually? I'd like to send them a PR (or of course, you can :) ).

afs:
(again - non-OSGi user question) Why do we need a customized version of dexx?  From a practical point of view, using the formal releases of another project is less work.

ajs6f:
OSGi requires some [metadata](https://en.wikipedia.org/wiki/OSGi#Bundles) to work, found in the `MANIFEST.MF` in `META-INF`. Most projects have that auto-created at packaging via tooling like Maven. If it's not there, it has to be added, as @jpullmann has done here. As I wrote [here](https://github.com/apache/jena/pull/134#issuecomment-205975853), we can (and should) try to get that upstream, to avoid exactly the special customization to which you refer here.

afs:
I don't understand why dexx-collections is different.

v0.2 has a MANIFEST.MF as does the v0.5 (but it's rather minimal)

Are you saying all our dependencies have that metadata except this one case?

ajs6f:
No, it's not the _presence_ of MAINFEST.MF (which I think tends to appear in pretty much any JAR or WAR). It's what's in it. There are specific headers in an OSGi version of MANIFEST.MF that support the sophisticated module-dependency management that OSGi does. For example, there are headers that explain what packages are exported from the module in question, and which are imported. If you don't need them (because you aren't using OSGi), no problem. You just ignore them, and most of us do, most of the time. But OSGi needs those headers to use the JAR as a module. As to whether all of our dependencies have it, I don't know, but it honestly wouldn't surprise me. It's so easy to include these days (autogenerated) that many, many projects do it by default. On the other hand, since `jena-osgi` is packaging up some dependencies internally, than it doesn't matter whether those guys do, because OSGi is not managing those guys. And those dependencies that `jena-osgi` is not packaging up internally certainly _do_ have the metadata (e.g. `httpcore-osgi`).
So one thing we could do here would be to have `jena-osgi` shade in the Dexx library the way it currently shades in, e.g. `xercesImpl`. But that's kind of hacky. I think the right thing to do here is work upstream and help Dexx get that little bit of metadata packed in, and then the job is done cleanly and forever.

ajs6f:
I've sent an [issue](https://github.com/andrewoma/dexx/issues/6) and [PR](https://github.com/andrewoma/dexx/pull/7) to Dexx Collections to get the OSGi metadata in there without our fancy footwork with moving Maven artifacts about.

ajs6f:
Dexx Collections has [merged](https://github.com/andrewoma/dexx/commit/7d0f242e64679b4f3fd4a5ee0598d29852df49ce) my PR to give OSGi metadata to their Java artifact. They are willing to make a Maven Central release, but they would like to make sure that the new OSGi metadata is good for us. @jpullmann, can you pull a fresh copy of `master` from Dexx, build it (it needs Gradle) and check it with this Jena PR? If that's annoying for you, just let me know. I can build you a Dexx artifact OSGi-fied according to my PR locally and get it to you.

jpullmann:
Sorry for being absent yesterday (all-day meeting)! The answer regarding OSGi given by @ajs6f is exhaustive. OSGi requires jars to explictily state their code (packages) and service dependencies (components) to become manageable by the framework. The former uses MANIFEST.MF headers. These were missing in Dexx collection (plain jar):

`Manifest-Version: 1.0`

 and I added them via creating a "Plug-in from Existing JAR Archives" project in Eclipse:

```
Manifest-Version: 1.0
Bundle-ManifestVersion: 2
Bundle-Name: Dexx Collection
Bundle-SymbolicName: com.github.andrewoma.dexx.collection
Bundle-Version: 0.4.0
Bundle-Vendor: https://github.com/andrewoma/dexx
Export-Package: com.github.andrewoma.dexx.collection
```
Some "internal" packages are apparently not intended for sharing, I removed them from the list of 
exports now. As suggested I'll build and test the Dexx collection bundle..

jpullmann:
Dexx collection (0.6.0-SNAPSHOT) works fine!

ajs6f:
Marvelous! I'll circle back with them and with any luck, we'll get a release eftsoons with the new metadata.

apache-jena-osgi/jena-osgi/src/main/java/org/apache/jena/osgi/Activator.java
```diff
@@ -0,0 +1,47 @@
...
+		t.start();
+	}
+
+	private void setUpJena() {
+		JenaSubsystemRegistry r = new JenaSubsystemRegistryBasic() {
+			@Override
+			public void load() {
+				super.load();
+			}
+		};
```
jpullmann: I did not consider changing Jena initialisation, but this exactly hits the point. Now, with 
JenaSubsystemRegistryBasic adapted: 

```
@Override
    public void load() {
        synchronized (registryLock) {
            // Find subsystems asking for initialization. 
            ServiceLoader<JenaSubsystemLifecycle> sl = ServiceLoader.load(JenaSubsystemLifecycle.class,this.getClass().getClassLoader()) ;
            sl.forEach(this::add) ;
        }
    }
```
the bundle's own class loader is used for SPI resloution. Great, it works out of the box!
I removed Aries SPI Fly dependency and headers from jena-osgi - they are not needed any more. 

The only remaining issue is embedding xerces. This is not recommended and yields to an error in the test runtime Apache Karaf 3.0.6 (with xerces exposed to all bundles via boot classloader): 

```
org.apache.xerces.impl.dv.DVFactoryException: Schema factory class org.apache.xerces.impl.dv.xs.SchemaDVFactoryImpl does not extend from SchemaDVFactory.
        at org.apache.xerces.impl.dv.SchemaDVFactory.getInstance(Unknown Source)
```
Suggestion: use "Import-Package" directive to share available org.apache.xerces.* packages.
I updated jena-osgi/pom.xml accordingly.

The integration test is rather rough and definitely needs elaboration (will do next week).

ajs6f: I agree that shading in `xercesImpl` is kind of… _shady_. (Hah!) But it would be worth making some effort to make plain to folks who have relied on it that they are going to have to make sure that it's available in their OSGi framework/container, just we don't have a bunch of people finding the same bundle resolution problem over and over.

jpullmann: The OSGi users are used to this process of a painfull dependcy resolution, luckily this is due only once at the beginning. While the package dependencies are obvoius from the manifest, we should support the resolution of the correct (OSGi-aware) Maven archives. This could be done by specifying a provisioning configuration, e.g. an [Apache Karaf Fature](https://karaf.apache.org/manual/latest/users-guide/provisioning.html) or [Equinox/Tycho](https://wiki.eclipse.org/Tycho/Reference_Card) feature. I'll need some time to prepare them and to set-up integration tests for these platforms (but this is a prerequisite to prove the portability).

ajs6f:
Dexx Collections has released a `0.6` with OSGi metadata, and I've updated https://github.com/apache/jena/pull/135 to refer to it, so if we order that PR before this one and rebase this one, I think we can entirely elide the special handling for Dexx Collections, right?

apache-jena-osgi/jena-osgi/src/main/java/org/apache/jena/osgi/Activator.java
ajs6f: Sure, I've been down enough roads with the package fandango to know it's just part of the journey. My suggestion is just towards documentation, because we would be making a change that would require action on the part of downstream folks. I think it would be great to offer provisioning artifacts, although I'm not sure they belong as a part of Jena itself (I'll shut up and let @afs and other committers speak to that, I'm just an interested contributor). If that does become part of the picture, inside Jena or next to it, there is previous work available in connection with Fedora Commons (for which I am a committer) that could be useful (Karaf features, actually).

jpullmann:
Right, the released Dexx collection 0.6 is fine and safe to use.

afs:
@ajs6f 's plan looks good to me.  #135 applied.

Changing `JenaSubsystemRegistryBasic` looks fine.

If this PR is rebased, I'll apply it.

afs:
I have applied this fix and pushed the result to branch "osgi-fix-init"

It builds with `mvn clean install` and because this fixes something that is otherwise broken, it seems better to me to apply it to master in the current state and then continue refining it if need be.

There are some changes and comment POM lines that it would be good to sort out:

I'd like help with:
* checking it works in an OSGi environment.
* assessing `jena-osgi-test`
* Can we remove the `<Require-Capability>`? The `JenaSubsystemRegistryBasic` has been changed.
* `<Embed-Dependency>` and `<Import-Package>` for Xerces.  Jena uses internal features of Xerces and is sentisite to the version (OK - it rarely changes these days). Also, Xerces can be a endorsed jar making version management by the app not work. Is it necessary to `Import` it?

The branch does not pass the tests in `jena-osgi-test`.  Note that `jena-osgi-test` is not in the build because of [JENA-913](https://issues.apache.org/jira/browse/JENA-913) and [JENA-1037](https://issues.apache.org/jira/browse/JENA-1037) so the test failures are probably unrelated.

For the record: steps take to integrate the diff:
* Add license header to `Activator.java`
* De-tabbed the pom files and sorted out trailing white space.
* Some formatting "bit rot" had set in.

jpullmann:
Thank you. I'll do the testing tomorrow, here my quick response:
Neither Require-Capability, nor SPI-* headers are needed any more. Same for Bundle activator, but I'll leave it for extension purposes.
In the former approach using SPI Fly there were issues with inline Xerces. Using an embedded version makes it more selfcontained, alernatively it may require import of its particular Xerces version.

afs:
Thanks for the quick response.  Things are looking good for this.

As Jena calls into the details of Xerces, not just the XML APIs, even a particular version can cause problems.

apache-jena-osgi/jena-osgi/src/main/java/org/apache/jena/osgi/Activator.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class Activator implements BundleActivator {
+	
+	/* 
+	 * Based on http://svn.apache.org/repos/asf/aries/trunk/spi-fly/spi-fly-examples/spi-fly-example-provider-consumer-bundle/src/main/java/org/apache/aries/spifly/pc/bundle/Activator.java
+	 * the Activator#start() waits for bundle extension by Aries SPI Fly, configures JenaSystem logging and requests for initialization.     
+	 */
+	public void start(BundleContext context) throws Exception {
+
+		Thread t = new Thread(new Runnable() {
```
ajs6f: Since we are in Java 8-land, this might read a bit more crisply with a lambda:
```
new Thread(() -> { 
    try { Thread.sleep(500); }
        catch (InterruptedException e) { }				
    setUpJena();
}).start();
```

afs: While we're in the area - why the thread and sleep 500 at all now that `ServiceLoader.load` passes in the class of `this`?

jpullmann: The asynchronous initialisation was a tweak to wait for SPI Fly to enhance the bundle and wire the service providers.  This is obsolate now, since JenaSubsystemRegistryBasic passes the bundle's private classloader to ServiceLoader for their resolution. The current code in bundle Activator now simply calls the default initialisation (more an example).

jpullmann:
I created a new folder containing the initial Apache Karaf feature definition (jena-osgi-features). We might think of providing different distributions in the future each one defined by own feature. It has been reused to define the bundle set for the integration test:
```
              ... features(maven().groupId("org.apache.jena")
			.artifactId("jena-osgi-features").type("xml")
			.classifier("features")
			.version("3.1.0-SNAPSHOT"), "jena") ...
```
While testing I was reminded harshly why I removed xerces from the bundle earlier: Apache Karaf (3* and 4*) ships with endorsed xerces 2.11.0 and exports it per default to any installed bundle via boot delegation. The bundle cant's suppress (not import) these classes, resulting in 2 versions available and tests failing with: 
```
Caused by: java.lang.VerifyError: (class: org/apache/xerces/impl/dv/xs/AbstractDateTimeDV, method: <clinit> signature: ()V) Bad type in putfield/putstatic
	at org.apache.xerces.impl.dv.xs.XSSimpleTypeDecl.<clinit>(Unknown Source)[:]
	at org.apache.xerces.impl.dv.xs.SchemaDVFactoryImpl.createBuiltInTypes(Unknown Source)[:]
	at org.apache.xerces.impl.dv.xs.SchemaDVFactoryImpl.<clinit>(Unknown Source)[:]
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)[:1.8.0_05]
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)[:1.8.0_05]
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)[:1.8.0_05]
	at java.lang.reflect.Constructor.newInstance(Constructor.java:408)[:1.8.0_05]
	at java.lang.Class.newInstance(Class.java:433)[:1.8.0_05]
	at org.apache.xerces.impl.dv.ObjectFactory.newInstance(Unknown Source)[:]

``` 
As expected, these configurations were proved to work:
1. Apache Karaf * with endorsed xerces-2.11.0 + plain Jena-osgi (without xerces)
2. Apache Karaf 4.0.4 without xerces-2.11.0 + Jena-osgi with inline xerces
3. Apache Karaf 4.0.4 without xerces-2.11.0 + plain Jena-osgi with "xerces" feature installed

In case 2,3 I removed "xerces-2.11.0.jar" from "lib/endorsed/" and from the configuration of packages implicitly provided by runtime (org.osgi.framework.bootdelegation) in  "etc/config.properties".
Case 3 seems to confirm, that importing xerces might work. Users may either install the "xerces" feature or resolve the indicated bundles manually:
```
<feature name="xerces" version="2.11.0">
<bundle>mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.xerces/2.11.0_1</bundle><bundle>mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.xmlresolver/1.2_5</bundle>
</feature>
```

afs:
Thank you!

Merged into master and a development deployment build done.

osgi-tests are disabled because they do not work from a preflight build  ("mvn clean verify"), only from "mvn clean install").  Preflight is used with a completely clean maven repo when preparing the releasable version.

PR #135, opened 2016-04-06
JENA-1160: Upgrade Dexx Collections version

ajs6f:


ajs6f:
This PR would pull in Kotlin's runtime to the dependencies, because Dexx Collections has moved to Kotlin for current versions. Obviously, that warrants discussion, minimally, and perhaps it's a blocker. Maybe I should start looking for an alternative persistent data structures library? Or perhaps it's even time to think about a simple purpose-built bit of code for the two applications we actually have for persistent structures right now. (`Map<Node, T>` and `Set<Node>`)

ajs6f:
Oops, disregard previous comment: I misunderstood some advice from the Dexx Collection maintainers and upgraded somehow to a Kotlin version of the library instead of the Java version. Interesting to note that it is indeed JVM compatible. But the issue of Kotlin's runtime is not an issue at all, sorry for the confusion.

PR #136, opened 2016-04-19
Unset prefixes

afs:
JENA-1165 : Support for handlign undefined prefixes, especially in command line tools.

The default setting everywhere is "off".

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/system/RiotLib.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/system/RiotLib.java
@@ -29,6 +29,8 @@ import static org.apache.jena.riot.writer.WriterConst.rdfNS ;
 import java.io.OutputStream ;
 import java.io.Writer ;
 import java.util.* ;
+import java.util.function.Function ;
+import java.util.function.Predicate ;
 
 import org.apache.jena.atlas.io.IndentedWriter ;
 import org.apache.jena.atlas.iterator.Iter ;
@@ -38,7 +40,10 @@ import org.apache.jena.graph.Node ;
 import org.apache.jena.graph.NodeFactory ;
 import org.apache.jena.graph.Triple ;
 import org.apache.jena.query.ARQ ;
-import org.apache.jena.riot.* ;
+import org.apache.jena.riot.Lang ;
+import org.apache.jena.riot.RDFLanguages ;
+import org.apache.jena.riot.SysRIOT ;
+import org.apache.jena.riot.WriterDatasetRIOT ;
 import org.apache.jena.riot.lang.LabelToNode ;
 import org.apache.jena.riot.tokens.Token ;
 import org.apache.jena.riot.tokens.Tokenizer ;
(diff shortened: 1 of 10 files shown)


jena-arq/src/main/java/org/apache/jena/riot/system/RiotLib.java
```diff
@@ -73,12 +78,37 @@ public static Node createIRIorBNode(String iri)
...
-    public static boolean isBNodeIRI(String iri)
-    {
+    /** Test whether a IRI is a ARQ-encoded blank node. */
+    public static boolean isBNodeIRI(String iri) {
         return skolomizedBNodes && iri.startsWith(bNodeLabelStart) ;
     }
     
+    // These two must be in-step.
+    /** Function applied to undefined prefixes to convert to a URI string */  
+    public static final Function<String,String> fixupPrefixes      = (x) -> "::"+x ;
```
ajs6f: This could be
```
private static final String URI_SKELETON = "::";
...
public static final Function<String,String> fixupPrefixes = URI_SKELETON::concat;
```
I find that clearer, because it names the string, but then, my tastes are odd.

jena-arq/src/main/java/org/apache/jena/riot/system/RiotLib.java
```diff
@@ -73,12 +78,37 @@ public static Node createIRIorBNode(String iri)
...
+
+    /** Function to test for undefined prefix URIs*/  
+    public static final Predicate<String> testFixupedPrefixURI     = (x) -> x.startsWith("::") ;
+    
+    /** Test whether a IRI is a ARQ-encoded blank node. */
+    public static boolean isPrefixIRI(String iri) {
+        return testFixupedPrefixURI.test(iri) ;
```
ajs6f: Again, just a style thing, maybe personal, but I would rather read:
```
/** Function to test for undefined prefix URIs*/  
    public static final Predicate<String> testFixupedPrefixURI  = RiotLib::isPrefixIRI ;
    
/** Test whether a IRI is a ARQ-encoded blank node. */
    public static boolean isPrefixIRI(String iri) {
        return iri.startsWith("::") ;
    }
```
or even just factor out `testFixupedPrefixURI` for the method reference `RiotLib::isPrefixIRI`.

jena-arq/src/main/java/org/apache/jena/riot/system/RiotLib.java
ajs6f: I'm probably missing something here, but it seems like this is only ever used unwrapped, via `fixupPrefixes.apply`, and that's down below in this class (where it is used instead of using `fixupPrefixIRI(String)`, but I'm not sure why). Maybe it would be clearer to have an ordinary method and in future, if a function is needed, to use a method reference?

jena-arq/src/main/java/org/apache/jena/riot/system/RiotLib.java
afs: The idea is to leave the possibility of making it configurable.  `// These two must be in-step.`

At the moment, they are final, but the structure is in place for different especially in dev (when tweaking the final is possible).

ajs6f: Ah, so you want to keep a "handle" on the test, to possibly swap in different tests, hence the `Predicate` is primary? That makes sense, cool.

PR #137, opened 2016-05-02
JENA-1172: restore support for blank nodes in jena-text

osma:
It seems that jena-text used to support blank nodes in the text index (there is some infrastructure to support this, particularly in TextQueryFuncs) but this was broken by JENA-999 (internal caching) or possibly even earlier. There were no unit tests involving blank nodes that would have caught this.

This PR restores support for blank node subjects. There are a couple of very basic unit tests that check the basic functionality.


osma:
I made this into a PR rather than merging directly because I thought I may have missed something obvious. But if it looks good to others, I will merge it.

afs:
Looks good to me.

PR #139, opened 2016-05-06
JSON-LD output

fpservant:
Hi,

here's my attempt at improving user's control over JSON-LD output and therefore, my request for comments.

Following discussions on jena-users list : [[1]](http://users.jena.apache.narkive.com/NF0pn3kq/controlling-json-ld-output) and [[2]](https://mail-archives.apache.org/mod_mbox/jena-users/201604.mbox/%3c6F2CAA77-7586-4330-8903-16A81C5AB60E@gmail.com%3e)

Mains functionalities:
- possibility to choose between expanded, compacted, flattened and framed output using different RDFFormat variants
- possibility to set the "@context" used to compute the output 
- possibility to substitute the "@context" used to compute the output by another one

Building on existing jena features, simple functionalities are easy to use, or so I hope: you just choose the RDFFormat to choose between expanded, compacted..: as in current jena master, a default context is computed for formats that need one.

Features that require passing an object to the JSON-LD API (a jsonld context or a frame) are built using the jena "Context" mechanism (``org.apache.jena.sparql.util.Context``).

The (jsonld) context and frame objects are passed as JSON strings (the user doesn't have to build JsonLD API objects).

The (jena) context parameter wasn't available in the write methods of ``org.apache.jena.riot.RDFDataMgr``however. I just added ``public static void write(OutputStream out, Model model, RDFFormat serialization, Context ctx)``to it. Should we also do it for the other write methods (write(StringWriter...))? Yes, probably.

TODO some documentation. Where should it be added? The best form would be small sample code. Easy to create from the existing tests.

QUESTIONS
- it is OK to have the Symbols used by the Context in JsonLDWriter? Should I define one subclass of Context instead?
- As there already were a JSONLD_PRETTY and a JSONLD_FLAT RDFFormat, I felt obliged to create 8 different RDFFormats for JSONLD (expanded, compact, etc. * 2 - on pretty and one flat for each, eg. JSONLD_EXPAND_FLAT). Without the 2 pre-existing ones, I would have created only 4, and use the context to pass a flag pretty/flat. But now, passing this flag would add complexity without decreasing much the number of different RDFformats. BTW: should I create "aliases" such as JSONLD_COMPACT, etc.?
- There is one more functionality that I developed, but I didn't include it in here: the possibility to choose to prefer prefixed forms for properties over localname based ones in the output. Not clear? Here is an example of what we get, giving the way the context is computed (when there is none): 

```
{
  "@id" : "_:b0",
  "@type" : "sh:Person",
  "name" : "Jane Doe",
  "url" : "http://www.janedoe.com",
  "@context" : {
    "url" : {
      "@id" : "http://schema.org/url"
    },
    "name" : {
      "@id" : "http://schema.org/name"
    },
    "sh" : "http://schema.org/"
  }
}
```

Note the "@type" : "sh:Person", while we have "name", etc. for props. Person and name are however in the same vocab.

We could choose to prefer to get:

```
{
  "@id" : "_:b0",
  "@type" : "sh:Person",
  "sh:name" : "Jane Doe",
  "sh:url" : "http://www.janedoe.com",
  "@context" : {
    "sh:url" : {
      "@id" : "http://schema.org/url"
    },
    "sh:name" : {
      "@id" : "http://schema.org/name"
    },
    "sh" : "http://schema.org/"
  }
}
```
Now we have "sh:name", etc, in line with the "sh:Person"
What do you think? 

Best Regards,

fps


jena-arq/src/main/java/org/apache/jena/riot/out/JsonLDWriter.java
```diff
@@ -79,26 +130,85 @@ public void write(OutputStream out, DatasetGraph dataset, PrefixMap prefixMap, S
...
-        addPrefixes(ctx, prefixMap) ;
+    private JSONLD_FORMAT getOutputFormat() {
+	  		if ((RDFFormat.JSONLD_COMPACT_PRETTY.equals(format)) || (RDFFormat.JSONLD_COMPACT_FLAT.equals(format))) return JSONLD_FORMAT.COMPACT;
+	  		if ((RDFFormat.JSONLD_EXPAND_PRETTY.equals(format)) || (RDFFormat.JSONLD_EXPAND_FLAT.equals(format))) return JSONLD_FORMAT.EXPAND;
+	  		if ((RDFFormat.JSONLD_FLATTEN_PRETTY.equals(format)) || (RDFFormat.JSONLD_FLATTEN_FLAT.equals(format))) return JSONLD_FORMAT.FLATTEN;
+	  		if ((RDFFormat.JSONLD_FRAME_PRETTY.equals(format)) || (RDFFormat.JSONLD_FRAME_FLAT.equals(format))) return JSONLD_FORMAT.FRAME;
+	  		throw new RuntimeException("Unexpected output format");
+    }
+    
+    private boolean isPretty() {
```
ajs6f: This may be down to personal preference, but I would find this a bit clearer with a constant collection, i.e.
```
private static final Set<RDFFormat> PRETTY_FORMATS = ImmutableSet.of(JSONLD_COMPACT_PRETTY, JSONLD_FLATTEN_PRETTY, etc.);
...
private boolean isPretty() { return PRETTY_FORMATS.contains(format); }
```

jena-arq/src/test/java/org/apache/jena/riot/writer/TestJsonLDWriter.java
```diff
@@ -0,0 +1,457 @@
...
+import org.apache.jena.riot.out.JsonLDWriter;
+import org.apache.jena.sparql.util.Context;
+import org.apache.jena.vocabulary.RDF;
+import org.apache.log4j.Logger;
+import org.junit.Test;
+
+import com.fasterxml.jackson.core.JsonParseException;
+import com.github.jsonldjava.utils.JsonUtils;
+
+public class TestJsonLDWriter extends BaseTest {
```
ajs6f: Again, maybe more of a style thing, but if you like, a lot of the building of simple RDF in this test could be done very concisely with the `static` methods in `org.apache.jena.sparql.sse.SSE`. They are used for a lot of other test data.

jena-arq/src/main/java/org/apache/jena/riot/out/JsonLDWriter.java
fpservant: Hi ajs6f, thanks for the comments. Yes, it is probably clearer.

jena-arq/src/test/java/org/apache/jena/riot/writer/TestJsonLDWriter.java
fpservant: Hi again. Well you know, old user of the Model API...

afs:
Apologies that this has taken so long.

> TODO some documentation. Where should it be added? The best form would be small sample code. 

Examples can go in `jena-arq/src-examples/arq/examples/riot/`. A new section about JSON-LD output for https://jena.apache.org/documentation/io/rdf-output.html, and use code links to apache/jena github, master branch. 

Symbols are fine but URI's please! (changing `RIOT.riotIRI` to "http://jena.apache.org/riot"):

e.g.  in JsonLDWriter
```
    public static final Symbol JSONLD_CONTEXT = Symbol.create("JSONLD_CONTEXT");
```

**RDFDataMgr**

The way to set things specially for a writing is to do:

```
    WriterDatasetRIOT w = RDFDataMgr.createDatasetWriter(RDFFormat.JSONLD_COMPACT_FLAT) ;
    w.write(System.out,  dataset.asDatasetGraph(), null, "http://base", RIOT.getContext()) ;
```
rather than add `RDFDataMgr.write(..., Context cxt)`

See [examples/riot/ExRIOT_2](https://github.com/apache/jena/blob/master/jena-arq/src-examples/arq/examples/riot/ExRIOT_2.java) for an example for a reader.

If you want, add a new API class specific to JSON output `JsonLdMgr` (to go in the same package as `RDFDataMgr`) with JSON-LD-specific functions to wrap common ways to call JSON-LD output. 

**Constants**

This could be tidied out with a subclass of `RDFFormatVariant` that carries JSON-LD information:

```
    // From JSON writer
    public static enum JSONLD_FORMAT {
        COMPACT, FLATTEN, EXPAND, FRAME
    }
    
    static class JSONLDFormatVariant extends RDFFormatVariant {
        private JSONLD_FORMAT option ;
        private boolean prettyJson ;

        public JSONLDFormatVariant(String name, boolean prettyJson, JSONLD_FORMAT option) { 
            super(name) ;
            this.options = option ;
            this.prettyJson = prettyJson ;
        }
        
        public boolean prettyJson() { return prettyJson ; }
        
        public boolean option(JSONLD_FORMAT fmt) {
            for ( JSONLD_FORMAT f : options ) {
                if ( fmt == f )
                    return true ; 
            }
            return false ;
        }
    }


    private static final RDFFormatVariant JSONLD_EXPAND_PRETTY      = new JSONLDFormatVariant("expand pretty", true, JSONLD_FORMAT.EXPAND) ;
    private static final RDFFormatVariant JSONLD_EXPAND_FLAT        = new JSONLDFormatVariant("expand flat", false, JSONLD_FORMAT.COMPACT) ;
    ...
```
then use in JsonLDWriter by casting to and JSONLDFormatVariant
```
        JSONLDFormatVariant format = (JSONLDFormatVariant)format ;
```
```        
        if ( format.option(COMPACT) ) {
          ...
        }
```

**Tests**

I ran the tests for JSONLD but got:

```
10:13:04 WARN  JsonLDWriter              :: JSONLD_CONTEXT value is not a String. Assuming a context object expected by JSON-LD JsonLdProcessor.compact or flatten
10:13:04 INFO  TestJsonLDWriter          :: Sorry to get this exception
org.apache.jena.riot.RiotException: com.github.jsonldjava.core.JsonLdError: loading remote context failed: http://schema.org/
    at org.apache.jena.riot.out.JsonLDWriter.serialize(JsonLDWriter.java:218)
    at org.apache.jena.riot.out.JsonLDWriter.write(JsonLDWriter.java:123)
    at org.apache.jena.riot.out.JsonLDWriter.write(JsonLDWriter.java:129)
    at org.apache.jena.riot.system.RiotLib$WriterAdapter.write(RiotLib.java:376)
    at org.apache.jena.riot.RDFDataMgr.write$(RDFDataMgr.java:1232)
    ...
```

We need to be able to run the tests without assuing that external network resources are availanle and functioning.

The tests and build should work if there is no extenral connectivity.

**Code organisation**

If you feel there classes, fell free to create a package org.apache.jena.riot.out.jsonld to put them together.

**Other**

Indentation: spaces, not tabs please.

afs:
JIRA created to track this: https://issues.apache.org/jira/browse/JENA-1208

fpservant:
Andy
> Apologies that this has taken so long.
no problem, really. Thanks for this answer. Give me some time to take all your valuable comments into account.

just one remark:
> I ran the tests for JSONLD but got:
>> 10:13:04 INFO  TestJsonLDWriter          :: Sorry to get this exception
>>org.apache.jena.riot.RiotException: com.github.jsonldjava.core.JsonLdError: loading remote context failed: http://schema.org/
...
> The tests and build should work if there is no extenral connectivity.

I understand the point, and I'll take it into account. But actually, I'm happy to see this, because it shows that one problem that I had at the time has been solved. As was noted in a comment in the code:
```
	} catch (Throwable e) {
		Logger.getLogger(getClass()).info("Sorry to get this exception",e);
	}

	// It seems to work, but there's a problem with httpclient version: the one used by jena
	// is not compatible with the one expected by JSONLD-java API. We get
	/*
java.lang.NoSuchMethodError: org.apache.http.impl.client.cache.CacheConfig.custom()Lorg/apache/http/impl/client/cache/CacheConfig$Builder;
```
Looks like the work on " (JENA-576) Upgrade Apache HTTP Client to 4.3" has solved the issue

fpservant:
Hi Andy,
getting back to this after, well, a long summer.
I have a remark regarding your following comment:
> RDFDataMgr
>
> The way to set things specially for a writing is to do:
>
>    WriterDatasetRIOT w = RDFDataMgr.createDatasetWriter(RDFFormat.JSONLD_COMPACT_FLAT) ;
>    w.write(System.out,  dataset.asDatasetGraph(), null, "http://base", RIOT.getContext()) ;
> 
> rather than add RDFDataMgr.write(..., Context cxt)

I understand the point of not adding new methods to RDFDataMgr.
However:
- there are read methods with a Context param (so, we would get some symetry between read and write methods)
- and I have the feeling that writing down a Model gets more convoluted for the user without the write methods with the Context param.

If we have a write with the Context param, one just has to:
```
private void writeExample(OutputStream out, Model m, RDFFormat f, Context jenaContext) {
      try {
		RDFDataMgr.write(out, m, f, jenaContext) ;
                ...
 ```

Without it:
```
private void writeExample(OutputStream out, Model m, RDFFormat f, Context jenaContext) {
      try {
		WriterDatasetRIOT w = RDFDataMgr.createDatasetWriter(f) ;
		DatasetGraph g = DatasetFactory.create(m).asDatasetGraph();
		PrefixMap pm = RiotLib.prefixMap(g);
		String base = null;
                w.write(out,  g, pm, base, jenaContext) ;
                ...
 ```

I've had to search a little bit to write it. I guess that other users also would. Would it justify the extra RDFDataMgr.write(..., Context cxt)? That's the question I am asking.

Best,

fps

afs:
I'm not opposed to adding "Context" variants on principle, just a concern about warping the API in the application writers view by many operations (not that there are a few at the moment!). There are 22 write(...) operations currently. Or maybe only for the 11 `RDFFormat` forms (the Lang/RDFFormat being one of the causes of so many options)

Adding the PrefixMap variants (4 new operations : 2 each for graph and datasetgraph writers).

This should work and it's a bit shorter: the machinery for DatasetGarph/Graph is built-in:

```
private void writeExample(OutputStream out, Model m, RDFFormat f, Context jenaContext) {
        WriterGraphRIOT w = RDFDataMgr.createGraphWriter(f) ;
        Graph g = m.getGraph() ;
        w.write(out, g, RiotLib.prefixMap(g), null, jenaContext);
    }
```

afs:
So how many additional 'write+Context' operations? Just the RDFFormat ones for now?

fpservant:
Andy,
I have reverted to an unmodified version of RDFDataMgr (without write methods with a context param). 

I am sorry that I take so long to finalize this stuff. Remember that one thing that we want to be able to do is to pass a (JSONLD) context (as a jena context) to the Writer, in order to have it used when computing the output. But there is a problem with JSONLD java (see [issue 184](https://github.com/jsonld-java/jsonld-java/issues/184)): it doesn't seem possible (or at least, easy) to let the user set the (JSONLD) context to something like ``"@context": "http://schema.org/"``. The best that I can do is to pass the content corresponding to the URL (downloaded and converted to a map) to the JSONLD API, and then to substitute the @context value with "http://schema.org/" at the end of the process (just before writing the output).

(I hadn't seen this problem when I created the pull request, because at this time, there was no way to have JSON-LD java download externally defined contexts, because of incompatible versions of HttpClient used by jena and json-ld java).

Regarding the way to expose the Symbols that can be set it the (jena) context, and the classes expected for the corresponding values, are you OK with a JSONLDWriteContext class (extending Context), with methods such as:
```
        public void setOptions(JsonLdOptions opts) {     
            set(JSONLD_OPTIONS_SYMBOL, opts);
        }
```

I try to finalize something quickly and I update this.

fpservant:
I updated this with some sample code, and it seems to me that this is as good as *I* can. The sample code is the best place to learn how it works and what it does. 

Andy, answering your remarks (Jul 10):

> Examples can go in jena-arq/src-examples/arq/examples/riot/.

done

> A new section about JSON-LD output for https://jena.apache.org/documentation/io/rdf-output.html

this still has to be done, but if I understand correctly, this is the documentation about the release, isn't it? (so should be updated later)

> Symbols are fine but URI's please!

done

> RDFDataMgr The way to set things specially for a writing is to do:...

ok. No new "write" methods with a Context argument (as there were in my frist attempt)

> If you want, add a new API class specific to JSON output ``JsonLdMgr ``

I created only a JSONLDWriteContext class (extending Context), with setters for the relevant Symbols

> Constants This could be tidied out with a subclass of ``RDFFormatVariant``

see ``public static class JSONLDVariant extends RDFFormatVariant`` inside RDFFormat.java, with methods such as ``isCompact()``, `isFratten()``etc.

> We need to be able to run the tests without assuing that external network resources

done

> Code organisation. If you feel there classes, fell free to create a package org.apache.jena.riot.out.jsonld 

very few classes, actually.

I moved ``jena/riot/out/JsonLDWriter.java`` (and ``JenaRDF2JSONLD``) to ``jena/riot/writer`` to have it with other writers. Only one new file``jena/riot/JsonLDWriteContext``


> Indentation: spaces, not tabs please.

done

Tell me if you want me to change things

Best

stain:
If you get the HTTPClient working again, then you should be able to use JSONLD-Java's
[Caching mechanism](https://github.com/jsonld-java/jsonld-java#loading-contexts-from-classpathjar) to avoid external retrieval of http://schema.org/  

(but note that a download of the schema.org context is covered by http://creativecommons.org/licenses/by-sa/3.0/ )

fpservant:
@stain yes, JSONLD-java can download a context from a URL, and cache it. The problem is, there is no way that I know of that would allow me to use that context in a write operation initiated from jena. I opened a JSONLD-java [issue](https://github.com/jsonld-java/jsonld-java/issues/184) about this.

acoburn:
While this isn't exactly a review, I would add that what @fpservant has implemented here is _exactly_ what I need for some projects I work on -- projects that make heavy use of Jena, but when it comes to JSON-LD output, I mostly have to use the `jsonld-java` directly -- especially for handling flattened and expanded profiles.

jena-arq/pom.xml
```diff
@@ -21,7 +21,7 @@
   <artifactId>jena-arq</artifactId>
   <packaging>jar</packaging>
   <name>Apache Jena - ARQ (SPARQL 1.1 Query Engine)</name>
-  <version>3.1.1-SNAPSHOT</version>
+  <version>3.1.2-SNAPSHOT</version>
```
afs: Needs to be 3.1.1-SNAPSHOT

jena-arq/pom.xml
```diff
@@ -85,8 +84,9 @@
     </dependency>
 
     <dependency>
-      <groupId>com.github.jsonld-java</groupId>
-      <artifactId>jsonld-java</artifactId>
+       <groupId>com.github.jsonld-java</groupId>
+        <artifactId>jsonld-java</artifactId>
+        <version>0.8.3</version>
     </dependency>
```
afs: `<version>` not needed - the dependency management in jena-parent controls the version.

jena-arq/pom.xml
```diff
@@ -85,8 +84,9 @@
...
 
     <dependency>
-      <groupId>com.github.jsonld-java</groupId>
-      <artifactId>jsonld-java</artifactId>
+       <groupId>com.github.jsonld-java</groupId>
```
afs: Indentation changed.

afs:
@acoburn - thanks - that sort of feedback is very useful.

fpservant:
@afs, 

> If possible, please see if the PR can be alighed to the current Jena master branch for a clean merge 

do you mean: "see if I can do something to get rid of the conflict" ?

Regarding the conflict, I tried the following: I replaced my local rio/out with the one from upstream/master (-> the old JsonLDWriter is back in my riot/out), and I committed/pushed it. (that was my "trying to solve conflict with master regarding JsonLDWriter" commit)
-> no more conflict.
Then I tried to delete the JsonLDWriter ("trying to solve conflict with master regarding JsonLDWriter, step2" commit)
-> conflict again
-> I don't know what I can do

jena-arq/pom.xml
fpservant: sorry. jena-arq/pom.xml now reverted to upstream/master

afs:
> do you mean: "see if I can do something to get rid of the conflict" ?

Yes. If you can't cause git(hub) to not complain, just log here what to do for the person doing the git merge on this pull request.

(with `.diff` files and especially with `patch` files, I've found that multiple commits to the same file cause GH confusion. This could be the same issue.)

fpservant:
@afs as I said in y previous mess, I don't seem to be able to get rid of the conflict (but I am not an expert with git and github) The only solution I see that would probably work would mean to move back ``JsonLDWriter`` from ``org.apache.jena.riot.writer`` to ``org.apache.jena.riot``.

The merge should be simple enough: the conflicting file is to be removed (there is a new version of it in ``jena-arq/src/main/java/org/apache/jena/riot/out/writer/JsonLDWriter.java``)

afs:
OK - I can work with that.

afs:
Merged! Thank you!

I'd like to announce it on dev@ (and maybe encourage some people on users@) to get feedback before we do the 3.1.1 release.

What should I link to to say "it's new - read this"?

fpservant:
@afs thanks to you.

> What should I link to to say "it's new - read this"?

good question. I think that the best is to link to the sample code, where I've tried to explain what can be done, and how: ``jena-arq/src-examples/arq/examples/riot/ExJsonLD.java``

I already had mentioned it, but note that I haven't done anything regarding the ["Writing RDF in Apache Jena" page]( https://jena.apache.org/documentation/io/rdf-output.html). The only way I saw is the " Improve this Page" link on the page, so this would mean that this can only be done after the release. Am I correct?

ajs6f:
@fpservant, No, using that button, you can send a patch for the page now, but it will not be applied until the trigger is pulled to republish the docs pages. That will happen for the release.

fpservant:
@ajs6f ah, OK, thank you. So I'll do it shortly

PR #141, opened 2016-05-10
Fix the complete Windows path, not just the filename

joe776:
After converting the String to an absolute path it can still contain backslashes, possibly resulting in compilation errors in the generated code. For example if one of the folder starts with "u" the path contains "\uXXXXX" which is usually not a valid Unicode escape and thus compilation fails.


afs:
Thanks!

PR #143, opened 2016-05-12
JENA-1175: Test function for Sprintf that is Timezone aware.

ales004:
I changed the test function for Sprintf so that now it should be timezone aware. It is difficult to test if it is working correctly as Jena is using the default timezone always but I think that the current implementation should solve the problem.

### DIFF
--- a/jena-arq/src/test/java/org/apache/jena/sparql/expr/TestFunctions.java
+++ b/jena-arq/src/test/java/org/apache/jena/sparql/expr/TestFunctions.java
@@ -30,6 +30,14 @@ import org.apache.jena.sparql.function.FunctionEnvBase ;
 import org.apache.jena.sparql.util.ExprUtils ;
 import org.junit.Test ;
 
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.Locale;
+import java.util.TimeZone;
+
 public class TestFunctions
 {
     private static final NodeValue INT_ZERO = NodeValue.makeInteger(0) ;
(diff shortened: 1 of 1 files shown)


jena-arq/src/test/java/org/apache/jena/sparql/expr/TestFunctions.java
```diff
@@ -81,28 +89,35 @@
...
+        stdFormatOut.setTimeZone(TimeZone.getDefault());
+        String outDate = stdFormatOut.format(dtDate);
+        assertEquals(s,outDate);
     }
     
     // Temporary fix for JENA-1175
     // Timezone -11:00 to any timezone can be a day ahead
-    @Test public void exprSprintf_05a() { test_exprSprintf_05("2005-10-14T14:09:43-11:00",  "10 14,2005", "10 15,2005") ; }
+    @Test public void exprSprintf_05a() { test_exprSprintf_05("2005-10-14T14:09:43-11:00") ; }
     // Timezone Z to any timezone can be a day behind or a day ahead
```
kinow: Looking at the changes, I think this PR may make that test pass. But the comment seems to indicate that we must test with Z suffix as well. Or update the comment if that makes sense.

afs:
`Z` is the same-value-as `+00:00`, testing both is ideal.

It is commenting that if the data is Z, which it is for `exprSprintf_05b`, the outcome can be a day behind (i.e. Hawaii direction) or ahead (New Zealand direction) or the same (UTC) depending on the exact time.

Timezones range from -14:00 to +14:00.

kinow:
Righto, so we just update the comment later. I got an error while testing in my current time zone (Pacific/Auckland).

With the code in the master branch, test receives 2005-10-14T10:09:43+11:00 as parameter. The possible values are [10 13,2005, 10 14,2005].

Inside the test, 2005-10-14T13:09:43Z is computed as: "10 15,2005", and the test fails.

Will test the pull request to confirm whether it is failing too or not.

afs:
`test_exprSprintf_05` is ignoring it's first argument.  Bother.

If NodeValue.makeDateTime is changed:

    String exprStr = "afn:sprintf('%1$tm %1$te,%1$tY',   +NodeValue.makeDateTime(**nodeStr**).toString()+")" ;

does it work?

kinow:
I think so. I noticed the first argument was not being used, tried that in Eclipse, and the test passed.

kinow:
FWIW, tested the pull request locally, the test seems to have passed (the build is still running, but will take a while. Arq tests passed fine).

ales004:
I had to sobstitute Z with 00:00 because the SimpleDateFormat is accepting either 
yyyy-MM-dd'T'HH:mm:ssXXX
or 
yyyy-MM-dd'T'HH:mm:ssZ
not both at the time.

afs:
I includes the new tests and kept the temporary tests so testing happens both ways.

Tests renumbered as well.

PR #144, opened 2016-05-14
JENA-508: Partial implementation of Xpath3 functions (fn:normalize-space, fn:normalize-unicode and fn:round-half-to-even)

ales004:
I implemented the three Xpath3 functions:
* fn:normalize-space
* fn:normalize-unicode
* fn:round-half-to-even

I tested the functionality taking the examples from the specs (xpath or unicode). 
I still have one problem with 2 tests exprStrNormalizeUnicode10 and exprStrNormalizeUnicode11 that are not passing. I wrote in the comment of the test my findings but I am not sure how to go on from here. Someone could help?

Also while creating fn:round-half-to-even I realized that fn:round-half is not following the Xpath3 specs. I thus changes also the fn:round-half implementation. Still as I saw that the function used for fn:round-half we used for SPARQL ROUND, I decided to separate the two functions. Not sure if also SPARQL ROUND should be implemented following the Xpath 3 spec.


afs:
Tests: exprStrNormalizeUnicode10, exprStrNormalizeUnicode11

I don't know why they are failing either. May be it's because the version of Unicodein Java is not new enough.

I've commented these out for the moment.

PR #146, opened 2016-05-15
JENA-1180: Add support for ComplexPhraseQueryParser to jena-text

adolby:
This PR expands on JENA-1134, which added support for new query parsers in jean-text. ComplexQueryParser is useful for performing wildcard or fuzzy search on terms in phrases.

I’ve attempted to follow the pattern for adding a query parser as in JENA-1134, but I may have duplicated some functionality in the unit test, when I should have restructured instead.

The test needs to specify the TTL spec analyzer and parser, which is available in TestDatasetWithConfigurableAnalyzer’s method makeSpec. However, TestDatasetWithConfigurableAnalyzer inherits from TestDatasetWithKeywordAnalyzer, which isn’t the analyzer the ComplexPhraseQuery needs to test.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/core/assembler/AssemblerUtils.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/core/assembler/AssemblerUtils.java
@@ -68,11 +68,16 @@ public class AssemblerUtils
         register(ConstAssembler.general(), r, a, DatasetAssembler.getType()) ;
     }
 
+    /** Register an assembler that creates a dataset */
+    static public void registerModel(Resource r, Assembler a) {
+        register(ConstAssembler.general(), r, a, JA.Model) ;
+    }
+
     /** Register an addition assembler */  
     static public void register(AssemblerGroup g, Resource r, Assembler a, Resource superType) {
         registerAssembler(g, r, a) ;
         if ( superType != null && ! superType.equals(r) ) 
-            modelExtras.add(r, RDFS.subClassOf, DatasetAssembler.getType()) ;
+            modelExtras.add(r, RDFS.subClassOf, superType) ;
     }
     
     /** register */ 
--- a/jena-cmds/pom.xml
+++ b/jena-cmds/pom.xml
@@ -87,9 +87,8 @@
       <groupId>log4j</groupId>
       <artifactId>log4j</artifactId>
     </dependency> 
-    
-    <!-- Testing -->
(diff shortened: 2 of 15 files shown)


osma:
Thanks for the PR!

The code looks good, as you say it's following the pattern established by JENA-1134.

You are right about the unit tests - the inheritance hierarchy of analyzer-related unit tests is rather messy. There is a very long inheritance chain and e.g. TestDatasetWithAnalyzingQueryParser  indirectly inherits from TestDatasetWithKeywordAnalyzer. There is actually a sort of logic here, because the tests specified in TestDatasetWithKeywordAnalyzer will also be run on the subclasses, as the analyzer configurations are compatible. This gives us some unit tests "for free".

I did some reorganizing of the test classes, namely, I abstracted out AbstractTestDatasetWithAnalyzer from TestDatasetWithKeywordAnalyzer (see https://github.com/apache/jena/commit/8a20f238ffc5aed23648b038132fe74cbca2c85c). Unless I'm missing ssomething, you should now be able to inherit from AbstractTestDatasetWithAnalyzer and avoid doing the full makeSpec in your test class. Would you mind switching over to that?

adolby:
Thank you for working on that refactoring! You're right, abstracting out makeSpec reduces the repetition for the ComplexPhraseQueryParser test. I've switched it over to inheriting from AbstractTestDatasetWithAnalyzer. Are there other changes you'd like to see in this PR? I'm not sure if/where documentation should be updated to reflect having the ComplexPhraseQueryParser as a QueryParser option.

osma:
I think the code is now good for merging. Thanks a lot!

@afs Can I just merge the code to master, or is there some legal policy that must be followed for contributions like this? I've never merged PRs initiated by others.

As for documentation, it lives outside this github repository (in the ASF SVN) and you can submit a patch via the web CMS - just click on "Improve this page" at the top of https://jena.apache.org/documentation/query/text-query.html

We can close this PR after the code is merged, then resolve the JIRA issue after the documentation has been updated.

afs:
You can merge it.

The principle is that any contribution needs to be clearly given by the original owner. Someone making sending a Pull Request is taking a step that clearly says "This is for the project" and so the Apache contribution is clear. 

We ask for additional paperwork when the contribution is large, and if they come by a different route to JIRA, or a PR to the project, we'll have to think about it.

 A PR like this can be dealt with as being clearly intended as being a contribution.

PR #148, opened 2016-06-03
JENA-508: implemented fn:adjust-datetime-to-timezone, fn:adjust-date-to-timezone and fn:adjust-time-to-timezone

ales004:
I implemented two new XPath 3 functions: fn:adjust-datetime-to-timezone, fn:adjust-date-to-timezone and fn:adjust-time-to-timezone. I used the examples of the specification for the tests + I decided to implement both using the same underlying function as it seems to me that in jena there is no distinction between date and datetime (date and datetime are using the same object and from the time we can also get the same object).

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/expr/nodevalue/XSDFuncOp.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/expr/nodevalue/XSDFuncOp.java
@@ -1423,9 +1423,10 @@ public class XSDFuncOp
         return NodeValue.makeString(dts.timezone) ;
     }
 
-    private static NodeValue fromTimezoneToDuration(DateTimeStruct dts){
+    public static NodeValue dtGetTimezone(NodeValue nv) {
+        DateTimeStruct dts = parseAnyDT(nv) ;
         if ( dts == null || dts.timezone == null )
-            return null;
+            throw new ExprEvalException("Not a datatype with a timezone: " + nv) ;
         if ( "".equals(dts.timezone) )
             return null ;
         if ( "Z".equals(dts.timezone) ) {
@@ -1454,13 +1455,6 @@ public class XSDFuncOp
         digitsTwo(s, idx, sb, 'M') ;
         idx += 2 ;
         return NodeValue.makeNode(sb.toString(), null, XSDDatatype.XSD + "#dayTimeDuration") ;
-    }
-
-    public static NodeValue dtGetTimezone(NodeValue nv) {
-        DateTimeStruct dts = parseAnyDT(nv) ;
-        if ( dts == null || dts.timezone == null )
-            throw new ExprEvalException("Not a datatype with a timezone: " + nv) ;
-        return fromTimezoneToDuration(dts);
     }
 
     private static void digitsTwo(String s, int idx, StringBuilder sb, char indicator) {
--- a/jena-arq/src/main/java/org/apache/jena/sparql/expr/nodevalue/XSDFuncOp.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/expr/nodevalue/XSDFuncOp.java
@@ -1590,6 +1590,11 @@ public class XSDFuncOp
             }
             Duration tzDuration = nv2.getDuration();
             tzOffset = tzDuration.getSign()*(tzDuration.getMinutes() + 60*tzDuration.getHours());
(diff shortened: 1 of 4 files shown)


afs:
Same function that checks for xsd:dateTime, xsd:date or xsd:time is the right thing to do.

afs:
It's a good thing it's summer where I am (British Summer Time)!

My reading of https://www.w3.org/TR/xpath-functions-3/#timezone.functions is that conversions where there is no provided timezone use local time, including daylight savings time.

In the tests `TimeZone.getDefault().getRawOffset()` returns the offset without regard for the local daylight savings time.

Also allowing for timezones which are negative offsets, I came up with:
```
 private String getDynamicDurationString(){
        int tzOffset = TimeZone.getDefault().getOffset(new Date().getTime())/(1000*60) ; 
        String x = "PT"+Math.abs(tzOffset)+"M";
        if ( tzOffset < 0 )
            x = "-"+x ;
        return x ;
    }
```

`XSDFuncOp.adjustDatetimeToTimezone` also needs this.

jena-arq/src/main/java/org/apache/jena/sparql/expr/nodevalue/XSDFuncOp.java
```diff
@@ -1554,4 +1551,55 @@ private static Duration valueCanonicalDuration(NodeValue nv) {
...
+
+        XMLGregorianCalendar calValue = nv1.getDateTime();
+        Boolean hasTz = calValue.getTimezone() != DatatypeConstants.FIELD_UNDEFINED;
+        int inputOffset = 0;
+        if(hasTz){
+            inputOffset = calValue.getTimezone();
+        }
+
+        int tzOffset = TimeZone.getDefault().getRawOffset() / (1000*60);
+        if(nv2 != null){
```
afs: `int tzOffset = TimeZone.getDefault().getOffset(new Date().getTime())/(1000*60) ;`

Ideally, don't set this with a call to `new Date()` until later -- see below.

jena-arq/src/main/java/org/apache/jena/sparql/expr/nodevalue/XSDFuncOp.java
```diff
@@ -1554,4 +1551,55 @@ private static Duration valueCanonicalDuration(NodeValue nv) {
...
+                    if(nv1.isDateTime())
+                        return NodeValue.makeDateTime(calValue);
+                    else if(nv1.isTime())
+                        return NodeValue.makeNode(calValue.toXMLFormat(),XSDDatatype.XSDtime);
+                    else
+                        return NodeValue.makeDate(calValue);
+                }
+                throw new ExprEvalException("Not a valid duration:" + nv2);
+            }
+            Duration tzDuration = nv2.getDuration();
```
afs:
```} else {
            tzOffset = TimeZone.getDefault().getOffset(new Date().getTime())/(1000*60) ;
        }
```
so `new Date` is only called if needed.

jena-arq/src/main/java/org/apache/jena/sparql/expr/nodevalue/XSDFuncOp.java
```diff
@@ -1554,4 +1551,55 @@ private static Duration valueCanonicalDuration(NodeValue nv) {
...
+            tzOffset = tzDuration.getSign()*(tzDuration.getMinutes() + 60*tzDuration.getHours());
+            if(tzDuration.getSeconds() > 0)
+                throw new ExprEvalException("The timezone duration should be an integral number of minutes");
+            int absTzOffset = java.lang.Math.abs(tzOffset);
+            if(absTzOffset > 14*60)
+                throw new ExprEvalException("The timezone should be a duration between -PT14H and PT14H.");
+        }
+        String tzSign = (tzOffset-inputOffset) > 0 ? "" : "-";
+        Duration durToAdd = NodeValue.makeDuration(tzSign+"PT"+java.lang.Math.abs(tzOffset-inputOffset)+"M").getDuration();
```
afs: Could this be done with something like `NodeValue.xmlDatatypeFactory.createDurationDayTime(tzOffset-inputOffset)`?

It avoids string parsing.

afs:
Other than my comments above - this looks good to include.

ales004:
Andy,
you are right! I was not thinking that with daylight savings the timezone offset can change. Thank you for your input!


I just push the changes you requested and all tests are passing.

PR #149, opened 2016-07-06
Jena 1206

barthanssens:
Added W3C DCAT, VoID, ORG and ROV vocabularies


afs:
+1 : Looks good to me.

afs:
I was assuming these vocabularies would be part of the main distribution and build.  If so, the POM file is not needed (nor invoked).

afs:
If we were starting from scratch, `vocabularies/w3c` makes sense.  We already have `RDF`, `RDFS`, `OWL`, `SKOS` which are W3C vocabularies and they are in `vocabularies/`.

barthanssens:
Ah OK, so should I go back  to the original proposal (just adding the classes to ...vocabularies), or use a separate package (as suggested in https://issues.apache.org/jira/browse/JENA-1206, e.g. "vocabularies.egov") ? Thanks.

afs:
I don't have any strong feelings - I'm just following the current pattern (for better or worse).

I doubt we will have name clashes (independent vocabs - not versions).

The minor downside of classifying by package is implicit domain. DCAT will be hopefully used much more widely than eGov.

PR #150, opened 2016-07-06
[JENA-1204] support for multiple BuiltinRegistry for different rule sets

paulhoule:
This patch introduces support for multiple BuiltinRegistry classes implemented in a simple way which is parallel to the existing architecture.  (i.e. the BuiltinRegistry is a property of the Rules Engine parser,  not the Reasoner)


afs:
Ping @der

der:
Thanks for the ping. I've seen it but not had time to look at yet.

der:
Looks reasonable to me, thanks for the contribution. Merging it.

PR #152, opened 2016-07-08
JENA-1158: Union default graph

afs:
This competes general handling of `urn:x-arq:UnionGraph`.


jena-arq/src/main/java/org/apache/jena/sparql/graph/GraphOps.java
```diff
@@ -29,60 +29,49 @@
...
+        if ( Quad.isUnionGraph(gn) )
             return unionGraph(dsg) ;
         return dsg.getGraph(gn) ;
     }
-    
-    public static Graph unionGraph(DatasetGraph dsg)
-    {
+
+    public static Graph unionGraph(DatasetGraph dsg) {
+        // Snapshot it now.
```
ajs6f: How come a snapshot here? I have no objection to it (other than the cost of rolling up), just wondering. Is it because of potential `ConcurrentModificationException`s?

afs: Yes, and it's just what was there before, though it does seem unlikely that CCME happens and the alternative of a dynamic union makes sense as well.

I'll update the PR so there are two operations.

PR #153, opened 2016-07-09
JENA-1090: Txn - a java8-centric transaction API

afs:


afs:
There is markdown documentation to go with this.

jena-arq/src/main/java/org/apache/jena/system/ThreadTxn.java
```diff
@@ -0,0 +1,148 @@
...
+    
+    private final Semaphore semaStart ;
+    private final Semaphore semaFinish ;
+    private final AtomicReference<RuntimeException> thrownRuntimeException = new AtomicReference<>(null) ; 
+    private final AtomicReference<Error> thrownError = new AtomicReference<>(null) ;
+    private final Runnable action ;
+    
+    private ThreadTxn(Runnable action) {
+        this.action = action ;
+        this.semaStart = new Semaphore(0, true) ;
```
ajs6f: I'm a little confused about the way these semaphores work. E.g. how do they function with 0 permits? Below, in `::run()` you release the start semaphore, but I can't see where you acquire it… on the theory (possibly a bad one) that other newbies will also find this opaque, can you comment this up a bit to explain the concurrency control at work here?

afs:
Semaphore hand out permits so 0 means no permit.

There is a `semaStart.release()` in `run()` and the corresponding `acquire` in the lambda in `create` which has comments. Renaming the local `semaStart` would help.

The `AtomicReference` can be combined in `AtomicReference<Throwable>`.

`ThreadTxn` came about for testing purposes - `Txn` is the main contribution.

ajs6f:
Thanks, @afs , that definitely helps. IIUC, `ThreadTxn` looks like kind of a "single-class" version of the mechanics in `Txn`?

ajs6f:
We usually use `try-finally` for the transaction cycle, i.e.
```
try {
    transactionalThing.begin(readWrite);
    //do some stuff
    transactionalThing.commit(); // if a write transaction
}
finally {
    transactionalThing.end();
}
```
Looking at `onThrowable` it looks like if the `Transactional txn` throws anything other than a `JenaTransactionException` out of `txn.abort()`, then `txn.end()` will never get called. Does that open a whole for `txn` to never actually finish aborting/ending?

afs:
If one of the transaction operations throws anything then the system is broken. It should not happen and would indicate an unrepairable error condition (e.g. NPE, broken journal). 

After that, all bets are off!

ajs6f:
Okay, so the appropriateness of `finally` depends on where you are (app vs. system code). Makes sense.

afs:
Merged but this code always benefits from ongoing code review!

PR #154, opened 2016-07-13
JENA-1209: Exclusive mode for TDB

afs:


jena-tdb/src/main/java/org/apache/jena/tdb/transaction/TransactionManager.java
```diff
@@ -436,8 +469,96 @@ public void notifyAbort(Transaction transaction)
...
+    private boolean acquireWriterLock(boolean canBlock) {
+        if ( ! canBlock )
+            return writersWaiting.tryAcquire() ;
+        try { 
+            writersWaiting.acquire() ; 
+            return true;
+        } catch (InterruptedException e) { throw new TDBTransactionException(e) ; }
+    }
+    
+    /** Block until no writers are active.
```
ajs6f: Just a "wordsmithing" suggestion: this might be better as `waitForWriters` or the like. `disableWriters` sounds a bit like the caller is going to have some kind of actual impact on the outstanding writers, but in truth the caller is just waiting for them to finish on their own, right? Same for the next method.

PR #159, opened 2016-07-24
JENA-1210: Call model.read

afs:


ajs6f:
This makes sense-- one question: currently `RdfDataMgr.read(Model m, etc)` falls through to `RdfDataMgr.read(m.getGraph(), etc)`. Should it be changed to call `Model.read`?

afs:
That will go in circles! Plain `model.read` goes to `RDFReaderRIOT` (The old RDFReader interface adapter to RIOT) which makes all calls go to `RDFDataMgr.read(model,...)`. If that calls `model.read` we have a loop.

`OntModel.read` overrides `Model.read` and does "get basic data" via a call to `super.read` (which is `ModelCom.read`) and then "load imports".

So the layering is that RDFDataMgr does the basic read operations, with the indirection of `StreamManager`/`LocationMapping`.



The key here is going via `FileManager.get().readModel` bypasses an overridable `model.read`.

ajs6f:
Okay, cool, in that case maybe we can put a couple of comments into `RdfDataMgr` to indicate that it can only ever do the "vanilla" stuff, and that if you expect action associated with a particular type of `Model`, use the `read` of an instance of that type. If you don't have time for that, NP, I'll do it with a nonfunctional commit after you merge this.

afs:
There's a comment in `AdapterFileManager.readModelWorker` about override of `Model.read`.

`RDFDataMgr` does say at the top "Read data from a location into a Model/Dataset etc" so it's not promising anything more.

ajs6f:
Sure, but unless you object, I don't think it can hurt to be even more specific, with some brief phrases like "The methods in this class treat all types of Model in the same way. For behavior specific to a type of Model, use the methods of that class."

afs:
No problem.

PR #164, opened 2016-08-21
JENA-1222 and JENA-1224: Flush queue when large by byte size or by pending commits.

afs:
This adds additional control over the queue in the journal.

If the journal gets large, as measured by the journal file size in bytes (JENA-1222) then try to flush the journal.

If the journal becomes excessive long in terms of pending commits (JENA-1224) then after a writer, flip to exclusive mode, flush the journal and flip back to normal mode.

### DIFF
--- a/jena-tdb/src/main/java/org/apache/jena/tdb/transaction/TransactionManager.java
+++ b/jena-tdb/src/main/java/org/apache/jena/tdb/transaction/TransactionManager.java
@@ -141,8 +141,9 @@ public class TransactionManager
     // "one exclusive, or many other" lock which happens to be called a ReadWriteLock
     private ReadWriteLock exclusivitylock = new ReentrantReadWriteLock() ;
     
-    // Delayes enacting transactions.
+    // Delays enacting transactions.
     private BlockingQueue<Transaction> queue = new LinkedBlockingDeque<>() ;
+    public long getQueueLength() { return queue.size() ; }
 
     private DatasetGraphTDB baseDataset ;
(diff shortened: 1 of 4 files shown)


jena-tdb/src/main/java/org/apache/jena/tdb/transaction/TransactionManager.java
```diff
@@ -613,21 +652,43 @@ public void flush() {
...
             processDelayedReplayQueue(txn) ;
     }
     
+    // Whether to try to flush the journal. We may stil find that we are blocked
+    // from doing so by another transaction.
+    private boolean checkForJournalFlush() {
+//        System.err.printf("checkForJournalFlush: queue size=%d; journal size = %d\n", queue.size(), journal.size()) ;
+//        System.err.printf("checkForJournalFlush: QueueBatchSize=%d; MaxQueueThreshold=%d; JournalThresholdSize=%d\n",
+//                          QueueBatchSize, MaxQueueThreshold, JournalThresholdSize) ;
+        if ( QueueBatchSize == 0 )
```
ajs6f: I'm probably missing something, but it seems like this is part of the next case (isn't `queue.size() >= 0` an invariant?). Is this called out just to emphasize that setting `0` => "no queuing"?

afs: True - remains of when JENA-1224 (excessive queue length was handled here ... bit that didn't work out).  Tiding up done in the next commit.

PR #165, opened 2016-08-29
JENA-1228: Expose each optimization transform step in std optimizer.

afs:
https://issues.apache.org/jira/browse/JENA-1228

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/Optimize.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/Optimize.java
@@ -18,298 +18,65 @@
 
 package org.apache.jena.sparql.algebra.optimize;
 
-import org.apache.jena.query.ARQ ;
 import org.apache.jena.sparql.ARQConstants ;
-import org.apache.jena.sparql.SystemARQ ;
-import org.apache.jena.sparql.algebra.* ;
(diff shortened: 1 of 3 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/Optimize.java
```diff
@@ -18,82 +18,58 @@
...
+ *  <pre>
+ *    Optimize.RewriterFactory f = (cxt)->new MyOptimize(cxt) ;
+ *    context.set(ARQConstants.sysOptimizerFactory, f) ;<pre>
+ */
+public class Optimize
 {
-    static private Logger log = LoggerFactory.getLogger(Optimize.class) ;
-
-    // A small (one slot) registry to allow plugging in an alternative optimizer
+    /** Factory for rewriters */
```
ajs6f: Should this maybe be `public static interface RewriterFactory`?

afs: What difference does it make? It's an interface; there is no static/enclosing distinction.

ajs6f: Right, right. Sorry, commented before I had my morning tea.

PR #166, opened 2016-08-31
Address Jena-1229, fuseki timeout queries not working properly with DISTINCT.

ehedgehog:
Remove incorrect call super.close() from the body of QueryIterDistinct.requestSubCancel()
as per JENA #1229.


afs:
You're lucky!

The additional first commits c88e008 and 18d28f5 don't contaminate the diff.

ehedgehog:
Did I manage to mess up commits somehow? That must be because of moving between
repos at home and at work. Sorry.

Is there something I could/should do at my end?

afs:
In the list of commits above, the first commit says "ehedgehog committed on GitHub on 27 Jul"  then a merge from jena/master 2 days ago, then the JENA-1229 change.  Just a guess: you created a branch, then merged jena/master, rather then getting your local master up to date then branching.  But it seems github works it out because if you go to the "166.diff" URL, then there is a short diff, no merge items.

ehedgehog:
(Moving from mailing list)

Andy had said:

>> (/me not entirely convinced that 'close' is always called for all
>> points where cancel triggers everywhere)

Yes. I've just had a look over some of the code and it
certainly wasn't obvious in that look that the query
ends up closing the result iterator. I could have missed
something easily. But if there's a problem there its
more general than the issue with QueryInterDistinct and
deserves its own JIRA? 

But in any case cancel() is [typically] called from the 
timeout thread, so it would surely be thread-unsafe for 
the cancel methods to go calling close() while the query
thread might be working on shared state, independently
of worries about close() always being called.

[A look at the codebase shows that there are three
classes with requestSubCancel() closing, not cancelling,
other iterators:

    AnstractIterHashJoin
    QueryIterDistinctReduced
    QueryIter2LoopOnLeft

which suggests these should be looked at in case there
might be a similar cancellation problem]

How should we make progress on this?

Chris

afs:
I have fixed those 3 requestSubCancel instances.

ehedgehog:
Thanks very much, Andy.

PR #167, opened 2016-09-02
Javadocs

ehedgehog:
Simplify the documentation of EnhGraph by stripping out references to Polymorphic and revising class comment.


afs:
There are 5 commits above but only 0eb58ba is for EnhGraph.

This has the changes for PR#166 in it.

afs:
I have manually applied the suggested changes - the diff of just EnhGraph didn't apply for me but it's easy enough to read.

PR #168, opened 2016-09-08
JENA-1231: Modernize executeInTransaction

afs:


afs:
The second commit, https://github.com/apache/jena/pull/168/commits/1d6a4758de9305d9fe25e08b7594bbda8bd36a83, uses execute/calculate style naming.

For discussion (the PR would be tidied up before merging into Jena).

jena-core/src/main/java/org/apache/jena/graph/impl/TransactionHandlerBase.java
```diff
@@ -36,21 +40,45 @@ public TransactionHandlerBase() {
...
-        catch (Throwable th) { e.addSuppressed(th); }
+    
+    /**
+     * Execute the supplier <code>action</code> within a transaction. If it completes normally,
+     * commit the transaction and return the result, otherwise abort the transaction.
+     */
+    @Override
+    public <T> T executeRtn( Supplier<T> action ) {
+        begin() ;
+        T result ;
```
ajs6f: I'm just wondering why this isn't 
```
try {
    T result = action.get() ;
    commit() ;
    return result ;
}
```
i.e. what is the need to declare `result` outside of the `try`?

afs: Yes - that would be better.  It was just old stuff, continued.

PR #171, opened 2016-09-25
Using library code in a few classes

ajs6f:
Just a couple of orthogonal commits that use library code in `Collections`, switch to using `default` methods where it seems appropriate, remove assignments with no effect, that sort of thing. If no one has any objection within a couple of days, I'll merge it myself.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/atlas/test/Gen.java
+++ b/jena-arq/src/main/java/org/apache/jena/atlas/test/Gen.java
@@ -22,7 +22,13 @@ import static org.apache.jena.atlas.lib.RandomLib.random ;
 
 import java.util.ArrayList ;
 import java.util.Arrays ;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List ;
+import java.util.Set;
+import java.util.StringJoiner;
+import java.util.function.IntSupplier;
+import java.util.stream.IntStream;
 
 /** Support for testing.  May be  generally useful */
 public class Gen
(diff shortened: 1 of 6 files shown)


afs:
Given PR #139 in-progress, can we hold back on the JSON-LD changes for now?

jena-arq/src/main/java/org/apache/jena/atlas/test/Gen.java
```diff
@@ -32,52 +38,35 @@
         return rand(numRand, low, high, false) ; 
     }
     
-    /** Generate a random sequence between low (inclusive) and high (exclusive) - choose whether to have duplicates or not */ 
+    /**
+     * Generate a random sequence between low (inclusive) and high (exclusive) - with duplicates or no
```
afs: Truncated javadoc -- was  "or not"

jena-arq/src/main/java/org/apache/jena/atlas/test/Gen.java
```diff
@@ -95,19 +84,13 @@
                 }
             if ( !found )
                 System.err.printf("Corrupted permute: [%s] [%s]\n", strings(x), strings(x2)) ;
-        }
+        }*/
```
afs: Delete commented out code? (we have git!)

ajs6f: Glad to!

jena-arq/src/main/java/org/apache/jena/atlas/test/Gen.java
ajs6f: Good catch.

ajs6f:
Yes, certainly I'll hold off-- I have no idea how I missed that, especially given that I just pointed out that JSON-LD PR to a colleague in a downstream project yesterday!

jena-arq/src/main/java/org/apache/jena/riot/out/JsonLDWriter.java
```diff
@@ -23,7 +23,6 @@
 import java.io.OutputStreamWriter ;
```
afs: This has moved to riot/writer/

jena-arq/src/main/java/org/apache/jena/riot/out/JsonLDWriter.java
```diff
@@ -98,10 +96,8 @@ private void serialize(Writer writer, DatasetGraph dataset, PrefixMap prefixMap,
             // Unclear as to the way to set better printing.
             obj = JsonLdProcessor.compact(obj, localCtx, opts) ;
 
-            if ( isPretty() )
-                JsonUtils.writePrettyPrint(writer, obj) ;
-            else
-                JsonUtils.write(writer, obj) ;
+            if ( isPretty() ) JsonUtils.writePrettyPrint(writer, obj) ;
+            else JsonUtils.write(writer, obj) ;
```
afs: I don't like if statements with their positive clause on the same line unless they have no else clause and are not part of the flow of the logic. (e.g. not part of the flow: `if ( DEBUG) System.err.println(....) ;`)

jena-arq/src/main/java/org/apache/jena/riot/out/JsonLDWriter.java
```diff
@@ -113,61 +109,45 @@ private void serialize(Writer writer, DatasetGraph dataset, PrefixMap prefixMap,
...
-
-        Iter.iter(graph.find(null, null, null)).apply(x) ;
+        };
+        graph.find(ANY)
+            .filterDrop(t -> {
+                Node pred = t.getPredicate();
+                return pred.equals(RDF.type.asNode()) || ctx.containsKey(pred.getLocalName());
+            })
+            .forEachRemaining(addToContext) ;
     }
```
afs: I think it was clearer to have the `if ( p.equals(RDF.type.asNode()) )` in teh Consumer because all the logic is in one place.

ajs6f:
Now that we have #139 in, I'll add a further commit addressing your points. Thanks!

ajs6f:
Made changes in response to your review, @afs . If you have no further objections, I'll merge this.

PR #172, opened 2016-09-29
Proposal to add 2 new getters in the Resource interface.

ramiromagalhaes:
Here is a proposal to solve issue [JENA-1238](https://issues.apache.org/jira/browse/JENA-1238).


ramiromagalhaes:
Hello,
I updated my push request with some test cases, but I'm having some problems compiling and running it. If the test case is good enough, you're able to run it, and it works properly, please do merge it. When I'm able to try to compile it in other circunstances, I'll retry to compile and run it.

afs:
It all looks fine - thanks.

The only change I made was to fix a javadoc warning.

PR #173, opened 2016-09-29
Solving a memory leak from LPRuleEngine

jfraudeau:
Each time a `LPRuleEngine` gets called to find a triple for which a cached `Generator` exists in `tabledGoals` a new `ConsumerChoicePointFrame` is registered as a consumer of this Generator.
The leak is that it is never unregistered from the consumer, the solution was simply to call the unregistering from the `close()` method of the `ConsumerChoicePointFrame`


der:
+1

Looks good, many thanks for the contribution, will merge.

PR #175, opened 2016-10-08
Added .travis.yml

ajs6f:
https://issues.apache.org/jira/browse/JENA-1241

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 sudo: false
-install: mvn go-offline
+install: mvn dependency:go-offline
 script: mvn -Pdev verify
 jdk:
   - oraclejdk8
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,4 @@ jdk:
   - oraclejdk8
 env:
   global:
-    - JAVA_OPTS="-Xmx8192M -Xms512M"
+    - JAVA_OPTS="-Xmx8192M -Xms512M -XX:+UseG1GC"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+jdk:
+  - oraclejdk8
+  - openjdk8
+env:
+  global:
+    - JAVA_OPTS="-Xmx1024M -Xms512M"
+notifications:
+  irc:
+    channels:
+      - "irc.freenode.org#jena"
```
afs: This is wasn't mentioned in the JIRA!

afs: Having the IRC channel notified.

And if anyone clones this and adds it to jarvis, we get it twice.

ajs6f: I just took that from [here](https://jena.apache.org/help_and_support/#other-channels) in our docs. I hang out in the Jena IRC room all the time and nothing ever gets notified into there. I thought it would be convenient (it's great for other projects on which I work), but it's hardly of the essence. I can take it out if it worries you. What is "jarvis"?

afs:
`openjdk8` is not provided. Use `oraclejdk8`.

https://docs.travis-ci.com/user/languages/java/
(other parts of the documentation may disagree but I just tried ... no `openjdk8`.)

ajs6f:
If it don't work, it don't work. I'll pull that line out, along with the IRC stanza.

ajs6f:
Fixed in Apache Git with f659ca8c16f030dcf4e1a1f547f2dfaefb092fdf

PR #178, opened 2016-10-14
Small optimization for writePrefixes()

OmarElabd:
I made a small change to the N3JenaWriterCommon writePrefixes function so that it does not need to make a call from the hashmap.


afs:
Thank you for the improvement. I've just merged it into the codebase.

The N3 writer is not as used as the Turtle writer in RIOT - I checked to make sure that uses `Entry`.

https://github.com/apache/jena/blob/master/jena-arq/src/main/java/org/apache/jena/riot/system/RiotLib.java#L274

PR #179, opened 2016-10-15
Reorganise control of a test Fuseki server for HTTP API testing.

afs:
JENA-1245


afs:
This is the first part of reorganizing the test server. Changes are made to Fuseki1, and to jena-jdbc-driver-remote, which uses a Fuseki1 server for its testing.

The changes for jena-jdbc-driver-remote mean it should just be a change of dependency to switch to using Fuseki2.

This PR is not yet complete - it shows how the test setup work for Fuseki1. The same changes will be made to Fuseki2, then switch jena-jdbc-driver-remote.

afs:
Commit 7c3a5cf applies the test sever style to Fuseki2. It also makes naming consistent across Fuseki1 and Fuseki2.  

 jena-jdbc-driver-remote has not been switched to using Fuseki2 yet. The "auth" tests will need changes and having them in a separate commit will be make them more easily reversible.

jena-fuseki1/src/test/java/org/apache/jena/fuseki/ServerCtl.java
```diff
@@ -18,39 +18,113 @@
...
+ * In the test suite, put:
+ * 
+ * <pre>
+ *  {@literal @BeforeClass} static public void beforeSuiteClass() { ServerCtl.ctlBeforeTestSuite(); } 
+ *  {@literal @AfterClass}  static public void afterSuiteClass()  { ServerCtl.ctlAfterTestSuite(); }
+ * </pre>
+ * <p>
+ * In the test class, put:
+ * <pre>
+ * {@literal @BeforeClass} public static void ctlBeforeClass() { ServerCtl.ctlBeforeClass(); }
```
ajs6f: Since these methods don't refer to any state, would it be possible to put them into an interface with `default` methods for convenience?

afs: I doubt JUnit would cope with that - super classes yes, mixins no.

1. JUnit4 depends on Java 1.5 (ceratinly not Java8) so it probably does not even check interfaces.
2. Even if it did, there are now ordering issues. Multiple _BeforeClass_ happen in super-class order. If they can be inherited via multiple routes ... it gets interesting.

More generally, the test have the effect of those calls - and may have to achieve to work with their own _BeforeClass_/_AfterClass_.

ajs6f: Actually, maybe this would make a good JUnit [ClassRule](https://github.com/junit-team/junit4/wiki/Rules#classrule). Not suggesting we worry too much about that now or block anything for it-- just thinking about next steps.

afs:
`AbstractFusekiTest` -- it isn't so simple though as there is only one super class and it ends up competing for that. And the Auth tests need to be different again.

So this PR is just Fuseki1, Fuseki2 and applied to our own code. Each Fuseki has its own needs.  jena-jdbc-drive-remote is the one instance of something using a Fuseki as a test server while not testing Fuseki.  When I realized that, it became clear why the idea of only one framework for both uses was proving a bit messy.

A framework to use Fuseki for testing use of SPARQL protocols can go in the embedded server (= no file footprint).

Testing Fuseki2 does need the full server, file footprint and all.

afs:
(Wrong button again)

PR #180, opened 2016-10-18
JENA-1248: Fuseki registry update

afs:
Also includes the test server control code ported from jena-fuseki-core.


afs:
(rebased to Apache master branch)

rvesse:
Looks good to me

My only concern is how Shiro behaves in a environment where each invaded server may have a different security configuration.

afs:
Embedded Fuseki does not use Shiro.

Fuseki (full) uses Shiro setup via the servlet context (not that Fuseki)full is aimed at multiple servers per JVM: better either 2 ports, one server, one JVM or multiple servers, multiple JVMs.

PR #181, opened 2016-10-21
SORT_ARGS behavior

ajs6f:
Allow env variable SORT_ARGS to be used for multiflag sort args
Do not override parallel flag if it is set on CLI in --sort-args or in env SORT_ARGS

### DIFF
--- a/apache-jena/bin/tdbloader2index
+++ b/apache-jena/bin/tdbloader2index
@@ -150,7 +150,9 @@ LOC=
 KEEP_WORK=0
 DEBUG=0
 JVM_ARGS=
-SORT_ARGS=
+if [ -n $SORT_ARGS ]; then
+  echo "Using SORT_ARGS: $SORT_ARGS"
+fi
 
 while [ $# -gt 0 ]
 do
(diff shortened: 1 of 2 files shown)


apache-jena/bin/tdbloader2index
```diff
@@ -150,7 +150,9 @@ LOC=
 KEEP_WORK=0
 DEBUG=0
 JVM_ARGS=
-SORT_ARGS=
+if [ -n $SORT_ARGS ]; then
```
afs: Need to use "" or use `[[ ]]`

apache-jena/bin/tdbloader2index
```diff
@@ -150,7 +150,9 @@ LOC=
 KEEP_WORK=0
 DEBUG=0
 JVM_ARGS=
-SORT_ARGS=
```
afs: Can use `SORT_ARGS="${SORT_ARGS:-}"` to set to environment or empty string.

ajs6f: Nice. I'll fix this.

apache-jena/bin/tdbloader2index
ajs6f: Ah, right. Will fix.

PR #184, opened 2016-10-27
JENA-1252: Tokenizer cleaup

afs:
Puts in place the structure for warnings not being tokenizer errors.

This is not enabled - it needs testing to see what knock-on effects it has. Ideally, it should be combined with parsers that can do some level of error recovery.

https://issues.apache.org/jira/browse/JENA-1252

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/tokens/TokenizerText.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/tokens/TokenizerText.java
@@ -61,31 +61,30 @@ public final class TokenizerText implements Tokenizer
     private boolean finished = false ;
     private TokenChecker checker = null ;
 
-    // The code assumes that errors throw exception and so stop parsing.
-    private ErrorHandler errorHandler = new ErrorHandler() {
-        @Override
-        public void warning(String message, long line, long col) {
+    private static class ErrorHandlerTokenizer implements ErrorHandler {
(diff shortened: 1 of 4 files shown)


jena-arq/src/main/java/org/apache/jena/riot/tokens/TokenizerText.java
```diff
@@ -60,6 +61,26 @@
     private boolean finished = false ;
     private TokenChecker checker = null ;
 
+    // The code assumes that errors throw exception and so stop parsing.
+    private ErrorHandler errorHandler = new ErrorHandler() {
```
ajs6f: NM, I see the mutator and accessor below. Still, the default instance could be factored out into a constant, I think.

afs: Will do. It is stateless though a stateless object is not very expensive to create!

The others are in ErrorHandlerFactory.

It maybe it becomes a parameter to the constructor.

Because letting dubious data through can lead to a lot of support load due to downstream database problems, this PR just adds mechanism and does not enable it. Hence it is keeping the default error handler in-file for now.

ajs6f: 👍 to the scheme (not making it too easy to let cruft through).

PR #185, opened 2016-10-28
Jena 1253 : PropertyImpl

afs:
Fix for JENA-1253.


afs:
Invalid - it has picked up changes from another branch.

PR #188, opened 2016-11-06
JENA-1197 return exit code 1 when there is a problem validating a file with riot

kinow:
A pull request with a simple approach for JENA-1197. When validate is checked, errors parsing a file will also be reported back, through a `CmdException`.

Quite hard to write a unit test for this, so did some simple regression test locally.

```
kinow@localhost:~/Development/java/jena/jena/apache-jena/bin$ JENA_HOME=/home/kinow/Desktop/JENA/apache-jena-3.1.1 ./riot --validate /tmp/invalid.ttl ; echo $?
23:34:46 ERROR riot                 :: [line: 1, col: 1 ] Out of place: [KEYWORD:I]
0
kinow@localhost:~/Development/java/jena/jena/apache-jena/bin$ JENA_HOME=../target/apache-jena-3.1.2-SNAPSHOT ./riot --validate /tmp/invalid.ttl ; echo $?
23:34:55 ERROR riot                 :: [line: 1, col: 1 ] Out of place: [KEYWORD:I]
1
kinow@localhost:~/Development/java/jena/jena/apache-jena/bin$ JENA_HOME=/home/kinow/Desktop/JENA/apache-jena-3.1.1 ./riot --validate ../../jena-sdb/testing/Assembler/graph-assembler.ttl ; echo $?
0
kinow@localhost:~/Development/java/jena/jena/apache-jena/bin$ JENA_HOME=../target/apache-jena-3.1.2-SNAPSHOT ./riot --validate ../../jena-sdb/testing/Assembler/graph-assembler.ttl ; echo $?
0
```

For invalid files, the new code prints the same as before, but exists with 1, rather than 0. For valid files, the behavior is consistent too.

Sending a pull request as we are in the middle of a release :-)

Cheers
Bruno

### DIFF
--- a/jena-cmds/src/main/java/riotcmd/CmdLangParse.java
+++ b/jena-cmds/src/main/java/riotcmd/CmdLangParse.java
@@ -285,6 +285,9 @@ public abstract class CmdLangParse extends CmdGeneral
             // System.err.println("++++"+ex.getMessage());
             if ( modLangParse.stopOnBadTerm() )
                 return ;
+            if ( modLangParse.validate() )
+                // The error handler will print the exception. Here we throw it so that the command exit with 1
+                throw new CmdException();
         } finally {
             // Not close the output - we may write again to the underlying output stream in another call to parse a file.  
             IO.close(in) ;


ajs6f:
I'm a little mistrustful of a boolean return value there because of an problem @osma dealt with recently wherein he wanted a parse to sort-of succeed but drop bad tuples. Maybe we could return there a marker  `interface ParseCompletion {}` (better name needed) and an out-of-the-box `enum StandardCompletion implements ParseCompletion {SUCCEEDED, FAILED}`? I know that's more complex, but not hugely so and it would bake in some room for expansion or extension.

afs:
The parser has already thrown RiotException by the time this PR comes into play. boolean is to note there is "exit(1)" and not "exit(1)" states.   Boolean, enums or exception are all 2-state returns for that condition; exception and message output or not. This is all within the single class of the command itself

Dropping bad triples needs parser changes. It is a much wider change.

Indeed, there is an argument not to design to a future feature until that feature is fully described!

ajs6f:
Well, the `enum` is just a convenience. It's the `interface` that matters, and the new condition is that parsing completed but tuples were dropped. `exit(n)`? But I admit that I don't understand all of what @osma would want anyway.

afs:
As long as that `n` is not a count! Exit codes are "enums", not a sneaky way to return a variable result.

ajs6f:
Oh, goodness no-- I just used `n` because I'm imagining a series of exit codes (e.g. 0 = success, 3 = partial success (tuples dropped for bad URIs, see STDERR for details), 4 = partial success (tuples dropped for bad syntax, see STDERR for details), etc.) None of which do I have actually planned or am I ready to write, so I guess I'm done talking. :)

afs:
Suggestion: we merge this as is, write a proposal on JENA-1197 that each call of a single file parse returns an object describing the outcome and `exec$()` aggregates the results into a final outcome, including setting the exit code.

kinow:
>Suggestion: we merge this as is, write a proposal on JENA-1197 that each call of a single file parse returns an object describing the outcome and exec$() aggregates the results into a final outcome, including setting the exit code.

Yup, agree. Here's the output when running a command with one valid file, followed by invalid files.

```
$ riot --validate /tmp/valid.ttl /tmp/invalid1.ttl /tmp/invalid2.ttl

INFO  File: /tmp/valid.ttl
INFO  File: /tmp/invalid1.ttl
ERROR [line: 1, col: 4 ] Unrecognized: FDS
```

The valid file is a cheeses ttl file found in Jena sources. The invalid file has some random data.

```
cat /tmp/invalid.ttl 
fdsfd
```

Maybe we could merge and fix JENA-1197, then file a new issue to accommodate the more complete changes suggested in this pull request.

PR #189, opened 2016-11-07
JENA-1258 NodeFmtLib calls JenaSystem initializer

stain:
Fixes [JENA-1258](https://issues.apache.org/jira/browse/JENA-1258) in `NodeFmtLib`

Note that the added `TestNodeFmtInit` test case would not normally find the bug ran from Maven as earlier tests would have caused initialization, but it can be tested from say Eclipse.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/out/NodeFmtLib.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/out/NodeFmtLib.java
@@ -34,6 +34,7 @@ import org.apache.jena.riot.system.* ;
 import org.apache.jena.shared.PrefixMapping ;
 import org.apache.jena.sparql.ARQConstants ;
 import org.apache.jena.sparql.core.Quad ;
+import org.apache.jena.system.JenaSystem;
 
 /** Presentation utilitiles for Nodes, Triples, Quads and more.
  * <p>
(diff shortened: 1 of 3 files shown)


afs:
This usage is calling into the middle of internal code.

Fixing output code seems wrong. It is better to initialize as objects get created, not on output. This shows the root cause is somewhere else.

stain:
Except for `impl` packages and suffixes it is never easy for Jena users to know what is "internal code". 

If it's in [Javadoc](https://jena.apache.org/documentation/javadoc/arq/org/apache/jena/riot/out/NodeFmtLib.html), and it is useful "outside" without much hazzle, then I don't see why it should be called "internal" just because it is also used internally.

Agree that it would be good to ensure Jena initialization from `NodeFactory`.

afs:
The root cause:

Commit 204f253 switched Context to using a `ConcurrentHashMap`, commit d67a426 added a check for null otherwise nothing has changed since 3.1.0.

afs:
The better fix is to initialize in `NodeFactory` (and `Node` for legacy reasons) only. I have tested this in a build.

There is little point in a specific test because tests run in the overall suite; it's almost confusing to have a test that doesn't. With initialization in `NodeFactory` this will get tested anyway!

Also, it would help to unroll ARQConstants (for RDF, RDFS, OWL, XSD) so it does not need JenaSystem.init and then it is, with care, init-safe.

See PR#192 for that all made real.

The original example works when run separately to ensure minimal system initialization.

stain:
Yes, your pull request looks like the best approach.

PR #193, opened 2016-11-16
JENA-1260

der:
Deterministic testcase for #JENA-1260. Tests the entangling of TransitiveEngine state by careful prodding of a TransitiveReasoner (avoiding full RDFS machinery) - fails prior to the fix for JENA-1260 and passes afterwards.

### DIFF
--- a/jena-core/src/test/java/org/apache/jena/reasoner/test/TestReasoners.java
+++ b/jena-core/src/test/java/org/apache/jena/reasoner/test/TestReasoners.java
@@ -18,26 +18,46 @@
 
 package org.apache.jena.reasoner.test;
 
-import junit.framework.TestCase;
-import junit.framework.TestSuite;
-
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 
-import org.apache.jena.graph.* ;
-import org.apache.jena.ontology.* ;
-import org.apache.jena.rdf.model.* ;
-import org.apache.jena.reasoner.* ;
+import org.apache.jena.graph.Factory;
+import org.apache.jena.graph.Graph;
+import org.apache.jena.graph.Node;
+import org.apache.jena.graph.NodeFactory;
+import org.apache.jena.graph.Triple;
+import org.apache.jena.ontology.OntClass;
+import org.apache.jena.ontology.OntModel;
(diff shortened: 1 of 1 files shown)


jena-core/src/test/java/org/apache/jena/reasoner/test/TestRDFSReasoners.java
```diff
@@ -95,6 +95,7 @@ public static TestSuite suite() {
     /**
      * Build a single named query test
      */
+    @SuppressWarnings("unused")
     private static void constructSingleQuerytests(TestSuite suite, String manifest, String test, ReasonerFactory rf, Resource config) throws IOException {
```
afs: No other change in this file.

And it introduces a warning when eclipse is set to not to have "unused" warnings. (Seems to be a "no win" situation in Eclipse.)  As Jena has a lot of warnings if "unused" warnigns are turned on, for now, can be just drop this annotation.

jena-core/src/test/java/org/apache/jena/reasoner/test/TestReasoners.java
```diff
@@ -333,6 +352,32 @@ public void testTransitiveBindReuse() {
     }
     
     /**
+     * Test that two transitive engines are independent.
+     * See JENA-1260
+     */
+    public void testTransitiveEngineSeparation() throws InterruptedException {
+        String NS = "http://example.com/test#";
```
afs: Test passes for me.

jena-core/src/test/java/org/apache/jena/reasoner/test/TestRDFSReasoners.java
der: Yes, sorry. I shouldn't have included that, as well as being cosmetic (and unsuccessfully cosmetic!) it's not relevant to this test case.

Would you like me to resubmit the PR without this?

afs: PRs are "live" - if the your branch changes so does the PR.

If you edit your branch and "push" it will change here as well.

To get rid of it, edit the file, then "git squash" or "git add" then "git commit --amend" followed by "git push --force" and it will disappear from the PR.

Or I can clean it after pulling it. This isn't a huge piece of work.

PR #195, opened 2016-12-04
JENA-1197: Record parse run outcomes

afs:
See discussion on PR #188.

This PR incorporates #188 and returns a record describing the outcome of each call of the parser.  The records are retained, then the command can decide how to return/exit based on all parser attempts.


afs:
And some clearing up / simplifying.

PR #196, opened 2016-12-11
JENA-1265: Setting for GraphView Capabilities.

afs:


afs:
@ajs6f This affects TIM.  I tested and TIM is term based but it's use of GraphView claims it handled literal datatyping.

(Ah - I see you beat me to it!)

ajs6f:
@afs Hm. [GraphInMemory](https://github.com/apache/jena/blob/master/jena-arq/src/main/java/org/apache/jena/sparql/core/mem/GraphInMemory.java) doesn't override for its own capabilities. Sounds like it should have been doing that, but now this change solves that problem (because of [this](https://github.com/apache/jena/pull/196/files#diff-d154f363df7630907dac5f0ba5d539a8R218)), right?

afs:
Yes - this makes overriding unnecessary. The second commit on this PR takes it out of GraphTDB for the same reason.

Ideally, maybe, GraphBase should provide a safer set (no iterator remove, no value indexing) and GraphMem / GraphMemBase add the specific capabilities.

Then the InfGraphs are probably inconsistent as their capabilities partially rely on their base graph.

More evidence Capabilities are not used for real.

PR #197, opened 2016-12-16
JENA-1267: RDFConnection

afs:
An API that brings together SPARQL APIs.

2 new modules:
- jena-rdfconnection
- jena-integration-tests


afs:
The "test" module is needed because the tests need a Fuseki test server but  jena-rdfconnection is before Fuseki in the build sequence.

Some tests in jena-fuseki-core that are DatasetAccess and query, update SPARQL API calls could similarly be moved to this new location.

afs:
Documentation: https://github.com/afs/jena/blob/rdf-connection/jena-rdfconnection/Documentation.md

pom.xml
```diff
@@ -155,6 +159,12 @@
         <module>jena-fuseki2</module>
         <module>jena-permissions</module>
 
+        <!-- Tests of artifacts that require
```
ajs6f: Is this comment missing some text?

ajs6f:
Is the basic idea of `jena-integration-tests` to provide a place for inter-module tests?

jena-integration-tests/README.md
```diff
@@ -0,0 +1,11 @@
+Jena Intrgration Tests
```
rvesse: Typo r -> e

jena-integration-tests/src/test/java/org/apache/jena/test/rdfconnection/AbstractTestRDFConnection.java
```diff
@@ -0,0 +1,382 @@
...
+import org.apache.jena.system.Txn ;
+import org.junit.Assume ;
+import org.junit.Test ;
+
+public abstract class AbstractTestRDFConnection extends BaseTest {
+    // Testing data.
+    static String DIR = "testing/RDFConnection/" ;
+    
+    protected abstract RDFConnection connection() ;
+    // Not all RDFConenction types support.
```
rvesse: Typo: conenction -> connection

jena-integration-tests/src/test/java/org/apache/jena/test/rdfconnection/AbstractTestRDFConnection.java
```diff
@@ -0,0 +1,382 @@
...
+    //  (may acquite RDFConnection.supportTransactionalAbort but ATM that isn't included)
```
rvesse: Typo: acquite -> acquire

jena-rdfconnection/Documentation.md
```diff
@@ -0,0 +1,205 @@
...
+for the transaction body.
+
+### Remote Transactions
+
+SPARQL does not define a remote transaction standard protocol. Each remote
+operation shuld be atomic (all happens or nothing happens) - this is the
+responsibility of the remote server.
+
+An `RDFConenction` will at least provide the client-side locking features.
+This means that overlapping operations that chnage data are naturally
```
rvesse: Typo: chnage -> change

jena-rdfconnection/Documentation.md
```diff
@@ -0,0 +1,205 @@
...
+## Query Usage
+
+`RDFConnection` provides methods for each of the SPARQL query forms (`SELECT`,
+`CONSTRUCT`, `DESCRIBE`, `ASK`) as well as a way to get the lower level
+`QueryExecution` for specialized configuration.
+
+When creating an `QueryExecution` explicitly, care shoud be taken to close
+it. If the application wishes to capture the result set from a SELECT query and
+retain it across the lifetime of the transaction or `QueryExecution`, then
+the application create a copy which is not attached to any external system
```
rvesse: Missing word: the application *should* create

jena-rdfconnection/Documentation.md
```diff
@@ -0,0 +1,205 @@
...
+         conn.update("DELETE DATA { ... }" ) ;
+         conn.load("data.ttl") ;
+         }) ;
+```
+
+## Dataset operations
+
+In addition to the SPARQL Graph Store Protocol, operations on whole
+datasets are provided for fetching (HTTP GET), adding data (HTTP POST) and
+settign the data (HTTP PUT) on a dataset URL.  This assumes the remote
```
rvesse: Typo: settign -> setting

jena-rdfconnection/Documentation.md
```diff
@@ -0,0 +1,205 @@
...
+operations are visible to the called code.
+
+* query via `SparqlQueryConnection`
+* update via `SparqlUpdateConnection`
+* graph store protocol `RDFDatasetAccessConnection` (read operations),
+   and `RDFDatasetConnection` (read and write operations).
+
+## Examples
+
+https://github.com/afs/jena-rdfconnection/tree/master/src/main/java/rdfconnection/examples
```
rvesse: TODO: update this to official repository after merging

jena-rdfconnection/src/main/java/org/apache/jena/rdfconnection/RDFConnectionFactory.java
```diff
@@ -0,0 +1,84 @@
...
+                                        String updateServiceEndpoint,
+                                        String graphStoreProtocolEndpoint) {
+        return new RDFConnectionRemote(queryServiceEndpoint, updateServiceEndpoint, graphStoreProtocolEndpoint) ;
+   }
+
+    
+    /** Create a connection to a remote location by URL.
+     * This is the URL for the dataset.
+     * 
+     *  This call requires specifying the names of the servicerelative to the dataset URL.
```
rvesse: Missing space

jena-rdfconnection/src/main/java/org/apache/jena/rdfconnection/RDFConnectionRemote.java
```diff
@@ -0,0 +1,478 @@
...
+        this(null, destination, sQuery, sUpdate, sGSP) ;
+    }
+    
+    /** Create connection, using URL of the dataset and short names for the services */
+    public RDFConnectionRemote(HttpClient httpClient, String destination, String sQuery, String sUpdate, String sGSP) {
+        this.destination = destination ;
+        this.svcQuery = formServiceURL(destination,sQuery) ;
+        this.svcUpdate = formServiceURL(destination,sUpdate) ;
+        this.svcGraphStore = formServiceURL(destination,sGSP) ;
+//        if ( httpClient == null )
```
rvesse: I am guessing that the null check is unnecessary because passing a null client to `HttpOp` causes the default client to be used?

jena-rdfconnection/Documentation.md
afs: Done, to avoid forgetting later.

afs:
Thank you for the reviews - types corrected.

afs:
> Is the basic idea of jena-integration-tests to provide a place for inter-module tests?

The driver is that testing remote operations needs a SPARQL server but that is not a dependency of the code, only of the tests.  However, Fuseki isn't built at this point in the build process.

jena-rdfconnection goes into the apache-jena-libs and apache-jena artifacts but Fuseki is built after those.

This must be a common need in system so if there is a better common pattern, then I'm happy to switch to that.

There are a number of ways this could be done:

1. Build modules in a different order - e.g. do building steps before delivery modules.
2. Put the tests in Fuseki
3. As here, a specific module.
4. Use a previous version of Fuseki.

(1) works in this case but it may not work out whereas the idea of an integration testing module, seems general. (2) is a less clean version of (1); this is already done in a small way - those can be moved now. (4) seems accident prone.

So (3) looks the best way and may be applicable for other situations. It may even to help unblock evolution of the legacy module/packing structure (the old Turtle reader in jena-core is needed for assembler tests - I haven't had a chance to try yet).

afs:
I wasn't happy with a single fixed way to use a local connection - I think all of copy (pseudo remote), no isolation and read-only (no copy so a pragmatic balance for larger datasets) make sense in different situations.

so I have added two isolation modes added to the documentation and added tests.

ajs6f:
I am completely cool with the idea of separate i-test modules-- I think that's very much the right idea. Is the thinking that we can move it later in the build as need be (i.e. if we wanted to include some i-tests for some later modules)?

afs:
Its location in the build shouldn't matter too much, just after all the modules it is testing.

ajs6f:
Right, cool. Just wanted to make sure we leave the door open to move it later in the build to include testing of later modules. +1

PR #199, opened 2016-12-23
Use some library functions

ajs6f:


ajs6f:
Just replacing some homegrown stuff with library functions that came out since, and pruning two classes, one for which an equivalent is now available in library code, and the other wasn't used at all and never has been AFAICT.

ajs6f:
Commits are orthogonal. I'll wait until the top of the week in case anyone has any comments/objections.

ajs6f:
Also will send a further commit getting rid of calls to now-deprecated functions.

jena-base/src/main/java/org/apache/jena/atlas/lib/StrUtils.java
```diff
@@ -33,32 +35,34 @@ private StrUtils() {}
...
+        return String.join("\n", args);
     }
 
-    /** Concatentate strings, using a separator */
+    /** 
+     * Concatentate strings, using a separator
+     * 
+     * @deprecated Prefer String.join(sep, args)
+     */
+    @Deprecated
```
afs: If we are going to deprecate this, then I really would like to see the changes propagated through the codebase, not left pending.

Currently, at least for me, Jena is warning free in the main modules so I can see when warnings appear. Having unfinished deprecations means that it is harder to notice warnings appearing.

jena-base/src/main/java/org/apache/jena/atlas/lib/StrUtils.java
```diff
@@ -33,32 +35,34 @@ private StrUtils() {}
...
     public static String strjoinNL(List<String> args) {
-        return join("\n", args);
+        return String.join("\n", args);
     }
 
-    /** Concatentate strings, using a separator */
+    /** 
+     * Concatentate strings, using a separator
+     * 
+     * @deprecated Prefer String.join(sep, args)
```
afs: Add to javadoc `This function will be removed - do not use.`

jena-base/src/main/java/org/apache/jena/atlas/lib/StrUtils.java
```diff
@@ -33,32 +35,34 @@ private StrUtils() {}
...
+        return String.join(sep, args);
     }
 
-    /** Concatentate strings, using a separator */
+    /**
+     * Concatentate strings, using a separator
+     * 
+     * @deprecated Prefer String.join(sep, args)
+     */
+    @Deprecated
```
afs: Ditto.

jena-base/src/main/java/org/apache/jena/atlas/lib/StrUtils.java
```diff
@@ -85,17 +89,20 @@ public static int strCompareIgnoreCase(String s1, String s2) {
...
     }
     
+    /**
+     * @param x
+     * @return &lt;null&gt; if x == null, otherwise, x.toString()
+     * 
+     * @deprecated prefer {@link Objects#toString(Object, String)} with "&lt;null&gt;" or just {@link Objects#toString(Object)} 
+     */
+    @Deprecated
     public static String str(Object x) {
```
afs: I suggest we leave this, and not `@Deprecated`.  `str()` is supposed to be reminiscent of SPARQL `STR`.  `str(...)` is used elsewhere so this is a pattern.

ajs6f:
@afs , yep there's another commit coming getting rid of those warnings. Just an oversight on my part, sorry.

jena-base/src/main/java/org/apache/jena/atlas/lib/StrUtils.java
ajs6f: all right, if that seems clearer.

ajs6f:
Weird, Travis blew out with No Heap Space...

https://api.travis-ci.org/jobs/186431648/log.txt?deansi=true

ajs6f:
Yeah, it does a perfectly nice full build (no profile) for me. Not sure what Travis' problem is. I never even noticed that `src-examples` dir-- I'll add a commit to clean that up.

ajs6f:
I'll wait for any further comment until Tuesday.

ajs6f:
Thanks, I got caught up with holiday stuff and forgot to merge this.

PR #200, opened 2016-12-24
JENA-1270: Deprecate some methods of Capabilities

afs:
This PR deprecates the following interface methods and clear up the code to align with these deprecations.

- `addAllowed( everyTriple )`
- `deleteAllowed( everyTriple )`
- `iteratorRemoveAllowed()`
- `canBeEmpty()`
- `findContractSafe()`

It does not touch `handlesLiteralTyping` at this stage.


afs:
If anyone can explain the issue noted in `xmloutput/impl/Abbreviated.java`, and especially provide a test case if it is still relevant, that would be great.

ajs6f:
👍 to deprecating the whole `Capabilities` idiom, so 👍 to this.

ajs6f:
All I could find re: 803804 and 858163 was [this](https://osdir.com/ml/misc.ontology.protege.owl/2004-04/msg00904.html). Something to do with the old, old confusion between XML Namespaces and RDF prefix management and numerical localnames?

afs:
It seems to be something to do with how `find` behaves if called multiple times with the same arguments as needed by pretty printing in XML. Turtle does not perform an equivalent test.  I don't recall that warning message ever being mentioned on users@ so I'm tempted to assume it is no longer relevant.

afs:
Commented capability checks removed.

PR #202, opened 2017-01-07
Formatting rampage

ajs6f:
Just some midwinter cleanup. Using library methods from `j.util.Collections` instead of Jena-specific code, removing some unused imports.


jena-base/src/main/java/org/apache/jena/atlas/lib/CollectionUtils.java
```diff
@@ -22,12 +22,7 @@
...
 public class CollectionUtils
 {   
-    public static <T> boolean disjoint(Collection<T> c1, Collection<T> c2)
-    {
-		return c1.stream().noneMatch(c2::contains);
-    }
-    
-    /** Test for same elements, regarless of cardinality */ 
+    /** Test for same elements, regardless of cardinality */ 
     public static <T> boolean sameElts(Collection<T> left, Collection<T> right) {
```
afs: Is there a better (library code) way to do the `sameElts` functionality?

jena-arq/src/main/java/org/apache/jena/riot/thrift/wire/RDF_ANY.java
```diff
@@ -23,7 +23,7 @@
   private static final org.apache.thrift.protocol.TStruct STRUCT_DESC = new org.apache.thrift.protocol.TStruct("RDF_ANY");
 
 
-  private static final Map<Class<? extends IScheme>, SchemeFactory> schemes = new HashMap<Class<? extends IScheme>, SchemeFactory>();
```
afs: Autogenerated file. I think we should leave them alone.

jena-core/src/main/java/org/apache/jena/rdfxml/xmlinput/impl/XMLHandler.java
```diff
@@ -374,7 +374,7 @@ public void initParse(String base, String lang) throws SAXParseException {
         idsUsed = 
         	ignoring(WARN_REDEFINITION_OF_ID)?
         			null:
```
afs: From here onwards the changes look good.

jena-base/src/main/java/org/apache/jena/atlas/lib/SetUtils.java
```diff
@@ -44,13 +42,26 @@ private SetUtils() {}
...
         Set<T> s3 = new HashSet<>(s1) ;
         s3.addAll(s2) ;
         return s3 ;
     }
 
-    /** union difference intersection : those elements in s1 or s2 but not both. */
+    /**
+     * @param s1 a {@link Set}
+     * @param s2 a {@link Set}
+     * @return eager union difference intersection : those elements in {@code s1} or {@code s2} but not both
```
afs: What does "eager union difference intersection" mean?

The difference is that this one isolates (calculates) the operation so is safe to future changes, and allows it to be further modified

the google.common one isn't "lazy" as I understand that term because it is a view (and can change), not a delayed calculation.

jena-base/src/main/java/org/apache/jena/atlas/lib/DS.java
```diff
@@ -1,45 +0,0 @@
-/*
```
afs: Good idea.

jena-base/src/main/java/org/apache/jena/atlas/lib/SetUtils.java
ajs6f: "Eager" is my insertion, to contrast with the Guava one, but "union difference intersection" comes from the previous code-- I have no idea what that means! :)

ajs6f:
I will back out the mods to autogen code-- certainly there's no point in tweaking that. I'm still very much learning what is and isn't autogen in ARQ and how it all works.

jena-base/src/main/java/org/apache/jena/atlas/lib/SetUtils.java
ajs6f: If you don't like the way I'm using those terms, I'll remove them. My real purpose is just to make sure people know they can choose between these guys (which do calculations immediately) and the Guava ones, which amortize the cost over accesses.

afs: Symmetric set difference == the-union set-difference the-intersection.

set-difference(A,B) is elts in A but not in B
Symmetric set difference(A,B) is elts in A or B but not both.

afs:
The auto-generated code has auto-generated comments to that effect at the top :-) ... which Eclipse then goes and hides :-(.

RAT seems to have some built-in detection code because these files don't trigger a validation failure.

jena-base/src/main/java/org/apache/jena/atlas/lib/SetUtils.java
ajs6f: Fixed.

ajs6f:
Autogen code is now back where it started.

jena-base/src/main/java/org/apache/jena/atlas/lib/CollectionUtils.java
ajs6f: You know, you would think there would be, but I didn't find anything that  ignores both cardinality and order and works for all `Collection`s. OTOH, I just check and nothing in Jena actually _uses_ this function except its own tests! So I wouldn't be averse to just dropping it. Anyone who was using it should have no difficulty in-lining it or inserting their own.

ajs6f:
@afs  Are you cool with this now?

jena-arq/src/main/java/org/apache/jena/sparql/algebra/OpVisitor.java
```diff
@@ -23,53 +23,53 @@
...
     // Op0
-    public void visit(OpBGP opBGP) ;
-    public void visit(OpQuadPattern quadPattern) ;
-    public void visit(OpQuadBlock quadBlock) ;
-    public void visit(OpTriple opTriple) ;
-    public void visit(OpQuad opQuad) ;
-    public void visit(OpPath opPath) ;
-    public void visit(OpTable opTable) ;
-    public void visit(OpNull opNull) ;
+    default void visit(OpBGP opBGP) {}
```
afs: I don't know about this change. 

There are `OpVisitor` impls that do not use `OpVisitorBase` and that has the desirable (and intended) effect that a change in `OpVisitor` forces implementation. e.g. `OpVisitorByType` e.g. `ExecutionDispatch`

Having default methods masks this so it makes things less robust. There is rarely a general implementation of an `visit` that will work for everything.

jena-base/src/main/java/org/apache/jena/atlas/lib/CollectionUtils.java
afs: Please leave.

jena-arq/src/main/java/org/apache/jena/sparql/algebra/OpVisitor.java
ajs6f: I figured the general implementation of `visit` is NOOP, but I will pull this out.

jena-base/src/main/java/org/apache/jena/atlas/lib/CollectionUtils.java
ajs6f: It has been left. It is not deleted. (?!)

afs: I was responding to the suggested of dropping it (in 8/Jan 15:33 GMT).

ajs6f: Ah, got it-- okay, I will pull out the changes to the `OpVisitor` type and then I think we are good?

afs:
Modified but is generated:
JSON_Parser, SPARQLParser10,  SPARQLParser11, SSE_ParserCore

OpVisitorBase is modified.

Various uses of "disjoint" have had the static import removed.

I have not checked:
testing_framework
jena-csv
jena-jdbc

ajs6f:
Fixed (reverted)  the additional autogen files.

PR #203, opened 2017-01-09
JsonLDReader: possibility to override the @context

fpservant:
following a question on the user list: "[Overriding @context when loading JSON-LD](http://jena.markmail.org/message/uf3zkzybv7gb4crt)",
a way using the jena Context mechanism to do just that.
This also allows to not depend on jsonld-java to download a context defined by a URI.


jena-arq/src/test/java/org/apache/jena/riot/TestJsonLDReader.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.jena.sparql.util.Context;
+import org.apache.jena.vocabulary.RDF;
+import org.junit.Test;
+
+import com.fasterxml.jackson.core.JsonGenerationException;
+import com.github.jsonldjava.utils.JsonUtils;
+
+public class TestJsonLDReader {
+
+    @Test public final void test() {
```
ajs6f: Perhaps a more descriptive name?

jena-arq/src/test/java/org/apache/jena/riot/TestJsonLDReader.java
```diff
@@ -0,0 +1,81 @@
...
+        try {
+            String jsonld = "{\"@id\":\"_:b0\",\"@type\":\"http://schema.org/Person\",\"name\":\"John Doe\",\"@context\":\"http://schema.org/\"}";
+            StringReader reader = new StringReader(jsonld);
+            Model m = ModelFactory.createDefaultModel();
+            m.read(reader, null, "JSON-LD");
+            m.write(System.out, "TURTLE");
+        } catch (Throwable e) {
+            e.printStackTrace();
```
ajs6f: Not so great to print stacktraces to STDOUT, even in an error situation.

jena-arq/src/test/java/org/apache/jena/riot/TestJsonLDReader.java
```diff
@@ -0,0 +1,81 @@
...
+
+public class TestJsonLDReader {
+
+    @Test public final void test() {
+        try {
+            String jsonld = "{\"@id\":\"_:b0\",\"@type\":\"http://schema.org/Person\",\"name\":\"John Doe\",\"@context\":\"http://schema.org/\"}";
+            StringReader reader = new StringReader(jsonld);
+            Model m = ModelFactory.createDefaultModel();
+            m.read(reader, null, "JSON-LD");
+            m.write(System.out, "TURTLE");
```
ajs6f: Printing test data to STDOUT is not a good thing. If the test needs to serialize data to prove out, perhaps a dummy `StringWriter` would serve?

jena-arq/src/test/java/org/apache/jena/riot/TestJsonLDReader.java
```diff
@@ -0,0 +1,81 @@
...
+        jenaCtx.set(JsonLDReader.JSONLD_CONTEXT, jsonldContextAsMap);
+        
+        // read the jsonld, replacing its "@context"
+        InputStream in = new ByteArrayInputStream(jsonld.getBytes(StandardCharsets.UTF_8));
+        Dataset ds = DatasetFactory.create();
+        ReaderRIOT reader = RDFDataMgr.createReader(Lang.JSONLD);
+        reader.read(in, null, null, StreamRDFLib.dataset(ds.asDatasetGraph()), jenaCtx);
+        
+        Model m = ds.getDefaultModel();
+        m.write(System.out, "TTL");
```
ajs6f: Again, writing data to STDOUT isn't ideal.

jena-arq/src/test/java/org/apache/jena/riot/TestJsonLDReader.java
```diff
@@ -0,0 +1,81 @@
...
+        
+        // pass the jsonldContext to the read using a jena Context
+        Context jenaCtx = new Context();
+        Object jsonldContextAsMap = JsonUtils.fromInputStream(new ByteArrayInputStream(jsonldContext.getBytes(StandardCharsets.UTF_8)));
+        jenaCtx.set(JsonLDReader.JSONLD_CONTEXT, jsonldContextAsMap);
+        
+        // read the jsonld, replacing its "@context"
+        InputStream in = new ByteArrayInputStream(jsonld.getBytes(StandardCharsets.UTF_8));
+        Dataset ds = DatasetFactory.create();
+        ReaderRIOT reader = RDFDataMgr.createReader(Lang.JSONLD);
```
ajs6f: This whole section seems complex. How about `org.apache.jena.riot.RDFDataMgr.read(Model, StringReader, String, Lang)`?

jena-arq/src/test/java/org/apache/jena/riot/TestJsonLDReader.java
```diff
@@ -0,0 +1,81 @@
...
+        // a subset of schema.org that can be used as @context for jsonld
+        String jsonldContext = "{\"name\":{\"@id\":\"http://schema.org/name\"},\"Person\": {\"@id\": \"http://schema.org/Person\"}}";
+        
+        // pass the jsonldContext to the read using a jena Context
+        Context jenaCtx = new Context();
+        Object jsonldContextAsMap = JsonUtils.fromInputStream(new ByteArrayInputStream(jsonldContext.getBytes(StandardCharsets.UTF_8)));
+        jenaCtx.set(JsonLDReader.JSONLD_CONTEXT, jsonldContextAsMap);
+        
+        // read the jsonld, replacing its "@context"
+        InputStream in = new ByteArrayInputStream(jsonld.getBytes(StandardCharsets.UTF_8));
```
ajs6f: It would be nice to use `try-with-resource` here, because not explicitly closing an `InputStream` can confuse IDEs which then throw warnings.

ajs6f:
This is good stuff-- thanks for taking a problem and turning it into an improvement!

jena-arq/src/test/java/org/apache/jena/riot/TestJsonLDReader.java
fpservant: Hmm, maybe I'm missing something obvious, but I need to pass the Context. I didn't find an easier way (all read methods in RDFDataMgr that take the Context as arg use the uri param to define what we read).

jena-arq/src/test/java/org/apache/jena/riot/TestJsonLDReader.java
fpservant: I changed that, but I get the ```org.apache.jena.riot.RiotException: loading remote context failed: http://schema.org/```here - that's bad, but not related to this PR (which, in fact, gives a way to get rid of the problem)

fpservant:
@ajs6f thanks for the review. I made most (all ?) of the changes that you suggested. Remains the problem "loading remote context failed: http://schema.org/" - unrelated to this PR per se.

Regarding the section that you find too complex: I'd be happy to learn about a better way to read data from an InputStream, that supports the passing of a Context

jena-arq/src/main/java/org/apache/jena/riot/lang/JsonLDReader.java
```diff
@@ -45,9 +45,20 @@
...
 
 public class JsonLDReader implements ReaderRIOT
 {
+    private static final String SYMBOLS_NS = "http://jena.apache.org/riot/jsonld#" ;
+    /**
+     * Symbol to use to pass the "@context" to be used when reading the jsonld
+     * (overriding the actual @context in the jsonld)
+     * Expected value: the value of the "@context", 
+     * as expected by the JSONLD-java API (a Map) */
+    public static final Symbol JSONLD_CONTEXT = Symbol.create(SYMBOLS_NS + "JSONLD_CONTEXT");
```
afs: Ideally, the constants go in `RIOT` so that application code just needs to import RIOT.  There aren't many - only an odd legacy feature - because Turtle/NTriples like writers don't have lots of configuration control.

ajs6f:
@fpservant Oh my goodness, you are quite right. There is no appropriate method in `RdfDataMgr`! @afs I think we could reasonably add such a method as part of this PR-- it would clearly be useful. What say ye?

afs:
With the changes that have happened, I'm not seeing the comment to which you refer.

fpservant:
@afs I think that @ajs6f refers to code:
```
InputStream in = new ByteArrayInputStream(jsonld.getBytes(StandardCharsets.UTF_8));
+        Dataset ds = DatasetFactory.create();
+        ReaderRIOT reader = RDFDataMgr.createReader(Lang.JSONLD);
(etc)
```
he commented:

> This whole section seems complex. How about org.apache.jena.riot.RDFDataMgr.read(Model, StringReader, String, Lang)?

and I replied:

> Hmm, maybe I'm missing something obvious, but I need to pass the Context. I didn't find an easier way (all read methods in RDFDataMgr that take the Context as arg use the uri param to define what we read).

and:
> I'd be happy to learn about a better way to read data from an InputStream, that supports the passing of a Context

ajs6f:
@fpservant @afs Yes, exactly. Having a method on `RdfDataMgr` that takes `InputStream` _and_ `Context` seems very natural.

jena-arq/src/main/java/org/apache/jena/riot/lang/JsonLDReader.java
fpservant: done

afs:
What about a `parse` method?

As discussed before, adding a lot extra functions just for one particular case of one particular language, seems to create unclarity for the general usage.  The read(...URL, Context) cases were added after discussion last time.

Local contents and overriding remote JSON-LD context does not seem to me to be a common need because you can change the local contents.  Test cases often do things that are not common.

(`RDFDataMgr.createReader` is there to expose specifics as well and is in the same style of old Jena readers.)

ajs6f:
You're saying use `org.apache.jena.riot.RDFDataMgr.parse(StreamRDF, InputStream, String, Lang, Context)`? That does make sense. I guess I was thinking of the general need to customize parsing behavior, but as long as there is a reasonably concise form, that's all I care about. @fpservant does that work for you?

jena-arq/src/main/java/org/apache/jena/riot/out/JenaRDF2JSONLD.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Override
+    public RDFDataset parse(Object object) throws JsonLdError {
+        RDFDataset result = new RDFDataset() ;
+        if ( object instanceof DatasetGraph )
+        {
+            DatasetGraph dsg = (DatasetGraph)object ;
+
+            Iterator<Quad> iter = dsg.find() ;
+            for ( ; iter.hasNext() ; )
```
ajs6f: Really more of a style thing, but you could replace these four lines with:
```
dsg.find().forEachRemaining(q->{
```
IOW using `forEachRemaining` with a lambda.

fpservant: Hmm, that's not code that I wrote, so I don't feel like taking the risk of adding a bug to something that works, just for aesthetic reasons...

ajs6f: Oh, that's fine. It shows as code that was added by this PR, strangely.

ajs6f:
This looks nice now. 👍 Please do either file a ticket for the new bug you found or find a ticket that already describes it (if there is one) and reference it in the comment you left in the test case with a `TODO`. That will help me remember to try to fix it! (Unless you want to do that too. :) )

fpservant:
@ajs6f are you speaking of the "org.apache.jena.riot.RiotException: loading remote context failed" thing? 

1) note that in the context of a test, jena doesn't want to depend on a connexion to the network (so told me @afs when I was working on JSON-LD writing - and getting, BTW, same kind of problem, for reasons which are supposed to have been solved since). So the catch Exception must probably stay in the code of the test.

2) I'd be curious to know whether someone else, who is completely sure about its settings, also gets this problem (problems with http connexions may be many different things!) (Other people experiment it, cf. question on jena-users that initiated this PR, but are they up-to-date?). So I would be happy to get confirmation of the problem from others.

I tried to track it down. It occurs to me at ```com.github.jsonldjava.core.DocumentLoader.loadDocument(String)``` line 29

where I get:
```org.apache.http.ConnectionClosedException: Premature end of Content-Length delimited message body (expected: 124346; received: 0```

**Comment by ajs6f** [comment]
@fpservant Yep, that's what I meant. I would be looking that as solvable by moving the location of the schema to a resource supported on the testing box, so it's not being resolved from the network. Note I'm not asking you to do that, just to make a note that someone should.

**Comment by stain** [comment]
For unit tests you can use a JAR-cached context by using the [jarcache.json](https://github.com/jsonld-java/jsonld-java#loading-contexts-from-classpathjar) approach and put it together with the JSON schema file in `src/test/resources/` - that would override the specified URL.

A similar approach can be used with `src/main/resources` - but then of course you would have to keep the schema cache up to date, so that's more for the end application to do to avoid external network dependencies.

Note that the actual [schema.org context file](http://schema.org/docs/jsonldcontext.json) is [licensed as CC BY SA](http://schema.org/docs/terms.html), but you could make a minimal version made just for the test which would not infringe copyright.

**Comment by afs** [comment]
External linkages in the tests is, in the long term, not good. At some point, the remote resource will be uncontactable. Sometimes people need to run the test offline as well. So for the long term robustness of the tests suite, I think the level of need for an external resource needs to be quite high. At least make the test respond to different errors - uncontactable vs a failure with the expected actions after the content is fetched. Even the latter is hard - the remote contents may change.

(this is what `StreamManager` is good for).

File: jena-arq/src/main/java/org/apache/jena/riot/out/JenaRDF2JSONLD.java (hunk 8 above)
Comment (afs) [inline jena-arq/src/main/java/org/apache/jena/riot/out/JenaRDF2JSONLD.java]: Converting loops to `forEachRemaining` is personal taste.

If a loop can be replaces by steram actions, then one thing, but this is turning a loop into a different loop!

`forEachRemaining` is reported to be slower. That might be just that the optimizer has not caught up but something needs to account for the needs of stack frames in case of an exception and that means extra work must be done.

**Comment by ajs6f** [comment]
@afs I think the point that @stain and I are making is that there is no need for an external resource linkage at all.

**Comment by fpservant** [comment]
Hi,

there are very different points in the discussion, all around one test that attempts to connect to an external resource.

1) It must first be noted that the test in question actually doesn't have anything to do with the functionality that the PR proposes. It can as well be removed, and that wouldn't change anything about the object of the PR (the functionality that it proposes to add), nor about the completeness of the testing of that functionality.

If you want, I can remove it.

2) this test attempts to parse some JSON-LD whose @context contains the URI of schema.org - probably the most obvious use-case for reading JSON-LD

    - BTW, I didn't see any test about reading JSON-LD. Are there any?

     - If this test cannot connect to schema.org because there are no external connection, (or for whatever reason) the test won't fail, because the RiotException is catched. Just one line is written to the console (and it is not by the test itself, but by the code that it calls). So this test with this external connection doesn't really cause trouble to the integration process (maybe a few seconds delay - the time for the HTTP client to stop trying to connect)

     - it shows me that I'm not able to read such JSON-LD on my machine (I get on the console: ```org.apache.jena.riot.RiotException: loading remote context failed: http://schema.org/```) - this, in a setting where it should work (standard jena 3.1.1, run from eclipse where I can connect to external resources from java, for instance using jersey client)

So: as far as I can tell (thanks to this test), current jena has a problem reading JSON-LD with @context set to a URI (at least, I have such a problem using current jena)

- it is probably possible to write a test that simulates connecting to an external resource without connecting to it (cf. "cache" mechanism in JSON-LD java), but
	- such a test would probably not detect a problem such as the one we seem to have here
	- it's out of the scope of current PR

**Comment by afs** [comment]
> BTW, I didn't see any test about reading JSON-LD. Are there any?

See `TestJsonLDReadWrite`.

**Comment by afs** [comment]
> current jena has a problem reading JSON-LD 

JsonLDReader catches an error from jsonld-java:
```
        catch (JsonLdError e) {
            errorHandler.error(e.getMessage(), -1, -1); 
            throw new RiotException(e) ;
        }
```
jsonld-java is reporting an error, all Jena does, and can do, is pass that error upwards.

fpservant:
>> current jena has a problem reading JSON-LD

> jsonld-java is reporting an error, all Jena does, and can do, is pass that error upwards.

reformulating: I have problem with current "jena as a whole, including jsonld-java" to read some JSON-LD

and therefore:

> I'd be curious to know whether someone else, who is completely sure about its settings, also gets this problem (or not)

afs:
`curl -v --header 'Accept: application/ld+json' http://schema.org/` returns zeros bytes. 

Is json-ld following redirects?  Where does jsonld-java's httpClient come from?

It is a 302 redirect to `http://schema.org/docs/jsonldcontext.json`. Does using that work? It does not for me but it's not clear it is the same error. (jsonld-java suppresses details).

If you agree this is from jsonld-java, could you close JENA-1276?  The jsonld-java developers aren't going to see it.

What about
* https://github.com/jsonld-java/jsonld-java/issues/188
* https://github.com/jsonld-java/jsonld-java/issues/186

fpservant:
@afs jsonld-java 0.9.0 solves the issue (and I have been able to run maven install on jena-arq with it).

How do you want to handle jsonld-java's version change? Other PR? Shouldn't I wait until then before closing JENA-1276?

afs:
I've done the upgrade to 0.9.0 on master for you. It's slightly easier to separate upgrades from other changes because they need noting in the release cycle.

fpservant:
@afs thanks. JENA-1276 now closed

PR #205, opened 2017-01-13
JENA-1277: don't use sorting in spatial queries, for much better performance

osma:
This PR proposes removing the `distSort` parameter from the Lucene spatial query performed by jena-spatial. Dropping the sorting gives a massive performance boost; in the Geonames example given in JENA-1277, the query time drops from over 20 seconds to less than 200 ms.

I suppose that the sorting is not necessary since jena-spatial results are just raw material for the SPARQL engine anyway.

### DIFF
--- a/jena-spatial/src/main/java/org/apache/jena/query/spatial/SpatialIndexLucene.java
+++ b/jena-spatial/src/main/java/org/apache/jena/query/spatial/SpatialIndexLucene.java
@@ -29,7 +29,6 @@ import org.apache.lucene.document.Document ;
 import org.apache.lucene.document.Field ;
 import org.apache.lucene.document.FieldType ;
 import org.apache.lucene.index.* ;
-import org.apache.lucene.queries.function.ValueSource ;
 import org.apache.lucene.search.* ;
 import org.apache.lucene.spatial.SpatialStrategy ;
 import org.apache.lucene.spatial.prefix.RecursivePrefixTreeStrategy ;
@@ -42,7 +41,6 @@ import org.apache.lucene.util.Version ;
 import org.slf4j.Logger ;
 import org.slf4j.LoggerFactory ;
 
-import com.spatial4j.core.shape.Point ;
 import com.spatial4j.core.shape.Shape ;
 
 public class SpatialIndexLucene implements SpatialIndex {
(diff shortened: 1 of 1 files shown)


osma:
Thanks @afs for the approval. I will wait a few more days and then merge this, unless there are objections.

The only scenario I'm mildly concerned about is if someone performs a spatial query for a large area with a low limit parameter, so that the result list gets cut off. Currently the top K results will (or at least should) be from near the center of the area, but with the sorting disabled, the selection of the top K results will become random.

I have to wonder why the sorting is so slow though. I don't know much about spatial data processing, but I suppose that calculating the distances of ~2500 coordinates to a given center point, and then sorting the list of 2500 results by those distances, shouldn't take 20 seconds. More like 20 milliseconds...

ajs6f:
@osma In the case you mention, shouldn't the user be explicitly ordering anyway?

osma:
@ajs6f I'm not sure whether you can use ORDER BY in a case like this. It would need a SPARQL function for calculating distances between coordinates, so you could say, in effect, "order by distance from point (x,y)". This is something that Lucene can do internally, though.

ajs6f:
@osma Yeah, that's what I meant tho'-- use a calculation (via `BIND`, I guess?) of the distance (and we have [√ available to use](https://jena.apache.org/documentation/query/library-function.html)) and order by that. To be clear, I'm not suggesting that that's a great solution (and I haven't thought at all about the complexities of diff coordinate systems), just that it's better than relying on undocumented impl behavior which can change (and WILL change with this PR! :) ).

osma:
@ajs6f It takes [a bit more](http://www.movable-type.co.uk/scripts/latlong.html) than just sqrt() to calculate great-circle distances between coordinates even within the same coordinate system, e.g. trigonometric functions. But yes, I agree.

ajs6f:
Sure, so my claims are 1) this is a good PR :) and 2) if we want to offer the kind of functionality that you are removing, that's cool, but we should do it in some appropriate, public, maintainable way (extension function?). I don't think I'm saying anything too profound here. :) Should we open a ticket to provide access to Lucene's spatial calculations as extension functions? I'm not saying I'm going to go work on that right away, but it would be worth not losing sight of your insight about people's expectations of being able to order  by distance easily.

rvesse:
We should already have support for most of the trigonometric functions built into ARQ as extension functions - https://github.com/apache/jena/tree/59e5e06c85c98c9c65a9c1addbc4252626e9d061/jena-arq/src/main/java/org/apache/jena/sparql/function/library/leviathan

For example:

```
sparql --data ~/Documents/Work/Datasets/minimal.nt "SELECT (<http://www.dotnetrdf.org/leviathan#sin>(90) AS ?x) { }"
------------------------
| x                    |
========================
| 0.8939966636005579e0 |
------------------------
```

Notes:

- Inverse functions are named as `sin-1` etc.
- function arguments are expected to be radians
- convenience functions `degreesToRadians` and `radiansToDegrees` are provided

PR #208, opened 2017-01-26
Logging fatal to error

afs:
This PR renames the helper `Log` operations from `fatal` to `error` as being the more usual naming these days.

It also contains adding a clone operation to `IndentedWriter` which makes it much easier necessary to create an `IndentedWriter` for stdout/stderr by cloning the standard ones. The clone can be used to indent etc without affecting the system provided ones.

(both these come from working with some complex multiline logging needs mixing 
the usual one line logging and multiline datastructures)


afs:
JENA-1282

PR #209, opened 2017-01-30
Close writers in NTtriple

rafaelcoutinho:
Not sure why writers are not being closed in the Ntriple RDFWriter.

I believe this is a good practice and (still under investigation) I suspect this is causing my application to have a memory leak (since I dump all the intermediary models on a server application).

Let me know if this is acceptable and i can make the same change on the other RDFWriters.

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/rdf/model/impl/NTripleWriter.java
+++ b/jena-core/src/main/java/org/apache/jena/rdf/model/impl/NTripleWriter.java
@@ -40,8 +40,9 @@ public class NTripleWriter extends Object implements RDFWriter {
     @Override
     public void write(Model model, OutputStream out, String base)
          {
+    	Writer w = null;
         try {
-            Writer w;
+            
             try {
                 w = new OutputStreamWriter(out, "ascii");
             } catch (UnsupportedEncodingException e) {
@@ -53,13 +54,19 @@ public class NTripleWriter extends Object implements RDFWriter {
 
         } catch (Exception ioe) {
             errorHandler.error(ioe);
+        }finally{
+        	try{
+        		w.close();
+        	}catch(IOException e){
+        		//nothing to be done
+        	}
         }
     }
     @Override
     public void write(Model model, Writer writer, String base)
          {
+    	PrintWriter pw = null;
         try {
-            PrintWriter pw;
             if (writer instanceof PrintWriter) {
                 pw = (PrintWriter) writer;
             } else {
@@ -81,6 +88,10 @@ public class NTripleWriter extends Object implements RDFWriter {
             pw.flush();
         } catch (Exception e) {
             errorHandler.error(e);
+        }finally{
+        	if(!(writer instanceof PrintWriter)){
+        		pw.close();     
+        	}   	
         }
     }
 


rvesse:
FYI this is the old writer code which for the most part is not used. The ARQ module contains the more modern API which is loaded in place of the old core implementations if both artefacts are present on the class path.

I'm not sure that we should be closing the writers here. Those method calls take in `OutputStream` instances and the writer is only a local wrapping object. If the caller is providing the output stream it is generally the caller's responsibility too close that appropriately. There are legitimate use cases where closing the provided stream can lead to issues e.g. if `System.out` was passed in

ajs6f:
I agree with @rvesse . One doesn't take control of the lifecycle of a parameter like this.

rafaelcoutinho:
Well i agree too, think i'm so confused with this mem leak that I mixed things out. Sorry

rafaelcoutinho:
Oh, odd... for some reason version 3.0.1 of jena is faster but uses 2x memory than version 3.1.1. I wonder if that's the issue i'm having.

afs:
As @rvesse said, this is not the parser applications should be using. It exist as legacy just for bootstrapping jena-core.

Are you using jena-core in it's own? Don't!

If you include the proper dependency `apache-jena-libs`, then you should get the parser in RIOT which is much faster then the old one in jena-core. "much faster" = "x5" or more triples per second on large files.

afs:
Space saving: intentional!

[JENA-1147](https://issues.apache.org/jira/browse/JENA-1147) and [background email](http://mail-archives.apache.org/mod_mbox/jena-dev/201603.mbox/%3C56DB0C83.2040609%40apache.org%3E).

While not free in terms of pure paring speeds, up to a 50% space saving is very useful.

And at Java9, there will be saving due to [JEP 254: Compact Strings](http://openjdk.java.net/jeps/254).

rafaelcoutinho:
Moved to apache-jena-libs dependency, it was still slow. Investigating it deeper I noticed that having namespaces with wrong URL (we do add those models names spaces programmatically and ignored the wrong url in the PREFIX part) makes the query slow, by changing them to correctly point to a model made a 160s query to take 30s. So I believe in the old 3.0.1 version I was using it was ignoring this too, while now it does not.

afs:
I am lacking context to understand your comment - it sounds like the query was searching through a lot of data that wasn't relevant and fixing the prefixes made the query more focused but with seeing a complete, minimal example, it is hard to tell.

rafaelcoutinho:
I know there is no query context or model, however one thing that I identified is that using different jena versions the executing time changes a lot. I don't think it's the query itself but it looks like jena avoids processing the entire model until needed, so when the query is executed jena seems to be performing the full inferences.

The odd thing is that version 3.0.1 is much faster than newer ones:

- v3.0.1 - takes 100s
- v3.1.1 - takes 222s
- v3.2.0 - takes 220

ajs6f:
Perhaps you could open an ordinary Jira issue with sample data and code?

afs:
As mentioned earlier, there was a change at 3.1.1 to reduce the memory foot print. It is a tradeoff - space vs time - for the default setup.  If you want pure speed, use a `ParserProfile` with a different `RDFFactory` - `RDFFactorStd` is the non-caching one, `RDFFactorCaching` is the caching one.

What would be useful to make any progress is to understand what you are trying to do. Raw parser performance is not necessarily the right metric because data is used for a purpose.  The figures you have don't match what I measured, and remeasured for this issue, and without seeing the test and data, it is not possible to draw any insight.

If you wish to explore improvements, it would be useful to use a Java profiler on the code. `RDFFactorCaching is the place to look. It use a cache from Google Guava so maybe there is a better choice there are no concurrency issues.

PR #210, opened 2017-01-30
Fixing JENA-1283 by using Maven properties to control versions

ajs6f:


acoburn:
I was able to install this successfully in a Karaf 4.x container. 👍

PR #211, opened 2017-01-31
writing jsonld: preserve id of blanknodes if they are valid jsonld identifiers

fpservant:
current process to output jsonld computes identifiers for the blanknodes: _:b0,  _:b1, etc.
Ids of blanknodes defined by the user (using AnonId.create()) are not used, even when they are valid jsonld blanknode identifiers.
Of course, this is just a cosmetic question - but if one took care to define valid bn identifiers, why wouldn't they be used in the output ?
This PR allows just that - with a limit: user defined identifiers beginning with "\_:b" are not preserved (this, to avoid clash with those generated by jena)


jena-arq/src/test/java/org/apache/jena/riot/writer/TestJsonLDWriter.java
```diff
@@ -460,15 +468,77 @@ private Model simpleModel(String ns) {
...
+        s = m2.createResource(AnonId.create("_:b0"));
+        m2.add(s, RDFS.label, "? in named graph");
+
+        // write to jsonld
+         
+        String jsonld = dataset2jsonld(ds.asDatasetGraph(), RDFFormat.JSONLD, null);
+        System.out.println(jsonld);
+        
+        // check we have kept the blanknodes identifiers that were valid jsonld bn identifiers
+        assertTrue(jsonld.indexOf("\"_:foo\"") > -1);
```
afs: ` if ( jsonld.contains("\"_:foo\"") )`

afs:
Is the intention to be able to allocate blank nodes in Jena with a specific label and have that label appear in the JSON-LD?

fpservant:
@afs yes

afs:
**NodeToLabel**

The general idea in RIOT is that a specific, special instance of `NodeToLabel` be used for the conversion step. Writing nodes elsewhere uses a `NodeFormatter` which for Turtle etc takes a `NodeToLabel` for the conversion policy.  There is such an object `labels` in JenaRDF2JSONLD.

Hard coding an algorithm for JSON-LD seems less than ideal, together the assumption that Jena blank nodes have a specific form which is a feature of the `NodeToLabel` algorithm.

The correct way is to have a specific `NodeToLabel` for this.
Example: `NodeToLabel.createBNodeByLabelAsGiven` which will use labels unchanged.

**Determining which label to preserve**

This could be a hint carried by `BlankNodeId` with method `boolean BlankNodeId.preserveHint()` seems the proper way to me.  This could extend to all formats and generalizes the "use label as given" option.

If it is necessary to have a way without calling a new constructor for `BlankNodeId`, a way to do that might be to require the string be marked, say, `|...|` (c.f. using quotes `"`!)

I am not sure if this should be some kind of option to writing, rather than the default.

** Parsing **

See also `LabelToNode` for the parsing direction.

fpservant:
@afs OK, that was a quick hack, based on the existing. I'll need more time, then

afs:
I'm working on a parser builder to make it possible to set up more detailed parser configurations (inspired by the work in CommonsRDF). No timescale but it is now at the stage of needing to tick off handling each feature.

I hope this will make setting up the details easier.

```
        StreamRDF destination = ...
        RDFParserBuilder.create()
            .source("http://example/data.ttl")
            .labelToNode(LabelToNode.createUseLabelAsGiven())
            .context(myContext)
            .parse(destination);
```
Code:
https://github.com/afs/jena-workspace/tree/master/src/main/java/parser

afs:
Parser builder : JENA-1306

afs:
Ping.
`RDFParserBuilder` supports this.

Is there anything in this PR not covered? (AKA can it be closed?)

fpservant:
@afs sorry for the time before coming back to this. I just tried in class ``JenaRDF2JSONLD`` to replace (first line of constructor)

```
 NodeToLabel labels = SyntaxLabels.createNodeToLabel() ;
```
by :
```
NodeToLabel labels = NodeToLabel.createBNodeByLabelAsGiven();
```

and it seems to work, but it is not really nice to have ids such as "_:b85f698a-0bfc-40e9-8458-4a200f6cf908" when none has been given by the user. It looks like mixing labels given by the user and labels such as "_:b1" otherwise would require to add a flag to the Node class, right?

I'm afraid I don't have time now to look into that. So maybe this PR should be closed for now.

afs:
Could you close it from you end?  These mirrored GH-repos don't behave exactly as if they were the master and we, the PMC, don't get a "close" button. We can close unmerged ones but it means committing to the code base.

I have found one issue - it should not be using `NodeToLabel labels` at all. In the clear-up of RIOT, the label mapping is passed in. I have fixed this in PR#234.

Side-effect - all nodes get created via `ParserProfile`/`FactoryRDF` which includes checking of URIs: this addresses JENA-1251.

However, it seems that java-jsonld is emitting bnodes without their original label - I get "_:b0" out despite the input having "_:BBB", when the blank node is created in `JsonLDReader.createNode(Map)`

fpservant:
closing this, as more work is still needed

afs:
In ` JenaRDF2JSONLD`

    NodeToLabel labels = NodeToLabel.createBNodeByLabelAsGiven();

works for me, as in it accurately outputs the blank node labels in the graph being written.

That leaves the problem of receiving remapped `\_:b0`, `\_:b1` from java-jsonld.

Easy way to see this is to have in `JsonLDReader`:

    private Node createBlankNode(String str) {
        if ( str.startsWith("_:") )
            str = str.substring(2);
        str = "ZZ"+str;
        return profile.createBlankNode(null, str, -1,-1);
    }

then `"_:ZZb0"` is in the output.

afs:
The `_:b0`, `_:b1` blank nodes from from jsonld-java.

It happens in `JsonLdApi.generateBlankNodeIdentifier`. The algorithm is non-normative in the JSON-LD 1.0 spec.

In the JSON-LD 1.1 spec: http://json-ld.org/spec/latest/json-ld-api/#generate-blank-node-identifier 

There is an issue with blank nodes generated indirectly, not from a label in the syntax, in case if there is an explicit label in the data of `b0` etc. there would be a clash.

So we can arrange that the output blank node labels reflect the data but these are lost when reading them in again.

PR #212, opened 2017-01-31
JENA-1284: Improvements for bulk graph operations in GraphUtil

afs:
(for Jena 3.3.0 development)

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/graph/GraphUtil.java
+++ b/jena-core/src/main/java/org/apache/jena/graph/GraphUtil.java
@@ -19,14 +19,14 @@
 package org.apache.jena.graph;
 
 import java.util.ArrayList;
-import java.util.Iterator ;
-import java.util.List ;
-import java.util.Set ;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
 
-import org.apache.jena.graph.impl.GraphWithPerform ;
-import org.apache.jena.util.IteratorCollection ;
-import org.apache.jena.util.iterator.ExtendedIterator ;
-import org.apache.jena.util.iterator.WrappedIterator ;
+import org.apache.jena.graph.impl.GraphWithPerform;
+import org.apache.jena.util.IteratorCollection;
+import org.apache.jena.util.iterator.ExtendedIterator;
+import org.apache.jena.util.iterator.WrappedIterator;
 
 /**
     An ad-hoc collection of useful code for graphs
@@ -102,22 +102,19 @@ public class GraphUtil
     }
     
     public static void add(Graph graph, Triple[] triples) {
-        if ( OldStyle && graph instanceof GraphWithPerform )
-        {
+        if ( OldStyle && graph instanceof GraphWithPerform ) {
             GraphWithPerform g = (GraphWithPerform)graph ;
             for (Triple t : triples )
                 g.performAdd(t) ;
             graph.getEventManager().notifyAddArray(graph, triples) ;
-        }
-        else
-        {
+        } else {
             for (Triple t : triples )
                 graph.add(t) ; 
         }
     }
         
     public static void add(Graph graph, List<Triple> triples) {
-        addIteratorWorkerDirect(graph, triples.iterator());
+        addIteratorWorkerDirect(graph, triples.iterator()) ;
         if ( OldStyle && graph instanceof GraphWithPerform )
             graph.getEventManager().notifyAddList(graph, triples) ;
     }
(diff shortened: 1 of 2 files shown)


rvesse:
I presume this was motivated by a real-world performance concern? It would be interesting to know how much difference it makes.

My only concern is what happens when there is only a slight difference between the size of the affected graphs? It's looks like when there is a slight difference then you potentially do twice the work because on one code path you do both a `find()` and a `delete()`/`add()` for every triple.

Would it be worth making the behaviour based a configurable percentage difference e.g. If the graphs are within 10% of each others size don't use the new path.

Also is materialising the list of triples a potential memory issue when the iterator is over a large amount of data? Click

ajs6f:
My 2¢ are inline, but I echo @rvesse's questions as well.

jena-core/src/main/java/org/apache/jena/graph/GraphUtil.java
```diff
@@ -115,55 +117,44 @@ public static void add(Graph graph, Triple[] triples) {
...
-        // Materialize to avoid ConcurrentModificationException.
-        List<Triple> s = IteratorCollection.iteratorToList(it) ;
-        if ( OldStyle && graph instanceof GraphWithPerform )
-        {
-            GraphWithPerform g = (GraphWithPerform)graph ;
-            for (Triple t : s)
-                g.performAdd(t) ;
+        if ( OldStyle && graph instanceof GraphWithPerform ) {
+            // Materialize for the notify.
+            List<Triple> s = IteratorCollection.iteratorToList(it) ;
```
ajs6f: I'm confused as to why we want to ensure notification this early-- why not wait for the actual work done below? Maybe we could have another line for the comment here on this?

jena-core/src/main/java/org/apache/jena/graph/GraphUtil.java
```diff
@@ -115,55 +117,44 @@ public static void add(Graph graph, Triple[] triples) {
...
-        {
-            for (Triple t : s )
-                graph.add(t) ;
+        addIteratorWorkerDirect(graph, s.iterator());
+    }
+    
+    private static void addIteratorWorkerDirect( Graph graph, Iterator<Triple> it ) {
+        if ( OldStyle && graph instanceof GraphWithPerform ) {
+            GraphWithPerform g = (GraphWithPerform)graph;
+            while (it.hasNext())
```
ajs6f: @ajs6f's usual carping: I really think "point-free" reads more clearly here and directly below: `it.forEachRemaining(g::performAdd)`.

jena-core/src/main/java/org/apache/jena/graph/GraphUtil.java
afs: Note: there is no change in behaviour here - this change are only due to addIteratorWorkerDirect/addIteratorWorker.

The notifcation requires a list so making that now the iterating on that is guaranteed CCME-free and so call `addIteratorWorkerDirect` with the necessary copy.

This preserves event semantics, for better or worse.

afs:
@rvesse : It came up when reviewing code; I don't have figures either way. We have situations of deleting small graphs from large ones and deleting large graphs from small ones (and empty ones). The size ratios can be "small" being down to 10 triples and large being 100k.

At the same time, this utility code must be robust again CCME on add/delete during iteration. Just because one graph is not identical to another, does not mean they are disconnected (e.g. inference graphs).

The code was always careful - if you can see cases where we can know non-materialization is possible, please point to them. I'd be very happy to exploit it.

afs:
> configurable percentage

For similar sized graphs, are the two choices very different in performance? I can't see it but I'm maybe too close to the code.

I believe/hope this PR is only ever an improvement but have I missed something?

afs:
> materialising the list of triples a potential memory issue

Yes :-( but the new code is supposed to materialize the smaller side of the operation so is a (small) improvement over the existing case.

rvesse:
My concern is that depending on the underlying storage of the data in the graphs a call to `contains()` might be quite expensive and it would be more efficient to simply call `delete()` directly.

You appear to have provided a flag to force the old behaviour so I think go ahead and merge this. Since if any user does encounter performance regressions we can advise them of the workaround and use a concrete report as the basis for any further optimisation.

afs:
@rvesse - so this is specific about `deleteFrom`.  Good point that `contains()` may be not low cost.

I have realized that we can use the fact that often there is no code listing to events by calling `Graph.getEventManager().listening()`. The code can have one path when events are needed, to get the perfect legacy contract, and another path for when events are not needed because no code is listening.

afs:
Revision to `deleteFrom`:  loop on src if (srcSize < 1000 || dstSize > 2*srcSize)

That is, when dst is small, do the loop/contains thing. 

This is more conservative and so less disturbing existing code but catches, I hope, the case of small destination (target) graph, typically in-memory, and large src graph (some large database).

1000 is a guess - 10,000 is better maybe?

PR #213, opened 2017-02-01
JENA-1285: One token type for strings

afs:
Having 4 separate string token types means that all four cases have to be considered, yet the more common case is that a string has been seen. Only sometimes does the exact form of the string matter.

This PR changes Jena to have one token type for STRING and separately record the string type seen for when it is needed.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/tokens/StringType.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/tokens/StringType.java
@@ -1,0 +1,22 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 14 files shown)


jena-arq/src/main/java/org/apache/jena/riot/tokens/StringType.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.jena.riot.tokens;
+
+/** Seen form of a {@link TokenType#STRING} */
+public enum StringType { STRING1, STRING2, LONG_STRING1, LONG_STRING2 }
```
ajs6f: It would be nice to have a quick comment explaining what the actual difference amongst these guys is.

ajs6f:
+1: this is clearer. But that's coming from someone w/o much familiarity with this part of the code!

PR #214, opened 2017-02-02
Small changes

retnor:
Omit the last '/' in line 32 will make this code work


ajs6f:
Looks good-- I'll merge this after we finish with the current release process. Thanks @retnor !

PR #215, opened 2017-02-02
JSON improvements

afs:
A number of small improvement to the JSON handling code, motivated by working with it in another project.


ajs6f:
A big +1 to the `return this;` idiom. It's very convenient in this kind of work.

afs:
Recorded as JENA-1287.

PR #218, opened 2017-02-25
JENA-1294: Don't use contants during initialization.

afs:
Fixes for RDFS - use functions for constants during initialization.

This fixes initialization as reported (TestParameterizedSparqlString).

RDF and OWL still to do in the same style for some rather unusual ways to trigger the same effect.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/query/ARQ.java
+++ b/jena-arq/src/main/java/org/apache/jena/query/ARQ.java
@@ -19,6 +19,7 @@
 package org.apache.jena.query;
 
 import org.apache.jena.riot.RIOT ;
+import org.apache.jena.riot.resultset.ResultSetLang;
 import org.apache.jena.riot.system.RiotLib ;
 import org.apache.jena.sparql.ARQConstants;
 import org.apache.jena.sparql.SystemARQ ;
@@ -579,7 +580,7 @@ public class ARQ
             JenaSystem.logLifecycle("ARQ.init - start") ;
             // Force constants to be set.  This should be independent of other initialization including jena core.
             ARQConstants.getGlobalPrefixMap();
-            
+            ResultSetLang.init();
             globalContext = defaultSettings() ;
             ARQMgt.init() ;         // After context and after PATH/NAME/VERSION/BUILD_DATE are set
             MappingRegistry.addPrefixMapping(ARQ.arqSymbolPrefix, ARQ.arqParamNS) ;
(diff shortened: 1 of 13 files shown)


jena-arq/src/main/java/org/apache/jena/query/ARQ.java
```diff
@@ -579,7 +580,7 @@ public static void init() {
             JenaSystem.logLifecycle("ARQ.init - start") ;
             // Force constants to be set.  This should be independent of other initialization including jena core.
             ARQConstants.getGlobalPrefixMap();
-            
+            ResultSetLang.init();
```
afs: This isn't directly related but I noticed that `ResultSetLang` init into RDFLangauge was happens later. Making it happen now was part of verification there were no other lurking problems.

jena-arq/src/main/java/org/apache/jena/sparql/core/assembler/AssemblerUtils.java
```diff
@@ -76,8 +76,11 @@ static public void registerModel(Resource r, Assembler a) {
     /** Register an addition assembler */  
     static public void register(AssemblerGroup g, Resource r, Assembler a, Resource superType) {
         registerAssembler(g, r, a) ;
-        if ( superType != null && ! superType.equals(r) ) 
-            modelExtras.add(r, RDFS.subClassOf, superType) ;
+        if ( superType != null && ! superType.equals(r) ) {
+            // This is called during Jena-wide initialization.
+            // Use function for constant (JENA-1294)
+           modelExtras.add(r, RDFS.Init.subClassOf(), superType) ;
```
afs: Example of use of the calculation rather than the eventual constant.

jena-core/src/test/java/org/apache/jena/test/TestSystemSetup.java
```diff
@@ -35,5 +42,24 @@ public void testRDF11() {
             fail("RDF 1.0 mode enabled in Jena3 test run") ;
     }
 
+    /** This test relies on forking a clean JVM */ 
+    public void testInitFromRDFS() throws IOException, InterruptedException {
+        String separator = System.getProperty("file.separator");
+        String classpath = System.getProperty("java.class.path");
+        String java = System.getProperty("java.home")
+                + separator + "bin" + separator + "java";
```
alexishuf: On windows, the java executable will end with ".exe". This should avoid any problem:
```java
import org.apache.commons.lang3.SystemUtils;
        if (SystemUtils.IS_OS_WINDOWS)
            java += ".exe";
```

alexishuf:
Tested here and it solved everything. Thanks!

Before this PR, Initializing from OWL was only a problem if TDB was also initialized. Maybe is it a good idea to add `TestSystemSetup`  to `org.apache.jena.tdb.sys.TS_Sys`?

afs:
".exe" append integrated in.

I had some problems in the build (because of a static method called "Property()") but all fixed now so this PR should be complete.  I'll merge it soon unless someone has an more comments.

PR #220, opened 2017-03-03
JENA-1301: Drop Solr support from jena-text and jena-spatial

osma:
See [JENA-1301](https://issues.apache.org/jira/browse/JENA-1301) for rationale.

Note that this is implemented after the Lucene 6 upgrade ([JENA-1250](https://issues.apache.org/jira/browse/JENA-1250)) so it has to be merged after that.


osma:
I tested this together with PR #219. I couldn't find any problems. I will merge this within the next 2 days unless anyone objects.

afs:
No objection here.

When done and a snapshot available, it would be good to send email to users@ to give interested parties notice of the upcoming change.

PR #223, opened 2017-03-09
JENA-1309: HttpOp should respect system properties for default client

ajs6f:


ajs6f:
Yeah, `HttpOp` is usually my first stop for this kind of thing because my understanding is that in theory, all ARQ routes all HTTP traffic through `HttpOp`.

afs:
Status?

ajs6f:
I was waiting to hear confirmation from the reporter that this is a fix, but it's been quite a while. I have no prob with merging this as-is.

ajs6f:
Fixed with https://git-wip-us.apache.org/repos/asf?p=jena.git;a=commit;h=711e9be838efdb056bfa4bccc253c58799affea0

PR #226, opened 2017-03-14
Refactor TextIndex and TextQueryPF in jena-text (preparation for JENA-1305)

osma:
In the development of JENA-1305 (Elasticsearch backend for jena-text) it surfaced that the current division of labor between TextQueryPF and TextIndex is awkward. The Lucene (or Solr) query string is mostly formed already on the TextQueryPF side, which makes it difficult to implement a backend that uses a different query language.

I have refactored the TextIndex interface, separating out the graph and language information away from the query string into separate method parameters, and changed the code in TextQueryPF and TextIndexLucene correspondingly.

I'm opening this PR to ask for comments on the change. If there is no opposition, I will merge it to master within a few days.

### DIFF
--- a/jena-text/src/main/java/org/apache/jena/query/text/DatasetGraphText.java
+++ b/jena-text/src/main/java/org/apache/jena/query/text/DatasetGraphText.java
@@ -93,12 +93,17 @@ public class DatasetGraphText extends DatasetGraphMonitor implements Transaction
 
     /** Search the text index on the text field associated with the predicate */
     public Iterator<TextHit> search(String queryString, Node predicate, int limit) {
+        return search(queryString, predicate, null, limit) ;
+    }
+
+    /** Search the text index on the text field associated with the predicate within graph */
+    public Iterator<TextHit> search(String queryString, Node predicate, String graphURI, int limit) {
         queryString = QueryParserBase.escape(queryString) ;
         if ( predicate != null ) {
             String f = textIndex.getDocDef().getField(predicate) ;
             queryString = f + ":" + queryString ;
         }
-        List<TextHit> results = textIndex.query(predicate, queryString, limit) ;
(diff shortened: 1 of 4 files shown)


anujgandharv:
@osma I do not see any Unit/Integration tests updated for Lucene for the changes you have done in this PR. Do they come in a separate PR? In the mean time, can you provide me some examples as to how the information will flow into the query endpoint. Will the lang be like: `en` or `lang:en`. I assume it would be `en. Is that the correct assumption.

osma:
The jena-text tests are written on a higher level (queries written as SPARQL) so they were unaffected by this PR and continue to work after the changes.

The query method now takes the plain query without field name prefix, language (eg "en" or "none" or null) and graph as URI string.

osma:
@anujgandharv forgot to ping you...see above

anujgandharv:
Thanks @osma . I learned from you that: _It's always good to have Unit Tests for the changes to make sure we are not missing any corner cases :)_

So just to clarify, if the Sparql query string is:

```
SELECT ?s
{ ?s text:query (rdfs:label 'word' 10) ; 
     rdfs:label ?label 
}
```
then the query method will receive the following attributes: `RDFS.label.asNode(), "word", null, null, 10`

And if the query string is: `?s text:query (rdfs:label 'word' 'lang:en' )`, then the query method will receive the following attributes: `RDFS.label.asNode(), "word", null, "en"`

And finally if the query string is: `?s text:query ('word' 'lang:en' )`, then the query method will receive the following attributes: `RDFS.label.asNode(), "word", null, "en"` , where the Node property is the default Node defined at the configuration time. 

Can you confirm all the scenarios above please. I will modify my test cases and logic accordingly.

NOTE: I say NULL for graph attribute, mainly because I don't care about it in the ES implementation.

osma:
@anujgandharv All correct! And if in doubt, just try it out...

Thank the Jena project, I've learned the lesson myself while working on Jena!

PR #230, opened 2017-03-31
JENA-733 add documentation for narrower data type case

kinow:
Tried to clarify what happens when users provide Long values that fit into Integers, for [JENA-733](https://issues.apache.org/jira/browse/JENA-733).

I think if the docs are good to be merged, then the issue in JIRA can be closed too.

Cheers
Bruno


rvesse:
+1 LGTM

kinow:
Merged. Will resolve the JIRA issue now. Thanks!

PR #235, opened 2017-04-09
jsonld: @vocab in @context for "" ns prefix

fpservant:
"" prefix are not allowed in jsonld's @context. "" ns prefix defined in a model were therefore ignored. This pull request proposes to define @vocab with the ns corresponding to such an "" prefix. Everything else remained unchanged.

cf. JENA-1292


afs:
I've pulled this into the work general cleaning up RIOT (PR#234). It'll get included when that is merged. 

Does this complete JENA-1292?

fpservant:
> Does this complete JENA-1292?

yes.

In JENA-1292's discussion I had suggested a way to automatically remove from the @context the properties that belong to the @vocab, but it is not a good idea. It adds unnecessary complexity, as same result can be achieved by a user with a few lines of code, and existing features (possibility to replace the @context by a given value in the jsonld output)

PR #236, opened 2017-04-10
No stacktrace on regex errors.  Cancel/close during sorting.

afs:
Changes for:

"runtime regex pattern and many many log messages"
https://lists.apache.org/thread.html/0d6042871739323716346ebfff76e64ede7c39a52eacbc0ab17d6d3c@%3Cdev.jena.apache.org%3E

"Bug in SortedDataBag?"
https://lists.apache.org/thread.html/be3ecf8f41b03a747f6bab78a072ca55b2ea1a07a6d7ae1e53600ac5@%3Cdev.jena.apache.org%3E

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/atlas/data/SortedDataBag.java
+++ b/jena-arq/src/main/java/org/apache/jena/atlas/data/SortedDataBag.java
@@ -94,6 +94,7 @@ public class SortedDataBag<E> extends AbstractDataBag<E> {
      */
     public void cancel() {
         comparator.cancel();
+        close();
     }
 
     /**
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterSort.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterSort.java
@@ -44,9 +44,7 @@ import org.apache.jena.sparql.engine.binding.BindingComparator;
  */
 
 public class QueryIterSort extends QueryIterPlainWrapper {
-    private final QueryIterator embeddedIterator; // Keep a record of the
-    // underlying source for
-    // .cancel.
+    private final QueryIterator embeddedIterator;
     final SortedDataBag<Binding> db;
 
     public QueryIterSort(QueryIterator qIter, List<SortCondition> conditions, ExecutionContext context) {
@@ -57,10 +55,8 @@ public class QueryIterSort extends QueryIterPlainWrapper {
             final ExecutionContext context) {
         super(null, context);
         this.embeddedIterator = qIter;
-
         ThresholdPolicy<Binding> policy = ThresholdPolicyFactory.policyFromContext(context.getContext());
         this.db = BagFactory.newSortedBag(policy, SerializationFactoryFinder.bindingSerializationFactory(), comparator);
-
         this.setIterator(new SortedBindingIterator(qIter));
     }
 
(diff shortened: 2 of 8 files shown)


ehedgehog:
Whoops, clicked too soon. 

In E_Regex.eval, we Log.warn(this,message). I see that using Log.warn would be
inappropriate, since it acts globally and would switch off all logging of whatever
we chose as a key, but as per the original problem it would be useful to
suppress identical log lines.

I wondered about replacing the Log.warn with

String m = ex.getMessage();
            	if (!m.equals(currentFailMessage)) {
            		currentFailMessage = m;
            		Log.warn(this, m); 
            	}

catching the case where the same pattern error keeps occuring.
This might be either of overkill or underkill.

Chris

afs:
This PR covers the "Bug in SortedDataBag?" report as well.  Waiting on tests for that.

ehedgehog:
I realised that there were no test cases for QueryIterTopN, which has
a failure mode like TestQuerySort. I added three tests to QueryIterSort
(because I reused the CallbackIterator from TestQueryIterSort and 
extracting that to a new file would make the diffs more complicated)
in the same style as the tests for QueryIterSort.close():

================================================

    @Test public void testTopNCloseClosesSource() {
		
		long numItems = 3;
		boolean distinct = false;
        Context context = new Context() ;
    	ExecutionContext ec = new ExecutionContext(context, (Graph) null, (DatasetGraph) null, (OpExecutorFactory) null);

		QueryIterTopN tn = new QueryIterTopN(iterator, comparator, numItems, distinct, ec);
		tn.close();
		assertTrue(iterator.isClosed());
	}
    
    @Test public void testTopNExhaustionClosesSource() {
		
		Callback nullCB = new Callback() { @Override public void call() {} };
		iterator.setCallback(nullCB);
		
		long numItems = 3;
		boolean distinct = false;
        Context context = new Context() ;
    	ExecutionContext ec = new ExecutionContext(context, (Graph) null, (DatasetGraph) null, (OpExecutorFactory) null);

		QueryIterTopN tn = new QueryIterTopN(iterator, comparator, numItems, distinct, ec);
		while (tn.hasNext()) tn.next();
		assertTrue(iterator.isClosed());
	}
    
    @Test public void testTopNCancelClosesSourceIterator() {
				
		long numItems = 3;
		boolean distinct = false;
        Context context = new Context() ;
    	ExecutionContext ec = new ExecutionContext(context, (Graph) null, (DatasetGraph) null, (OpExecutorFactory) null);

		QueryIterTopN tn = new QueryIterTopN(iterator, comparator, numItems, distinct, ec);
		try {
			while (tn.hasNext()) tn.next();
			fail("QueryCancelledException not caught");
		} catch (QueryCancelledException q) {
			assertTrue("cancellation did not close source iterator", iterator.isClosed());			
		}
    }

==========================================

It turns out that the third of these tests fails because an exception
thrown during QueryIterTopN's initializeIterator does not close
the source (or any other) iterator. So I wrapped the body of the
initializeIterator in a try-catch-QueryCancelledException and closed
both the initializeIterator nad the QueryIterTopN.
===============================================

   private Iterator<Binding> sortTopN(final QueryIterator qIter, final Comparator<Binding> comparator) {
        return new IteratorDelayedInitialization<Binding>() {
            @Override
            protected Iterator<Binding> initializeIterator() {
            	try {
            		while ( qIter.hasNext() ) {
	                    Binding binding = qIter.next() ;
	                    if ( heap.size() < limit )
	                        add(binding) ;
	                    else {
	                        Binding currentMaxLeastN = heap.peek() ;
	                        if ( comparator.compare(binding, currentMaxLeastN) < 0 )
	                            add(binding) ;
	                    }
	                }
	                qIter.close() ;
	                Binding[] y = heap.toArray(new Binding[]{}) ;
	                heap = null ;
	                Arrays.sort(y, comparator) ;
	                return asList(y).iterator() ;
	            } catch (QueryCancelledException e) {
	            	QueryIterTopN.this.close();
	            	this.close();
	            	throw e;
	            }
            }
        } ;
    }

================================================
 The new tests now pass.

ehedgehog:
Sorry for the layout mess.

afs:
Split into PR#238 (regex) and PR#239 (cancel sort)

PR #240, opened 2017-04-13
JENA-1321: Improved error handling so response body is not masked.

rmorrise:


afs:
To check my understanding: the root problem is that the `QueryExceptionHTTP` does not have the `HttpException.getResponse` available.

`QueryExceptionHTTP` can't be an `HttpException` and `QueryException` (separation of HTTP code from query handling). 

If so, maybe it would be better to add the response body to the `QueryExceptionHTTP` - the removal of the `HttpException` level of exception removes a layer of unhelpful stacktrace (the local source of the `HttpException` is not relevant).  By adding the error response to `QueryExceptionHTTP`, the response would be available still.

ajs6f:
If that understanding is correct (and I now share it) then I agree with @afs's proposal, and I can make that change. @rmorrise ?

rmorrise:
That works.  Personally I prefer to see the original root cause but I guess people sometimes get distracted by long stack traces. If you are going to do that can you check the original exception's hierarchy for other props that might be needed like headers etc.?

ajs6f:
`org.apache.jena.atlas.web.HttpException` only has `getResponseCode()`, `getStatusLine()`, and `getResponse()`. Wouldn't it work to just add those three to `QueryExceptionHTTP`?

rmorrise:
Sounds good. :)

ajs6f:
@rmorrise If you are satisfied with our work on JENA-1321, would you close this?

PR #241, opened 2017-04-17
JENA-1321: Exception rewrapping in HttpQuery masks error response from the server

ajs6f:
It's a bit odd-feeling to couple `HttpException` so tightly, but given that it is an Atlas type, I don't feel terrible about it.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/http/QueryExceptionHTTP.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/http/QueryExceptionHTTP.java
@@ -18,16 +18,19 @@
 
 package org.apache.jena.sparql.engine.http;
 
+import org.apache.jena.atlas.web.HttpException;
 import org.apache.jena.query.QueryException ;
 
 /** Exception class for all operations in the SPARQL client library.
  *  Error codes are as HTTP status codes. */
 public class QueryExceptionHTTP extends QueryException
 {
-    private static final long serialVersionUID = 99L;  // Serilizable.
+    private static final long serialVersionUID = 99L;  // Serializable.
     public static final int noResponseCode = -1234 ;
     private int responseCode = noResponseCode ;
-    private String responseMessage = null ;
+    private final String responseMessage ;
+    private String statusLine ;
+    private String response;
 
     // Codes for extra errors.  We use HTTP error codes so
     // these are negative to avoid clashes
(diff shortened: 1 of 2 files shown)


ajs6f:
@rmorrise, will this suit you?

rmorrise:
Code looks good!

I am new to git, but if I understood correctly, you took my change plus adding the responseCode, but did not take my changes to TestService. I think this will fail the unit tests? Please correct me if I am reading the diff wrong.

Do I need to do anything for the changes to be applied to the main repo?

ajs6f:
No, actually, I did not start with your code-- if you look at what I did, I added fields and accessors and a constructor to `QueryExceptionHTTP` instead as we discussed. You need do nothing further. I will merge this in and it will appear in our next release. (In the meantime, you can just `git pull` the changes from the main repo (once I merge them).)

ajs6f:
The tests pass fine, because as @afs prescribed, we aren't changing the wrapping of exceptions, just adding useful information to "outer" ones.

rmorrise:
My mistake! I didn't notice the call to httpEx.getCause() in the constructor.

jena-arq/src/main/java/org/apache/jena/sparql/engine/http/QueryExceptionHTTP.java
```diff
@@ -63,11 +66,21 @@ public QueryExceptionHTTP(int responseCode)
...
      */  
     public String getResponseMessage() { return responseMessage ; }
 
+    /** The response for this exception
+     * @return response
+     */  
+    public String getResponse() { return response ; }
+
+    /** The status line for the response for this exception
+     * @return status line
```
afs: "The status line for the response if available from HTTP"

jena-arq/src/main/java/org/apache/jena/sparql/engine/http/QueryExceptionHTTP.java
```diff
@@ -63,11 +66,21 @@ public QueryExceptionHTTP(int responseCode)
...
     
     
-    /** The messge for the reason for this exception
+    /** The message for the reason for this exception
      * @return message
      */  
     public String getResponseMessage() { return responseMessage ; }
 
+    /** The response for this exception
+     * @return response
```
afs: Note that this can be null if the original exception was not from the HTTP protocol, but from e.g. an IOException why doing an HTTP operation.

"The response for this exception if available from HTTP"

PR #243, opened 2017-04-19
JENA-1324: Check for spaces in URIs (if RDF 1.1)

afs:
Set the IRIFactory to the IRI version, if parsing via RIOT and in RDF 1.1 mode.


afs:
Symmetry,driven by the fact that I tried out handling the situation in this routine, which needed the error handler which is an object field.

PR #245, opened 2017-04-22
jena-text-es: Separate module for ElasticSearch text search

afs:
This PR makes the ElasticSearch a separate module, that depends on jena-text.

Fuseki incorporates jena-text, not jena-text-es, and the server jar is 25M.

This is a proof-of-concept for discussion.


afs:
See JENA-1328.

PR #247, opened 2017-04-27
Query string in URLs; uniform HttpClient/HttpContext usage.

afs:
JENA-1330: pass HttpClient/HttpContext in query and update calls.

JENA-1331: Support additional query string in URLs for services.

Add operations for HttpClient/HttpContext to 
QueryExecutionFactory and UpdateExecutionFactory.

Use HttpContext in HtpQuery, then HttpClientContext.adapt as necessary.

Tests for URL generation.


afs:
Based on the discussion on JENA-1331, commit 1e51ed0 makes some private->protected and adds some javadoc. It makes no other changes.

PR #248, opened 2017-04-28
JENA-1329: Define and use QueryIterator.isJoinIdentity.

afs:
Don't test for class `QueryIterRoot`.
Don't use `QueryIterRoot` to hold a binding with set variables.


afs:
This should wait until after 3.3.0 to give significant time before a release.

While it is a fix, it is an old feature so may have knock-on effects.

PR #251, opened 2017-05-22
JENA-1341: Fuseki jetty upgrade

afs:
Jetty dependency updates for Fuseki2 and Fuseki1.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/atlas/web/TypedInputStream.java
+++ b/jena-arq/src/main/java/org/apache/jena/atlas/web/TypedInputStream.java
@@ -31,6 +31,7 @@ public class TypedInputStream extends FilterInputStream
     // May be different from the URI used to access the resource 
     // e.g. 303 redirection, mapped URI redirection 
     private String baseURI ;
+    private boolean isClosed = false ;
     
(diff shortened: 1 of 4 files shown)


jena-arq/src/main/java/org/apache/jena/atlas/web/TypedInputStream.java
```diff
@@ -31,6 +31,7 @@
     // May be different from the URI used to access the resource 
     // e.g. 303 redirection, mapped URI redirection 
     private String baseURI ;
+    private boolean isClosed = false ;
```
ajs6f: Maybe `volatile`, just in case?

afs: Unnecessary in this case because it is safe to call close on the wrapped `InputStream` anyway.  The flag goes from false to true only.  The flag is useful (e.g. debugging) but is safe without. (I don't like over using `volatile` on principle.)

ajs6f: Ok, cool. I tend to be liberal with it, unless I am worried about a specific performance problem.

PR #255, opened 2017-05-28
JENA-1350: Use Fuseki embedded server for testing

afs:


jena-fuseki2/jena-fuseki-embedded/src/test/java/org/apache/jena/fuseki/embedded/FusekiTestAuth.java
```diff
@@ -0,0 +1,182 @@
...
+    public static final String serviceQuery()       { return "http://localhost:"+port()+datasetPath()+"/query" ; }
+    public static final String serviceGSP()         { return "http://localhost:"+port()+datasetPath()+"/data" ; }
+    
+    private static FusekiEmbeddedServer server ;
+    
+    public static void setupServer(boolean updateable, SecurityHandler sh) {
+        // Clean datasets.
+        dsgTesting = DatasetGraphFactory.createTxnMem();
+        server = FusekiEmbeddedServer.create()
+            .add(datasetPath(), dsgTesting)
```
ajs6f: Looks like all of the test server instances will be in-memory datasets, which is great for speed. Will there be any places where the test server instances touch the disk?

afs: Speed but also for clear-up.  No files get left on disk to interfere with other tests; especially if the server takes awhile to stop properly and release all its resources.

We could add `void setupServer(boolean updateable, DatasetGraph dsg, SecurityHandler sh)` for the general case.

`FusekiTestServer` is also a in-memory transactional dataset and there it is much more difficult to inject a specific implementation because the lifecycle is driven by JUnit.

The example on the JIRA shows it is not necessary to use these helper classes.  They are helpful when a lot of small tests are run, ensuring a balance between careful clean-up and speed.

PR #257, opened 2017-05-29
Upgrade some dependencies to next minor versions

ajs6f:


jena-parent/pom.xml
```diff
@@ -59,10 +59,10 @@
...
     
     <ver.commonslang3>3.4</ver.commonslang3>
-    <ver.commonscsv>1.3</ver.commonscsv>
-    <ver.dexxcollection>0.6</ver.dexxcollection>
+    <ver.commonscsv>1.4</ver.commonscsv>
+    <ver.dexxcollection>0.7</ver.dexxcollection>
 
-    <ver.httpclient>4.5.2</ver.httpclient>
+    <ver.httpclient>4.5.3</ver.httpclient>
     <ver.httpcore>4.4.4</ver.httpcore>
```
afs: This needs to be 4.4.6 to align with httpclient 4.5.3 (from http://central.maven.org/maven2/org/apache/httpcomponents/httpcomponents-client/4.5.3/httpcomponents-client-4.5.3.pom)

afs:
Adding JENA-1353 for linkage.

jena-parent/pom.xml
ajs6f: It passed the tests at 4.4.4, which is interesting, but I am happy to upgrade that too!

afs:
The maven property only gets used to pick up the right httoClient OSGi artifact IIRC.  Otherwise, dependency management should pick the right one.

ajs6f:
Yeah, I was going to look at that at some point-- transitive dependency management, yes, please!

PR #259, opened 2017-06-03
JENA-1357: Add support for a StreamManager when building an RDFParser.

afs:


kinow:
Everything builds fine locally, all tests pass. Code looks good, and the few parts I am familiar in the change seem OK too. +1 lgtm

PR #262, opened 2017-06-14
JENA-1313: Some cleanup around NodeValueSortKey

afs:
- NodeValue.sameAs (complete switch statement)
- Add NodeValue.getSortKey() and use instead of casts.
- javadoc fixes

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/expr/NodeValue.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/expr/NodeValue.java
@@ -586,7 +586,9 @@ public abstract class NodeValue extends ExprNode
                 raise(new ExprEvalException("Unknown equality test: "+nv1+" and "+nv2)) ;
                 throw new ARQInternalErrorException("raise returned (sameValueAs)") ;
             }
-            
+            case VSPACE_SORTKEY:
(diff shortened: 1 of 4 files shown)


kinow:
+1 !!!

Just so I know when to use it, what's `SystemARQ.ValueExtensions` for?

Thanks

afs:
`SystemARQ.ValueExtensions` is a very old flag (that I'd complete forgotten about!) that will make ARQ only deal with the minimum set of datatypes in SPARQL, so no xsd:date, xsd:g*, language tags, and others appearing in the built-in operators for "same value" and "compare".

Looking at it, the treatment I can see that the list isn't right - I'll push a fix as part of this PR and tidy up some comments.

It is enabled in "strict mode", which is only useful for investigating specs, and is unlikely to be perfect so check the query and results to confirm.

The set of "value spaces" is more detailed than that in XDM (XQuery and XPath Data Model) - xsd:date and xsd:dateTimes can be compared using special rules, duration needed a java fix, so {{NodeValue.compare}} has several special cases.

kinow:
Thanks for the explanation Andy! I hadn't a clue that would be necessary, but will try to keep it in mind next time I have a change for ARQ.

kinow:
And had a look at the second commit, looks OK from what I could see. Keeping my +1 lgtm for it! :+1:

PR #263, opened 2017-06-17
Writer updates

afs:
JENA-369: Some lists in WHERE blocks are printed in ()-form - in particular, property functions are likely to be done this way.

JENA-1361: Fix bug in update output.

The existing testing already covers the changes in `TestSerialziation`.


afs:
Printing basic graph patterns prettily for queries is a different problem to printing Turtle if it is to retain the triple ordering of the BGP.

This PR conservatively spots the layout the parser generates and puts back pretty lists for at least flat lists.

It does not apply to property paths.

PR #266, opened 2017-06-23
JENA-1367: Improve apf:strSplit corner cases

cygri:
This is a patch to improve the behaviour of `apf:strSplit` in cases where it is invoked with inputs other than the obvious node types.

The general syntax of the property function is:

    ?parts apf:strSplit (?input ?regex)

The patch changes various cases that previously would have thrown `ExprEvalException` or `NotLiteralException`:

- If `?input` is an unbound variable, return no match
- If `?input` is an IRI or blank node, return no match
- If `?regex` is not a literal, return no match
- If `?parts` is a plain literal that equals one of the tokens of the parts of `?input`, return true (one match, no new variable bindings)
- If `?parts` is anything else but not a variable, return no match
- If the object is a list but does not have two elements, throw `QueryBuildException`

The patch also adds fairly exhaustive tests for the function. The test class may still need to be hooked into the test suite in some way.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/pfunction/library/strSplit.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/pfunction/library/strSplit.java
@@ -39,11 +39,13 @@ import org.apache.jena.sparql.util.IterLib;
 import org.apache.jena.vocabulary.XSD;
 
 /**
- * Property function that requires the subject to be unbound, and the object to
+ * Property function that requires the object to
  * contain a list of two items, the first of which is a string to be split, and
- * the second is a regular expression denoting the split point. The subject
- * variable is bound for each result of the split, and each result has the
- * whitespace trimmed from it.
+ * the second is a regular expression denoting the split point. If the subject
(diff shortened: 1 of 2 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/pfunction/library/strSplit.java
```diff
@@ -44,27 +47,58 @@
...
+            
+            final Var subjectVar = Var.alloc(subject);
+
+            Iterator<Binding> it = Iter.map(
+                    tokens.iterator(),
+                    item -> BindingFactory.binding(binding, subjectVar,
+                            NodeFactory.createLiteral(item)));
+            return new QueryIterPlainWrapper(it, execCxt);
+            
+        } else if (subject.isLiteral()
```
kinow: I think this doesn't match the current Javadoc? Maybe this change requires some changes to the text in the class Javadoc, that explain the new behaviour.

cygri: You are right. I updated the javadoc to match the new behaviour.

PR #267, opened 2017-06-28
JENA-1369: TDB cleanup

afs:
This PR is the result of having to investigate what was going on inside TDB1.

There are no changes to TDB behaviour in this PR.


jena-tdb/src/main/java/org/apache/jena/tdb/transaction/DatasetBuilderTxn.java
```diff
@@ -159,6 +160,22 @@ public BlockMgr buildBlockMgr(FileSet fileSet, String ext, IndexParams params) {
         }
     }
 
+    // Object files currently, don't need journalling. Because they aer apend only, they
```
kinow: typo (aer/are) but not important. Internal comment only.

afs: Fixed!

PR #268, opened 2017-07-07
JENA-1371: Fuseki renames

afs:


afs:
@kinow - thanks for the prompt review.

This came from using embedded Fuseki with multiple servers in a single JVM for testing complex scenarios of start-stop-restart.

PR #271, opened 2017-07-27
JENA-1378 : Set the accept header in the default use case.

afs:
...while still respecting any direct HttpClient setup.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/RDFParserBuilder.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/RDFParserBuilder.java
@@ -26,7 +26,6 @@ import java.util.*;
 
 import org.apache.http.Header;
 import org.apache.http.client.HttpClient;
-import org.apache.http.impl.client.cache.CachingHttpClientBuilder;
 import org.apache.http.message.BasicHeader;
 import org.apache.jena.atlas.lib.IRILib;
 import org.apache.jena.graph.BlankNodeId;
@@ -35,6 +34,7 @@ import org.apache.jena.riot.lang.LabelToNode;
 import org.apache.jena.riot.system.*;
 import org.apache.jena.riot.system.stream.StreamManager;
 import org.apache.jena.riot.web.HttpNames;
+import org.apache.jena.riot.web.HttpOp ;
 import org.apache.jena.sparql.core.DatasetGraph;
 import org.apache.jena.sparql.util.Context;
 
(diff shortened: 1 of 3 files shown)


jena-arq/src/main/java/org/apache/jena/riot/RDFParserBuilder.java
```diff
@@ -504,8 +504,10 @@ private HttpClient buildHttpClient() {
             Header header = new BasicHeader(k, v);
             hdrs.add(header);
         });
-        HttpClient hc = CachingHttpClientBuilder.create().setDefaultHeaders(hdrs).build();
-        return hc;
+        HttpClient hc = HttpOp.createPoolingHttpClientBuilder()
```
ajs6f: Should we maybe just put the default headers into the clients built by `HttpOp` as options (e.g. `HttpOp.createPoolingHttpClientBuilder(boolean useDefaultAcceptHeaderForRdf)` or something like that?

afs: A possibility though I like that HttpOp is neutral as to its usage - it is not RDF-specific and the default headers are none.  There are different headers for graph, datasets and result sets. Responsibility for the customization is in the RDFParser which here (this is the case of the app making detailed settings) captures those choices for a reusable parser process.

Adding to the complexity of HttpOp.createPoolingHttpClient seems a slippery slope that goes against the builder pattern (IMO). It would be easier if there were an HttpComponents way to say "make a builder that would produce this HttpClient" capturing the internal settings details of an HttpClient.

afs: Merged, which fixes the bug as reported. That does not preclude refining it some more.

ajs6f: For future refining, my understanding of `HttpClientBuilder` is that it does exactly what you describe above. IOW, you can call `build()` on it and get a client but keep it around to call `build()` on it later for another same-config client, over and over again.

afs: What is required is a builder that is based on another but isolated from it by cloning (or in some way) so that adding new builder settings only changes the clone, not also the original (the current active default in `HttpOp` in this case).`HttpClientBuilder` is a setter only style, you can't get settings from it (builder builder needed ?!). `RDFparserBuilder` has `clone()`.

PR #273, opened 2017-08-17
JENA-1372: fn:apply and fn:collation-key

afs:


kinow:
+1 code looks good!

Are fn:collation-key and fn:apply a replacement for that other custom function for collation? I vaguely remember reading in the mailing list about something that existed in the SPARQL spec for collations... are these functions related to that?

Thanks!

afs:
This is just completing the coverage for F&O 3.1.

I don't think `fn:apply` replaces any custom function. There is a syntax extension `CALL` (little know, undocumented; a toy from a long time ago) which is very similar.

"FN_Collation" is related `fn:collation-key` has a specific, different result type specified by the standard.  (there isn't a `fn:collation` in Functions and Operators 3.1 (F&O) so it is oddly named - it's wired to an ARQ URI which is correct)  Can I rename FN_Collation as "collation" to make the naming systematic? (FN_* are F&O functions).

With `fn:apply` it would be nice is to be able to call `op:` things e.g. `op:numeric-add` which aren't callable functions in F&O so by my reading, aren't covered.

As all the operation code (`E_*`) are not inheriting from "Function", a bit of adapter code is needed to register them.   Some messing around with arities.

kinow:
Thanks for the explanation Andy!

jena-arq/src/main/java/org/apache/jena/sparql/function/library/FN_Apply.java
```diff
@@ -0,0 +1,96 @@
...
+ * {@code fn:apply(function, args)}
+ */
+public class FN_Apply extends FunctionBase {
+    // Assumes one object per use site. 
+    private Cache<String, Function> cache1 = CacheFactory.createOneSlotCache();
+    
+    @Override
+    public void checkBuild(String uri, ExprList args) {
+        if ( args.isEmpty() )
+            throw new ExprBuildException("fn:apply: no function to call (minimum number of args is one)");
```
ajs6f: I may be misunderstanding the semantics of `apply`, but would it useful to have some kind of check here to see that the parameter actually names a registered function (or op)?

jena-arq/src/main/java/org/apache/jena/sparql/function/library/FN_Apply.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+    @Override
+    public NodeValue exec(List<NodeValue> args) {
+        if ( args.isEmpty() )
```
ajs6f: Does `checkBuild` not have to get called?

jena-arq/src/main/java/org/apache/jena/sparql/function/library/FN_Apply.java
```diff
@@ -0,0 +1,96 @@
...
+            throw new ExprBuildException("fn:apply: no function to call (minimum number of args is one)");
+        NodeValue functionId = args.get(0);
+        List<NodeValue> argExprs = args.subList(1,args.size()) ; 
+        ExprList exprs = new ExprList();
+        argExprs.forEach((a)->exprs.add(a));
```
ajs6f: Maybe `argExprs.forEach(exprs::add);`?

 It's not clear to me why not `ExprList exprs = new ExprList(argExprs);`. I'm sure there's a reason-- I must be missing something in the execution flow? Why copy instead of view?

jena-base/src/main/java/org/apache/jena/atlas/lib/cache/Cache0.java
```diff
@@ -39,7 +39,13 @@ public V getIfPresent(K key) {
 
     @Override
     public V getOrFill(K key, Callable<V> callable) {
-        return null ;
+        try {
+            return callable.call() ;
+        }
+        catch (Exception e) {
+            e.printStackTrace();
```
ajs6f: `printStackTrace()`? Isn't it better to use a logger?

jena-arq/src/main/java/org/apache/jena/sparql/function/library/FN_Apply.java
afs: Yes - it would be better exception.  Added.

jena-arq/src/main/java/org/apache/jena/sparql/function/library/FN_Apply.java
afs: Ideally, but no harm checking. `build` is only guaranteed to happen when called from parsed code. The build pass is a query thing to move errors from runtime to setup time. Otherwise, you get many, many runtime errors.

jena-arq/src/main/java/org/apache/jena/sparql/function/library/FN_Apply.java
afs: Limitation of Java generics.  `List<NodeValue>` isn't a `List<Expr>` so there is no constructor. 
`ExprList` can't implement a wildcard-extends in `Iterable<Expr>`.

jena-arq/src/main/java/org/apache/jena/sparql/function/library/FN_Apply.java
ajs6f: Okay, so the checks could be factored out and either called from both `checkBuild` and `exec`, right?

jena-arq/src/main/java/org/apache/jena/sparql/function/library/FN_Apply.java
ajs6f: ARGH OMG I hate you Java generics. I've gotten used to Scala's flexibility.

PR #274, opened 2017-08-19
JENA-1381: Use all information in the cache key (text queries).

afs:
See https://issues.apache.org/jira/browse/JENA-1381 for details.

### DIFF
--- a/jena-text/src/main/java/org/apache/jena/query/text/TextQueryPF.java
+++ b/jena-text/src/main/java/org/apache/jena/query/text/TextQueryPF.java
@@ -29,6 +29,7 @@ import org.apache.jena.atlas.lib.CacheFactory ;
 import org.apache.jena.atlas.logging.Log ;
 import org.apache.jena.datatypes.RDFDatatype ;
 import org.apache.jena.datatypes.xsd.XSDDatatype ;
+import org.apache.jena.graph.Graph ;
 import org.apache.jena.graph.Node ;
 import org.apache.jena.ext.com.google.common.collect.LinkedListMultimap;
 import org.apache.jena.ext.com.google.common.collect.ListMultimap;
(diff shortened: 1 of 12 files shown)


afs:
This addresses the cache-key part of the issue, not graph naming part.

afs:
This adds `NamedGraph` - a graph wrapper that carries around a graph name.

The problem is now that graphs in a general dataset only have an associated name by virtue of the slot in the map they occupy, unlike ones from a quad dataset where the graph is a `GrahView`, a restriction of the "G" slot of the quad to a specific name.

The updated test case on [JENA-1381](https://issues.apache.org/jira/browse/JENA-1381) shows this - it tests three kinds of dataset in memory.

| Dataset factory | Implementation | Characteristics | Status (before PR) |
| --------------- | -------------- | --------- | ---- |
| `create()`               | `DatasetGraphMap` | Copy-in on add graph | No graph index |
| `createGeneral()`  | `DatasetGraphMapLink` | Link to added graphs | No graph index |
| TIM | Quads |   `getGraph` returns a `GraphView` | OK |
| TDB | Quads |   `getGraph` returns a `GraphView` | OK |

**`create()`** : `DatasetGraphMap` can change to use a  `NamedGraph` by changing the `GraphMaker` interface to take a name and then using `NamedGraphWrapper` aroudn an in-memeory graph.

**`createGeneral()`** : `DatasetGraphMapLink` can either change to return a `NamedGraph` which mean 
the graph returned by `getGraph` is not the graph added, or can remain not supporting test query graph. The general dataset is already less aligned with the other implementations and can be considered "special purposes only" - usually complex mixtures of graph types or inference graphs.

The javadoc for `DatasetFactory` is supposed to discourage use of the linking `createGeneral`.

The recent commits provide this and update the javadoc in `DatasetFactory`.

afs:
That would be contrary to what every other place does. Having a name for the default graph unifies the design.  Else, we'd end up with `Optional<Node>` all over the place, or compensatory code.

ajs6f:
If we ever want to use `Optional` at all (and I would, I think it is clear and avoids special names in many cases), we have to start somewhere (or we have to make a massive sudden change to the Graph SPI, maybe 4.0?). I don't want to make a fuss, I would just like to be able to gradually introduce it. Maybe not on this PR, and maybe gradual isn't better than a big change at 4.0.

afs:
Here, the design should align with `Quad`.

What's wrong with special names _for graphs_ anyway? There are 3 special names already, and it's an extensible approach for other names - semantic web/linked data is about naming things.

`Optional` is for a different situation, null means absent or unknown, not "different".

ajs6f:
I'm not that worried about this case. (Although I would actually have fewer special graph names and more types, but that's just my taste; I'm not arguing that we should change that now.) It was more your first remark about "Else, we'd end up with `Optional<Node>` all over the place" and the fact that I don't feel like we have a clear way to make any changes at all to the core SPI and API. This (PR) isn't really the right place for the larger discussion-- I'll take it to dev@.

jena-text/src/main/java/org/apache/jena/query/text/TextQueryPF.java
```diff
@@ -240,6 +240,36 @@ private QueryIterator concreteSubject(Binding binding, Node s, Node score, Node
...
+            if ( log.isDebugEnabled())
+                log.debug("Text query: {} ({})", queryString,limit) ;
+        } else {
+            Explain.explain(execCxt.getContext(), "Text query <"+graphURI+">: "+queryString) ;
+            if ( log.isDebugEnabled())
+                log.debug("Text query: {} <{}> ({})", queryString, graphURI, limit) ;
+        }
+        
+        // Cache-key does not matter if lang or graphURI are null
+        String cacheKey = limit + " " + property + " " + queryString + " " + lang + " " + graphURI ;
```
afs: @osma If you have a moment, please could you check I haven't misunderstood the design here and that changing the cache-key will not cause other problems.

osma: Thanks for pinging @afs. It's been a while since I looked at this code, but I think that the fix is the correct one here. I'm not sure I understand the comment "Cache-key does not matter if lang or graphURI are null" though - what's that intended to mean?

afs: Thanks for checking.

String "+" works on nulls (adds the word 'null') so we don't need to special case it.

PR #275, opened 2017-08-20
JENA-1383: Improve handling of bad character encoding.

afs:
https://issues.apache.org/jira/browse/JENA-1383

### DIFF
--- a/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ActionSPARQL.java
+++ b/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ActionSPARQL.java
@@ -23,6 +23,7 @@ import static org.apache.jena.fuseki.server.CounterName.RequestsBad ;
 import static org.apache.jena.fuseki.server.CounterName.RequestsGood ;
 
 import java.io.InputStream ;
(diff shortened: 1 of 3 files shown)


jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ActionSPARQL.java
```diff
@@ -205,7 +206,11 @@ public static void parse(HttpAction action, StreamRDF dest, InputStream input, L
                 .lang(lang)
                 .base(base)
                 .parse(dest);
-        } 
+        } catch (RuntimeIOException ex) {
+            if ( ex.getCause() instanceof CharacterCodingException )
+                throw new RiotException("Character Coding Error: "+ex.getMessage());
```
ajs6f: maybe `throw new RiotException("Character Coding Error: "+ex.getMessage(), ex.getCause());`  to keep more context?

afs: This is an error caught and handled as an HTTP servlet response and is a bad request - the client's mistake. It is not an error in the server to track down so I see this as normal processing from the server's point of view. The message is the information needed for the HTTP response.

PR #278, opened 2017-09-17
[JENA-1392] Add DynamicDatasets support to QueryEngineSDB for FROM an…

grahamtriggs:
…d FROM NAMED queries


afs:
Good idea to treat FROM etc this way in SDB.

It should be the same as pattern as TDB does (I'm not at a development computer ATM - I'll look when I'm near an IDE).

sarbajitdutta:
Thanks for merging this.

PR #279, opened 2017-09-23
JENA-1394: Add operations to serialize to and deserialize from byte arrays

afs:
https://issues.apache.org/jira/browse/JENA-1394


afs:
And some other minor cleaning up.

PR #280, opened 2017-10-01
Deprecate Jena's Callback<T> in favor of Java API's Consumer<T>

ajs6f:
Seems like we could deprecate `Callback<T>` in the next release and remove in the following, unless I am missing something about the contract for callbacks.


afs:
+1

PR #282, opened 2017-10-02
JENA-1393: Format prefix names

afs:
This PR includes the patch on [JENA-1393](https://issues.apache.org/jira/browse/JENA-1393) and also applies the same to Turtle writing.

":" in prefix local names is used on output which is spec-legal.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/util/FmtUtils.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/util/FmtUtils.java
@@ -535,6 +535,7 @@ public class FmtUtils
     {
         if ( Character.isLetterOrDigit(ch) ) return true ;
         if ( ch == '.' )    return true ;
+        if ( ch == ':' )    return true ;
         if ( ch == '-' )    return true ;
         if ( ch == '_' )    return true ;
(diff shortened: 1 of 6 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/util/FmtUtils.java
```diff
@@ -535,6 +535,7 @@ private static boolean validPNameChar(char ch)
     {
         if ( Character.isLetterOrDigit(ch) ) return true ;
         if ( ch == '.' )    return true ;
+        if ( ch == ':' )    return true ;
```
ajs6f: This is getting long enough that it might read better as a `switch`/`case`.

afs: If only `switch` accepted `Character.isLetterOrDigit` ... (this is not `match`!)

ajs6f: Oh, missed that. Yeah, I was reading some Scala the other day and it made me sad to come back to ADT-less Java. 😞

PR #285, opened 2017-10-11
JENA-1398: Marked archaic FOAF terms in the vocabulary as deprecated.

shirren:
@afs The approach was to download the latest FOAF.rdf from the W3C website from which the FOAF.ttl and FOAF.java file was generated. "Archaic" terms from the spec were then marked as @Deprecated in the Java code. All tests pass. Please provide feedback or comments, thanks in advance.


jena-arq/src/main/java/org/apache/jena/sparql/vocabulary/FOAF.java
```diff
@@ -19,242 +19,282 @@
...
+
+    /** <p>A checksum for the DNA of some thing. Joke.</p> */
+    @Deprecated
+    public static final Property dnaChecksum = M_MODEL.createProperty( "http://xmlns.com/foaf/0.1/dnaChecksum" );
+
+    /** <p>The family name of some person.</p> */
+    public static final Property familyName = M_MODEL.createProperty( "http://xmlns.com/foaf/0.1/familyName" );
+
+    /** <p>The family name of some person.</p> */
+    @Deprecated
```
rvesse: For deprecated terms like this where there is a new alternative can we explicitly point to the new term to help users migrate forwards

shirren: @rvesse that's a good idea. I'll make it so.

shirren:
@rvesse I've made those changes. How does the Jena team schedule releases, and do you know when this PR will be released? Thank you so much for working on this projects, it's been invaluable to our team.

rvesse:
Thanks for the updates, looks good to me

We aim to release approximately every three months, the last release was July so we are probably due another release soon. However everything is done on volunteer time so it will really depend on peoples availability

shirren:
@rvesse understand, thanks once again.

PR #287, opened 2017-10-12
JENA-1391: adding isEmpty method to Dataset

ajs6f:
One of the asks from JENA-1391. Not by any means the whole ticket. :)

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/query/Dataset.java
+++ b/jena-arq/src/main/java/org/apache/jena/query/Dataset.java
@@ -115,9 +115,7 @@ public interface Dataset extends Transactional
     public void close() ;
     
     /**
-     * @return Whether this {@code Dataset} is empty of graphs. Be aware of the semantic looseness inherent in
-     * <a href="https://www.w3.org/TR/2014/REC-rdf11-concepts-20140225/#h_note_4">the definition
-     * of RDF Datasets</a>; whether a named graph exists if nothing is in it is implementation-specific.
+     * @return Whether this {@code Dataset} is empty of triples, whether in the default graph or in any named graph.
      */
     boolean isEmpty();
 }


jena-arq/src/main/java/org/apache/jena/query/Dataset.java
```diff
@@ -113,4 +113,11 @@
      *  The dataset can not be used for query after this call.
      */
     public void close() ;
+    
+    /**
+     * @return Whether this {@code Dataset} is empty of graphs. Be aware of the semantic looseness inherent in
+     * <a href="https://www.w3.org/TR/2014/REC-rdf11-concepts-20140225/#h_note_4">the definition
+     * of RDF Datasets</a>; whether a named graph exists if nothing is in it is implementation-specific.
+     */
+    boolean isEmpty();
```
afs: "is empty of graphs" 

Firstly, there is always at least one graph (the default graph).

Secondly, "isEmpty" on a DSG means no triples in any graph (default implement is `!contain(ANY,ANY,ANY,ANY)` which includes the default graph).

jena-arq/src/test/java/org/apache/jena/sparql/core/AbstractTestDataset.java
```diff
@@ -108,4 +108,12 @@
...
         assertTrue(model2.isIsomorphicWith(ds.getNamedModel(graphName))) ;
     }
+    
+    @Test public void dataset_06()
+    {
+        String graphName = "http://example/" ;
+        Dataset ds = createDataset() ;
+        ds.addNamedModel(graphName, model1) ;
+        assertFalse("Dataset should not be empty after a named graph has been added!", ds.isEmpty());
+    }
```
afs: This is a test passes because `model1` is not empty - not because it is used as a named graphs.

```
        DatasetGraph dsg = DatasetGraphFactory.createTxnMem(); // or createGeneral
        Graph g = GraphFactory.createGraphMem();
        dsg.addGraph(NodeFactory.createURI("http;//example/g"), g);
        boolean b = dsg.isEmpty();
        System.out.println(b) ;
```
prints "true".

jena-arq/src/main/java/org/apache/jena/query/Dataset.java
ajs6f: Oh, fudge. Then we really can't have a default impl of this, can we?

jena-arq/src/test/java/org/apache/jena/sparql/core/AbstractTestDataset.java
ajs6f: See above-- so do we need a different impl of `isEmpty` for every kind of `DatasetGraph`?

jena-arq/src/main/java/org/apache/jena/query/Dataset.java
afs: Sure we can.  Change the javadoc to "returns true if the dataset contains no statements in any model". That's what it does, and what the test does anyway, because the impl passes down to `DatasetGraph.isEmpty`. That is the useful operation.

ajs6f: I didn't do that at first because it felt like a bit of a conflict against the rest of the API for `Dataset`, which discusses models/graphs and not tuples. But if you're okay with it, it doesn't bother me.

jena-arq/src/main/java/org/apache/jena/query/Dataset.java
```diff
@@ -113,4 +113,9 @@
      *  The dataset can not be used for query after this call.
      */
     public void close() ;
+    
+    /**
+     * @return Whether this {@code Dataset} is empty of triples, whether in the default graph or in any named graph.
```
ajs6f: @afs  Better?

afs: Yes!
+1 to merging.

PR #288, opened 2017-10-13
JENA-1401 (fuseki backup) Don't use Jetty code in war servlets.

afs:


jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/async/AsyncPool.java
```diff
@@ -51,7 +51,13 @@ public AsyncTask submit(Runnable task, String displayName, DataService dataServi
         synchronized(mutex) {
             String taskId = Long.toString(++counter) ;
             Fuseki.serverLog.info(format("Task : %s : %s",taskId, displayName)) ;
-            Callable<Object> c = Executors.callable(task) ;
+            Callable<Object> c = ()->{
+                try { task.run(); } 
+                catch (Throwable th) {
+                    Fuseki.serverLog.warn(format("Exception in task %s execution", taskId), th);
```
ajs6f: Does this qualify as an error? (Logging-wise)

afs: Maybe. It is a development bug!

PR #289, opened 2017-10-21
Version bumps for 3.5

ajs6f:
There are 5 commits here, the first 4 of which are (I think) non-controversial. In the last, to get from Commons Lang 3.4 to 3.5 (and thence to 3.6) I had to change test code.

I think the changes are kosher-- using `Z` instead of `+00:00` is legit according to the [XSD Dataypes spec](https://www.w3.org/TR/xmlschema11-2/#nt-tzFrag). But it might cause some surprises.

### DIFF
--- a/jena-project/pom.xml
+++ b/jena-project/pom.xml
@@ -80,7 +80,7 @@
 
     <ver.elasticsearch>5.2.2</ver.elasticsearch>
     <!-- elasticsearch uses this by reflection -->
-    <ver.log4j2>2.7</ver.log4j2>
+    <ver.log4j2>2.9.1</ver.log4j2>
 
     <ver.spatial4j>0.6</ver.spatial4j>
 
--- a/jena-project/pom.xml
+++ b/jena-project/pom.xml
@@ -92,7 +92,7 @@
 
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.build.timestamp.format>yyyy-MM-dd'T'HH:mm:ssZ</maven.build.timestamp.format>
-    <ver.contract.tests>0.1.5</ver.contract.tests>
+    <ver.contract.tests>0.1.7</ver.contract.tests>
   </properties>
 
   <profiles>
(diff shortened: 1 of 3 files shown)


afs:
I don't think we should be updating versions this late in the release cycle unless they fix identified bugs. They don't get tested by SNAPSHOT use.

Both the thrift change (not incremental - do we need to rebuild from IDL? Binary changes?) and the commons-lib change need closer inspection. Probably code elsewhere (TDBx NodeId encoding?) for +00:00/Z differences (syntax vs value).

afs:
And no JIRA :-)  People do look sometimes.

ajs6f:
Okay, let's let it hang until after the release then. And I'll file a Jira.

jena-project/pom.xml
```diff
@@ -1,867 +1,828 @@
...
-      <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>findbugs-maven-plugin</artifactId>
-        <version>3.0.1</version>
-      </plugin>
-
-    </plugins>
-
-  </reporting>
+<!-- Licensed to the Apache Software Foundation (ASF) under one or more contributor
```
afs: This change is reported as: +831 −869 because tabs have been used.

1. Can't see what has materially changed.
1. It changes the licences header
1. Please - no tabs - display can be bad on the web (worse it if it gets mixed)
1. It asks in CONTRIBUTING.md :grin:

ajs6f:
Wait, something has gone wrong here. The only changes I made were to the values of a few properties in that `pom.xml`. I have no idea why it's doing such a giant diff. I need to figure that out.

jena-project/pom.xml
ajs6f: Wait, something has gone wrong here. The only changes I made were to the values of a few properties in that `pom.xml`. I have no idea why it's doing such a giant diff. I need to figure that out.

afs:
I've picked out the Thrift update and done it because it may affect external data. 

JENA-1417

ajs6f:
I'm closing this PR because for such a small changeset, it's not worth figuring out how my delta gots screwed up with the tabs in the `pom.xml`. I'll just open a fresh PR in a day or three, for JENA-1418.

PR #294, opened 2017-10-25
Fix and tests for possible NPE (JENA-1405)

rvesse:
Fixes a possible NPE in AdaptorFileManager by aligning its behaviour
with FileManager to throw NotFoundException

Adds specific tests for these cases

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/adapters/AdapterFileManager.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/adapters/AdapterFileManager.java
@@ -21,6 +21,7 @@ package org.apache.jena.riot.adapters;
 import java.io.InputStream ;
 import java.util.Iterator ;
 
+import org.apache.jena.atlas.logging.FmtLog;
 import org.apache.jena.atlas.web.TypedInputStream ;
 import org.apache.jena.rdf.model.Model ;
 import org.apache.jena.riot.Lang ;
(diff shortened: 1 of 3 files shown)


jena-arq/src/main/java/org/apache/jena/riot/adapters/AdapterFileManager.java
```diff
@@ -285,6 +286,12 @@ protected Model readModelWorker(Model model, String filenameOrURI, String baseUR
         if ( baseURI == null )
             baseURI = SysRIOT.chooseBaseIRI(filenameOrURI) ;
         try(TypedInputStream in = streamManager.openNoMapOrNull(mappedURI)) {
+            if ( in == null )
+            {
+                if ( log.isDebugEnabled() )
+                    log.debug("Failed to locate '"+mappedURI+"'") ;
```
ajs6f: As I understand it, `log.debug("Failed to locate '{}'", mappedURI) ` [avoids the need](https://www.slf4j.org/faq.html#logging_performance) to explicitly check `isDebugEnabled()`.

rvesse: Yes, I just copied and pasted this from the parent class where the same logic is present

afs: More recently, `FmtLog.debug(logger, fmtString, args...)`.

PR #301, opened 2017-10-30
JENA-1413: increase Elasticsearch startup timeout to 60 seconds

osma:
This attempts to fix intermittent build problems with jena-text-es by increasing the startup timeout.
See https://issues.apache.org/jira/browse/JENA-1413


kinow:
Sounds like a good idea. I got a build on Windows stuck, in one of the -text projects. After sending a CTRL+C, noticed a pending java process. Looking at the command line, I could see it had some elasticsearch mention in it. Killed manually, and executed the `-Pdev` and `-Pbootstrap` to build the project. Wonder if this would fix the issue I had?

PR #303, opened 2017-11-04
JENA-1408: Quicker -Pdev; simplify profiles.

afs:
This simplifies and speeds up the quick development profile.

* Don't skip jena-iri and jena-shared-guava in -Pdev. Skipping does not always seem to work when there are new Jenkins built snapshots as well. The time difference is small.
* Remove -Pbootstrap because it is now the same as -Pdev
* Don't build javadoc when running -Pdev

(see the JIRA for some timings)

### DIFF
--- a/jena-project/pom.xml
+++ b/jena-project/pom.xml
@@ -115,6 +115,23 @@
         </plugins>
       </build>
     </profile>
+
+    <profile>
+      <id>dev</id>
+      <build>
+        <plugins>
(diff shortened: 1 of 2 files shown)


ajs6f:
What's here looks good, but I don't see where `jena-iri` and `jena-shaded-guava` are being added in?

afs:
Sorry - missed that copying over the test setup. Now done.

PR #304, opened 2017-11-04
JENA-1418: Upgrading minor dependencies and plugins

ajs6f:


jena-project/pom.xml
```diff
@@ -813,7 +833,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-report-plugin</artifactId>
-        <version>2.18.1</version>
+        <version>2.20.1</version>
         <configuration>
```
afs: From the parent? We may have had to set this to override in the past but that has moved on.

jena-project/pom.xml
```diff
@@ -740,18 +754,17 @@
...
           <configuration>
             <overWriteReleases>false</overWriteReleases>
             <overWriteIfNewer>true</overWriteIfNewer>
           </configuration>
         </plugin>
 
         <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-eclipse-plugin</artifactId>
-          <version>2.9</version>
```
afs: Can we remove this plugin all together? Now Eclipse has built-in maven support, I don't think this plugin is necessary. Need to check more widely first.

jena-project/pom.xml
```diff
@@ -668,8 +682,8 @@
         </plugin>
 
         <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-source-plugin</artifactId>
+          <version>3.0.1</version>
```
afs: Isn't this inherited from the apache parent POM? Ditto others.

jena-project/pom.xml
```diff
@@ -74,13 +75,14 @@
...
+    
+    <ver.shiro>1.4.0</ver.shiro>
 
-    <ver.commons-codec>1.10</ver.commons-codec>
     <ver.lucene>6.4.1</ver.lucene>
 
     <ver.elasticsearch>5.2.2</ver.elasticsearch>
     <!-- elasticsearch uses this by reflection -->
-    <ver.log4j2>2.7</ver.log4j2>
+    <ver.log4j2>2.9.1</ver.log4j2>
```
afs: Given the warning in the comment, sticking to 2.7 is safer (it is what ES depends on).

jena-permissions/pom.xml
```diff
@@ -19,9 +19,7 @@
 
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
-  <groupId>org.apache.jena</groupId>
   <artifactId>jena-permissions</artifactId>
-  <version>3.6.0-SNAPSHOT</version>
```
afs: This isn't "Upgrading minor dependencies".

IIRC Having the version in the module is, (or was - maven changes) necessary so that modules can be built on their own during development.

This is the style elsewhere in the project - one change and not others seems odd.

PR #306, opened 2017-11-12
Algorithms for JENA-1414

afs:
This PR rstructures `GraphUtil.deleteFrom` to put the decision on whether to loop on the target (_dst_) graph or the source (_src_) graph.

There are 3 algorithms for discussion:

1. Use the size of the graphs - this is the current Jena 3.5.0 policy
2. Use the size of the _src_ and iterate on the _dst_ to compare sizes
3. Iterator on both _src_ and _dst_ to compare sizes. `Graph.size` is not used at all.

The cost of `Graph.size()` can be small (already known) or large (needs to be calculated to be accurate). The latter is bad for large persistent graphs.

The PR also adds javadoc to explain how to call a specific algorithm (_src_ loop or _dst_ loop).

### DIFF
--- a/jena-base/src/main/java/org/apache/jena/atlas/iterator/Iter.java
+++ b/jena-base/src/main/java/org/apache/jena/atlas/iterator/Iter.java
@@ -351,6 +351,22 @@ public class Iter<T> implements Iterator<T> {
         return filter(iter, Objects::nonNull) ;
     }
 
+    /** Step forward up to {@code steps} places.
+     * <br/>Return number of steps taken.
+     * 
+     * @apiNote
+     * The iterator is moved at most {@code steps} places with no overshoot.
+     * The iterator can be used afterwards.
+     */ 
+    public static int step(Iterator<?> iter, int steps) {
+        for ( int i = 0 ; i < steps; i++) {
+            if ( ! iter.hasNext() ) 
+                return i;
+            iter.next();
+        }
+        return steps;
+    }
+
     /** Take the first N elements of an iterator - stop early if too few */
     public static <T> List<T> take(Iterator<T> iter, int N) {
         iter = new IteratorN<>(iter, N) ;
--- a/jena-core/src/main/java/org/apache/jena/graph/GraphUtil.java
+++ b/jena-core/src/main/java/org/apache/jena/graph/GraphUtil.java
@@ -23,6 +23,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 
+import org.apache.jena.atlas.iterator.Iter;
 import org.apache.jena.graph.impl.GraphWithPerform;
 import org.apache.jena.util.IteratorCollection;
 import org.apache.jena.util.iterator.ExtendedIterator;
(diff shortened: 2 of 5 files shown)


jena-core/src/main/java/org/apache/jena/graph/GraphUtil.java
```diff
@@ -223,6 +258,116 @@ public static void deleteFrom(Graph dstGraph, Graph srcGraph) {
...
+        if ( step2 <= 0 )
+            throw new IllegalArgumentException("step size 'step2' is not positve: "+step2);
+        
+        int x = 0; 
+        for ( ;; ) {
+            x++ ;
+            boolean ended1 = iter1.hasNext();
+            boolean ended2 = iter2.hasNext();
+            
+            if ( ended1 )
```
rvesse: Shouldn't these be `! ended1` and `! ended2`?

afs: Good catch. It should be `boolean ended1 = ! iter1.hasNext();`

This work needs to test the worker functions themselves because the wrong execution policy decision isn't going to show up in the test suite. That seems better than adding large scale testing which isn't black/white on this sort of thing.

afs:
Revised version.

The co-iteration algorithm (choice 3) is complicated. Determining the right thing to do each time is expensive and still a pragmatic choice with graph probing. The app may well know what to do so call the appropriate method as already documented.

My suggestion is algorithm 2 (a slightly improved version of the patch on  JENA-1414), document what to do if src is a large persistent database and remove the complicate algorithm 3.

If that is our decision, I will produce a much simpler commit for algorithm 2.

jena-base/src/main/java/org/apache/jena/atlas/iterator/Iter.java
```diff
@@ -351,6 +351,22 @@ public void remove() {
         return filter(iter, Objects::nonNull) ;
     }
 
+    /** Step forward up to {@code steps} places.
+     * <br/>Return number of steps taken.
```
ajs6f: `@return number of steps taken`

afs:
> @afs I'm not sure which algo is number 3?

_2. Use the size of the src and iterate on the dst to compare sizes
3. Iterator on both src and dst to compare sizes. Graph.size is not used at all._

jena-core/src/main/java/org/apache/jena/graph/GraphUtil.java
```diff
@@ -246,43 +282,214 @@ private static void deleteIteratorWorkerDirect(Graph graph, Iterator<Triple> it)
...
+     * Delete triples in the destination (arg 1) as given in the source (arg 2).
+     *
+     * @implNote
+     *  This is designed for the case of {@code dstGraph} being comparable or much larger than
+     *  {@code srcGraph} or {@code srcGraph} having a lot of triples to actually be
+     *  deleted from {@code dstGraph}. This includes large, persistent {@code dstGraph}.
+     *  <p>  
+     *  It is not designed for a large {@code srcGraph} and large {@code dstGraph} 
+     *  with only a few triples in common delete from {@code dstGraph}. It is better to
+     *  calculate the difference in someway, and copy into a small graph to use as the {@srcGraph}.
```
ajs6f: typo: some way

ajs6f:
Okay, now I get it. Agreed that number 3 is "trying too hard" and on the proposal to provide number 2 and document appropriate usage.

afs:
Force-pushed the reduced code - algorithm 2 is the active one.  Algorithm 1 is there but not active.

PR #307, opened 2017-11-13
JENA-1418: Upgrade some versions

afs:
This extracts the suggested version upgrades from PR#304.

It removes [maven-eclipse-plugin](http://maven.apache.org/plugins/maven-eclipse-plugin/) which is retired and no longer maintained.

### DIFF
--- a/jena-csv/pom.xml
+++ b/jena-csv/pom.xml
@@ -150,16 +150,6 @@
           <overWriteIfNewer>true</overWriteIfNewer>
         </configuration>
       </plugin>
-      <plugin>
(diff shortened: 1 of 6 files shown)


afs:
I'd like to get version changes in so they have more time to bed down.

I took the proposed version changes in PR #304 which seem uncontroversial.

PR #289 also has Commons Lang 3.4 -> 3.6 which I haven't had time to investigate yet.

ajs6f:
👍 the Commons Lang jump is more complicated because of the issues around ISO date format.

PR #308, opened 2017-11-14
JENA-1384: Canonical literals: lexical form and langTags

afs:
RDFParser options to handle canonical lexcial form for values ("+0123" becomes "123" for an xsd:integer) and options for handling language tags (to RFC-canonical form; to lowercase).

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/process/normalize/CanonicalizeLiteral.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/process/normalize/CanonicalizeLiteral.java
@@ -39,6 +39,9 @@ public class CanonicalizeLiteral implements Function<Node, Node>
 
     private CanonicalizeLiteral() {}
     
+    /**
+     * Canonicaize a literal, both lexical form and language tag (RFc canonical). 
+     */
     @Override
     public Node apply(Node node) {
         if ( ! node.isLiteral() )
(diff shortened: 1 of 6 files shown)


eroux:
Thanks a lot for the pull request!

jena-arq/src/main/java/org/apache/jena/riot/process/normalize/CanonicalizeLiteral.java
```diff
@@ -73,6 +76,36 @@ public Node apply(Node node) {
...
+            return node ;
+        return n2 ;
+    }
+    
+    /** Convert the language tag of a lexical form to a canonical form if one of the known datatypes,
+     * otherwise return the node argument. (same object; compare by {@code ==})  
+     */
     private static Node canonicalLangtag(String lexicalForm, String langTag) {
         String langTag2 = LangTag.canonical(langTag);
         if ( langTag2.equals(langTag) )
```
rvesse: Shouldn't we be returning `node` not `null` in the subsequent line?

afs: Here, node isn't passed in so it can't be returned. Style thing. Node is already known to have a language tag so I don't like passing in a Node which can be wrong e.g.through mis-call from somewhere else.. Passing lex+lang forces it to be the information for a language tagged literal.

It's tested at line 74
```
        if ( n2 == null )
            return node ;
```
and elsewhere conversion also sometimes returns `null` for "no conversion" which means no new node is needed which is more efficient (meaureably).

PR #310, opened 2017-11-15
JENA-1422:  Transactions and switchable graphs

afs:


osma:
FYI, testing this locally.

osma:
At least the loading into graph (JENA-1422 original problem) seems to work now. I tested both with a small and a large-ish data set.

PR #311, opened 2017-11-16
JENA-1424: load http

afs:
There is also a has a minor change to `RDFParser`/`RDFParserBuilder` to put the convenience operations for parse-to-graph and parse-to-dataset on a `RDFParser`, not just on `RDFParserBuilder`.


osma:
Wow @afs, you're super quick!

afs:
This includes efficiency improvement as well. For a non-SILENT LOAD into one of the fully transactional stores, it is done direct, no read buffering to deal with pesky syntax errors. If it goes wrong, the exception propagates up and the transaction aborts as normal (if `Txn` is used or the transaction cycle is written out long form properly).

PR #313, opened 2017-11-20
JENA-1430: Read quads for ja:data by filename

afs:
Add support for {{ja:data "filename"}}.

Strings as filenames and URIs as file names resolve different (the JVM cwd and the assembler file location respectively) which can be confusing for filenames.


afs:
Closed in favour of PR #314.

PR #315, opened 2017-11-23
Add --stop-warnings to riot (JENA-1434)

rvesse:
Add the option to allow riot to abort on warnings

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/system/ErrorHandlerFactory.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/system/ErrorHandlerFactory.java
@@ -270,11 +270,11 @@ public class ErrorHandlerFactory
         }
         
         public boolean hadErrors() {
-            return this.errorCount > 1;
+            return this.errorCount > 0;
         }
         
         public boolean hadWarnings() {
-            return this.warningCount > 1;
+            return this.warningCount > 0;
         }
         
         public boolean hadIssues() {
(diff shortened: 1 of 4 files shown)


rvesse:
Reworked based on @afs's feedback in the JIRA

jena-arq/src/main/java/org/apache/jena/riot/system/ErrorHandlerFactory.java
```diff
@@ -221,6 +224,64 @@ public void fatal(String message, long line, long col) {
         }
     }
     
+    /** An error handler that logs message for errors and warnings and throw exceptions on either */ 
+    public static class ErrorHandlerTracking extends ErrorLogger implements ErrorHandler {
+        private final boolean failOnError, failOnWarning;
+        private long errorCount, warningCount;
+        
+        public ErrorHandlerTracking(Logger log, boolean failOnError, boolean failOnWarning) {
```
afs: Suggestion: `ErrorHandlerTracking` wraps another `ErrorHandler` as argument and forwards all warning/error/fatal calls to that error handler after incrementing its counters.  This makes the choice of how to respond to warning or error dependent on the argument `ErrorHandler`.

jena-cmds/src/main/java/riotcmd/CmdLangParse.java
```diff
@@ -195,7 +202,7 @@ protected void exec() {
         
         // exit(1) if there were any errors.
```
afs: `// exit(1) if there were any errors or warnings.`

PR #316, opened 2017-11-25
JENA-1435: Add service extensibilty to Fuseki2.

afs:
This PR rewrite how request are dispatched so that routing to extensions happens. The standard services use the new mechanism for request dispatch.

The PR also does some clearing up.


afs:
Example custom service, setup and execution to follow.

afs:
Not for 3.6.0.

PR #317, opened 2017-11-27
JENA-1440: TDB2 - transform bytes to NodeIds directly.

afs:


afs:
Not for Jena 3.6.0.

jena-db/jena-dboe-base/src/main/java/org/apache/jena/dboe/base/buffer/RecordBufferIteratorMapper.java
```diff
@@ -0,0 +1,105 @@
...
+                nextIdx = decodeIndex(nextIdx) ;
+        }
+        
+        maxRec = maxRecord ; 
+    }
+
+    private void finish()
+    {
+        rBuff = null ;
+        nextIdx = -99 ;
```
ajs6f: Might be nice to call this out as a constant, like `NO_NEXT_INDEX` or the like.

afs:
I had time for some more testing.

PR #318, opened 2017-11-28
JENA-1441 jena-text tracing for development

xristy:
This adds tracing in `TextQueryPF` and `TextIndexLucene` to track calls to `exec` and subsequent processing as well as the add, update and delete of Lucene documents.

Also adds configuration item `text:cacheQueries` on `text:TextIndexLucene` which allows to bypass use of the `queryCache` in `TextQueryPF.query(...)` during development. This configuration item is not for general use and need not be documented further. `text:cacheQueries` defaults to `true` so there is no net change in the behavior of jena-text.

There is also tidying of code to use `doc.get()` rather than `doc.getValues() followed by vals[0]` in `TextLuceneIndex.query$`.


xristy:
Closing to reissue without the JENA-1438 commit

afs:
FYI: On a PR branch, you can do all sorts of git stuff on your local repo like reset to remove a commit, rebasing, and squashing.  Then do a force push to your PR branch at GH and it will all appear here with the updated state. GH even tracks revisions and hides comments no longer relevant.  (Use all this with care!)

PR #319, opened 2017-11-29
Jena 1441 tracing

xristy:
Opening new to not include JENA-1438 related commit.

This adds tracing in TextQueryPF and TextIndexLucene to track calls to exec and subsequent processing as well as the add, update and delete of Lucene documents.

Also adds configuration item text:cacheQueries on text:TextIndexLucene which allows to bypass use of the queryCache in TextQueryPF.query(...) during development. This configuration item is not for general use and need not be documented further. text:cacheQueries defaults to true so there is no net change in the behavior of jena-text.

There is also tidying of code to use doc.get() rather than doc.getValues() followed by vals[0] in TextLuceneIndex.query$.


jena-text/src/main/java/org/apache/jena/query/text/TextIndexLucene.java
```diff
@@ -204,7 +204,10 @@ public void close() {
 
     @Override public void updateEntity(Entity entity) {
         if ( log.isDebugEnabled() )
-            log.debug("Update entity: " + entity) ;
+            if (log.isTraceEnabled() && entity != null)
```
ajs6f: Better than checking `log.isDebugEnabled()` is to use the idiom `log.debug("Update entity: {}", entity))`, same with other log statements that have guards and are calling `toString` implicitly. Using it for `trace` here makes sense because you aren't just calling `toString` and sadly, SLF4J doesn't yet accept `Supplier`s for messages.

jena-text/src/main/java/org/apache/jena/query/text/TextQueryPF.java
```diff
@@ -224,12 +234,16 @@ private QueryIterator resultsToQueryIterator(Binding binding, Node s, Node score
     }
 
     private QueryIterator variableSubject(Binding binding, Node s, Node score, Node literal, StrMatch match, ExecutionContext execCxt) {
+        if (log.isTraceEnabled())
+            log.trace("variableSubject: {}", match) ;
```
ajs6f: You don't need to guard with `isTraceEnabled()` if you are using this form of `Logger.trace`.

xristy: I'm happy to amend by removing the unnecessary `log.isTraceEnabled()`. I was intending to follow the coding practices elsewhere in TextQueryPF and TextIndexLucene, should I amend the unneeded occurrences of `log.isDebugEnabled()` as well?

jena-text/src/main/java/org/apache/jena/query/text/assembler/TextIndexLuceneAssembler.java
```diff
@@ -144,6 +144,17 @@ public TextIndex open(Assembler a, Resource root, Mode mode) {
                 storeValues = svNode.asLiteral().getBoolean();
             }
 
+            // use query cache by default
+            boolean cacheQueries = true;
+            Statement cacheQueriesStatement = root.getProperty(pCacheQueries);
+            if (null != cacheQueriesStatement) {
+                RDFNode cqNode = cacheQueriesStatement.getObject();
+                if (! cqNode.isLiteral()) {
+                    throw new TextIndexException("text:cacheQueries property must be a string : " + cqNode);
```
ajs6f: This is maybe a little picky, but maybe better: `"text:cacheQueries property must be a boolean`?

ajs6f:
Logging statements look good to me!

jena-text/src/main/java/org/apache/jena/query/text/assembler/TextIndexLuceneAssembler.java
xristy: Thanks. Best to be accurate.

PR #320, opened 2017-11-29
JENA-1438 rdf:langString args

xristy:
JENA-1438 allow rdf:langString arguments to text:query

### DIFF
--- a/jena-text/src/main/java/org/apache/jena/query/text/TextQueryPF.java
+++ b/jena-text/src/main/java/org/apache/jena/query/text/TextQueryPF.java
@@ -295,6 +295,14 @@ public class TextQueryPF extends PropertyFunctionBase {
         return results;
     }
     
+    private boolean noLang(String lang) {
+        return (lang == null || lang.isEmpty() || " ".equals(lang));
+    }
+    
+    private String fixLang(String lang) {
+        return lang.isEmpty() || " ".equals(lang) ? null : lang;
+    }
+    
     /** Deconstruct the node or list object argument and make a StrMatch 
      * The 'executionTime' flag indicates whether this is for a build time
      * static check, or for runtime execution.
@@ -309,14 +317,18 @@ public class TextQueryPF extends PropertyFunctionBase {
                 return null ;
             }
 
+            String lang = o.getLiteralLanguage();
             RDFDatatype dt = o.getLiteralDatatype() ;
-            if (dt != null && dt != XSDDatatype.XSDstring) {
-                log.warn("Object to text query is not a string") ;
-                return null ;
+            if (noLang(lang)) {
+                if (dt != null && dt != XSDDatatype.XSDstring) {
+                    log.warn("Object to text query is not a string") ;
+                    return null ;
+                }
             }
+            lang = fixLang(lang);
 
             String qs = o.getLiteralLexicalForm() ;
-            return new StrMatch(null, qs, null, -1, 0) ;
+            return new StrMatch(null, qs, lang, -1, 0) ;
         }
 
         List<Node> list = argObject.getArgList() ;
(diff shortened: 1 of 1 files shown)


jena-text/src/main/java/org/apache/jena/query/text/TextQueryPF.java
```diff
@@ -295,6 +295,14 @@ private String chooseGraphURI(ExecutionContext execCxt) {
         return results;
     }
     
+    private boolean noLang(String lang) {
+        return (lang == null || lang.isEmpty() || " ".equals(lang));
```
ajs6f: I don't know much about the text indexing, but this single space-- does that have some special meaning, or is it just a simple check for whitespace? I.e., could we see more than one space char here?

xristy: The single space is the response from `o.getLiteralLanguage()` where `Node o` and `o.isLiteral()` at lines 344 and 386 of `TextQueryPF.java`. It appears to be the response when the literal has no explicit language tag. I thought the response was supposed to be an empty string, but the single space is what I was getting.

afs: Yes - no lang tag is a return of empty string.  If you get `<SPC>`, there is a bug somewhere.

afs: Other than this issue (" " for lang tag), the rest looks good.

jena-text/src/main/java/org/apache/jena/query/text/TextQueryPF.java
```diff
@@ -319,13 +311,13 @@ private StrMatch objectToStruct(PropFuncArg argObject, boolean executionTime) {
...
             RDFDatatype dt = o.getLiteralDatatype() ;
-            if (noLang(lang)) {
+            if (lang.isEmpty()) {
                 if (dt != null && dt != XSDDatatype.XSDstring) {
                     log.warn("Object to text query is not a string") ;
                     return null ;
                 }
             }
-            lang = fixLang(lang);
+            lang = lang.isEmpty() ? null : lang;
```
ajs6f: I like [`emptyToNull`](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Strings.html#emptyToNull(java.lang.String)) for this, but this isn't a problem or anything.

xristy: Agreed and changed. Thanks

jena-text/src/main/java/org/apache/jena/query/text/TextQueryPF.java
```diff
@@ -309,14 +310,18 @@ private StrMatch objectToStruct(PropFuncArg argObject, boolean executionTime) {
                 return null ;
             }
 
+            String lang = o.getLiteralLanguage();
             RDFDatatype dt = o.getLiteralDatatype() ;
-            if (dt != null && dt != XSDDatatype.XSDstring) {
-                log.warn("Object to text query is not a string") ;
-                return null ;
+            if (lang.isEmpty()) {
```
afs: Comment: Testing for datatype `RDF.dtLangString` is also possible.

jena-text/src/main/java/org/apache/jena/query/text/TextQueryPF.java
```diff
@@ -347,11 +352,15 @@ private StrMatch objectToStruct(PropFuncArg argObject, boolean executionTime) {
                 log.warn("Text query string is not a literal " + list) ;
             return null ;
         }
-        
-        if (x.getLiteralDatatype() != null && !x.getLiteralDatatype().equals(XSDDatatype.XSDstring)) {
```
afs: Comment:
In RDF 1.1, the data type of a literal is never null.

There are functions like `Util.isLangString` (in RDF 1.1: rdf:langString) and `Util.isSimpleString` (in RDF 1.1, xsd:string) that mask the RDF1.0/RDF 1.1 mode setting (not that Jena claims to support RDF 1.0 any more).

xristy: Thank you for the tips. I tried to avoid changing code that was present. I plan to raise another issue or two in jena-text and will be glad to visit some of these cases to try and use RDF functionality rather than bare Java functionality. It certainly will make the code clearer.

xristy:
Thank you for approving. I'll be traveling this coming week and won't be able to add test cases until week after at the earliest. I'd like to fold that into a future issue if that works.

PR #324, opened 2017-12-03
JENA-1443: Use DatasetGraph for GraphView TransactionHandler

afs:


ajs6f:
`GraphInMemory extends GraphView`, so this should apply automatically to TIM, right?

afs:
Yes, it does; see `TestGraphView`.

ajs6f:
Nice, thanks!

PR #325, opened 2017-12-09
fix JENA-1439  graph queries fail 'lang:xx'

xristy:
Fixes JENA-1439 graph queries fail to preserve text:query 'lang:xx' arg. TextQueryPF.extractArg(...) removed the arg unnecessarily from the input list, side-effecting the list which was saved in an OpPropFunc for repeated use on graph iterations and thus the argument was only present on the first graph in the iteration


osma:
Great detective work @xristy ! It would be nice to have a unit test that checks this, though the fix seems rather obvious after finding out the culprit, and the bug unlikely to surface again.

PR #331, opened 2017-12-17
JENA-1450:  fn:error

afs:


jena-arq/src/main/java/org/apache/jena/sparql/function/library/FN_Error.java
```diff
@@ -0,0 +1,66 @@
...
+            throw new QueryBuildException("Function '"+Lib.className(this)+"' takes zero or one arguments") ;
+    }
+    
+    @Override
+    public NodeValue exec(List<NodeValue> args) {
+        if ( args == null )
+            // The contract on the function interface is that this should not happen.
+            throw new ARQInternalErrorException("FunctionBase1: Null args list") ;
+        
+        int size = args.size();
```
ajs6f: Did you mean to use this `size` var below? It doesn't seem to get used at all...

PR #332, opened 2017-12-20
JENA-1452: GraphUnionRead; Union of one.

afs:


jena-arq/src/main/java/org/apache/jena/sparql/graph/GraphUnionRead.java
```diff
@@ -82,16 +94,29 @@ protected PrefixMapping createPrefixMapping() {
...
+            // Avoid needing distinct.
+            return dataset.getGraph(graphName).find(m);
         }
         // Only certain graphs.
         IteratorConcat<Triple> iter = new IteratorConcat<>() ;
         forEachGraph((g) -> iter.add(g.find(m))) ;
         return WrappedIterator.createNoRemove(Iter.distinct(iter)) ;
     }
     
+    private static ExtendedIterator<Triple> quadsToTriples(DatasetGraph dsg, Node graphName, Triple m) {
```
ajs6f: Is this different from `dsg.getGraph(graphName).find(m)`?

afs: It avoids churning graph objects. Graphs will eventually have to call down to core dataset-centric operations. Also this avoids an infinite recursion risk.

(This code fragment is out of sync.)

ajs6f: Ah, that makes more sense, thanks!

PR #335, opened 2017-12-28
Jena 1453 reduce docs

xristy:
resolves [Jena 1453](https://issues.apache.org/jira/browse/JENA-1453). 

Removes redundant fields from Lucene documents.

Adds graph output arg to TextQueryPF

Adds unit tests.

### DIFF
--- a/jena-text/src/test/java/org/apache/jena/query/text/TestTextGraphIndexExtra2.java
+++ b/jena-text/src/test/java/org/apache/jena/query/text/TestTextGraphIndexExtra2.java
@@ -1,0 +1,335 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.jena.query.text;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import java.io.Reader ;
+import java.io.StringReader ;
+import java.util.Arrays ;
+import java.util.HashMap ;
+import java.util.HashSet ;
+import java.util.Map ;
+import java.util.Set ;
+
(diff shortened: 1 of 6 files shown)


xristy:
Assuming the PR is accepted I'll update the jena-text doc to reflect that there is an additional output arg for `text:query`. An example is:

    select ?g ?s ?lit ?sc
    where {
       (?s ?sc ?lit ?g) text:query (skos:altLabel "one" 100 "lang:en") .
    }

where the `?g` reports the graph in which the matching triples occur. This is likely to be rather more performant than iterating over all graphs or collecting the graph URIs after the fact.

jena-text/src/test/java/org/apache/jena/query/text/TestTextGraphIndexExtra2.java
```diff
@@ -0,0 +1,335 @@
+/**
```
afs: Minor - license headers aren't javadoc so they should be `/*` not `/**`

The codebase all over the place seem to have quite a few `/**` - not just this PR.

afs:
Does this change the on-disk format? I do't think it does but confirmation would be good.

xristy:
Ah! I should have written on this.

Upgrading to this PR does not affect an existing text index.

The changes to the Lucene documents that are indexed will affect triples that are added - they will have fewer fields and if the graph field is enabled then a single stored graph field will be present rather than several instances. 

This PR removes redundant fields or unreferenced fields when indexing new triple documents. 

The triple/document deletion functionality will behave as before if it was enabled when the text index was created.

The graph return feature will function with older indexes provided that the graph field was enabled when the text index was created.

Re-indexing should generally reduce the size of the text index.

ajs6f:
This is a naive question, so feel free to answer it as briefly as it deserves, but since for some instances reindexing may be a really considerable amount of work, would it be useful/possible to include a very simple migrator class that can strip out the no-longer-used fields and produce a new leaner text index that could then be swapped in?

xristy:
@ajs6f I haven't thought about trying to do an in-place update of a text index; however, perhaps one could use jena/textindexer with an  assembler file modified to create a Lucene index off-to-the-side and then halt the main server and swap in the new index.

ajs6f:
@xristy That's what I meant-- I didn't mean to suggest an in-place op, sorry.

xristy:
I've submitted an update to the jena-text documentation to reflect the graph output argument for `text:query`.

afs:
A few words for the release announcement, especially about non needing to rebuild indexes but if you do, hey are smaller would be most helpful.

xristy:
Happy to. Here's a brief statement:

This release includes updates to the Jena Lucene integration that reduces the size of the documents indexed by Lucene and reduces the size of the resulting indexes. Re-indexing is not necessary as the changes are compatible with existing indexes. Additionally, there is an optional output argument for `text:query` that allows to retrieve the graph that contains a result triple. See the updated [jena text documentation](http://jena.apache.org/documentation/query/text-query.html).

afs:
Presumably this was ready to merge. I've noted the announcement text as well, thanks, and will sort out the documentation (unless someone beats me to it).

xristy:
Yes it was ready to merge. The documentation updates are queued in the anonymous "improve this page" commit that I made a week ago.

afs:
Documentation changes applied. Thanks!

PR #336, opened 2018-01-01
JENA-1458: Transaction promotion

afs:
This PR provides promotion  (read mode goes to write mode as needed) for TIM, TDB1 and TDB2.

Transactions have a type (new) and a current mode.

* TxnMode (enum `ReadWrite` - name retained for compatibility)
* `TxnType` - `WRITE`, `READ`, `READ_PROMOTE`, `READ_COMMITTED_PROMOTE`

Transactions can go from "read" to "write", if their initial setup allows it, when an update operation is attempted (implicit promotion) or if the app asks for the transaction to be promoted (explicit promotion). 

`Transactional` and hence the DatasetGraph/Dataset interfaces get new operations:

WRITE and READ are the same as the previous use of ReadWrite for W and R transactions.

The two different types of promote differ in whether strict transaction isolation is
maintained (`READ_PROMOTE`) and no changes from other writers
are seen during the transaction, or (`READ_COMMITTED_PROMOTE`) whether at the point of promotion, changes become visible.  
 
`READ_PROMOTE` can fail (implicit promote - transaction aborts; explicit promote, `promote()` returns false) whereas `READ_COMMITTED_PROMOTE` does not fail - it's like ending a READ transaction and starting a fresh WRITE transaction; changes from other writers can be seen across the change.
 
The new operations are:
```
public void begin()
public void begin(TxnType type)
public boolean promote()
public ReadWrite transactionMode();
public TxnMode transactionType();
```
and `public void begin(ReadWrite mode)` remains with the same semantics as before. No application needs to change.

`begin()` is `begin(TxnType.READ_PROMOTE)`, which can cause a system abort.

This is a choice - it could be `begin(TxnType.READ_COMMITTED_PROMOTE)` which does not cause system aborts but any reads before the promote point are not no longer guaranteed to be the state of the data (the transaction is "read committed", not strong isolation properties) leading to possible inconsistent data update.

Graph transactions handlers call `begin()`.

To do:

* Update `Txn`
* More checking


jena-arq/src/main/java/org/apache/jena/query/TxnType.java
```diff
@@ -0,0 +1,77 @@
...
+     * exception.
+     * 
+     * <li>{@code READ_COMMITTED_PROMOTE}: Use this with care. The promotion will succeed but 
+     * changes from oher transactions become visible.
+     * </ul>
+     * 
+     * Read committed: at the point transaction attempts promotion from "read" to
+     * "write", the sytem checks if the datset has chnage since the trsnaction started
+     * (called {@code begin}). If {@code READ_PROMOTE}, the dataset must not have
+     * changed; if {@code READ_COMMITTED_PROMOTE} anyh intermediate changes are
```
ajs6f: typo : `anyh` -> `any`

jena-arq/src/main/java/org/apache/jena/query/TxnType.java
```diff
@@ -0,0 +1,77 @@
...
+package org.apache.jena.query;
+
+import java.util.Objects;
+
+import org.apache.jena.sparql.JenaTransactionException;
+
+public enum TxnType {
+    /** Transaction mode:
+     * <ul>
+     * <li>{@code WRITE}: this gaurantees a WRITE will complete if {@code commit()} is
```
ajs6f: typo: `gaurantees` -> `guarantees`

afs:
That's it for now - the last commits have been converting and tidying up the test machinery. This in turn showed up some corner cases.

There is one design issue with TDB2 where `TransactionalComponentLifecycle` should probably split lifecycles into read and write but the main components (`BPTree` and `TransBinaryData`) do this them selves and now so does the classed used for testing `TransCounter`); or may I've forgotten why it is the way it is (because of promotion?).  Needs investigation sometime.  But it all seems to work as is.

afs:
(:clap: :clap: :clap:).  :grinning:

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
```diff
@@ -146,25 +162,84 @@ private void _begin(ReadWrite readWrite) {
...
+        } catch (JenaTransactionException ex) {
+            return false ;
+        }
+    }
+    
+    private void _promote(boolean readCommited) {
+        //System.err.printf("Promote: version=%d generation=%d\n", version.get() , generation.get()) ;
+        
+        // Outside lock.
+        if ( ! readCommited && version.get() != generation.get() )  {
```
ajs6f: Just checking that I understand the algo here...

1. Every transaction pulls a version number from the current generation number.
2. A commit increments the generation number.
3. When a transaction attempts to promote, it checks to see that the generation hasn't incremented out from under it (indicating that some other transaction committed since this one began). If that happened, this transaction will throw an exception as shown below.
4. Otherwise, it's okay to promote.

So two questions:
1. Is that understanding accurate?
2. Is that essentially how this is implemented in other dataset impls, or do I need to go read them as well?

afs: Yes- that's what happens. This part of the code isn't new - only moved. The check is done twice, once before the lock, once again after to avoid unnecessary `transactionLock` taking.

It's the same algorithm in TDB and TDB2.  TDB and TDB2 have explicit `Transaction` objects that carry the start version. The shared tests cover all cases.

ajs6f: Yeah, I knew this stuff was here but I didn't read it carefully when you put it in months ago because we weren't turning it on. Okay, I'm :+1: . Let 'er rip!

afs: I'll take that as an "approved" :-) and merge it.

This PR is big so we can keep reviewing after the merge if people want to make further comments.

PR #341, opened 2018-01-11
JENA-1462: Tests RDF/XML parsing newer URI schemes

stain:
Tests for [JENA-1462](https://issues.apache.org/jira/browse/JENA-1462)

RIOT parsing RDF/XML with a base URI different from http/https/file, such as `ssh://example.com/nested/`, fails.

Note as JENA-1462 is not fixed by this PR, this only adds the unit tests and test files.

This test also highlights a bug in parsing URIs like `file://example.com/etc/passwd` as described in [JENA-1463](https://issues.apache.org/jira/browse/JENA-1463)

### DIFF
--- a/jena-arq/testing/RIOT/URISchemes/app.nt
+++ b/jena-arq/testing/RIOT/URISchemes/app.nt
@@ -1,1 +1,0 @@
-<app://2dee5b0a-6100-470a-a67f-1399518cb470/nested/foo.txt> <http://www.w3.org/2000/01/rdf-schema#seeAlso> <app://2dee5b0a-6100-470a-a67f-1399518cb470/bar.txt> .
--- a/jena-arq/testing/RIOT/URISchemes/app.nt
+++ b/jena-arq/testing/RIOT/URISchemes/app.nt
@@ -1,0 +1,1 @@
+<app://2dee5b0a-6100-470a-a67f-1399518cb470/nested/foo.txt> <http://www.w3.org/2000/01/rdf-schema#seeAlso> <app://2dee5b0a-6100-470a-a67f-1399518cb470/bar.txt> .
--- a/jena-arq/testing/RIOT/URISchemes/file-base.rdf
+++ b/jena-arq/testing/RIOT/URISchemes/file-base.rdf
@@ -1,5 +1,5 @@
 <rdf:RDF
-    xml:base="http://2dee5b0a-6100-470a-a67f-1399518cb470/nested/"
+    xml:base="file://example.com/nested/"
     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
     xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#">
   <rdf:Description rdf:about="foo.txt">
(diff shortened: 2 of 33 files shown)


jena-arq/testing/RIOT/URISchemes/file-base.rdf
```diff
@@ -0,0 +1,8 @@
+<rdf:RDF
+    xml:base="http://2dee5b0a-6100-470a-a67f-1399518cb470/nested/"
```
afs: Wrong URI scheme, was "file://example.com/nested/" intended?

jena-arq/testing/RIOT/URISchemes/app.nt
```diff
@@ -0,0 +1 @@
+<app://2dee5b0a-6100-470a-a67f-1399518cb470/nested/foo.txt> <http://www.w3.org/2000/01/rdf-schema#seeAlso> <app://2dee5b0a-6100-470a-a67f-1399518cb470/bar.txt> .
```
afs: N-Triples is handled differently - there is no IRI base resolution in normal parsing; this is intentional.

The command line "riot" specially adds IRI checking for N-Triples. Please remove the NT tests.

afs:
To be clear: RIOT based parsing does not enforce IANA checking. It was just that the RDF/XML parser didn't pick up the right IRIFactory via RIOT.

https://github.com/apache/jena/pull/342/commits/aecfaac0d631995120b77f22059291ffe6b9f06f

jena-arq/testing/RIOT/URISchemes/file-base.rdf
stain: Commented in Java file, it was just to verify that it was not a UUID-like hostname that was causing the problem. It could be valid HTTP URL - with a local DNS search domain at least :)

jena-arq/testing/RIOT/URISchemes/app.nt
stain: I included this to verify that the IRIs were acceptable for parsing overall, precisely as N-Triples has no base resolution.

stain: Should I still remove `*.nt`? For all we know future code might use different iri factories for relative resolution and NT/NQ parsing, as all the other formats use `file:///current/directory` as base by default.

jena-arq/testing/RIOT/URISchemes/file-base.rdf
afs: I don't see a comment about file-base.rdf being a different pattern from other "base.rdf" files. file-base.rdf is the same as http-base.rdf. 

All the other "X-base.rdf" have "X:" for xml:base.  Otherwise there is no xml:base= file: testing.

stain: Bah, sorry.. silly me, I thought the comment was on http-base.rdf! Fixed in 33063f335d0cb9224a64dc2f77dbaa9963d92ecb

`fileBaseRDF()` now passes after your JENA-1463 fix 1d037e80dfd8a91e9c4042f65695cadec4b17097 from #342  :-)  -- only `fileRefRDF()` still fails (passing the base to `RDFDataMgr.read`)

```turtle
<file:///example.com/nested/foo.txt> <http://www.w3.org/2000/01/rdf-schema#seeAlso> <file:///example.com/bar.txt> .
```

jena-arq/testing/RIOT/URISchemes/app.nt
stain: Removed *.nt tests in 5200369653704be86e277774e6ce6d856369eec6

PR #343, opened 2018-01-21
Fixes for JENA-1457

Claudenw:
I had some issues with merging into my branch so I created this pull to ensure that only expected files were updated.  

As part of addition ot JENA-1457 this updates junit-contracts to version 0.2.0

Also added .recommenders directory to RAT exclude. (it was already in .gitignore)


afs:
Yes - there are problems with JSONInput.  This is visible in the {{.patch}} form.  It looks like changes were copied from Jena master and separately committed.

jena-core/src/test/java/org/apache/jena/graph/GraphWithPerformContractTest.java
```diff
@@ -68,7 +68,13 @@ public void testPerformAdd_Triple() {
 		g.performAdd(triple("S3 P3 O3"));
 		txnCommit(g);
 		GL.assertEmpty();
+		txnRun( g, new Runnable() {
```
afs: There is no need to explicitly have a `Runnable` here or in the other uses:
```
txnRun( g, ()->assertTrue(g.contains(triple("S3 P3 O3"))) );
```
makes the tests shorter and clearer.

PR #349, opened 2018-01-30
Cleanup comments and code around txn

afs:
* Comment that is out of date
* Code that has legacy use only


jena-tdb/src/main/java/org/apache/jena/tdb/store/GraphNonTxnTDB.java
```diff
@@ -32,25 +34,58 @@
...
+    {
+        private final Graph graph;
+
+        public TransactionHandlerTDB(GraphTDB graph) {
+            this.graph = graph ;
+        }
+
+        @Override
+        public void abort() {
+            // Not the Jena old-style transaction interface
```
ajs6f: Not quite sure what this comment means-- is there a specific type you are mentioning?

afs: Comment removed.

Remember if the app gets to this code it is heaving for deep trouble!

For legacy reasons, touching the GraphTransactionHandler can't switch to transaction mode because the rest of the app is probably trying to work old-style, non-transactionally.

PR #350, opened 2018-02-02
JENA-1475: Build with Java9

afs:
Top level POM changes to be able to build with Java9, output for Java8.


afs:
This PR is currently for discussion and is not ready to be merged because it switches out Elephas.

afs:
Elephas is enabled.

PR #351, opened 2018-02-03
JENA-1476: Switch javax.xml.bind.DatatypeConverter to commons-codec

afs:


afs:
The use by Jena is quite simple, string encoding for base64 and hex binary, and is covered by commons-codec.

This PR adds commons-codec as a dependency for jena-core - it was already a managed dependency, used in jena-text.

PR #353, opened 2018-02-05
Refactoring complex methods of classes FBRuleInfGraph and LPRuleStore

santiago-a-vidal:
Related to no particular issue. This is a behavior-preserving refactoring.

Summary of this pull request:

•	We are evaluating a research prototype called Bandago that assists in the refactoring of complex methods. Bandago is an Eclipse plugin that automatically identifies and refactors a type of code smell called Brain Method. A Brain Method centralizes the intelligence of a class and manifests itself as a long and complex method that is difficult to understand and maintain We have applied Bandago to 2 complex methods of your project, and we would like to receive feedback.
•	Bandago is very conservative and you should not observe many source code changes (only in the affected class).
•	The source code (after the refactoring) should behave equivalently to the original one.
•	As a sanity check, we have run tests before and after Bandago performed the refactoring(s) on the project. All tests passed.
•	The goal of the refactorings applied is to improve the legibility of the refactored method.
•	In this case, Bandago refactored the method FBRuleInfGraph.prepare extracting fragments of its code into the new methods initializeDeductions, initializeTGCCache, and callPreprocessingHook. Also, the method LPRuleStore.compileAll() was refactored extracting fragments of its code into the new methods addWildCardRules and buildTwoLevelIndices.

Thanks in advance for your help in this evaluation!

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/reasoner/rulesys/FBRuleInfGraph.java
+++ b/jena-core/src/main/java/org/apache/jena/reasoner/rulesys/FBRuleInfGraph.java
@@ -387,17 +387,54 @@ public class FBRuleInfGraph  extends BasicForwardRuleInfGraph implements Backwar
         if (fdata != null) data = fdata.getGraph();
         
         // initilize the deductions graph
-        if (fdeductions != null) {
-            Graph oldDeductions = (fdeductions).getGraph();
-            oldDeductions.clear();
-        } else {
-            fdeductions = new FGraph( createDeductionsGraph() );
-        }
-        dataFind = (data == null) ? fdeductions :  FinderUtil.cascade(fdeductions, fdata);
-        Finder dataSource = fdata;
+        Finder dataSource = initializeDeductions(data);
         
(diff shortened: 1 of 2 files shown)


ajs6f:
I simply don't know enough about our inferencing gear to comment on this… @der, can you say anything?

afs:
Thank you to @santiago-a-vidal for this work! It is really interesting to have Bandago prototype applied the code base.   Given the age of Jena, and evolution of contributors skills, I'm sure that many places would be written somewhat differently by today's standards, my own contributions especially.

The code here has not changed in 4 years (and that was a patch) so there is a balance to be discussed between changing it and leaving proven code as-is.

santiago-a-vidal:
Hi, thanks a lot for the feedback. I agree with you. The refactoring of complex methods in a class with no change since its initial implementation (and not expected to be modified in the future) may have low  priority when compared to a complex method in a class that received modifications in the last 10 most significant revisions. 

Thanks again.

Cheers,

PR #354, opened 2018-02-08
JENA-1482: Binding checks

afs:


afs:
This has got POM early work on changes for Apache-v19 mixed into it.  i will close this and submit a clean PR.

PR #355, opened 2018-02-09
JENA-1483: Promote with explicit promotion mode.

afs:
Includes a fix for JENA-1478.


pom.xml
```diff
@@ -735,6 +735,7 @@
           <artifactId>maven-compiler-plugin</artifactId>
           <version>3.7.0</version>
           <configuration>
+            <showDeprecation>false</showDeprecation>
```
afs: This crept in. We don't build with `-Xlint:deprecation` and all the output is
```
[INFO] ....java: Some input files use or override a deprecated API.
[INFO] ....java: Recompile with -Xlint:deprecation for details.
```
so it does not given much information. Being an old codebase, and valuing legacy continuity, using/overriding deprecation happens.

PR #357, opened 2018-02-10
JENA-1482: Binding checks

afs:


afs:
Sorry about that - closing and redoing the PR lost the split.

From the previous PR#354, see commit for 6abd485868376 for the null checks and 8f17867269f5e for the reformatting.

PR #360, opened 2018-02-12
Make sure to use datasets transactionally (JENA-1486)

rvesse:
textindexer did not use datasets transactionally which could fail for
dataset implementations that require transactional usage.  And it is
safer to always use transactions where provided anyway.

### DIFF
--- a/jena-text/src/main/java/jena/textindexer.java
+++ b/jena-text/src/main/java/jena/textindexer.java
@@ -24,6 +24,7 @@ import java.util.Set ;
 
 import org.apache.jena.graph.Node ;
 import org.apache.jena.query.Dataset ;
+import org.apache.jena.query.ReadWrite;
 import org.apache.jena.query.text.* ;
 import org.apache.jena.sparql.core.Quad ;
 import org.slf4j.Logger ;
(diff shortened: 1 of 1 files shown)


jena-text/src/main/java/jena/textindexer.java
```diff
@@ -104,39 +105,57 @@ protected String getSummary() {
...
-
-        for ( Node property : properties )
-        {
-            Iterator<Quad> quadIter = dataset.find( Node.ANY, Node.ANY, property, Node.ANY );
-            for (; quadIter.hasNext(); )
+        try {
+            // JENA-1486 Make sure to use transactions if supported
+            // The supportsTransactions() check should be strictly unecessary as we should always be using a
+            // DatasetGraphText which is transactional but just for future proofing we check anyway
+            if (dataset.supportsTransactions()) {
```
afs: Comment, rather than change request : 
https://github.com/apache/jena/blob/master/jena-cmds/src/main/java/arq/query.java#L220
```
 Transactional transactional = (dataset != null && dataset.supportsTransactions()) 
                                                 ? dataset
                                                 : new TransactionalNull() ;
Txn.executeRead(transactional, ()->{
    ....
});
```
is an idiom that can be useful to (1) ensure the transaction is finished, and (2) avoid repeated `if (dataset.supportsTransactions())`.

afs:
This is in the code base now.

Can we close it please? 

To do it after the event, you can do an empty commit with message "This closes #360."

PR #361, opened 2018-02-15
Small improvements

afs:
1. Yet another check for JENA-1482
2. Put some cut-and-paste internal code as compiled java so it stay up to date.

The (2) code is because Jena does not do multiple inheritance.

### DIFF
--- a/jena-tdb/src/main/java/org/apache/jena/tdb/solver/BindingTDB.java
+++ b/jena-tdb/src/main/java/org/apache/jena/tdb/solver/BindingTDB.java
@@ -26,6 +26,7 @@ import org.apache.jena.riot.out.NodeFmtLib ;
 import org.apache.jena.sparql.core.Var ;
 import org.apache.jena.sparql.engine.binding.Binding ;
 import org.apache.jena.sparql.engine.binding.BindingBase ;
+import org.apache.jena.tdb.TDBException ;
 import org.apache.jena.tdb.store.NodeId ;
 import org.apache.jena.tdb.store.nodetable.NodeTable ;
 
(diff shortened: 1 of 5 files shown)


jena-tdb/src/main/java/org/apache/jena/tdb/solver/BindingTDB.java
```diff
@@ -120,6 +121,8 @@ public Node get1(Var var)
             if ( id == null )
                 return null ; 
             n = nodeTable.getNodeForNodeId(id) ;
+            if ( n == null )
+                throw new TDBException("No node in NodeTable for NodeId "+id);
```
ajs6f: Do we know all/most of the circumstances that might cause this? Is it basically an interrupted write? If so, it might be worth including some error logging with that hint, but if there are more than one or two potential causes, never mind. I mention it because getting from "a missing entry in the `NodeTable`" to "that was the aftereffect of a failed write" is going to bring some users to the list when they might be able to figure out what happened with a simple hint.

afs: I don't know all the reasons for getting it.  The use  reports don't tend to tell us when things are fixed so I don't for sure whether my suggestions were of any help.

One possible case is early exit, non-transactional use of TDB so the triple or quad table has written back (by natural OS managed file caching) but the NodeTable has not.  Exit without sync (not necessarily a crash) and they are out of step.

Maybe with this we'll get better information and be able to refine it - at the moment, coudl well be a false trail.

ajs6f: Right-o, sounds like it's much better to leave it be.

PR #366, opened 2018-02-24
JENA-1494: Reader properties

afs:
Also includes:
* Additional operations for RDFParser and RDFWriter
* RDF/XML Writer properties example.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/SysRIOT.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/SysRIOT.java
@@ -20,6 +20,7 @@ package org.apache.jena.riot;
 
 import org.apache.jena.atlas.lib.IRILib ;
 import org.apache.jena.riot.system.IRIResolver ;
+import org.apache.jena.sparql.util.Context;
 import org.apache.jena.sparql.util.Symbol ;
 import org.apache.jena.util.FileUtils ;
 import org.slf4j.Logger ;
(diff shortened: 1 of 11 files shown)


jena-arq/src/main/java/org/apache/jena/riot/SysRIOT.java
```diff
@@ -50,14 +51,20 @@
...
      */
-
-    /** Context key for old style RDFWriter properties */ 
     public static final Symbol sysRdfWriterProperties      = Symbol.create(riotBase+"rdfWriter_properties") ;
-    
+
+    /**
+     * Context key for old style RDFReader properties. The value of this in a
+     * {@link Context} must be a {@code Map<String, Object>}. The entries of the
+     * map are used to set reader properties before the Jena legalacy
```
rvesse: Typo - `legalacy`

jena-arq/src/main/java/org/apache/jena/riot/lang/ReaderRIOTRDFXML.java
```diff
@@ -116,17 +154,29 @@ public void parse() {
...
             arp.getOptions().setIRIFactory(IRIResolver.iriFactory());
 
+        if ( context != null ) {
+            try { 
+                @SuppressWarnings("unchecked")
+                Map<String, Object> p = (Map<String, Object>)(context.get(SysRIOT.sysRdfReaderProperties)) ;
+                if ( p != null )
+                    p.forEach((k,v) -> oneProperty(arpOptions, k, v)) ;
+            } catch (Throwable ex) {
+                Log.warn(AdapterRDFWriter.class, "Problem setting properties", ex);
```
rvesse: Should this message more clearly state that some/all properties may have been ignored?

afs: Will revise. The important item being covered is the class cast (maybe it should throw a meaningful exception?)  If `oneProperty` throws an exception, it can propagate as normal.

PR #368, opened 2018-03-03
JENA-1495 Return Model from PrefixMapping methods.

jaco0646:


jena-core/src/main/java/org/apache/jena/rdf/model/Model.java
```diff
@@ -1051,4 +1051,12 @@ Remove all the statements matching (s, p, o) from this model.
         Answer true iff .close() has been called on this Model.
     */
     public boolean isClosed();
+
+    // Override return type for methods inherited from PrefixMapping
```
ajs6f: We should use `@Override` for explicitness on these changes.

ajs6f:
This seems like a useful change, but it is not back-compatible, so in theory we should hold it for a major version release.

ajs6f:
This looks basically good to me: with an additional commit addressing https://github.com/apache/jena/pull/368#pullrequestreview-101006841, I'll be happy to merge this.

jena-core/src/main/java/org/apache/jena/rdf/model/Model.java
jaco0646: Done.

afs:
Closing this PR. I see these commits in the codebase under different SHA1 ids (b8b482ecaa and 036ebce8ed).

@jaco0646 - thanks!

PR #369, opened 2018-03-03
JENA-1492: Pass down transactions from general datasets

afs:


afs:
"Skolemization function" is not related; it is a small clarification done in passing.

ajs6f:
There is some new `Assembler` stuff and reformatting, so I'm not totally clear what the thrust of this PR is: is it to ensure that `Model`s wrapping other `Model`s (e.g. for inferencing) correctly pass transaction control calls into their wrapped models (which should pass them into their underlying `Dataset`)?

afs:
Yes, this is models passing down transactions.  It only became possible with dataset-transaction promotion (new to 3.7.0).  

This all has a lot of knock on consequences. The wrap/create distinction is quite important and that touches a lot of files; it is not reformatting. "wrap" a model and the dataset can not have named graphs added, "create" with a model and it starts with that model and can have named graphs added. We only used to have "create" which has overhead and complexity as well as occasionally leading to unexpected behaviour with SPARQL Updates (updates work but the new NG are in-memory and disappear on restart - better to have a closed dataset and catch the issue at update time).

While testing, I found lack of test coverage and integration testing, which is the root of the users@ email that triggered this, was the place to put this. There as a lack of assemblers for certain other special cases so I included them as well as they arose naturally from checking test coverage.

jena-arq/src/main/java/org/apache/jena/sparql/core/TxnDataset2Graph.java
```diff
@@ -0,0 +1,240 @@
...
+ * Best practice is to change the graph membership outside of any transaction,
+ * ideally at setup time of the object using this class. (Caution: SPARQL Update
+ * can create graphs.   
+ * @See {@link DatasetGraphMapLink}
+ * @See {@link DatasetGraphOne}
+ */
+public class TxnDataset2Graph extends TransactionalLock {
+    /**
+     * Control whether to pass down transactions from the dataset to the graph in the
+     * dataset. This should be set to "true"; setting it "false" causes the onld,
```
ajs6f: `onld` => `old`

jena-arq/src/main/java/org/apache/jena/sparql/core/TxnDataset2Graph.java
```diff
@@ -0,0 +1,240 @@
...
+        TransactionHandler th = graph.getTransactionHandler();
+        if ( ! th.transactionsSupported() )
+            return;
+        Object key = calcKey(graph);
+        if ( th.transactionsSupported() )
+            handlers.put(key, th) ;
+    }
+
+    // Determine the key - an object that is the unit of transactions.
+    // For two graphs form the same DatasetGraph, i.e. GraphView, there should be one transaction.
```
ajs6f: `form`=>`from`

jena-arq/src/main/java/org/apache/jena/sparql/core/TxnDataset2Graph.java
```diff
@@ -0,0 +1,240 @@
...
+    private static Object calcKey(Graph graph) {
+        if ( graph instanceof GraphView )
+            // Use the database as the key so that transactions are started once-per-storage.
+            // This the case of a graph from some storage being plavced in a general dataset.  
+            return ((GraphView)graph).getDataset();
+        if ( graph instanceof InfGraph )
+            // InfGraph does actual pass done in its TransactionHandler.
```
ajs6f: Not sure what this comment means?

jena-arq/src/main/java/org/apache/jena/sparql/core/TxnDataset2Graph.java
```diff
@@ -0,0 +1,240 @@
...
+        toBeRemoved.forEach(g->removeHandler(handlers, g));
+    }
+    
+    // TransactionalLock.begin(ReadWrite) calls begin(TxnType)
+    @Override
+    public void begin(TxnType type) {
+        super.begin(type);
+        // Whatever the type. Graph Transactions do not allow for "read-only".
+        start();
+        handlers(h->h.begin());
```
ajs6f: No big deal, but this and things like it could be `handlers(TransactionHandler::begin)`, which I think reads a bit easier.

jena-arq/src/main/java/org/apache/jena/sparql/core/TxnDataset2Graph.java
```diff
@@ -0,0 +1,240 @@
...
+ * when a new memory graph is added mid-transaction.
+ * <p>
+ * This is not "nested transactions" - theer is no overall "commit" or "abort". If
+ * failure/restart occurs, some graphs may have commited and others not. It is the best
+ * that can be done given for an arbitrary collection of graphs, backed by different
+ * storage and having different capabilities.
+ * <p>
+ * Best practice is to change the graph membership outside of any transaction,
+ * ideally at setup time of the object using this class. (Caution: SPARQL Update
+ * can create graphs.
```
ajs6f: typo missing `)`

ajs6f:
Okay. `::wrap` vs. `::create` is not what I am referring to as reformatting. It's stuff like [this](https://github.com/apache/jena/pull/369/files#diff-27cb8b74e951d2872a91a48066930e2d) or [this](https://github.com/apache/jena/pull/369/files#diff-9d27f4f9095b5d190b6fe8977b6c393b).

From what I understand, this all makes sense. Datasets that are "collected" from multiple graphs can, at best, pass the transaction through to their graphs, those that are in control of their own data can retain the responsibility for transactions themselves. That does seem to be the least surprising behavior and the one that veers towards "dataset as the fundamental unit" which we have been working towards.

jena-arq/src/main/java/org/apache/jena/sparql/core/TxnDataset2Graph.java
afs: I don't think it always does :-) `h.` means the IDE method finding works.

ajs6f: It's fine not to, but what kind of IDE are you using that can't trace method references? Eclipse certainly can.

afs: Finding for completion on the lambda argument.  Complete on `handlers(` is `Consumer<TransactionHandler> action`.

ajs6f: Yeah, Eclipse (Cmd-space) gives me that, no problem.

PR #372, opened 2018-03-06
Reformatting + two RiotChars functions.

afs:
Minor : reformatting around blank node allocation and two library functions, triggered by checking blank nodes formats.

* `RiotChars.isHexCharLC`
* `RiotChars.isHexCharUC`

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/lang/BlankNodeAllocatorGlobal.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/lang/BlankNodeAllocatorGlobal.java
@@ -41,12 +41,7 @@ public class BlankNodeAllocatorGlobal implements BlankNodeAllocator
 
     @Override
     public Node alloc(String label) {
-        Node b = map.get(label);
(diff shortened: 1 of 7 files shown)


jena-arq/src/main/java/org/apache/jena/riot/lang/BlankNodeAllocatorGlobal.java
```diff
@@ -18,41 +18,39 @@
...
     @Override
-    public Node alloc(String label)
-    {
-        Node b = map.get(label) ;
-        if ( b == null )
-        {
-            b = create() ;
-            map.put(label, b) ;
+    public Node alloc(String label) {
+        Node b = map.get(label);
```
ajs6f: Maybe just `return map.computeIfAbsent(x->create())`?

afs: Good idea. (Some of this code is moderately old.)

PR #373, opened 2018-03-06
JENA-1454: Fix to handle text output format.

afs:
Bug fix to JENA-1454.

### DIFF
--- a/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ResponseResultSet.java
+++ b/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ResponseResultSet.java
@@ -40,7 +40,6 @@ import org.apache.jena.query.ResultSet ;
 import org.apache.jena.query.ResultSetFormatter ;
 import org.apache.jena.riot.Lang;
 import org.apache.jena.riot.WebContent ;
-import org.apache.jena.riot.resultset.ResultSetLang ;
 import org.apache.jena.riot.resultset.ResultSetWriterRegistry;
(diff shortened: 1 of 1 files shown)


jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ResponseResultSet.java
```diff
@@ -131,26 +130,30 @@ public static void doResponseResultSet(HttpAction action, ResultSet resultSet, P
...
         // (2) ?stylesheet=
         //
         // Thrift is special because
         // (1) charset is meaningless
         // (2) there is no boolean result form.
-        //
-        // Text plain is special because of the formatting by prologue.
-        
+
+        if ( Objects.equals(serializationType, contentTypeTextPlain) ) {
```
ajs6f: Just out of curiosity-- why `Objects.equals` here instead of `contentTypeTextPlain.equals` or `serializationType.equals`?

ajs6f: Oh, I see-- this was cut from down below. NM.

afs: Style, no more :  `Objects.equals` copes with null in the same way the back-to-front `constant.equals(variable)` is also a null-defending idiom. Not that there will be nulls.  It could be a switch on strings. It should be `lang` driven but there is no text result set writer ATM and text is odd because it uses prefixes and no other result set format does.

This is a quick fix to restore what become broken in the JENA-1454 redesign.

PR #375, opened 2018-03-08
JENA-1501: add systemd unit file for Fuseki

osma:
This adds a systemd unit file to the Fuseki2 distribution.

This is the same file that was circulated on the users list (in response to a question by @jneubert), but with reduced default memory setting (2GB), an ASF license header and some hopefully helpful comments about how to set it up.

See https://issues.apache.org/jira/browse/JENA-1501

### DIFF
--- a/jena-fuseki2/apache-jena-fuseki/assembly-dist.xml
+++ b/jena-fuseki2/apache-jena-fuseki/assembly-dist.xml
@@ -86,6 +86,7 @@
       <includes>
         <include>log4j.properties</include>
         <include>fuseki</include>
+        <include>fuseki.service</include>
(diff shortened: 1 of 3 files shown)


jneubert:
Thank you very much, Osma!
One small suggestion about the wording of the comment: Instead of ``Edit the line below to adjust memory size`` it could be perhaps more clear to write ``Edit the line below to adjust the amount of memory for Fuseki`` for people who are not familiar with Java memory management and could assume the total memory size of their machine.

osma:
Thank you for your comments @ajs6f and @jneubert ! I made the changes you suggested.

I will merge this early next week unless there are further suggestions or objections.

PR #377, opened 2018-03-09
JENA-1503: Bug in XSDDateTime toString

ajs6f:


jena-core/src/test/java/org/apache/jena/graph/test/TestTypedLiterals.java
```diff
@@ -989,7 +991,17 @@ public void testXSDanyURI() {
         Node node2 = NodeFactory.createLiteral("http://example/", XSDDatatype.XSDstring) ;
         assertFalse(node1.sameValueAs(node2)) ;
     }
-    
+
+    /**
+     * Test a user error report concerning date/time literals from JENA-1503
+     */
+    public void testDateTimeBug3() {
+        final RDFDatatype XSDdateTime = TypeMapper.getInstance().getSafeTypeByName(XSD.dateTime.getURI());
```
afs: More FYI than anything else: There is a constant `XSDDatatype.XSDdateTime`.

ajs6f: Hey, let's model good behavior! (Says the father of two toddlers.) I'll add a commit right now.

PR #379, opened 2018-03-09
Refactor tests; don't test for GregorianYear< 0

afs:
Follow on from JENA-1503.

This splits up some `TestTypeLiteral` tests and removes the old testing for GregorianYear< 0 which didn't test correctly, and couldn't.


afs:
Added zero.  I split one mega test (`testDateTime`) into parts. It was a sequence of independent parts. The same testing is done so that it is easier to work on any failures. `testDateTime_9` contains a chag e- don't touch years < 0 because of the GregorianCalendar issue.

ajs6f:
Ok, `testDateTime_9` LGTM and if the others are just cut-and-paste, it should all be good.

PR #384, opened 2018-03-15
JENA-1158: Make the union graph a graphView(Quad.unionGraph).

afs:
Don't rely on the default implementation in DatasetGraphBase.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/core/DatasetGraphBase.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/core/DatasetGraphBase.java
@@ -69,9 +69,11 @@ abstract public class DatasetGraphBase implements DatasetGraph
     @Override
     public Graph getUnionGraph() {
         // Implementations are encouraged to implement an efficent
-        // {@code DatasetGraphBase.findQuadsInUnionGraph} or
+        // named graph for Quad.unionGraph, and this operation that
+        // does not require the full "distinct()" used by the general purpose
+        // GraphUnionRead. See also
+        // {@code DatasetGraphBase.findQuadsInUnionGraph} and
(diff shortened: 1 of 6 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/core/DatasetGraphBase.java
```diff
@@ -69,9 +69,11 @@ public boolean containsGraph(Node graphNode) {
     @Override
     public Graph getUnionGraph() {
         // Implementations are encouraged to implement an efficent
-        // {@code DatasetGraphBase.findQuadsInUnionGraph} or
+        // named graph for Quad.unionGraph, and this operation that
```
ajs6f: Does this belong in a Javadoc, like maybe `@apiNote`?

jena-arq/src/main/java/org/apache/jena/sparql/core/GraphView.java
```diff
@@ -146,7 +148,7 @@ private static Node graphNode(Node gn) {
     public void performAdd( Triple t ) { 
         Node g = graphNode(gn) ;
         if ( Quad.isUnionGraph(g) )
-            throw new GraphViewException("Can't update the default union graph of a dataset") ; 
+            throw new AddDeniedException("Can't update the union graph of a dataset") ;
```
ajs6f: Holy moly, I had no idea we had these. II'm going to remember these, they are way more specific than `UnsupportedOperationException`.

afs: It's pushing the "contract". "Denied" suggests to me that it could work in other circumstances but the tests said this was expected so this is what you get.

PR #385, opened 2018-03-15
defined filters and tokenizers for ConfigurableAnalyzer

xristy:
This PR implements [JENA-1506](https://issues.apache.org/jira/browse/JENA-1506). See the issue for details.


rvesse:
Looks like a really nice extension to the current machinery!

I notice there is a unit test for the assembler support but no unit tests for using some of these new features, would it be possible to include some of those to verify that this actually allows addressing the use cases that led to this PR?

One other thought is whether the explicit param type definitions are necessary for the simpler cases (integer, boolean) or whether we could just rely on literal datatypes to reduce the verbosity of the configuration for those simpler cases?  For the complex cases like sets, token streams, analysers etc then explicit types makes total sense.

rvesse:
Btw as you are now a committer you can follow the instructions at https://reference.apache.org/committer/github to get added to the Apache GitHub org so you'll show up as `Member` and not just as `Contributor` on here

osma:
Wow, quite a big patch for a "straightforward" change :) Great work @xristy !

+1 to @rvesse's suggestion to simplify by making use of literal data types, if it's possible.

Thanks for the tip @rvesse , I also added my GH username to the ASF profile.

xristy:
@rvesse  On the matter of the tests, I just updated the ticket on that matter:

> There are already a suite of tests associated with TestDatasetWithConfigurableAnalyzer that exercise the new logic for identifying tokenizers and filters and instantiating them so this part of the extension is tested there.

As for using literal datatypes for String, int, and boolean, I think that's doable. I tend to get slavish in following a pattern so I'm inclined to permit either the explicit parameter types as currently implemented and add the use of literal datatypes as an alternative when applicable.

I added github id per your suggestion, thanks.

rvesse:
@xristy Yes I don't think there is any harm in supporting both the explicit form as well as supporting literal datatypes directly

kinow:
+1 @rvesse comments. And +1 for the great PR @xristy !!! Tested with the WIP `SelectiveFoldingFilter` for JENA-1488, and it worked like a charm!!!!! :tada: :tada: :tada: 

First created a normal JenaText configuration, and everything worked with no issues. Then replaced my `analyzer` by a `DefinedAnalyzer`... had a few hiccups forgetting to create a tokenizer, and had to update the filter to use a `CharArraySet` (much better that way I think).

But in the end got the query working as expected. Confirmed with Luke that the contents were indexed correctly with the new filter... did a quick search in Luke, then used Fuseki, and everything worked as expected.

Once this one is merged, we'll be ready for another PR for JENA-1488. In the meantime, I will start working on documentation and tests !

Thanks!
Bruno

xristy:
The parameter types: `int`, `boolean` and `String` may be derived from the `text:paramValue`. See [JENA-1506](https://issues.apache.org/jira/browse/JENA-1506) for details.

xristy:
I'm guessing that I'm supposed to perform the pull request but when I try to push to apache/jena master on github my github user, xristy, doesn't have permission for apache/jena.git.

What is the proper procedure?

ajs6f:
Hey, @xristy, you want to push instead to the `master` branch at Apache:

https://git-wip-us.apache.org/repos/asf/jena.git

ajs6f:
To be clear, what is here at Github is essentially a read-only mirror. What's at Apache is the real deal. We've talked about reversing that mirroring relationship (so that we could merge PRs here at Github) but there are other consequences to so doing, mostly involving integrations with other services like Jira.

afs:
Workflow doucmentation: https://cwiki.apache.org/confluence/display/JENA/Commit+Workflow+for+Github-ASF

PR #387, opened 2018-03-19
fix for JENA-1509

Claudenw:
fixed string issue and added tests


jena-rdfconnection/src/main/java/org/apache/jena/rdfconnection/RDFConnectionRemote.java
```diff
@@ -280,7 +280,7 @@ private void updateExec(UpdateRequest update, String updateString ) {
         }
         // Use the query string as provided if possible, otherwise serialize the query.
```
afs: s/query/update/g

PR #390, opened 2018-04-04
Bugfix Tutorial 6: Missing "/" at the end of an URI resulted in a Nul…

supergitacc:
I found a bug in tutorial 6 and propose a fix for it.

Without "/" at the end of the URI of JohnSmith (http://somewhere/JohnSmith/) an exception occurs in line 51 because no resource vcard could be found. The exception was: "Exception in thread "main" java.lang.NullPointerException". Adding the slash at the end fixes the issue because   <rdf:Description rdf:about="http://somewhere/JohnSmith/">

### DIFF
--- a/jena-core/src-examples/jena/examples/rdf/Tutorial06.java
+++ b/jena-core/src-examples/jena/examples/rdf/Tutorial06.java
@@ -29,7 +29,7 @@ import java.io.*;
 public class Tutorial06 extends Object {
     
     static final String inputFileName = "vc-db-1.rdf";
-    static final String johnSmithURI = "http://somewhere/JohnSmith";
+    static final String johnSmithURI = "http://somewhere/JohnSmith/";
     
     public static void main (String args[]) {
         // create an empty model


afs:
Hi there - thanks for the PR. There is a problem somewhere.

Looking at the RDF tutorial, it uses "http\://somewhere/JohnSmith" (no /) so I think the problem in the data  file vc-db-1.rdf (and other data files) where they use `http://somewhere/JohnSmith/` (with /).

kinow:
Note: these files are not in GitHub, but in SVN. For example: http://svn.apache.org/repos/asf/jena/site/trunk/content/tutorials/sparql_data/vc-db-1.rdf

@supergitacc would you be interested in reviewing the code & data from the RDF tutorial? We can create a ticket in JIRA (happy to do that) and you can attach the SVN patch.

ajs6f:
Not to hijack the thread, but do we also want to open a ticket to get all of the relevant (tutorial) files into Git alongside the main codebase? Or should we just subsume that into the larger project of getting the site into Git?

kinow:
@ajs6f good point. I remember some projects in ASF keep the example code in the main code repository (e.g. [Commons Imaging](https://github.com/apache/commons-imaging/tree/master/src/test/java/org/apache/commons/imaging/examples)). Or we can move the site to git?

Some time ago we discussed in the mailing list gitbox, gitpusub... I believe the old git-wip are being slowly transitioned to gitbox (see [this](http://mail-archives.apache.org/mod_mbox/commons-dev/201804.mbox/%3cCAOGo0VY+nds2aftyStaz96chsbLKHMzqzzeAw8u_gdeqbW+Fow@mail.gmail.com%3e) and [this](http://mail-archives.apache.org/mod_mbox/commons-dev/201804.mbox/%3cCAE9L6G2zEst9_roT_6riJDQPCqEzJyiwDASffUP9NVxy0twMiA@mail.gmail.com%3e) for recent discussion in Commons dev).

So perhaps we could resurrect that old e-mail thread and discuss perhaps moving the site to gitpubsub, and a gitbox repository? This way the code would be hosted in git, making contributions a bit easier.

ajs6f:
@kinow Yep, those Commons threads are exactly what brought it back to my mind. Okay, I suggest that we leave this particular issue as @kinow suggests (Jira ticket and go from there) and I will (sometime today or tomorrow) restart those threads on dev@ about GitBox, etc.

afs:
Hi @supergitacc -- the data has been fixed in all the vc-db files.

Thanks for the report.

PR #391, opened 2018-04-04
JENA-1516:  ObjectFileStorage fix

afs:
The main commit is the second one.

Commit 1 is general removal of the alloc-complete writer cycle.

Commit 2 is the core fix - simpler wring of objects.

Commit 3 is reformatting and removal of some dev code.


jena-tdb/src/main/java/org/apache/jena/tdb/lib/NodeLib.java
```diff
@@ -53,26 +52,32 @@
     // Characters in IRIs that are illegal and cause SSE problems, but we wish to keep.
     final private static char MarkerChar = '_' ;
     final private static char[] invalidIRIChars = { MarkerChar , ' ' } ; 
+    final private static int SIZE = 1024;
+    // Marshalling space.
+    final private static  ByteBuffer workspace = ByteBuffer.allocate(SIZE);
```
rvesse: This buffer is a `static` and there is no thread safety or synchronisation around its usage.  Is this safe?

afs: No problem about the review timing - any PR is as far as I'm concerned on permanent review. The more eyes on concurrency the better.

It is safe.  It gets called from `NodeTableNative.writeNodeToTable` which itself (see its comment) is inside synchronization. Several things need to happen together - write the node and update the node cache - so the sync is further out in `NodeTableNative`.

The comment on `encodeStore` where the byte buffer is used applies "Callers must synchronize to ensure writing is not concurrent."

For decode, the ByteBuffer comes from `ObjectFileStorage.read`.

I can add a comment to the `workspace` declaration. (I'd _like_ scoped statics inside methods.)

PR #393, opened 2018-04-09
JENA-1517: Implement missing override in OpVarsPattern visitor

jeremy-coulon:


rvesse:
+1

jeremy-coulon:
I will try to add a unit test in order to make sure that MINUS actually works with OpQuadBlock.

jeremy-coulon:
It seems that I fixed the MINUS queries but FILTER EXISTS/NOT EXISTS are still not working.

jeremy-coulon:
EXISTS/NOT EXISTS UTs are working. I probably have another bug in my application.

PR #394, opened 2018-04-10
JENA-1519: OpWalkerVisitor with custom operators

jeremy-coulon:


jena-arq/src/main/java/org/apache/jena/sparql/algebra/OpWalker.java
```diff
@@ -139,6 +139,7 @@ protected void visitN(OpN op) {
         @Override
         protected void visitExt(OpExt op) {
             before(op) ;
+            super.visitExt(op);
```
afs: Not sure about this. It is visiting both the effective op and the real OpExt. Shoudn't it be the visitor deciding that? ie. visiting `op.effectiveOp()` is required?

jeremy-coulon: Visiting the real OpExt is not useful for me but I can't say for other people.

afs: There are probably both cases. Its an extension point so it is hard to be categorical. Only the effective is walkable unless the OpExt is comprised of known Ops in which case it does not need to be an OpExt.

Let's leave it as per the PR - we might have to return to it but without a concrete alternative usage, it feels like slipping towards guessing.

PR #396, opened 2018-04-12
JENA-1520: tdb2.tdbstats: cmd and fix for rdf:type

afs:


jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/stats/StatsCollectorNodeId.java
```diff
@@ -18,39 +18,39 @@
...
     }
-    
-    private static NodeId findRDFType(NodeTable nodeTable2)
-    {
-        return nodeTable2.getAllocateNodeId(NodeConst.nodeRDFType) ;
+
+    private static NodeId findRDFType(NodeTable nodeTable) {
+        // It may not exist.
+        NodeId nodeId = nodeTable.getNodeIdForNode(NodeConst.nodeRDFType);
+        if ( NodeId.isDoesNotExist(nodeId) )
```
ajs6f: Jut for my own education-- am I right in understanding this to be the fix for the "no type triples in the data" problem?

afs: Yes, that's right.  The stats engine gathers details for `rdf:type` specially so it needs to know the trigger for that. `StatsCollectorBase` is a generic and works on an index scan, no touching the node table for every tuple of a triple/quad. Here, it's a `NodeId`. if it isn't in the node table, it is not in the data.

As it was, it tried to update the node table to add the `rdf:type`, but it is in a read transaction so `NodeTable.getAllocateNodeId` fails.

PR #398, opened 2018-04-13
JENA-1523: Allow internal variables names in variable parsing.

afs:


jena-arq/src/main/java/org/apache/jena/riot/tokens/TokenizerText.java
```diff
@@ -739,16 +739,21 @@ private String readWordSub(boolean leadingDigitAllowed, boolean leadingSignAllow
         return readCharsAnd(leadingDigitAllowed, leadingSignAllowed, extraCharsWord, false);
     }
 
-    static private char[] extraCharsVar = new char[]{'_', '.', '-', '?', '@', '+'};
+    // This array adds the other characters that can occurs in an internal variable name.
+    // Variable a creeated with SPARQL-illegal syntax to encsure they do not clash with
```
ajs6f: typos, can be line-changed to:
```
Variables are created with SPARQL-illegal syntax to ensure that they do not clash with
```

PR #400, opened 2018-04-15
[JENA-1525] Add <Automatic-Module-Name> to <manifestEntries> for each jar module

christopher-johnson:
See comments here: 
https://github.com/apache/jena/pull/399

It is possible that there are more artifacts that might want this.

FYI: This does not include the following top level projects:
apache-jena, apache-jena-libs, apache-jena-osgi, jena-elephas, jena-examples, jena-extras, jena-fuseki1, jena-fuseki, jena-integration-tests, jena-iri, jena-jdbc, jena-maven-tools, jena-permissions, jena-sdb, jena-shaded-guava

### DIFF
--- a/jena-db/pom.xml
+++ b/jena-db/pom.xml
@@ -32,7 +32,11 @@
     <version>3.8.0-SNAPSHOT</version>
     <relativePath>..</relativePath>
   </parent>
-  
+
+  <properties>
+    <automatic.module.name>org.apache.jena.db</automatic.module.name>
+  </properties>
+
   <licenses>
     <license>
       <name>Apache License 2.0</name>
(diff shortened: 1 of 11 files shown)


jena-db/pom.xml
```diff
@@ -75,6 +79,13 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestEntries>
+              <Automatic-Module-Name>${automatic.module.name}</Automatic-Module-Name>
```
afs: My knowledge of JPMS is a bit weak. This module (jena-db) is a POM. This means that Automatic-Module-Name with this setting will apply to all the jars underneath this module (the jena-db-*  modules) with the same name. 

If a number of jars have the same Automatic-Module-Name, are they in the same module? (which would be convenient).

Otherwise, I can fix up the submodules when merging it. No need to round again.

christopher-johnson: yes.  the Automatic-Module-Name is like a qualified jar name.  All of the packages in the jena-db project will be visible with Automatic-Module-Name: org.apache.jena.db.  You would not want to split the components of `org.apache.jena.dboe` into separate modules, so it makes sense to unify them.  It generally seems to be coherent to align the module name with a top level package name when there is one package tree per project.  It might be advisable to make a separate jar configuration for `org.apache.jena.tbd2` that has a matching module name (if granularity is a requirement).

PR #401, opened 2018-04-15
 [JENA-1524] org.apache.jena.system is split by org.apache.jena.arq and org.apache.jena.core

christopher-johnson:
per @afs 

> this renames of "jena-core:org.apache.jena.system" as "jena-core:org.apache.jena.sys", and puts an adapter class `JenaInit` in "jena-arq:org.apache.jena.system" that has two functions to call the real init and shutdown. 
> 
> `JenaInit` will be deprecated prior to removal quite quickly (in a very few release cycles).
> 
> This way the impact is lessened: `org.apache.jena.system.Txn` is not repackaged. The adapter class approach does not work for calls from jena-core to jena-arq without reflection.

### DIFF
--- a/apache-jena-osgi/jena-osgi/src/main/resources/META-INF/services/org.apache.jena.system.JenaSubsystemLifecycle
+++ b/apache-jena-osgi/jena-osgi/src/main/resources/META-INF/services/org.apache.jena.system.JenaSubsystemLifecycle
@@ -1,4 +1,4 @@
-org.apache.jena.system.InitJenaCore
+org.apache.jena.sys.InitJenaCore
 org.apache.jena.riot.system.InitRIOT
 org.apache.jena.sparql.system.InitARQ
 org.apache.jena.tdb.sys.InitTDB
--- a/jena-arq/src/main/java/org/apache/jena/system/JenaInit.java
+++ b/jena-arq/src/main/java/org/apache/jena/system/JenaInit.java
@@ -16,27 +16,25 @@
  * limitations under the License.
  */
 
-package org.apache.jena.riot ;
+package org.apache.jena.system;
 
-import org.apache.jena.riot.system.IO_JenaReaders ;
-import org.apache.jena.riot.system.IO_JenaWriters ;
-import org.apache.jena.sys.JenaSystem ;
+import org.apache.jena.sys.JenaSystem;
 
-public class IO_Jena
-{
-    static { JenaSystem.init(); }
-    
-//    private static String      riotBase               = "http://jena.apache.org/riot/" ;
-//    private static String      streamManagerSymbolStr = riotBase + "streammanager" ;
-//    public static Symbol       streamManagerSymbol    = Symbol.create(streamManagerSymbolStr) ;
+/**
+ * This is a temporary adapter for implementations to transition to [JENA-1524]
+ */
+public class JenaInit {
 
-    public static void wireIntoJena() {
-       IO_JenaReaders.wireIntoJena() ;
-       IO_JenaWriters.wireIntoJena() ;
+    /**
+     * Initialize Jena.
+     */
+    public static void init() {
+        JenaSystem.init();
     }
 
-    public static void resetJena() {
-        IO_JenaReaders.resetJena() ;
-        IO_JenaWriters.resetJena() ;
+    /** Shutdown subsystems */
+    public static void shutdown() {
+        JenaSystem.shutdown();
     }
+
 }
--- a/apache-jena-osgi/jena-osgi/pom.xml
+++ b/apache-jena-osgi/jena-osgi/pom.xml
@@ -145,7 +145,7 @@
     </dependency>
     <!--
     NOTE: Also update 
-    src/main/resources/META-INF/services/org.apache.jena.system.JenaSubsystemLifecycle
+    src/main/resources/META-INF/services/org.apache.jena.sys.JenaSubsystemLifecycle
     to the concatenation of the different Jena modules' JenaSubsystemLifecycle
     (order does not matter). See JENA-1139 for details.
     -->
--- a/apache-jena-osgi/jena-osgi/src/main/java/org/apache/jena/osgi/Activator.java
+++ b/apache-jena-osgi/jena-osgi/src/main/java/org/apache/jena/osgi/Activator.java
@@ -18,7 +18,7 @@
 
 package org.apache.jena.osgi;
 
-import org.apache.jena.system.JenaSystem;
+import org.apache.jena.sys.JenaSystem;
 import org.osgi.framework.BundleActivator;
 import org.osgi.framework.BundleContext;
 
--- a/jena-arq/src/main/java/org/apache/jena/query/ARQ.java
+++ b/jena-arq/src/main/java/org/apache/jena/query/ARQ.java
@@ -37,7 +37,7 @@ import org.apache.jena.sparql.pfunction.PropertyFunctionRegistry ;
 import org.apache.jena.sparql.util.Context ;
 import org.apache.jena.sparql.util.MappingRegistry ;
 import org.apache.jena.sparql.util.Symbol ;
-import org.apache.jena.system.JenaSystem ;
+import org.apache.jena.sys.JenaSystem ;
 import org.slf4j.Logger ;
 import org.slf4j.LoggerFactory ;
 
--- a/jena-arq/src/main/java/org/apache/jena/query/Query.java
+++ b/jena-arq/src/main/java/org/apache/jena/query/Query.java
@@ -40,7 +40,7 @@ import org.apache.jena.sparql.syntax.Element ;
 import org.apache.jena.sparql.syntax.PatternVars ;
 import org.apache.jena.sparql.syntax.Template ;
 import org.apache.jena.sparql.util.FmtUtils ;
-import org.apache.jena.system.JenaSystem ;
+import org.apache.jena.sys.JenaSystem ;
 
 /** The data structure for a query as presented externally.
  *  There are two ways of creating a query - use the parser to turn
--- a/jena-arq/src/main/java/org/apache/jena/query/ResultSetFormatter.java
+++ b/jena-arq/src/main/java/org/apache/jena/query/ResultSetFormatter.java
@@ -45,7 +45,7 @@ import org.apache.jena.sparql.resultset.ResultsFormat;
 import org.apache.jena.sparql.resultset.TextOutput;
 import org.apache.jena.sparql.resultset.XMLOutput;
 import org.apache.jena.sparql.serializer.SerializationContext ;
-import org.apache.jena.system.JenaSystem;
+import org.apache.jena.sys.JenaSystem;
 
 /** ResultSetFormatter - Convenience ways to call the various output formatters.
  *  in various formats.
--- a/jena-arq/src/main/java/org/apache/jena/riot/RDFDataMgr.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/RDFDataMgr.java
@@ -45,7 +45,7 @@ import org.apache.jena.sparql.core.Quad ;
 import org.apache.jena.sparql.graph.GraphFactory ;
 import org.apache.jena.sparql.util.Context ;
 import org.apache.jena.sparql.util.Symbol ;
-import org.apache.jena.system.JenaSystem ;
+import org.apache.jena.sys.JenaSystem ;
 import org.slf4j.Logger ;
 import org.slf4j.LoggerFactory ;
 
--- a/jena-arq/src/main/java/org/apache/jena/riot/RDFWriterRegistry.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/RDFWriterRegistry.java
@@ -25,7 +25,7 @@ import org.apache.jena.riot.system.RiotLib ;
 import org.apache.jena.riot.thrift.WriterDatasetThrift ;
 import org.apache.jena.riot.thrift.WriterGraphThrift ;
 import org.apache.jena.riot.writer.* ;
-import org.apache.jena.system.JenaSystem ;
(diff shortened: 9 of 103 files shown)


jena-arq/src/main/java/org/apache/jena/system/JenaInit.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package org.apache.jena.system;
+
+import org.apache.jena.sys.JenaSystem;
+
+/**
+ * This is a temporary adapter for implementations to transition to [JENA-1524]
+ */
+public class JenaInit {
```
afs: I'd prefer this to be called "JenaSystem" then init() is 

```
  @Deprecated
  public static void init() {
         org.apache.jena.sys.JenaSystem.init();
   }
```

because this will work with existing code.  If that's OK, I will do that as part of the merge.

apache-jena-osgi/jena-osgi/src/main/resources/META-INF/services/org.apache.jena.system.JenaSubsystemLifecycle
```diff
@@ -1,4 +1,4 @@
-org.apache.jena.system.InitJenaCore
+org.apache.jena.sys.InitJenaCore
```
afs: Ouch! Yes, the initialization services classes will be impacted.

That means any customer extensions change but an intermediate subclass of JenaSubsystemLifecycle in the new "systems" package should provide a bridge.

Or we can be honest and announce to custom extensions there are changes they need to be aware of.

ajs6f: +1 to honesty. I hear tell it's the best policy :) In all seriousness, the alternative seems to me to build up debt.

afs: I hope this debt isn't too bad because it can be signalled via deprecation. As ever, there is a balance.

The services filename ATM is `org.apache.jena.system.JenaSubsystemLifecycle`. I'm happy to and have a interface `...system.JenaSubsystemLifecycle` to make the transition not too abrupt.

(
Speculative thought, maybe trying too hard:

Can we actually move initialization all to jena-arq and just leave `Serializer` behind? `InitARQ` would call to be non-lifecycle managed initialization code in jena-core. That fixes the jena-core/jena-arq relationship. This may be less debt in the sense that it is a relationship the code depends on anyway. 

People using only jena-core aren't impacted - we actively discourage that but it happens. jena-core initialization happens via services currently but it can be dropped as it is using "services" to call itself, so that flexibility is unnecessary and it can just call itself in a fixed fashion.
)

christopher-johnson:
In working with the automatic modules, I have also found a split with`org.apache.jena.atlas`.  This is between `org.apache.jena.arq` and `org.apache.jena.base`.  

Would it make sense to create a separate project and artifact for `org.apache.jena.atlas`?  

I realize that this sort of package refactoring is a slippery slope.  However, the new module idiom was to enforce a semantic relationship between the project structure, the artifact(s) it produces and the package(s) it provides.  I have found furthermore that it is consistent to name a project in accordance with the qualified module name (without the top level group).  So, the traditional "-" naming semantic could be replaced in archives with a "."

ajs6f:
That sounds like a set of changes (e.g. module renaming) much better suited for 4.0.

afs:
I don't see a split package in `org.apache.jena.atlas` - it has classes and sub-packages in `jena-base` and separate sub-packages in `jena-arq` but I don't see a package being split.  That should not be a problem, should it?

Given this is retrofitting, we have to balance change with continuity for existing users.  Calling it `org.apache.jena.base` follows the general automatic module style. "base" is code that is RDF-independent and is more for internal Jena use.

christopher-johnson:
the module system does not allow two distinct modules (e.g. `org.apache.jena.arq` and `org.apache.jena.base`) to export the same package (e.g. `org.apache.jena.atlas`).  

There are two options in this situation, 
1. consolidate the package into an existing module.
2. separate the package into a new module.

Option 2 is only viable when the package has no dependencies to another module that depends on it (i.e. a cyclical relationship).  I think that  `org.apache.jena.atlas` is not a candidate for separation from `org.apache.jena.arq` as  `org.apache.jena.riot` depends on it, and  `org.apache.jena.atlas` depends on `org.apache.jena.riot`, so they must be in the same module (given the existing code).   

I agree that there is a balance between organizational convenience and radical (presumably hierarchical) package retrofitting to support strong encapsulation.  The concept of one-way dependency vectors governs modularity, but this does not necessarily agree with component semantics when implemented with package names.

afs:
jena-arq does not need to export `org.apache.jena.atlas` - it needs to export `org.apache.jena.atlas.csv`, org.apache.jena.atlas.data`  etc.

`org.apache.jena.atlas` does not depend on `riot`, it only has two classes, both exceptions.

This is the case for `org.apache.jena` in "jena-core". `org.apache.jena` subpackages are everywhere.

afs:
I've tried out the PR and it passes the build and Fuseki starts up. I renamed the ServiceLoader files as `org.apache.jena.sys.JenaSubsystemLifecycle` (I have a script for that now,I can do that after the PR is merged).

One other change is class `system.JenaInit` => `...system.jenaSystem` for compatibility.

This is the JDI version, not the "speculative thought" above. That works, but does risk changing initialization order between jena-arq and jena-core which is the most sensitive part of initialization. That does not cause problems in the test suite but deployed applications are a big unknown. Given the risks and impacts, I'm currently inclined to go straight  to the change to `META-INF/services/org.apache.jena.sys.JenaSubsystemLifecycle`.

Putting in an adapter `org.apache.jena.system.JenaSubsystemLifecycle` ends up as replicating the machinery to find services.

It's a bit bumpy for people working from development - we can let them know via dev@ that they need to do a clearup and rebuild.

So - looking good!

christopher-johnson:
re: `org.apache.jena.atlas` dependents in `org.apache.jena.arq` (namely `org.apache.jena.riot`)

I attach an excerpt from a `jdep`s output looking at `jena-arq-3.7.0.jar`, sorted by package dependency.  
The attached just includes the packages that depend on `org.apache.jena.atlas` in `org.apache.jena.arq`
This is convenient to analyse interpackage relationships (that you probably know of).  It shows non-module deps as "not found", but indicates the package name.  
[arq-atlas-jdeps.txt](https://github.com/apache/jena/files/1939950/arq-atlas-jdeps.txt)

Perhaps this is useful.  I guess the atlas question is a new issue?

afs:
What is the issue here? Does anything break?

oaj = org.apache.jena

> the module system does not allow two distinct modules (e.g. org.apache.jena.arq and org.apache.jena.base) to export the same package (e.g. org.apache.jena.atlas).

jena-arq does not contain any classes of `oaj.atlas` so the package isn't split.

jena-arq would  export the packages like `oaj.atlas.web` that it has but not `oaj.atlas` itself.

jena-base would export `oaj.atlas`, and each subpackage by name. (No use of `*`)

It is the same situation as jena-core having `org.apache.jena` and other jars having `oaj.query`etc.
That is not a split package. 

In an ideal world, the naming might be different but the question is what does the current situation break?

The dependency is:
{noformat}
[INFO] org.apache.jena:jena-arq:jar:3.8.0-SNAPSHOT
[INFO] +- org.apache.jena:jena-core:jar:3.8.0-SNAPSHOT:compile
[INFO] |  \- org.apache.jena:jena-base:jar:3.8.0-SNAPSHOT:compile
{noformat}

> "org.apache.jena.atlas.json.io.parser"  | "org.apache.jena.riot.tokens (jena-arq-3.7.0.jar)";
> Could this be changed?

To what?

They are both in jena-arq - I'm not seeing any problem with that. Are you suggesting better naming or because something breaks?

(The name "atlas" is no more than a holder as a "lib".)

Arguably, the JSON package should have a different hierarchical name but I agree with @ajs6f shuffling package names around impacts users heavily, and few use java9 yet. The end target naming isn't clear until the whole of Jena is considered.

afs:
I ran:

     /usr/lib/jvm/java-9-openjdk-amd64/bin/jdeps -cp 'lib/*' lib/jena-arq-3.8.0-SNAPSHOT.jar 

and the only split packages reported are for Xerces/xml-apis. Various packages under javax.xml, org.w3c.dom and org.xml.sax split (clashing) with the java runtime libraries.

afs:
jdeps complains about oaj.dboe so having a common automatic-moulde-name seems not to work. 

Clearing that out and xerces, then 

    /usr/lib/jvm/java-9-openjdk-amd64/bin/jdeps -R --module-path libx -filter:module -m org.apache.jena.arq 

shows (reformatted from jdeps)

    org.apache.jena.atlas.data -> org.apache.jena.atlas                              org.apache.jena.base

christopher-johnson:
Compilation in JPMS requires a `module-info.java` and is distinct from compiling in JDK 9+.  

> What is the issue here? Does anything break?

A quick background of the use case might help here.  I have been developing JPMS support for the TrellisLDP project recently and have encountered several blocking dependency issues resulting from `jena-osgi`, which is a [transitive dependency ](https://github.com/apache/commons-rdf/blob/master/commons-rdf-jena/pom.xml#L68-L72) resulting from `org.apache.commons.rdf.jena` that is a direct dependency of the `org.trellisldp.app.triplestore` implementation.  JPMS compilation is strict and cannot resolve a class-containing package sourced from multiple "parent modules" (even if the package is private).[1]   A good question here that you raise is what happens if a top level package contains no classes and only exports subpackages.  This requires further investigation and I do not know whether this semi-split, semi-private package, multi-module configuration works, but it is probably not advisable according to the requirements of the [specification](http://openjdk.java.net/projects/jigsaw/spec/reqs/#non-interference).

Regarding this PR, I am running tests now.  BTW, for me, the shaded-guava jar does not work.   I do not know whether it is my IDE, JRE or the maven build configuration.  At any rate, I can run tests if I replace the shaded-guava with a direct dependency.  Also, `guava 21.0` does not work (without hacking) in a JDK10 because it depends on  `javax.annotation.Nullable`.  (it seems that `guava 24.1-jre` removes the problematic jsr-305 annotations).

There seems to be an initialization/lifecycle issue that can be seen when running `TestFactoryRDF`.  The first test to fail there is `factoryRDF_triple_01` with this exception `java.lang.ExceptionInInitializerError`.
also 
`com.google.common.util.concurrent.ExecutionError: java.lang.NoClassDefFoundError: Could not initialize class org.apache.jena.riot.system.RiotLib` is reported.

The preceding tests pass.  It could be my issue, though as an initialization related, it could be a consequence of a the changes in this PR:

`java.lang.ExceptionInInitializerError` is also produced in a test runtime use of `org.apache.jena.arq` by a `org.trellisldp.vocabulary` test:
```
java.lang.ExceptionInInitializerError
       org.apache.jena.riot.RDFParserBuilder.buildFactoryRDF(RDFParserBuilder.java:597)
       org.apache.jena.riot.RDFParserBuilder.build(RDFParserBuilder.java:569)
       org.apache.jena.riot.RDFParserBuilder.parse(RDFParserBuilder.java:512)
       org.trellisldp.vocabulary.AbstractVocabularyTest.getVocabulary(AbstractVocabularyTest.java:56)
       org.trellisldp.vocabulary.AbstractVocabularyTest.testVocabulary(AbstractVocabularyTest.java:62)
```

[1] https://stackoverflow.com/questions/40616739/java-9-overlapping-non-exported-packages

afs:
> Compilation  in JPMS requires a module-info.java and is distinct from compiling in JDK 9+.

The issue is for trellisldp to solve surely? The transitivity is an indirect jena-osgi dependency and a direct jena dependency.

Or a PR to improve `jena-osgi` (which does not get much use or at least not much user@ questions). You may also wish to take that up with CommonsRDF. I'm not sure OSGi is helping here.

`jena-osgi` does not affect the oaj.atlas issue between `jena-arq` and `jena-base`.

This PR is for the org.system split package. This seems completed. I'd like to get this merged and other matters taken up in separate issues/JIRA so we can make incremental progress.

christopher-johnson:
Indeed, these dependency questions are issues that I need to solve... I would prefer to simplify the über-jar `jena-osgi` dependency of `commons.rdf.jena` by creating lean modular jena artifacts (with BND manifests), that can be managed by a downstream application more easily.

I have a quick related question about the shaded-guava dependency.  Would you know why `org.apache.jena.atlas.lib.cache.CacheGuava` would still be looking for `org/apache/jena/ext/com/google/common/cache/RemovalNotification` even though there are absolutely no references to the shadow redirection in the code or the build scripts?  This reference to the shadow redirection is buried somewhere (perhaps in my filesystem), and I know that it is not using any local maven cache now as I built `org.apache.jena.base` in offline mode.  Thanks.

afs:
`RemovalNotification` is used in the constructor for `CacheGuava`.

afs:
PR merged and email sent to dev@jena to let everyone know this has an impact on ServiceLoader files.

afs:
See PR #408 for updating the shaded Google guava.

christopher-johnson:
a quick note here about the prospect of further restructuring to support jpms.  I have learned today that if a module (e.g. `org.apache.jena.arq`) does not contain a single top level package this creates an "ambiguous module".  The expectation is that the module contains a hierarchy of packages and this does not exist in the primary jena library modules,  It is possible for the top package to not contain any classes, but the `module-info.java`  then must explictly export every subpackage it wants to make public.   With a new package hierarchy, it does not seem possible to maintain compatibility with existing implementations _and_ support modules with jena master.  One possibility might be to create a new branch and produce distinct artifacts that could be used downstream.  It could also be used for testing and making changes to support the new jdk methods without impacting LTS.  btw, I have noticed in running the tests that there are a few issues due to jdk changes (`java.nio.ByteBuffer`, for example).  This might be a question for the dev list.

PR #406, opened 2018-04-23
JENA-1532 | Added support for escaping special characters in search strings

anujgandharv:
Added support for escaping special characters in search strings

### DIFF
--- a/jena-text-es/src/main/java/examples/JenaESTextExample.java
+++ b/jena-text-es/src/main/java/examples/JenaESTextExample.java
@@ -73,7 +73,8 @@ public class JenaESTextExample {
         String qs = StrUtils.strjoinNL
                 ( "SELECT * "
 //                        , " { ?s text:query (rdfs:comment 'this' 'lang:en') ;"
-                        , " { ?s text:query ('this' 'lang:en*') ;"
+//                        , " { ?s text:query ('this' 'lang:en*') ;"
+                        ,"{ ?s text:query ('2016-12-01T15:31:10-05:00') ;"
 //                        , " { ?s text:query ('this' 'lang:en-GB') ;"
 //                        , " { ?s text:query (rdfs:comment 'this' 'lang:en-GB') ;"
 //                        , " { ?s text:query (rdfs:comment 'this' 'lang:*') ;"
--- a/jena-text-es/src/main/java/org/apache/jena/query/text/es/TextIndexES.java
+++ b/jena-text-es/src/main/java/org/apache/jena/query/text/es/TextIndexES.java
@@ -23,6 +23,7 @@ import org.apache.jena.graph.Node;
 import org.apache.jena.graph.NodeFactory;
 import org.apache.jena.query.text.*;
 import org.apache.jena.sparql.util.NodeFactoryExtra;
+import org.apache.lucene.queryparser.classic.QueryParserBase;
 import org.elasticsearch.action.admin.indices.exists.indices.IndicesExistsRequest;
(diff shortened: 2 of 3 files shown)


jena-text-es/src/main/java/org/apache/jena/query/text/es/TextIndexES.java
```diff
@@ -422,6 +422,27 @@ public EntityDefinition getDocDef() {
     }
 
     private String parse(String fieldName, String qs, String lang) {
+        //Escape special characters if any in the query string
+        qs = qs.replaceAll("\\:", "\\\\:")
```
ajs6f: This may be a question of style, but might this read a bit better as a constant `Pattern`? (I don't know enough to think very usefully about the performance implications, but I would think a `Pattern` would be at _least_ as performant.)

anujgandharv: Awesome. I will wait it to be merged in master.

jena-text-es/src/main/java/examples/JenaESTextExample.java
```diff
@@ -73,7 +73,8 @@ public static void queryDataWithoutProperty(Dataset dataset)
         String qs = StrUtils.strjoinNL
                 ( "SELECT * "
 //                        , " { ?s text:query (rdfs:comment 'this' 'lang:en') ;"
-                        , " { ?s text:query ('this' 'lang:en*') ;"
+//                        , " { ?s text:query ('this' 'lang:en*') ;"
```
osma: Does the example code have to be changed?

jena-text-es/src/main/java/org/apache/jena/query/text/es/TextIndexES.java
```diff
@@ -422,6 +422,27 @@ public EntityDefinition getDocDef() {
...
+                .replaceAll("\\(", "\\\\(")
+                .replaceAll("\\)", "\\\\)")
+                .replaceAll("\\{", "\\\\{")
+                .replaceAll("\\}", "\\\\}")
+                .replaceAll("\\]", "\\\\]")
+                .replaceAll("\\[", "\\\\[")
+                .replaceAll("\\^", "\\\\^")
+                .replaceAll("\\~", "\\\\~")
+                .replaceAll("\\?", "\\\\?");
+
```
osma: What about escaping preexisting backslashes in the query string? I suppose that would have to be done first.

jena-text-es/src/main/java/examples/JenaESTextExample.java
anujgandharv: All the scenarios are covered in the Unit Test. This example is more of a convenient class to give user an idea as to how to use the ES Integration. It was not meant to be exhaustive.

jena-text-es/src/main/java/org/apache/jena/query/text/es/TextIndexES.java
afs: The [Lucene escape code|https://github.com/apache/lucene-solr/blob/master/lucene/queryparser/src/java/org/apache/lucene/queryparser/classic/QueryParserBase.java#L971].

It includes `\\` and does a simple pass over the string.

The thing to watch for with `replaceAll` is that it creates a regex every call which can mount up even for these fixed string regexps.

anujgandharv:
I agree with you @osma. I have made the changes to use QueryParserBase instead of custom logic. Let me know if it is OK.

afs:
Are we good to merge this now?

osma:
Looks good now. Sorry for not noticing the previous update. +1 for merging.

PR #407, opened 2018-04-24
Cleanup

afs:
A number of small things noticed while investigating a TDB2 report.

Includes tweak for JENA-1428 (cleanup) and fix for JENA-1531.


jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/store/DatasetGraphSwitchable.java
```diff
@@ -111,7 +112,8 @@ public Graph getDefaultGraph() {
     
     @Override
     public Graph getGraph(Node gn) {
-        return ngCache.getOrFill(gn, ()->GraphViewSwitchable.createNamedGraph(this, gn));
+        Node key = ( gn != null ) ? gn : Quad.defaultGraphNodeGenerated;
+        return ngCache.getOrFill(gn, ()->GraphViewSwitchable.createNamedGraph(this, key));
```
rvesse: Shouldn't the first `gn` reference on this line also be changed to `key` otherwise you are adding a potentially `null` key to the cache

PR #413, opened 2018-05-01
JENA-1537: Remove dependency on Apache Xerces.

afs:
The dependency of Apache Xerces 2.11.0 can be removed by extracting the necessary datatype validation code from Xerces and using the JDK XML parser. The Xerces release jars `xercesImpl-2.11.0.jar` and `xml-apis-1.4.01.jar` are no longer needed which has OSGi and JPMS advantages.

Impacts:

* Switch to using the JDK built-in XML parser (this affects any use of XML in an application using Jena)
* Drop `XSDDatatype.loadUserDefined` - the necessary code isn't available via JDK APIs
* Add package tree `org.apache.jena.ext.xerces` for the extracted datatype validation and regex code (a SPARQL corner case)
* Remove Xerces from `pom.xml` and `jena-core/pom.xml`
* Remove Xerces from `jena-osgi/pom.xml`
* No checking of encoding mismatches between reader and XML charset declaration (access to the XML Declaration is not available, at least not in the same way)

The extracted code is only slightly cleaned up to keep some degree of alignment with the original Xerces source. That code originates from a long time ago and has a lot of warnings which have been suppressed.

There will also need to be a change to NOTICE to reflect NOTICE from Xerces (it is already in the NOTICE for the distribution).


stain:
+1 - And as the extracted code is under `org.apache.jena.ext.*` so the modified OSGI export looks good.

stain:
Would be good for the history if this pull request (or ideally git commit) referenced which particular commit/version of Xerces this was extracted from - even if we are "importing from ourself" as Xerces is ASF.

afs:
It is from the released artifacts (specifically, the source artifact).

The Xerces `Version` class has `getVersion()` -> "2.11.0-jena".
The last Xerces dependency is in the JIRA ticket.
And the commit has the change to the dependency management in the POM.

PR #414, opened 2018-05-06
JENA-1542: Integrate Lucene index in transaction lifecycle (TDB1, TDB2).

afs:
Includes the fix for JENA-1302.

Integrate a lucene text index into the transaction systems for TDB1 and DBOE/TDB2.  This way, the prepare/commit for Lucene happens during database commit and the synchronization dance (JENA-1302) does not happen.

JENA-1302 only effects TDB1, not TDB2. Only TDB1 has a commit queue that can build up under load and need to be flushed, which in turn causes a thread to take the exclusive lock and so cause the JENA-1302 lockup.
Includes the fix for JENA-1302.

Integrate a Lucene text index into the transaction systems for TDB1 and DBOE/TDB2.  This way, the prepare/commit for Lucene happens during database commit and the synchronization dance (JENA-1302) does not happen.

JENA-1302 only effects TDB1, not TDB2 or any other DatasetGraph implementation. Only TDB1 has a commit queue that can build up under load and need to be flushed, which in turn causes a thread to take the exclusive lock and so cause the JENA-1302 lockup.

### DIFF
--- a/jena-db/jena-dboe-transaction/src/main/java/org/apache/jena/dboe/transaction/txn/TransactionCoordinator.java
+++ b/jena-db/jena-dboe-transaction/src/main/java/org/apache/jena/dboe/transaction/txn/TransactionCoordinator.java
@@ -131,7 +131,7 @@ public class TransactionCoordinator {
      * This must be setup before recovery is attempted. 
      */
     public TransactionCoordinator add(TransactionalComponent elt) {
-        checklAllowModication() ;
+        checklAllowModification() ;
         components.add(elt) ;
         return this ;
     }
@@ -141,7 +141,7 @@ public class TransactionCoordinator {
      * @see #add 
      */
     public TransactionCoordinator remove(TransactionalComponent elt) {
-        checklAllowModication() ;
+        checklAllowModification() ;
         components.remove(elt.getComponentId()) ;
         return this ;
     }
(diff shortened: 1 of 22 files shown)


jena-db/jena-dboe-transaction/src/main/java/org/apache/jena/dboe/transaction/txn/TransactionCoordinator.java
```diff
@@ -142,41 +141,56 @@ public TransactionCoordinator add(TransactionalComponent elt) {
...
-        synchronized(coordinatorLock) {
-            components.remove(elt.getComponentId()) ;
-        }
+        checklAllowModification() ;
+        components.remove(elt.getComponentId()) ;
         return this ;
     }
+    
+    /** 
+     * Perform modification of this {@code TransaxctionCoordiator} after it has been started.
```
rvesse: Typo in the `{@code }` here?

jena-text/src/test/java/org/apache/jena/query/text/TestTextTxn.java
```diff
@@ -0,0 +1,341 @@
...
+import org.junit.Test ;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+/** Text dataset tests using datasets transactionally, named unionDefaultGraph.
+ * <p>
+ * Note that in Lucene, writes are not visible to a reader of the {@code Directory}
+ * until committed. A special {@link IndexReader} is needed. See
+ * {@link DirectoryReader#open(IndexWriter)}. jena-text does not currentyl do this.
```
rvesse: Typo - `currentyl`

PR #416, opened 2018-05-07
JENA-1545: ParserProfileWrapper

afs:


jena-arq/src/main/java/org/apache/jena/riot/system/ParserProfile.java
```diff
@@ -49,6 +49,12 @@
     /** Create a triple */
     public Triple createTriple(Node subject, Node predicate, Node object, long line, long col);
 
+    /** Check a triple */
```
ajs6f: Might be nice to include a Javadoc note declaring what to do for bad tuples. Presumably `throw` something, but is there a specific type to throw that would be best?

afs:
Extensively revised : calling `checkTriple` inside `createTriple` doesn't delegate properly in the proxy pattern. Instead, and in the interests of simplicity, any `ParserProfile` exposes enough setup state to allow a proxy to provide an alternative `createTriple` operation: 

This way provides the immediate need for PR #299 without limiting future possibilities for stream based checking (e.g. to have a checking policy passed to a ParserProfile) to include line and column numbers.  It is inline with the current javadoc for `ParserProfile` which already says a `FactoryRDF` is used.

Example, to allow literals as subjects:

        ParserProfile pp = RiotLib.profile(Lang.TTL, "http://base/");
        ParserProfile pp2 = new ParserProfileWrapper(pp) {
            @Override public Triple createTriple(Node subject, Node predicate, Node object, long line, long col) {
               if ( subject == null || (!subject.isURI() && !subject.isBlank() && !subject.isLiteral()) ) {
                    getErrorHandler().error("Subject is not a URI, blank node or literal", line, col);
                    throw new RiotException("Bad subject: " + subject);
                }
                if ( predicate == null || (!predicate.isURI()) ) {
                    getErrorHandler().error("Predicate not a URI", line, col);
                    throw new RiotException("Bad predicate: " + predicate);
                }
                if ( object == null || (!object.isURI() && !object.isBlank() && !object.isLiteral()) ) {
                    getErrorHandler().error("Object is not a URI, blank node or literal", line, col);
                    throw new RiotException("Bad object: " + object);
                }
                getErrorHandler().warning("Create triple!", line, col);
                return getFactorRDF().createTriple(subject, predicate, object);
            }
        };
        // Example of literal subject, calling machinery directly.
        Tokenizer t = TokenizerFactory.makeTokenizer(new StringReader("123 <p> <o>"));
        LangTurtle parser = new LangTurtle(t, pp2, StreamRDFLib.writer(System.out));
        parser.parse();
    }

PR #421, opened 2018-05-12
Escape @xerces.internal annotations

afs:
Replace "@xerces.internal" with "{@literal @xerces.internal}".

This suppresses javadoc warning about unknown javadoc tags.


ajs6f:
I've never seen `@xerces.internal` before, but it seems to denote types internal to Xerces, right? My one question is: do we need to keep it on all types or can we stick it on a package comment? (Since we're not planning on maintaining a separate "branch" of Xerces as a whole.)

ajs6f:
Okay, either way works for me. This is already done, so easier!

PR #426, opened 2018-06-02
JENA-1552: Phased loader

afs:
This has become a general purpose loader (`LoaderMain`) which is controlled by a `LoaderPlan`. LoaderMain is both the parallel loader and the phased loader of JENA-1552.


jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/loader/parallel/LoaderPlan.java
```diff
@@ -0,0 +1,57 @@
...
+/** 
+ * A {@code LoaderPlan}
+ * <p>
+ * For triples and for quads there is a first phase to parse the input, 
+ * convert to tuples of {@link NodeId NodeIds}, including allocating the ids,
+ * and do at least one tuple index for each of triples quads to capture the input.
+ * <p>   
+ * After that, a number of phases builds the other indexes. 
+ * <p>
+ * The {@code mulithreadedInput} f;ag indicates whether the first phase is
```
rvesse: Typo `f;ag` -> `flag`

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/loader/parallel/LoaderPlan.java
```diff
@@ -0,0 +1,57 @@
...
+        this.loadGroup3 = loadGroup3;
+        this.loadGroup4 = loadGroup4;
+        this.secondaryGroups3 = secondaryGroups3;
+        this.secondaryGroups4 = secondaryGroups4;
+    }
+    public InputStage dataInputType()   { return dataInput; }
+    public String[] primaryLoad3()          { return loadGroup3; }
+    public String[] primaryLoad4()          { return loadGroup4; }
+    public String[][] secondaryIndex3()     { return secondaryGroups3; }
+    public String[][] secondaryIndex4()     { return secondaryGroups4; }
```
rvesse: The naming of these seems somewhat esoteric, I assume that the 3 and 4 refers to Triples and Quads respectively?

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/loader/parallel/LoaderPlan.java
```diff
@@ -0,0 +1,57 @@
...
+ * convert to tuples of {@link NodeId NodeIds}, including allocating the ids,
+ * and do at least one tuple index for each of triples quads to capture the input.
+ * <p>   
+ * After that, a number of phases builds the other indexes. 
+ * <p>
+ * The {@code mulithreadedInput} f;ag indicates whether the first phase is
+ * done in parallel (threads for parer, node table building and primary indexes)
+ * or as a single threaded process.
+ */
+public class LoaderPlan {
```
rvesse: I really like this design, since the class has a public constructor end-users could potentially test out alternative strategies if they so desired?

afs:
Long term (!), I'd like to get some aspects TDB1 `tdbloader2` into the TDB2 but in Java for portability. `tdbloader2` is breaking up the task of building a single index into machine-resource sized chunks. `LoaderMain` does not consider multiple tasks to build one index; this also means `LoaderPlan`s are not completely stable as an API.

In doing work on loading, I can see more possibilities which need experimentation to see if they are a good idea or a mad idea. This PR is breaking the "wait until perfect" effect. This loader should be good for a 100's millions triples (I haven't tested a billion yet) so it is a significant step forward in itself.

afs:
All technical work is finished for this iteration. I'll do some renaming ("parallel" is still used as a general name whereas it is now a profile of "LoaderMain") then this should be good to go.

PR #428, opened 2018-06-06
Update OSGi imports

acoburn:
Resolves: JENA-1557

This adds an exclusion for `org.checkerframework.checker.*`, which is now pulled in by `jena-shaded-guava`. This also adds guava (jsonld-java depends on this) and commons-compress to the `features.xml` file.

In addition, this reformats the `Import-Package` definition to make it easier to read (and maintain).

It is worth noting that the Guava version (24.1-jre) is hard-coded in the `features.xml` file. I was reluctant to add a variable for this in the root `pom.xml` file given [the note](https://github.com/apache/jena/blob/master/jena-shaded-guava/pom.xml#L46) in jena-shaded-guava. However, there may be a better way to structure that value so that it is obvious that it is tied to the jsonld-java dependency.

### DIFF
--- a/apache-jena-osgi/jena-osgi-features/src/main/resources/features.xml
+++ b/apache-jena-osgi/jena-osgi-features/src/main/resources/features.xml
@@ -21,31 +21,24 @@
 	<feature name="jena" version="${project.version}">
 		<bundle>mvn:org.apache.jena/jena-osgi/${project.version}</bundle>
 		<feature version="${project.version}">jena_osgi_dependencies</feature>
-		<!-- This is normally exposed by Apache Karaf via root classloader -->
-		<!--<feature version="2.11.0">xerces</feature>-->
 	</feature>
 
 	<feature name="jena_osgi_dependencies" version="${project.version}">
-		<bundle>mvn:com.github.andrewoma.dexx/collection/${ver.dexxcollection}</bundle>
-		<bundle>mvn:com.github.jsonld-java/jsonld-java/${ver.jsonldjava}</bundle>
-		<bundle>mvn:com.fasterxml.jackson.core/jackson-core/${ver.jackson}</bundle>
-		<bundle>mvn:com.fasterxml.jackson.core/jackson-databind/${ver.jackson}</bundle>
-		<bundle>mvn:com.fasterxml.jackson.core/jackson-annotations/${ver.jackson}</bundle>
-		<bundle>mvn:org.apache.httpcomponents/httpcore-osgi/${ver.httpcore-osgi}</bundle>
-		<bundle>mvn:org.apache.httpcomponents/httpclient-osgi/${ver.httpclient-osgi}</bundle>
-		<bundle>mvn:org.apache.commons/commons-compress/${ver.commons-compress}</bundle>
-		<bundle>mvn:org.apache.commons/commons-csv/${ver.commonscsv}</bundle>
-		<bundle>mvn:org.apache.commons/commons-lang3/${ver.commonslang3}</bundle>
-		<bundle>mvn:commons-codec/commons-codec/${ver.commons-codec}</bundle>
-		<bundle>mvn:commons-io/commons-io/${ver.commonsio}</bundle>
-		<bundle>mvn:org.apache.thrift/libthrift/${ver.libthrift}</bundle>
+		<bundle dependency="true">mvn:com.github.andrewoma.dexx/collection/${ver.dexxcollection}</bundle>
+		<bundle dependency="true">mvn:com.github.jsonld-java/jsonld-java/${ver.jsonldjava}</bundle>
+		<bundle dependency="true">mvn:com.fasterxml.jackson.core/jackson-core/${ver.jackson}</bundle>
+		<bundle dependency="true">mvn:com.fasterxml.jackson.core/jackson-databind/${ver.jackson}</bundle>
+		<bundle dependency="true">mvn:com.fasterxml.jackson.core/jackson-annotations/${ver.jackson}</bundle>
+		<bundle dependency="true">mvn:org.apache.httpcomponents/httpcore-osgi/${ver.httpcore-osgi}</bundle>
+		<bundle dependency="true">mvn:org.apache.httpcomponents/httpclient-osgi/${ver.httpclient-osgi}</bundle>
+		<bundle dependency="true">mvn:org.apache.commons/commons-compress/${ver.commons-compress}</bundle>
+		<bundle dependency="true">mvn:org.apache.commons/commons-csv/${ver.commonscsv}</bundle>
+		<bundle dependency="true">mvn:org.apache.commons/commons-lang3/${ver.commonslang3}</bundle>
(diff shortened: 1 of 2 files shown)


afs:
We ought to exclude all the tools that Guava adds as it's own dependencies (it would be nice if they were marked optional).

acoburn:
I believe the Guava dependencies are brought in as part of the shading process. Those dependencies are not typically required for Guava itself (they are only test-scoped dependencies). So perhaps there is something that can be done in the `jena-shaded-guava` project to exclude those at that point.

afs:
We can, and for clarity should, exclude them but when I look in the jena shaded jar, I don't see them. There is a filter-includes in Jena to pick only `com.google.common` and `com.google.thirdparty`. Maybe we don't even need "thirdparty".

In the jsonld-java 0.12.0 jar, I only see `com.github.jsonldjava.shaded.com.google.common....`.

acoburn:
I am definitely no expert on shaded jars. The above comment was based on running `mvn dependency:tree` and seeing these compile-scoped dependencies:

```
[INFO] +- org.apache.jena:jena-shaded-guava:jar:3.8.0-SNAPSHOT:compile
[INFO] |  \- com.google.guava:guava:jar:24.1-jre:compile
[INFO] |     +- com.google.code.findbugs:jsr305:jar:1.3.9:compile
[INFO] |     +- org.checkerframework:checker-compat-qual:jar:2.0.0:compile
[INFO] |     +- com.google.errorprone:error_prone_annotations:jar:2.1.3:compile
[INFO] |     +- com.google.j2objc:j2objc-annotations:jar:1.1:compile
[INFO] |     \- org.codehaus.mojo:animal-sniffer-annotations:jar:1.14:compile
```

Also, when trying to provision `jena-osgi` in Karaf, those dependent modules are needed unless they are explicitly excluded (as in this PR).

As for the jsonld dependency, the OSGi manifest headers include:

```
[IMPEXP]
Import-Package
  ...
  com.google.common.cache                {version=[24.1,25)}
  com.google.common.collect              {version=[24.1,25)}
```

so perhaps that is something to bring up with the json-ld project.

afs:
We can exclude for jena shading - it makes the `dependency:tree` cleaner but (I have tried) does not affect the shaded jar because the jar is filtered by an include.

We could try making the dependency "optional" so it is not picked up transitively (I have not tried yet). Would that help jena-osgi, if it works?  Jena and jsonld-java do not have a real dependency of the guava jar.

We can add the same excludes to the dependency in jena-arq that brings in jsonld-java and PR that and/or the <optional> upstream when we have it all proven.

For `jena-osgi`, do we need `<Import-Package...com.google.errorprone.annotations.concurrent...`/ -- presumablu not.

`jena-osgi-features` - because Jena now has it own, cut down Xerces, presumably the `<feature name="xerces" version="2.11.0">` can be removed.

afs:
JENA-1558 / PR #429.

`<optional>true`seems to work - I've added the exclusions as well so that `dependency:tree` run in `jena-shared-guava` looks right.  The test suite passes. If it looks decent I can push it into the development code to make it easier to try out for OSGi.

Putting all the OSGi changes in one is PR is fine by me - whatever is easier for you.

acoburn:
I built Jena with your change in #429 along with the changes from this PR, and for some reason the `jena-osgi` module still wants to import the `com.google.errorprone.annotations` and `org.checkerframework.checker` packages -- unless, of course, they are explicitly excluded, as in this PR. So, unless there is some other way to achieve this, I think it will be necessary to explicitly exclude those guava-related transitive dependencies.

In the second commit, I removed the `xerces` feature. I also added the `dependency="true"` flag to all of the dependencies. Background on the `dependency="true"` attribute is available in the [Karaf documentation](https://karaf.apache.org/manual/latest/provisioning).

I tested this in two ways. First, simply provisioning the `jena` feature in Karaf 4.2.0 works via `feature:install jena`. I also was able to provision and run a complete OSGi-based CXF application that makes use of Jena.

One further change that could be made to the `features.xml` file is to remove the `jena_osgi_dependencies` feature, and instead put the content of that directly into the `jena` feature definition. That would result in `jena-osgi-features` making only a single feature available. That would simplify the file a bit, but that change also isn't strictly necessary.

apache-jena-osgi/jena-osgi-features/src/main/resources/features.xml
```diff
@@ -21,28 +21,24 @@
...
+		<bundle dependency="true">mvn:org.apache.httpcomponents/httpclient-osgi/${ver.httpclient-osgi}</bundle>
+		<bundle dependency="true">mvn:org.apache.commons/commons-compress/${ver.commons-compress}</bundle>
+		<bundle dependency="true">mvn:org.apache.commons/commons-csv/${ver.commonscsv}</bundle>
+		<bundle dependency="true">mvn:org.apache.commons/commons-lang3/${ver.commonslang3}</bundle>
+		<bundle dependency="true">mvn:commons-codec/commons-codec/${ver.commons-codec}</bundle>
+		<bundle dependency="true">mvn:commons-io/commons-io/${ver.commonsio}</bundle>
+		<bundle dependency="true">mvn:org.apache.thrift/libthrift/${ver.libthrift}</bundle>
+		<!-- Guava 24.1 is required by jsonld-java -->
+		<bundle dependency="true">mvn:com.google.guava/guava/24.1-jre</bundle>
 	</feature>
```
afs: I don't think this is true any more. At version 0.12.0, they shade guava and put in their jar so it does not need com.google.guava:guava. At least, that's the theory! If you have found that in practice it is needed, that has more weight.

afs:
Thanks for trying out  PR #429 - it is now merged and there is a development build with it in. All I can think of for the `com.google.errorprone.annotations` and `org.checkerframework.checker` packages is that the OSGi build is taking from the repos but that's nothing more than speculation.

apache-jena-osgi/jena-osgi-features/src/main/resources/features.xml
acoburn: You are correct that jsonld-java does shade guava, but their OSGi bundle still imports the `com.google.common` packages: this is from the jsonld-java OSGi metadata:

```
Import-Package
  ...
  com.google.common.cache                {version=[24.1,25)}
  com.google.common.collect              {version=[24.1,25)}
```

And so in practice, adding guava as a dependency is still necessary. I think this is an error with the jsonld-java OSGi metadata, and I will bring this up with that project.

For example, when trying to load just the `jsonld-java` bundle in Karaf -- once I add the Jackson dependencies -- I get this error if guava is not also installed:

```
Unable to resolve com.github.jsonld-java [46](R 46.0): missing requirement [com.github.jsonld-java [46](R 46.0)] osgi.wiring.package; (&(osgi.wiring.package=com.google.common.cache)(version>=24.1.0)(!(version>=25.0.0)))
```

...which means that the guava packages need to be installed. Once guava is added, the jsonld-java bundle installs and starts just fine.

afs: Ah - OK - that explains it. All looks to go.

PR #432, opened 2018-06-09
JENA-1560: PrefixMappingUtils

afs:
Some utility code that can be useful.

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/util/PrefixMappingUtils.java
+++ b/jena-core/src/main/java/org/apache/jena/util/PrefixMappingUtils.java
@@ -1,0 +1,364 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.jena.util;
+
+import java.util.* ;
+import java.util.function.Consumer ;
+import java.util.stream.Collectors ;
+
+import org.apache.jena.atlas.lib.SetUtils ;
+import org.apache.jena.atlas.lib.Trie ;
+import org.apache.jena.graph.Graph ;
+import org.apache.jena.graph.Node ;
+import org.apache.jena.graph.Triple ;
+import org.apache.jena.graph.impl.WrappedGraph;
+import org.apache.jena.rdf.model.Model ;
+import org.apache.jena.shared.PrefixMapping ;
+import org.apache.jena.shared.impl.PrefixMappingImpl ;
+
+public class PrefixMappingUtils {
+    /**
+     * Return a read-only graph that has the same data (RDF triples) as the one given, but has a
+     * prefix mapping that only includes "in use" prefixes as calculated by
+     * {@link #calcInUsePrefixMapping(Graph, PrefixMapping)}.
+     * <p>
+     * The prefix mappings of the two graphs are not connected. 
+     * Later changes to the prefix mapping of the original graph are not reflected in the returned graph.
+     * Modifications to the triples contained in the underlying graph are reflected.   
+     */
+    public static Graph graphInUsePrefixMapping(Graph graph) {
+        final PrefixMapping prefixMapping = calcInUsePrefixMapping(graph) ;
+        prefixMapping.lock() ;
+        Graph graph2 = new WrappedGraph(graph) {
+            @Override
+            public void performAdd(Triple triple)
+            { throw new UnsupportedOperationException() ; }
+            
+            @Override
+            public void performDelete(Triple triple)
+            { throw new UnsupportedOperationException() ; }
+            
+            @Override
+            public PrefixMapping getPrefixMapping() {
+                return prefixMapping ;
+            }
+        } ;
+        return graph2 ;
+    }
+    
+    /**
+     * Analyse the graph to see which prefixes of the graph are in use.
+     * <p>
+     * In the case of overlapping prefixes (where one prefix declaration is has an initial
+     * URI string which matches another prefix declaration), all are included, though
+     * they may not be used when printing (that depends on the output process). In effect,
+     * this process has "false positives".
+     * <p>
+     * This function does not calculate new prefixes.
+     * 
+     * @see #calcInUsePrefixMappingTTL(Graph)
+     */
+    public static PrefixMapping calcInUsePrefixMapping(Graph graph) {
+        PrefixMapping prefixMapping = graph.getPrefixMapping() ;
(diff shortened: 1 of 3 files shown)


jena-core/src/main/java/org/apache/jena/util/PrefixMappingUtils.java
```diff
@@ -0,0 +1,361 @@
...
+            }
+        } ;
+        return graph2 ;
+    }
+    
+    /**
+     * Analyse the graph to see which prefixes of the graph are in use.
+     * <p>
+     * In the case of overlapping prefixes (where one prefix declaration is has an initial
+     * URI string which imatches another prefix declaration), all are included, though
```
kinow: s/imatches/it matches

?

jena-core/src/main/java/org/apache/jena/util/PrefixMappingUtils.java
```diff
@@ -0,0 +1,361 @@
...
+            return null ;
+        return calcInUsePrefixMapping(graph, prefixMapping) ;
+    }
+    
+    /**
+     * Analyse the graph to see which prefixes of the given {@link PrefixMapping} are in
+     * use.
+     * <p>
+     * In the case of overlapping prefixes (where one prefix declaration is has an initial
+     * URI string which imatches another prefix declaration), all are included, though
```
kinow: s/imatches/it matches
?

jena-core/src/main/java/org/apache/jena/util/PrefixMappingUtils.java
```diff
@@ -0,0 +1,361 @@
...
+         * 
+         * Exit early if every prefix is accounted for. 
+         */
+       
+        // Map prefix to URI.
+        Map<String, String> pmap = prefixMapping.getNsPrefixMap() ;
+        
+        // Map URI to prefix, with partial lookup (all uri keys that partly match the URI) 
+        Trie<String> trie = new Trie<>() ; 
+        // Chnage this to "add(uri, uri)" to calculate the uris.
```
kinow: s/Chnage/Change

jena-core/src/main/java/org/apache/jena/util/PrefixMappingUtils.java
```diff
@@ -0,0 +1,361 @@
...
+        Set<String> inUse = new HashSet<>() ;
+        
+        // Process to apply to each node
+        Consumer<Node> process = (node)->{
+            if ( ! node.isURI() )
+                return ;
+            String uri = node.getURI() ;
+            // Get all prefixes whose URIs are candidates 
+            List<String> hits = trie.partialSearch(uri) ;
+            if ( hits == null || hits.isEmpty() )
```
kinow: I think `partialSearch` never returns `null`, but harmless to leave the check here.

afs:
Thanks - all review points done. I amended to PR to keep it to one commit.

jena-core/src/main/java/org/apache/jena/util/PrefixMappingUtils.java
```diff
@@ -0,0 +1,362 @@
...
+import org.apache.jena.shared.impl.PrefixMappingImpl ;
+
+public class PrefixMappingUtils {
+    /**
+     * Return a read-only graph that has the same data (RDF triples) as the one given, but has a
+     * prefix mapping that only includes "in use " prefixes.
+     * <p>
+     * The prefix mappings of the two graphs are not connected. 
+     * Later changes to the prefix mapping of the original graph are not reflected in the returned graph.
+     * Modifications to the triples conatained in the underlying graph are reflected.
```
ajs6f: Typo "conatained" => "contained"

jena-core/src/main/java/org/apache/jena/util/PrefixMappingUtils.java
```diff
@@ -0,0 +1,362 @@
...
+import org.apache.jena.graph.Triple ;
+import org.apache.jena.graph.impl.WrappedGraph;
+import org.apache.jena.rdf.model.Model ;
+import org.apache.jena.shared.PrefixMapping ;
+import org.apache.jena.shared.impl.PrefixMappingImpl ;
+
+public class PrefixMappingUtils {
+    /**
+     * Return a read-only graph that has the same data (RDF triples) as the one given, but has a
+     * prefix mapping that only includes "in use " prefixes.
```
ajs6f: Might want to put a link here to `calcInUsePrefixMapping` below just in case someone wants to the precise meaning of "in use".

jena-core/src/main/java/org/apache/jena/util/PrefixMappingUtils.java
```diff
@@ -0,0 +1,362 @@
...
+     * <p>
+     * This function attempts to process each URI in the graph as if it were to be printed
+     * in Turtle. Only prefixes that lead to valid output strings are returned. This is
+     * more expensive than {@link #calcInUsePrefixMapping(Graph, PrefixMapping)}.
+     * <p>
+     * This function does not calculate new prefixes.
+     * 
+     * @see #calcInUsePrefixMapping(Graph, PrefixMapping)
+     */
+    public static PrefixMapping calcInUsePrefixMappingTTL(Graph graph, PrefixMapping prefixMapping) {
```
ajs6f: Maybe leave a TODO to factor out common logic between this and `calcInUsePrefixMapping`?

jena-core/src/main/java/org/apache/jena/util/PrefixMappingUtils.java
```diff
@@ -0,0 +1,362 @@
...
+    private static void processFull(Trie<String> trie, Set<String> prefixesInUse, Node node) {
+        if ( ! node.isURI() )
+            return ;
+        String uri = node.getURI() ;
+        // Shorten to "/" or "#" or ":"
+        String pref = uri ;
+        
+        // Get all under the pref
+        List<String> hits = trie.partialSearch(pref) ;
+        if ( hits == null || hits.isEmpty() )
```
ajs6f: Looking at `Trie::partialSearch`, I don't think `hits` can be `null` here. I guess why you don't check for `null` above?

jena-core/src/main/java/org/apache/jena/util/PrefixMappingUtils.java
```diff
@@ -0,0 +1,362 @@
...
+        if ( prefixURIs.contains(prefixUri) )
+            inUse.add(prefixUri) ;
+//        String ns = node.getNameSpace() ;
+//        if ( ns == null )
+//            return ;
+//        if ( prefixURIs.contains(ns) )
+//            inUse.add(ns) ;
+    }
+    
+    private static void print(Set<String> inUsePrefixURIs, PrefixMapping prefixMapping) {
```
ajs6f: Should we leave this in? It looks like debugging code?

jena-core/src/main/java/org/apache/jena/util/PrefixMappingUtils.java
afs: Done

jena-core/src/main/java/org/apache/jena/util/PrefixMappingUtils.java
afs: The algorithm took a while to get right and being about to print the state was useful.  I'll add a comment that's it for development assistance. Maybe it will never needed.

afs: The algorithm took a while to get right and being about to print the state was useful.  I'll add a comment that's it for development assistance. Maybe it will never needed.

jena-core/src/main/java/org/apache/jena/util/PrefixMappingUtils.java
afs: Just being defensive.

PR #433, opened 2018-06-09
JENA-1561: Enable TDB2 in Fuseki UI and on the command line.

afs:
Both full and basic server command line code.

### DIFF
--- a/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/mgt/ActionDatasets.java
+++ b/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/mgt/ActionDatasets.java
@@ -24,10 +24,7 @@ import java.io.IOException ;
 import java.io.InputStream ;
 import java.io.OutputStream ;
 import java.io.StringReader ;
-import java.util.HashMap ;
-import java.util.Iterator ;
-import java.util.List ;
-import java.util.Map ;
+import java.util.*;
 
 import javax.servlet.http.HttpServletRequest ;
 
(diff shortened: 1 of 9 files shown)


jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/mgt/ActionDatasets.java
```diff
@@ -351,15 +356,13 @@ private static void assemblerFromForm(HttpAction action, StreamRDF dest) {
...
         
-        String template = null ;
-        if ( dbType.equalsIgnoreCase(tDatabasetTDB))
-            template = TemplateFunctions.templateFile(Template.templateTDBFN, params, Lang.TTL) ;
-        if ( dbType.equalsIgnoreCase(tDatabasetMem))
-            template = TemplateFunctions.templateFile(Template.templateMemFN, params, Lang.TTL) ;
-        RDFParser.create().source(new StringReader(template)).base("http://base/").lang(Lang.TTL).parse(dest);
+        String template = dbTypeToTemplate.get(dbType.toLowerCase(Locale.ROOT));
+        if ( template == null )
+                ServletOps.errorBadRequest(format("dbType can be only '%s' or '%s'", tDatabaseTDB, tDatabaseMem)) ;
```
rvesse: This PR now allows three possibilities so the error message is now inaccurate

afs: Good point! Done.

PR #434, opened 2018-06-10
JENA-1562: Fix for Graph.size() for TDB2 graphs

afs:


jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/store/GraphViewSwitchable.java
```diff
@@ -79,16 +83,34 @@ public TransactionHandler getTransactionHandler() {
...
+        if ( isDefaultGraph() )
+            return getDSG().getDefaultGraphTDB().size();
+        return getDSG().getGraphTDB(getGraphName()).size();
+    }
+
+    private DatasetGraphTDB getDSG() {
+        return ((DatasetGraphTDB)(getx().get()));    
+    }
+    
+    private static Function<Tuple<NodeId>, Tuple<NodeId>> project4TupleTo3Tuple = item -> {
```
ajs6f: This seems kind of out of place...

jena-arq/src/main/java/org/apache/jena/sparql/core/GraphView.java
```diff
@@ -165,6 +165,9 @@ public void performDelete( Triple t ) {
         Node o = t.getObject() ;
         dsg.delete(g, s, p, o) ;
     }
+    
+    // Subclasses may wish to provide graphBaseSize otherwise GraphBase uses find()
```
ajs6f: Maybe this is just where I like to find things, but I'd rather `@Override protected int graphBaseSize() { return super.graphBaseSize(); }` with this as a Javadoc comment.

afs: Done.

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/store/GraphViewSwitchable.java
afs: Don't understand.  It is used in the method above and pulled out for clarity. It makes the handling of union graph clearer IMO with a named function at that point, not inline code.

ajs6f: I'm totally confused. The method above reads:
```
private DatasetGraphTDB getDSG() {
        return ((DatasetGraphTDB)(getx().get()));    
}
```
Where is `project4TupleTo3Tuple` used there?

afs: OK - got it - it's a stray from GraphTDB. This one can be cleaned out.

PR #438, opened 2018-06-22
Fix javadoc in OntModel

fabriziocucci:
Fixed order of parameters description as they appear in the corresponding method signature (first `uri`, then `cls`).


afs:
Thanks!

PR #439, opened 2018-06-25
JENA-1565: Print the throwable and stacktrace

afs:


rvesse:
Do you want this in 3.8.0?

afs:
This is a direct fix of the problem. While the JUL support isn't used by Jena itself, I have used it elsewhere for simple logging for testing and for servers.

For more controlled logging, a specific package would be used to provide the slf4j implementation for Jena to see if, long term, it can be a simple default for Jena, dropping the now discontinued log4j in cases where Jena needs an output channel.

afs:
> Do you want this in 3.8.0?

It is not vital but it would be helpful. I have been using it elsewhere as a try-out for whether it's viable for Jena itself as a minimal default.

rvesse:
@afs If you want to merge this now I was about to cut the RC for 3.8.0

PR #440, opened 2018-06-27
Fix SSE parsing on table with boolean values (JENA-1566)

rvesse:
JENA-1566 reports an issue with round tripping through SSE of tables that contain boolean literals i.e. `true` and `false`

This was due to overly strict checking in `BuilderBinding` which has been removed although a missing test for non-blank-ness was added.  Various unit tests are added around this issue.

### DIFF
--- a/jena-arq/src/test/java/org/apache/jena/sparql/syntax/TestSSE_Builder.java
+++ b/jena-arq/src/test/java/org/apache/jena/sparql/syntax/TestSSE_Builder.java
@@ -22,6 +22,7 @@ package org.apache.jena.sparql.syntax;
 import junit.framework.TestCase ;
 import org.apache.jena.graph.Node;
 import org.apache.jena.sparql.algebra.Op ;
+import org.apache.jena.sparql.algebra.TableFactory;
 import org.apache.jena.sparql.algebra.op.OpLabel ;
 import org.apache.jena.sparql.algebra.op.OpNull ;
 import org.apache.jena.sparql.algebra.op.OpTable ;
@@ -29,9 +30,12 @@ import org.apache.jena.sparql.core.Var;
 import org.apache.jena.sparql.expr.E_IsNumeric ;
 import org.apache.jena.sparql.expr.E_SameTerm ;
 import org.apache.jena.sparql.expr.Expr ;
+import org.apache.jena.sparql.graph.NodeConst;
 import org.apache.jena.sparql.sse.Item ;
 import org.apache.jena.sparql.sse.SSE ;
 import org.apache.jena.sparql.sse.builders.BuilderNode ;
+import org.apache.jena.sparql.sse.builders.ExprBuildException;
+import org.apache.jena.vocabulary.XSD;
 import org.junit.Test ;
 
 public class TestSSE_Builder extends TestCase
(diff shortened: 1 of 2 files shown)


jena-arq/src/test/java/org/apache/jena/sparql/syntax/TestSSE_Builder.java
```diff
@@ -233,4 +247,54 @@ public void testBuildExpr_14() {
...
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    public void testBuildTableBad_01() {
+        try {
+            SSE.parseOp("(table (vars ?x) (row (?x (table unit))))");
+            Assert.fail("Parsing should fail");
+        } catch (ExprBuildException e) {
+            // OK
```
afs: `@Test(expected=ExprBuildException.class)`

jena-arq/src/test/java/org/apache/jena/sparql/syntax/TestSSE_Builder.java
```diff
@@ -233,4 +247,54 @@ public void testBuildExpr_14() {
...
+        }
+    }
+
+    @Test
+    public void testBuildTableBad_02() {
+        try {
+            SSE.parseOp("(table (vars ?x) (row (?x _:anon)))");
+            Assert.fail("Parsing should fail");
```
afs: Shouldn't this parse OK?

Even though SPARQL syntax does not allow blank nodes in VALUES, it seem to be no harm to have the `(table)` parsing be more general. The `Table` class is a list of `Binding`s and `Binding` can have blank nodes. `Table` can used for intermediate results like in the ref engine.

rvesse: Yeah I was unsure on this, I added a new check for this but I am happy to allow this to parse if desired

jena-arq/src/test/java/org/apache/jena/sparql/syntax/TestSSE_Builder.java
rvesse: Yeah I know, wasn't working for me for some reason

rvesse: I get the following when using the preferred form:

```
[-1,-1]: Not a node: (table ...)
[1,27]: Only concrete nodes as table values: ANY: _
[ERROR] Tests run: 139, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 0.036 s <<< FAILURE! - in org.apache.jena.sparql.syntax.TS_SSE
[ERROR] testBuildTableBad_01(org.apache.jena.sparql.syntax.TestSSE_Builder)  Time elapsed: 0.006 s  <<< ERROR!
org.apache.jena.sparql.sse.builders.ExprBuildException: [-1,-1]: Not a node
	at org.apache.jena.sparql.syntax.TestSSE_Builder.testBuildTableBad_01(TestSSE_Builder.java:282)

[ERROR] testBuildTableBad_02(org.apache.jena.sparql.syntax.TestSSE_Builder)  Time elapsed: 0 s  <<< ERROR!
org.apache.jena.sparql.sse.builders.ExprBuildException: [1,27]: Only concrete nodes as table values: ANY
	at org.apache.jena.sparql.syntax.TestSSE_Builder.testBuildTableBad_02(TestSSE_Builder.java:288)
```

Any idea why?

jena-arq/src/test/java/org/apache/jena/sparql/syntax/TestSSE_Builder.java
```diff
@@ -233,4 +249,42 @@ public void testBuildExpr_14() {
...
+    }
+    
+    @Test
+    public void testBuildTable_04() {
+        // Can't test for equality because can't create a BNode in a way that equality will
+        // succeed because OpTable does strict equality and ignores NodeIsomorphismMap
+        SSE.parseOp("(table (vars ?x) (row (?x _:test)))");
+    }
+
+    @Test(expected = ExprBuildException.class)
```
rvesse: For some reason `expected` is not working as expected ;)

rvesse: Apparently it's because we extend `TestCase` so JUnit assumes a JUnit 3.8 test which doesn't support expected, adding `@RunWith(JUnit4.class)` to the test class resolves this

See https://stackoverflow.com/a/50867198/107591

PR #442, opened 2018-06-28
JENA-1568: Tidy SSE tests - don't use TestCase or BaseTest

afs:
.. and nearby tests.


afs:
Added clean-up of other uses of JUnit3 TestCase in jena-arq - see discussion on https://issues.apache.org/jira/browse/JENA-1568.

PR #447, opened 2018-07-12
JENA-1574: Resolve named graph URIs.

afs:


kinow:
Code looks good, and I think this is for a recent e-mail thread in one of the mailing lists. Are you going to add any unit tests as well? Not sure if it's easy/doable to test that part as well. Otherwise will approve as-is.

afs:
Yes - it is a bit tricky to do deep testing. This is partially offset by the fact the command line tools are ways to call the core functionality so are small. I think most commands can now run "in-JVM" and they don't call e.g. `System.exit` but I'm not sure how uniform this is.

PR #448, opened 2018-07-12
JENA-1575: tdbloader - unnecessarily specific class cast.

afs:


jena-tdb/src/main/java/org/apache/jena/tdb/TDBLoader.java
```diff
@@ -219,11 +211,14 @@ public final void setGenerateStats(boolean generateStats)
 
     // These are the basic operations for TDBLoader.
 
-    private static void loadGraph$(GraphNonTxnTDB graph, InputStream input, boolean showProgress, boolean collectStats) {
+    private static void loadGraph$(GraphTDB graph, InputStream input, boolean showProgress, boolean collectStats) {
+        
+        DatasetGraphTDB dsgtdb = TDBInternal.getBaseDatasetGraphTDB(graph.getDatasetGraphTDB());
```
ajs6f: Just for my education, why is `getBaseDatasetGraphTDB` coming in here? What case would it meet that `graph.getDatasetGraphTDB()` by itself wouldn't?

afs: In TDB1, `DatasetGraphTDB` has refs to the indexes and node tables. The TDB1 transaction mechanism is to have different, layered indexes and node tables so a `DatasetGraphTDB` can be holding either the base storage indexes/node table or the in-transaction indexes/node table.

`getDatasetGraphTDB` is the base if non-transactional and transaction version if transactional. i.e it switches.

The loader wants the base storage version.

PR #449, opened 2018-07-27
JENA-1578

GregAlbiston:
- Subclass ValueReplacement and methods added to allow substitution of varNames with collections of RDFNodes. This is used in SPARQL queries to provide inline data with the VALUES keyword.
- Supports multiple values for a single variable, sets of values for multiple variables and multiple sets of values for multiple values.
- Optional inclusion of parenthesis for the first two cases.
- Tests included for use cases.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
+++ b/jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
@@ -1744,7 +1744,7 @@ public class ParameterizedSparqlString implements PrefixMapping {
      * Assign a VALUES varName with a multiple items.<br>
      * Can be used to assign multiple values to a single variable or single
      * value to multiple variables (if using a List) in the SPARQL query.<br>
-     * See setGroupedValues to assign multiple values to multiple variables.<br>
+     * See setRowValues to assign multiple values to multiple variables.<br>
      * Using "var" with list(prop_A, obj_A) on query "VALUES (?p ?o) {?var}"
      * would produce "VALUES (?p ?o) {(prop_A obj_A)}".
      *
@@ -1774,7 +1774,7 @@ public class ParameterizedSparqlString implements PrefixMapping {
      * Sets a map of VALUES varNames and their items.<br>
      * Can be used to assign multiple values to a single variable or single
      * value to multiple variables (if using a List) in the SPARQL query.<br>
-     * See setGroupedValues to assign multiple values to multiple variables.
+     * See setRowValues to assign multiple values to multiple variables.
      *
      * @param itemsMap
      */
@@ -1789,11 +1789,11 @@ public class ParameterizedSparqlString implements PrefixMapping {
      * (prop_B obj_B)}".
      *
      * @param varName
-     * @param groupedItems
+     * @param rowItems
      */
-    public void setGroupedValues(String varName, Collection<List<? extends RDFNode>> groupedItems) {
-        groupedItems.forEach(collection -> collection.forEach(item -> validateParameterValue(item.asNode())));
-        this.valuesReplacements.put(varName, new ValueReplacement(varName, groupedItems, true));
+    public void setRowValues(String varName, Collection<List<? extends RDFNode>> rowItems) {
+        rowItems.forEach(collection -> collection.forEach(item -> validateParameterValue(item.asNode())));
+        this.valuesReplacements.put(varName, new ValueReplacement(varName, rowItems, true));
     }
 
     private String applyValues(String command) {
@@ -1822,21 +1822,6 @@ public class ParameterizedSparqlString implements PrefixMapping {
         return targetVars;
     }
 
-    protected static boolean checkParenthesis(String command, String varName) {
-        boolean isNeeded;
-
-        int varIndex = command.indexOf(varName);
-        if (varIndex > -1) {
-            String subCmd = command.substring(0, varIndex).toLowerCase(); //Truncate the command at the varName. Lowercase to search both types of values.
-            int valuesIndex = subCmd.lastIndexOf(VALUES_KEYWORD);
-            int parenthesisIndex = subCmd.indexOf("(", valuesIndex + VALUES_KEYWORD.length());
-            isNeeded = parenthesisIndex > -1;
-        } else {
-            isNeeded = false;
-        }
-        return isNeeded;
-    }
-
     /**
      * Performs replacement of VALUES in query string.
      *
@@ -1845,26 +1830,26 @@ public class ParameterizedSparqlString implements PrefixMapping {
 
         private final String varName;
         private final Collection<? extends RDFNode> items;
-        private final Collection<List<? extends RDFNode>> groupedItems;
-        private final Boolean isGrouped;
+        private final Collection<List<? extends RDFNode>> rowItems;
+        private final Boolean isRows;
 
         public ValueReplacement(String varName, Collection<? extends RDFNode> items) {
             this.varName = varName;
             this.items = items;
-            this.groupedItems = new ArrayList<>();
-            this.isGrouped = false;
+            this.rowItems = new ArrayList<>();
+            this.isRows = false;
         }
 
-        public ValueReplacement(String varName, Collection<List<? extends RDFNode>> groupedItems, Boolean isGrouped) {
+        public ValueReplacement(String varName, Collection<List<? extends RDFNode>> rowItems, Boolean isRows) {
             this.varName = varName;
             this.items = new ArrayList<>();
-            this.groupedItems = groupedItems;
-            this.isGrouped = isGrouped;
+            this.rowItems = rowItems;
+            this.isRows = isRows;
         }
 
         public String apply(String command) {
 
-            if (items.isEmpty() && groupedItems.isEmpty()) {
+            if (items.isEmpty() && rowItems.isEmpty()) {
                 return command;
             }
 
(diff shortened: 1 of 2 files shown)


jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
```diff
@@ -1734,4 +1740,243 @@ public String toString() {
...
+        setValues(varName, Arrays.asList(item), false);
+    }
+
+    /**
+     * Sets a map of varNames and their items.
+     *
+     * @param valuesItems
+     */
+    public void setValues(Map<String, Collection<? extends RDFNode>> valuesItems) {
+        for (String varName : valuesItems.keySet()) {
```
ajs6f: Just a style thing, really, but I believe this method could be just `valuesItems.forEach(this::setValues)`.

jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
```diff
@@ -1734,4 +1740,243 @@ public String toString() {
...
+    /**
+     * All varNames in the map will use the same approach to parenthesis.
+     *
+     * @param valuesItems
+     * @param isParenthesisNeeded
+     */
+    public void setValues(Map<String, Collection<? extends RDFNode>> valuesItems, Boolean isParenthesisNeeded) {
+        for (String varName : valuesItems.keySet()) {
+            Collection<? extends RDFNode> items = valuesItems.get(varName);
+            setValues(varName, items, isParenthesisNeeded);
```
ajs6f: Similarly, `valuesItems.forEach((varName, item)->setValues(varName, items, isParenthesisNeeded))`, and the same below.

afs:
@ajs6f Will take a look.

A particular thing to be careful of with ParameterizedSparqlString in general is injection attacks.

cc @rvesse

jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
```diff
@@ -1734,4 +1739,237 @@ public String toString() {
...
+
+        /**
+         * Insert the SPARQL representation of the RDF node.
+         *
+         * @param item
+         * @return
+         */
+        private String createInsert(RDFNode item) {
+            String insert;
+            if (item.isLiteral()) {
```
afs: Literals with language tags need treating specially. Lang tag literals have datatype rdf:langString, which should not be printed, and the @lang must be printed. 

Literal lexical forms need escaping for any ", newline and tabs in the string.

Other forms could be made to be more natural : strings without "^^xsd:string", integers as `123` etc.

The good news is that there is code to help at the node level so the literal and URI cases can be replaced by `FmtUtils.stringForNode(rdfNode.asNode(), (PrefixMapping)null)` - must use the "(PrefixMapping)null" form otherwise a helpful (for display) prefix mapping is used. This is the code used by the SPARQLserializer.

This includes the delimiters e.g `<...>` or string `"..."`.

The URI may need care in case "`...> finish URI early + something to follow`" is injected.

jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
```diff
@@ -1734,4 +1739,237 @@ public String toString() {
...
+                Literal lit = item.asLiteral();
+                insert = "\"" + lit.getLexicalForm() + "\"^^" + lit.getDatatypeURI();
+            } else if (item.isResource()) {
+                insert = "<" + item.asResource().getURI() + ">";
+            } else {
+                insert = item.asResource().getId().getLabelString();
```
afs: Doesn't this need the "_:" put in? A test case will pin this down.

Can the label string be an injection attack with a space or newline to end the label? Probably need to restrict labels to `alpha-alphanumeric*`. (Mostly checking happens in the parsers in tokensing which is pretty heavy for this, though there may be code somewhere else that validates blank node labels.)

GregAlbiston:
I've made updates to try and address the comments made so far. 

- Now using _FmtUtils.stringForNode_ for conversion of RDFNode to replacement string.
- Local method _validateParameterValue_ now used as values are being set to prevent injection attack.
- Local method _validateSafeToInject_ now used when the query is being parsed to prevent injection attack. This is called each of the target variables for each relevant item. i.e. Not the varName supplied for the substitution but the variable in the VALUES clause of the query that will be evaluated.
- The need for parenthesis is now determined when the query is being parsed based on number of target variables in the VALUES clause and then, for a single target variable, the presence of parenthesis in the query.
- Updated tests, removed methods no longer required and added additional JavaDoc comments.

Thanks,

Greg

afs:
There is no need to use the un-parenthesed form - it is a syntactic short form and the full form can always be used. This may make the code generation easier and have less at risk of misalignment.
```
SELECT * {
       VALUES (?x) { (1) (2) (3) }
       }
```
The query formatter does much the same - if it one variable, it uses the short form, but it does not worry about the data being mis aligned:

It is the case of (illegal) `VALUES ?x { (1 2) (3) }` becoming treated as `VALUES ?x { 1 2 3 }`.

jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
```diff
@@ -1734,4 +1739,250 @@ public String toString() {
...
+                } else {
+                    if (targetVars.length > 1) {
+                        if (items instanceof List) {
+                            //Multiple vars with items in an ordered list. Each var is checked against the item.
+                            List<? extends RDFNode> listItems = (List<? extends RDFNode>) items;
+                            RDFNode item = listItems.get(i);
+                            validateSafeToInject(command, targetVar, item.asNode());
+                        } else {
+                            //Multiple vars with items not in an ordered list. This is parsing error.
+                            throw new ARQException("Multiple VALUES variables (" + String.join(", ", targetVars) + ") being used without an ordered list of items: " + items.toString());
```
rvesse: Can this error be caught sooner at the time the user makes the original `setValues()` call?

jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
```diff
@@ -1734,4 +1739,250 @@ public String toString() {
...
+     * See setGroupedValues to assign multiple values to multiple variables.<br>
+     * Using "var" with list(prop_A, obj_A) on query "VALUES (?p ?o) {?var}"
+     * would produce "VALUES (?p ?o) {(prop_A obj_A)}".
+     *
+     *
+     * @param varName
+     * @param items
+     */
+    public void setValues(String varName, Collection<? extends RDFNode> items) {
+        items.forEach(item -> validateParameterValue(item.asNode()));
```
rvesse: Later code requires that `items` be a `List` so this code should either enforce it here or otherwise sanitise `items` into a `List` at this point

jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
GregAlbiston: This error isn't relevant anymore as the `List` is being enforced by the `setValues` method. A check is performed that the number of target vars equals the number of 'items' in a 'row' with exception thrown. The check is performed late as the 'command' is not final in the class and can be changed using `setCommandText`. Could be performed earlier but the existing approach is to wait until `toString` is called.

jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
```diff
@@ -1734,4 +1739,250 @@ public String toString() {
...
+            int bracesIndex = subCmd.lastIndexOf("{");
+            String vars = command.substring(valuesIndex + VALUES_KEYWORD.length(), bracesIndex);
+            targetVars = vars.replaceAll("[(?)]", "").trim().split(" ");
+        } else {
+            targetVars = new String[]{};
+        }
+        return targetVars;
+    }
+
+    protected static boolean checkParenthesis(String command, String varName) {
```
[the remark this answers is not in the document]
GregAlbiston: Parenthesis are now always applied.

jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
```diff
@@ -1734,4 +1739,250 @@ public String toString() {
...
+
+    protected static String[] extractTargetVars(String command, String varName) {
+        String[] targetVars;
+
+        int varIndex = command.indexOf(varName);
+        if (varIndex > -1) {
+            String subCmd = command.substring(0, varIndex).toLowerCase(); //Truncate the command at the varName. Lowercase to search both types of values.
+            int valuesIndex = subCmd.lastIndexOf(VALUES_KEYWORD);
+            int bracesIndex = subCmd.lastIndexOf("{");
+            String vars = command.substring(valuesIndex + VALUES_KEYWORD.length(), bracesIndex);
```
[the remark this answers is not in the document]
GregAlbiston: Additional checking has been added to catch missing `VALUES` or braces. The query will be unchanged if these or the valueName are missing. The existing design is that syntax errors are caught by the `Query` or `UpdateRequest` stage.

jena-arq/src/test/java/org/apache/jena/query/TestParameterizedSparqlString.java
```diff
@@ -1925,4 +1926,190 @@ public void test_param_string_bug_04() {
...
+        String cmd = "SELECT * WHERE { VALUES ?p {?props} VALUES ?o {?objs} ?s ?p ?o }";
+        String varName = "objs";
+        String[] res = ParameterizedSparqlString.extractTargetVars(cmd, varName);
+        String[] exp = new String[]{"o"};
+
+        //System.out.println("Exp: " + exp);
+        //System.out.println("Res: " + res);
+        Assert.assertArrayEquals(exp, res);
+    }
+
```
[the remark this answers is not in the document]
GregAlbiston: Test cases added for incorrect number of values in a row for the variables.

Query that does not contain the `VALUES` keyword will now be unchanged. The existing design is that syntax errors are caught by the `Query` or `UpdateRequest` stage.

jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
```diff
@@ -1734,4 +1739,250 @@ public String toString() {
...
+    /**
+     * Allocate multiple lists of variables to a single VALUES varName.<br>
+     * Using "vars" with list(list(prop_A, obj_A), list(prop_B, obj_B)) on query
+     * "VALUES (?p ?o) {?vars}" would produce "VALUES (?p ?o) {(prop_A obj_A)
+     * (prop_B obj_B)}".
+     *
+     * @param varName
+     * @param groupedItems
+     */
+    public void setGroupedValues(String varName, Collection<List<? extends RDFNode>> groupedItems) {
```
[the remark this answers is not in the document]
GregAlbiston: 'setGroupedValues' now renamed to 'setRowValues'. Replaced 'group' with 'row' and 'varName' with 'valueName'.

PR #456, opened 2018-08-05
JENA-1585: Fuseki core reorg

afs:
Class reorganisation within the the jena-fuseki-core module.

jena-fuseki-embedded and jena-fuseki-basic changes show the cleaner split - no mention of the webapp or mgt packages where the webapp code now resides.


jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/build/FusekiBuildLib.java
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.jena.fuseki.webapp.SystemState;
+import org.apache.jena.query.* ;
+import org.apache.jena.rdf.model.Literal ;
+import org.apache.jena.rdf.model.Model ;
+import org.apache.jena.rdf.model.RDFNode ;
+import org.apache.jena.rdf.model.Resource ;
+import org.apache.jena.shared.PrefixMapping ;
+import org.apache.jena.vocabulary.RDFS ;
+
+public class FusekiBuildLib {
```
kinow: Is `FusekiBuildLib` going to replace `FusekiLib`? Would it be worth a one/two lines comment explaining what each one does? Just so one does not create a method in one, while the method was supposed to be in the other class... 

From a quick peek at the new classes, I think `FusekiLib` is for methods used by  Fuseki server, and `FusekiBuildLib` is more for handling nodes/resources. In which case maybe `FusekiLib#addDataInto` methods sit in between, but maybe could be in `FusekiBuildLib` too?

jena-fuseki2/jena-fuseki-core/src/test/java/org/apache/jena/fuseki/package-info.java
```diff
@@ -16,38 +16,17 @@
...
-    public void close() { /* .close() ;*/ }
-	@Override
-    public void flush() { /* .flush() ;*/ }
-
-	// Need to implement this one.
-	@Override
-    public void write(int b) { /* .write(b) ;*/ }
-	@Override
-    public void write(byte b[]) { /* this.write(b, 0, b.length) ; */}
+ * This package has the Fuseki full server tests.  ServerCtl manages a full server for tetsing.
```
kinow: s/tetsing/testing

jena-fuseki2/jena-fuseki-embedded/src/main/java/org/apache/jena/fuseki/embedded/FusekiServer.java
```diff
@@ -246,6 +248,13 @@ public Builder enableStats(boolean withStats) {
             return this;
         }
 
+        /** Add the "/$/ping" servlet that responds to HTTP very efficiently.
+         * This is useful for testign whether a server is alove, for example, from a load balencer.
```
kinow: s/testign/testing & s/balencer/balancer

kinow:
Great stuff! Noticed a few misspellings, and also had a question about when to use the new  `FusekiBuildLib`. But other than that looks great! Thanks for the new comments, and removing unnecessary code (e.g. that one that simply did a `return;`). :tada: 

I also realized that I submitted my review at the same time you updated the branch with more commits I think (at least I received an e-mail from GitHub right after that). So apologies if my comments are now out of sync. Thanks

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/build/FusekiBuildLib.java
afs: Good idea. And, yes, it is the build-specific operations.

Done.

afs:
No problem on timing - I found I'd tripped myself up moving java resources around and tested against a setup that had old and new location materials.  TravisCI caught this for me.

PR #458, opened 2018-08-10
JENA-1481, JENA-1586: Delete databases and expel from JVM

afs:


jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
```diff
@@ -364,4 +369,31 @@ public static String uniqueFilename(String directory, String base, String ext) {
...
     }
+    
+    /** Delete everything from a {@code Path} start point, including the path itself.
+     * This function works on files or directories.
+     * This function does not follow symbolic links.
+     */  
+    public static void deleteAll(Path start) {
+        // Walks down the tree and deletes directories on the way backup.
+        try { 
+            Files.walkFileTree(start, new SimpleFileVisitor<Path>() {
```
ajs6f: Could this `new SimpleFileVisitor` be broken out as a `static` inner class?

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/mgt/ActionDatasets.java
```diff
@@ -288,28 +294,89 @@ protected void execDeleteItem(HttpAction action) {
...
+                if ( tryToRemoveFiles && Files.exists(pDatabase)) {
+                    try { 
+                        if ( Files.isSymbolicLink(pDatabase)) {
+                            action.log.info(format("[%d] Database is a symbolic link, not removing files", action.id, pDatabase)) ;
+                        } else {
+                            IO.deleteAll(pDatabase);
+                            action.log.info(format("[%d] Deleted database files %s", action.id, pDatabase)) ;
+                        }
+                    } catch (RuntimeIOException e) {
+                        e.printStackTrace();
```
ajs6f: Should this be sent to a logger?

jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
ajs6f: I'm not worried about it. I'm always a bit suspicious on anon inner classes because of the occasional GC problems, but I think I'm letting my suspicions run away from me.

PR #460, opened 2018-08-11
JENA-1590: fix if statement in RDFWriter#output

kinow:
Eclipse alerted for some Dead code in this class, and after having a look, I thought we actually had to check the `lang` in the second `if`, not `ct`?

Very hard to write a unit test, as the map backing the languages registry will pretty much always return a language, once you have the content type (I think).


kinow:
Merging (amended my e-mail in the commit, replacing the no-reply from GitHub by my apache.org one)

PR #461, opened 2018-08-11
JENA-1591: use boolean value in if expression

kinow:
Another one found in Eclipse while I was reading some other code for UUID's. The value returned from the `Set` object will be a `boolean`, so the `== null` won't ever eval to true.

This pull request uses the expression `! x` as that's what I thought we actually wanted to test. But happy to update it if I made the wrong assumption.

Could not find an easy way to generate a unit test, but if there's some simple approach, please let me know and I will update the pull request.

Thanks
Bruno


kinow:
Rebased and amended to use my apache e-mail in the commits. Merging. Thanks!!!

PR #463, opened 2018-08-17
Cleanup

afs:
This is a collection of small items, accumulated while working with the source code attached to a project.

* Various comments/javadoc added and fixed and other code cleanup.
* make subclassing of the Fuseki query servlet better
* building Fuseki embedded using the command line syntax
* deprecate TDB ops "isBackedBy" in favour of "isTDB" for better naming.

I've left it as each individual commit for the details.


jena-arq/src/main/java/org/apache/jena/sparql/util/NodeUtils.java
```diff
@@ -18,12 +18,9 @@
 
 package org.apache.jena.sparql.util;
 
-import java.util.Collection ;
-import java.util.HashSet ;
-import java.util.Iterator ;
-import java.util.Objects;
-import java.util.Set ;
+import java.util.*;
```
kinow: Isn't it a bit better to keep what we have in import statements, expanding to what's used instead of importing the whole package?

jena-arq/src/main/java/org/apache/jena/sparql/util/NodeUtils.java
```diff
@@ -105,15 +102,23 @@ public String accept(Node x) {
...
         return conv ;
     }
-
-    /** Convert IRI String to Node */  
-    public static Set<Node> convertToNodes(Collection<String> uris) {
-        Set<Node> nodes = new HashSet<>() ;
-        for ( String x : uris )
-            nodes.add(NodeFactory.createURI(x)) ;
-        return nodes ;
+
```
kinow: Nit-pick! Some spaces were left behind.

jena-base/src/main/java/org/apache/jena/atlas/lib/ListUtils.java
```diff
@@ -34,49 +34,40 @@
...
+    public static <T> List<T> unique(List<T> list) {
+        return toList(list.stream().distinct());
     }
-    
-    public static
-    List<Integer> asList(int... values)
-    {
-        List<Integer> x = new ArrayList<>(values.length) ;
+
+    public static List<Integer> asList(int...values) {
```
kinow: Couldn't we replace this method by something like `Arrays.asList(1, 2, 3);` ? Or maybe deprecate it?

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ServiceRouter.java
```diff
@@ -120,10 +120,12 @@ protected boolean allowQuads_RW(HttpAction action) {
...
 
-        // Test whether there is a configuration that allows this action as the operation
-        // given.
-        // Ignores the operation in the action (set due to parsing - it might be "quads"
-        // which is the generic operation when just the dataset is specificed.
+        /**
+         * Test whether there is a configuration that allows this action as the operation
+         * given. Ignores the operation in the action which is set due to parsing - it
+         * might be "quads" which is the generic operation when just the dataset is
+         * specificed.
```
kinow: s/specificed/specified

jena-fuseki2/jena-fuseki-embedded/src/main/java/org/apache/jena/fuseki/embedded/FusekiServer.java
```diff
@@ -20,10 +20,9 @@
 
 import static java.util.Objects.requireNonNull;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
```
kinow: Ditto first comment, about maintaining the current standard in the code (I think)

jena-fuseki2/jena-fuseki-embedded/src/main/java/org/apache/jena/fuseki/embedded/FusekiServer.java
```diff
@@ -455,30 +527,36 @@ private static void setMimeTypes(ServletContextHandler context) {
...
+
             if ( staticContentDir != null ) {
                 DefaultServlet staticServlet = new DefaultServlet();
                 ServletHolder staticContent = new ServletHolder(staticServlet);
                 staticContent.setInitParameter("resourceBase", staticContentDir);
                 context.addServlet(staticContent, "/");
             }
         }
-
+
```
kinow: Nit-picking again! Unnecessary spaces...

jena-base/src/main/java/org/apache/jena/atlas/lib/ListUtils.java
afs: The signature of `Arrays.asList` is  `List<T> asList(T... a)` and auto-boxing does not work when the argument is `int[]` or `int ...`).

jena-fuseki2/jena-fuseki-embedded/src/main/java/org/apache/jena/fuseki/embedded/FusekiServer.java
kinow: >simply turning on the save action to remove trailing spaces is likely to cause such formatting changes on files with functional changes.

Agreed, I used the save action in Eclipse for a while and regretted later.

>Previously, keeping formatting changes separate from functional changes has been requested. (I'm neutral.)

I'm ok with either too. But my point here was that this case wasn't like the save action... at least from looking at the github diff, it appears there was no spaces before, and these were added in this commit only. But as I said, nit-pick. And nit-picks like this one can cause distraction, so feel free to ignore or hide :-)

jena-base/src/main/java/org/apache/jena/atlas/lib/ListUtils.java
kinow: Oh, my mistake :) :+1:

jena-fuseki2/jena-fuseki-embedded/src/main/java/org/apache/jena/fuseki/embedded/FusekiServer.java
afs: These leading space arise in Eclipse. After a closing "}" the cursor is padded to the same depth as the "}". Is there a way to change that without using a save action?

jena-arq/src/main/java/org/apache/jena/sparql/core/DynamicDatasets.java
```diff
@@ -56,8 +55,8 @@ public static DatasetGraph dynamicDataset(DatasetDescription description, Datase
     	if (description.isEmpty() )
     		return dsg;
     	
-        Set<Node> defaultGraphs = NodeUtils.convertToNodes(description.getDefaultGraphURIs()) ; 
-        Set<Node> namedGraphs = NodeUtils.convertToNodes(description.getNamedGraphURIs()) ;
+        Collection<Node> defaultGraphs = NodeUtils.convertToNodes(description.getDefaultGraphURIs()) ;
```
rvesse: Isn't the point of using sets here to remove duplicates or is that now done in the `NodeUtils.convertToNodes()` method?

afs: There are various places duplicate removal may happen but it is confusing and with all the routes may be depending on here. Better to maintain compatibility.

PR #464 restores the explicit behaviour by having "to list" and "to set" versions and a deprecated "NodeUtils.convertToNodes" with the Set semantics.

Does that look good?

PR #470, opened 2018-09-17
Rename README to README.md

nirajpatel:


ajs6f:
I'm not sure that our `README` actually _is_ Markdown. Is there a reason you would particularly like to see this file ending? Perhaps some further integration?

nirajpatel:
The only reasoning behind renaming is to have it render nicely in the GitHub UI and IntelliJ

Before:
https://github.com/apache/jena/blob/master/README

After: https://github.com/apache/jena/blob/5f6fd0cb710420ce17e87c28236bbfa056a485e3/README.md

ajs6f:
That does look more pleasant! I'll wait a few days in case there's some reason this is a bad idea that I'm not seeing, and then I'll merge. Thanks for the contribution @nirajpatel !

afs:
@nirajpatel  - done - thanks!

PR #473, opened 2018-10-01
Extendable versions of classes for printing result sets

hartig:
This PR modifies the classes `CSVOutput` and `TSVOutput` to make it possible to extend them and override their behavior in subclasses. I want to implement such subclasses for my SPARQL* extension of Jena.

A related question: Is there a reason why the mapping from `ResultsFormat` objects to `ResultSetLang`object (i.e., the [`mapResultsFormatToLang` in `ResultsFormat.java`](https://github.com/apache/jena/blob/aef35ee50eef9e0820d23d5ba184906b177a5c15/jena-arq/src/main/java/org/apache/jena/sparql/resultset/ResultsFormat.java#L190)) does not map `ResultsFormat.FMT_TEXT` to `ResultSetLang.SPARQLResultSetText`?
Without this, the command line program `arq.query` falls back to the old `ResultSetWriter` and there is no elegant way for my SPARQL* extension of this program to use my SPARQL*-aware version of `ResultSetWriter` for text output. Hence, it would be great to add the following line into code that sets up `mapResultsFormatToLang`.

```
mapResultsFormatToLang.put(ResultsFormat.FMT_TEXT, ResultSetLang.SPARQLResultSetText) ;
```

Alternatively, the could be a method that can be used for adding something to `mapResultsFormatToLang`.

If you are fine with any of these two options, I would like to add it to this PR.


afs:
On the question of mapping `ResultsFormat.FMT_TEXT`: IIRC it is because text is only an output format and not an input format (but then CSV is also only an output format).

When Jena 3.9.0 is completed, I can come back and look at this PR. In principle, there is no reason not have have them extensible.

What syntax are you using for TSV extensions?

hartig:
> When Jena 3.9.0 is completed, I can come back and look at this PR. In principle, there is no reason not have have them extensible.

Sounds good. Thanks!

> What syntax are you using for TSV extensions?

My idea was to simply extend the TSV format that you have in Jena with a possibility to also represent `Node_Triple` objects (as solution mappings in SPARQL* may map variables not only to URIs, literals, and bnodes, but also to triples).

> On the question of mapping `ResultsFormat.FMT_TEXT`: IIRC it is because text is only an output format and not an input format (but then CSV is also only an output format).

I see. If you are hesitant to add `ResultsFormat.FMT_TEXT` to `mapResultsFormatToLang` directly in the Jena code base, my alternative proposal of introducing a method for adding something to `mapResultsFormatToLang` from external code (like mine) would work for me too.

afs:
Recorded as https://issues.apache.org/jira/browse/JENA-1640.

afs:
Sorry this has taken so long to get round to.

afs:
I got sidetracked :-) last time.  The code in this area is ancient.

Text format isn't like the other formats because (1) it is isn't self-contained, It needs prefixes for abbreviation and (2) there isn't an input version.

Where are you wanting to output text format?

`QueryExecutils.outputResultSet` handles it - making `ResultSetFormatter.output(OutputStream, ResultSet, ResultsFormat)` do the same (lookup properly for first-class results formats, and have some adapter code to handle the other ways) is one way. Would that work or you?

hartig:
> Where and how are you wanting to output text format?

Where? In the command line tool for executing SPARQL* queries (see [`ExecuteSPARQLStar` in RDFstarTools](https://github.com/RDFstar/RDFstarTools/blob/master/src/main/java/se/liu/ida/rdfstar/tools/ExecuteSPARQLStar.java)).

How? As it is done by the `TextOutput` class in Jena, plus added support for the case in which the `Node` for a variable is of type `Node_Triple` (which is possible in results of SPARQL* queries). In this case, the triple wrapped by that `Node_Triple` should be printed Turtle*-style, i.e., enclosed within `<<` and `>>`. I have already implemented the necessary extension of the `TextOutput` class: see [`TextOutputStar` in RDFstarTools](https://github.com/RDFstar/RDFstarTools/blob/master/src/main/java/se/liu/ida/rdfstar/tools/sparqlstar/resultset/TextOutputStar.java).

> `QueryExecutils.outputResultSet` handles it - making `ResultSetFormatter.output(OutputStream, ResultSet, ResultsFormat)` could do the same (lookup properly for first-class results formats, and have some adapter code to handle the other ways) is one way. Would that work or you?

Not really I think. I am indirectly using `QueryExecUtils.outputResultSet` already. That is, my command line tool for executing SPARQL* queries (`ExecuteSPARQLStar`) ends up in exactly that method, simply because `ExecuteSPARQLStar` extends `arq.query` (which uses `QueryExecUtils.executeQuery(Prologue, QueryExecution, ResultsFormat)`). Then, `QueryExecUtils.outputResultSet` calls `ResultSetFormatter.out(OutputStream, ResultSet, Prologue)`.
The problem is that all of these are static methods and there is no way for me to specify which `TextOutput` object should be used by `ResultSetFormatter.out(OutputStream, ResultSet, Prologue)`. I want it to use my `TextOutputStar` (see above) instead of `TextOutput`.

So, again, I think the easiest solution would be if my command line tool for executing SPARQL* queries (`ExecuteSPARQLStar`) could simply call a to-be-created method of `ResultsFormat` that allows my program to add something to `ResultsFormat.mapResultsFormatToLang`. Then, my program would simply add `ResultsFormat.FMT_TEXT` -> `ResultSetLang.SPARQLResultSetText` (which would be used only within the context of my program), and `ResultsWriter` would find and use my `ResultSetWriterFactory` (which is created and registered in [`ResultSetWritersSPARQLStar`](https://github.com/RDFstar/RDFstarTools/blob/master/src/main/java/se/liu/ida/rdfstar/tools/sparqlstar/resultset/ResultSetWritersSPARQLStar.java)). Hence, all that would be needed is to add the following code to the `ResultsFormat` class of Jena.
```
public static void addResultsFormatToLangPair(ResultsFormat fmt, Lang lang) {
       mapResultsFormatToLang.put(fmt, lang) ;
}
```
I believe such a method can be helpful for other use cases as well (e.g., assume someone wants to introduce a new result format language). What do you say?

afs:
Jena is already extensible for results formats. See `ResultSetWriterRegistry` and `ResultSetReaderRegistry`. I'm sorry there is old code around but, well, time ... . 

Adding `ResultsFormat.FMT_TEXT` to `mapResultsFormatToLang` breaks text output. The text format needs prefix mapping for abbreviation and that is bypassed if that change is made to `mapResultsFormatToLang`. Give it a try! For `<< >>` output, abbreviated is good!

It would be like:

    ResultsWriter.create()
        .lang(ResultSetLang.SPARQLResultSetText)
        .build()
        .write(System.out, resultSet);

If `ResultSetWriterRegistry.writerText` used the `Context` to get a optional prologue; that should work out if setting the prologue is rolled out into `arq.query`.

It might be better for the long term is to break up `query.queryExec` into steps with protected methods for the steps as well.

afs:
@hartig - see PR #507 ; that should allow you to `ResultSetWriterRegistry.register(SPARQLResultSetText, <YourCodeFactory>)` to set your own text output engine.

For grabbing the prologue, see:
https://github.com/apache/jena/pull/507/files#diff-6efebeadd02ab276d49ddc9bf99f0758R117

hartig:
@afs thanks for your clarification and for PR #507 ; yes, that will work!

PR #474, opened 2018-10-01
JENA-1609 Jena master does not build under JDK10

lewismc:
This PR addresses https://issues.apache.org/jira/browse/JENA-1609

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,10 @@ buildNumber.properties
 .project
 .settings
 .recommenders
-.metadata
+.metadat
+**/*/META-INF
+**/.externalToolBuilders/
+**/maven-eclipse.xml
 
 # Idea IntelliJ
 *.ipr
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,7 @@ buildNumber.properties
 .project
 .settings
 .recommenders
-.metadat
+.metadata
 **/*/META-INF
 **/.externalToolBuilders/
 **/maven-eclipse.xml
--- a/jena-arq/pom.xml
+++ b/jena-arq/pom.xml
@@ -216,7 +216,6 @@
             </group>
           </groups>
           <bottom>Licenced under the Apache License, Version 2.0</bottom>
-          <additionalOptions>-html5</additionalOptions>
         </configuration>
       </plugin>
     </plugins>
--- a/jena-arq/pom.xml
+++ b/jena-arq/pom.xml
@@ -216,6 +216,7 @@
             </group>
           </groups>
           <bottom>Licenced under the Apache License, Version 2.0</bottom>
+          <additionalOptions>-html5</additionalOptions>
         </configuration>
       </plugin>
     </plugins>
--- a/jena-elephas/jena-elephas-mapreduce/src/main/java/org/apache/jena/hadoop/rdf/mapreduce/filter/AbstractNodeTupleFilterMapper.java
+++ b/jena-elephas/jena-elephas-mapreduce/src/main/java/org/apache/jena/hadoop/rdf/mapreduce/filter/AbstractNodeTupleFilterMapper.java
@@ -28,9 +28,7 @@ import org.apache.jena.hadoop.rdf.types.AbstractNodeTupleWritable;
 /**
  * Abstract mapper implementation which helps in filtering tuples from the
  * input, derived implementations provide an implementation of the
- * {@link #accepts(TKey, T)}
- * 
- * 
+ * {@link #accepts(Object, AbstractNodeTupleWritable)}
  * 
  * @param <TKey>
  *            Key type
@@ -39,7 +37,6 @@ import org.apache.jena.hadoop.rdf.types.AbstractNodeTupleWritable;
  * @param <T>
  *            Writable tuple type
  */
-@SuppressWarnings("javadoc")
 public abstract class AbstractNodeTupleFilterMapper<TKey, TValue, T extends AbstractNodeTupleWritable<TValue>> extends
         Mapper<TKey, T, TKey, T> {
 
--- a/jena-jdbc/jena-jdbc-core/pom.xml
+++ b/jena-jdbc/jena-jdbc-core/pom.xml
@@ -39,17 +39,28 @@
 			<version>3.10.0-SNAPSHOT</version>
 		</dependency>
 
-    <!-- Jena Driver configures Log4j directly -->
-    <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-log4j12</artifactId>
-    </dependency>
+		<!-- Jena Driver configures Log4j directly -->
+		<dependency>
+			<groupId>org.slf4j</groupId>
+			<artifactId>slf4j-log4j12</artifactId>
+		</dependency>
 
-    <dependency>
-     <groupId>log4j</groupId>
-      <artifactId>log4j</artifactId>
-   </dependency>
-    
+		<dependency>
+			<groupId>log4j</groupId>
+			<artifactId>log4j</artifactId>
+		</dependency>
+
+		<dependency>
+			<groupId>org.apache.logging.log4j</groupId>
+			<artifactId>log4j-api</artifactId>
+		</dependency>
+
+		<!-- Intercept direct use of log4j2 -->
+		<dependency>
+			<groupId>org.apache.logging.log4j</groupId>
+			<artifactId>log4j-to-slf4j</artifactId>
+		</dependency>
+
 		<!-- Test Dependencies -->
 		<dependency>
 			<groupId>junit</groupId>
@@ -97,6 +108,7 @@
 					<encoding>UTF-8</encoding>
 					<charset>UTF-8</charset>
 					<docencoding>UTF-8</docencoding>
+					<additionalOptions>-html5</additionalOptions>
 				</configuration>
 				<executions>
 					<execution>
--- a/jena-jdbc/jena-jdbc-core/pom.xml
+++ b/jena-jdbc/jena-jdbc-core/pom.xml
@@ -108,7 +108,6 @@
 					<encoding>UTF-8</encoding>
 					<charset>UTF-8</charset>
 					<docencoding>UTF-8</docencoding>
-					<additionalOptions>-html5</additionalOptions>
 				</configuration>
 				<executions>
 					<execution>
--- a/jena-base/pom.xml
+++ b/jena-base/pom.xml
@@ -166,7 +166,6 @@
           <windowtitle>Apache Jena Base </windowtitle>
           <doctitle>Apache Jena Base ${project.version}</doctitle>
           <bottom>Licenced under the Apache License, Version 2.0</bottom>
-          <additionalOptions>-html5</additionalOptions>
         </configuration>
       </plugin>
       
--- a/jena-base/pom.xml
+++ b/jena-base/pom.xml
@@ -166,6 +166,7 @@
           <windowtitle>Apache Jena Base </windowtitle>
           <doctitle>Apache Jena Base ${project.version}</doctitle>
           <bottom>Licenced under the Apache License, Version 2.0</bottom>
+          <additionalOptions>-html5</additionalOptions>
         </configuration>
       </plugin>
       
--- a/jena-cmds/pom.xml
+++ b/jena-cmds/pom.xml
@@ -196,6 +196,7 @@
           <windowtitle>Apache Jena ARQ</windowtitle>
           <doctitle>Apache Jena ARQ ${project.version}</doctitle>
           <bottom>Licenced under the Apache License, Version 2.0</bottom>
+          <additionalOptions>-html5</additionalOptions>
         </configuration>
       </plugin>
 
--- a/jena-cmds/pom.xml
+++ b/jena-cmds/pom.xml
@@ -196,7 +196,6 @@
           <windowtitle>Apache Jena ARQ</windowtitle>
           <doctitle>Apache Jena ARQ ${project.version}</doctitle>
           <bottom>Licenced under the Apache License, Version 2.0</bottom>
-          <additionalOptions>-html5</additionalOptions>
         </configuration>
       </plugin>
 
--- a/jena-core/pom.xml
+++ b/jena-core/pom.xml
@@ -204,7 +204,6 @@
               <packages>org.apache.jena.assembler</packages>
             </group>
           </groups>
-          <additionalOptions>-html5</additionalOptions>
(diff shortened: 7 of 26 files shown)


.gitignore
```diff
@@ -14,7 +14,10 @@ buildNumber.properties
 .project
 .settings
 .recommenders
-.metadata
+.metadat
```
ajs6f: Is this correct? `.metadata` => `.metadat`?

jena-arq/pom.xml
```diff
@@ -216,6 +216,7 @@
             </group>
           </groups>
           <bottom>Licenced under the Apache License, Version 2.0</bottom>
+          <additionalOptions>-html5</additionalOptions>
```
ajs6f: Is this in some way needed for Java 10 builds?

jena-elephas/jena-elephas-mapreduce/src/main/java/org/apache/jena/hadoop/rdf/mapreduce/filter/AbstractNodeTupleFilterMapper.java
```diff
@@ -28,9 +28,7 @@
 /**
  * Abstract mapper implementation which helps in filtering tuples from the
  * input, derived implementations provide an implementation of the
- * {@link #accepts(TKey, T)}
- * 
- * 
+ * {@link #accepts(Object, AbstractNodeTupleWritable)}
```
ajs6f: This just looks wrong-- the method below clearly is `accepts(TKey key, T tuple)`, and the concrete class isn't there at all.

jena-jdbc/jena-jdbc-core/pom.xml
```diff
@@ -39,17 +39,28 @@
...
+			<groupId>org.slf4j</groupId>
+			<artifactId>slf4j-log4j12</artifactId>
+		</dependency>
+
+		<dependency>
+			<groupId>log4j</groupId>
+			<artifactId>log4j</artifactId>
+		</dependency>
+
+		<dependency>
```
ajs6f: Why are you pulling in the `log4j` API?

jena-arq/pom.xml
lewismc: If it is not included there is a ```WARNING``` message as follows 
```
javadoc: warning - You have not specified the version of HTML to use. The default is currently HTML 4.01, but this will change to HTML5 in a future release. To suppress this warning, please specify the version of HTML used in your documentation comments and to be generated by this doclet, using the -html4 or -html5 options.
```
The -html5 additional option cleans up presence of ```WARNING```'s during Maven build.

jena-elephas/jena-elephas-mapreduce/src/main/java/org/apache/jena/hadoop/rdf/mapreduce/filter/AbstractNodeTupleFilterMapper.java
lewismc: I agree... it was strange when it turned out that this seemed to be the fix. Without the code augmentation, I was unable to generate the Javadocs for the project. It should be noted that Javadoc generation is activated by default during an ```install``` target.

acoburn:
It is possibly worth noting here that these changes also make it possible to build Jena on JDK 11.

lewismc:
Thats good to know @acoburn thanks for chiming in.

jena-arq/pom.xml
afs: This setting in the parent POM. It is commented out because the build fails with Java8.

1. Is setting in the parent enough? If it does, it is better to have in one place because ...
2. Has this PR been run with Java8? The build exits with an error in jena-iri:javadoc for me.

afs:
See https://builds.apache.org/view/H-L/view/Jena/job/Jena_JDK11/ which builds with -Pdev and has been running for a while. It does not build jena-jdbc or jena-elephas both of which are need looking at.

jena-jdbc/jena-jdbc-core/pom.xml
lewismc: This being said, what would you rather see? This issue is titled _Jena master does not build under JDK10_ so I wonder if getting Jena to build under JDK 10 is the immediate response. Someone can then go and work on the underlying logging issue. Right now I have not investigated it enough or else I would have provided the fix.

lewismc:
Thanks @afs I looked at the build and built locally using the following parameters
```
mvn clean verify -Pdev
```
... it results in SUCCESSFUL build.

jena-jdbc/jena-jdbc-core/pom.xml
rvesse: `JenaDriver` can optionally configure Log4j, this is because JDBC drivers can be used in contexts where the end user can't directly configure logging e.g. when dropped into other applications.  In those contexts users can add `logging=<log4j-config>` to have the driver configure logging for them.

However the default behaviour is to not configure anything and all the code internally uses SLF4J so in environments where users can control logging configuration they are free to do that.  So in environments where the user fully controls things they can configure logging as appropriate.

jena-jdbc/jena-jdbc-core/pom.xml
```diff
@@ -39,17 +39,28 @@
...
+			<groupId>org.apache.logging.log4j</groupId>
+			<artifactId>log4j-api</artifactId>
+		</dependency>
+
+		<!-- Intercept direct use of log4j2 -->
+		<dependency>
+			<groupId>org.apache.logging.log4j</groupId>
```
rvesse: I think this is unnecessary, the only Log4j direct usage is to configure Log4j, the code itself always uses SLF4J for its logging

afs:
I'm trying to get a single build configuration for both java10 and java8.

I can't get Jena to build with java8 if `<additionalOptions>-html5</additionalOptions>` is set.

It is a problem in the full build because skipping javadoc isn't an option - the distributions include javadoc by dependency.

With java 10.0.2 (Ubuntu latest with the placeholder java11 - still waiting for real java11): jena-elephas-common fails because:
```
[ERROR] Failed to execute goal on project jena-elephas-common: Could not resolve dependencies for project org.apache.jena:jena-elephas-common:jar:3.10.0-SNAPSHOT: Could not find artifact jdk.tools:jdk.tools:jar:1.6 at specified path /usr/lib/jvm/java-11-openjdk-amd64/../lib/tools.jar -> [Help 1]
```
(it also used MRUnit which is a retired project)

A pure maven build will put the java11 runtime (no tools.jar) on the compile classpath.

@lewismc - am I right in concluding it worked for you in the IDE because the java8 runtime was on the build path?

rvesse:
@afs For some projects where I have added JDK 9/10 support I have had to create separate profiles activated by JDK version for the Javadoc plugin because there are breaking changes between the plugin versions and supported flags for the underlying `javadoc` tool

Also if you have updated the Javadoc plugin version to a 3.x release then you need to add an `<option>` sub-element within `<additionalOptions>` rather than just specifying the options directly

jena-jdbc/jena-jdbc-core/pom.xml
lewismc: Ok thanks @rvesse. The logic seems clear from what you wrote, it doesn't however take away from the fact that the codebase will not build under JDK10 (and possibly JDK9) without the presence of the log4j library containing org.apache.log4j.PropertyConfigurator as currently used within jena/jena-jdbc/jena-jdbc-core/src/main/java/org/apache/jena/jdbc/JenaDriver.java.
The issue here is to either refactor the logging or to add the dependency... I would think there is no other scenario under consideration.
Is that a fair statement?

jena-jdbc/jena-jdbc-core/pom.xml
lewismc: OK so we agree on this however the code doesn't build due to the missing dependency. I wasn't able to build without the additions as proposed.

lewismc:
@afs how about we propose the change to
```
<additionalOptions>-html4</additionalOptions>
```
This would produce JDK WARNING's under several environment's however would at least enable successful build under this environments.

lewismc:
I have not tried this build under JDK11 yet.

afs:
@lewismc 
> @afs how about we propose the change to
> 
> ```
> <additionalOptions>-html4</additionalOptions>
> ```
> This would produce JDK WARNING's under several environment's however would at least enable successful build under this environments.

The problem is not html4/html5 (we can clean up the javadoc) but compatible with java8 javadoc tool. Java8 build causes:

    Exit code: 1 - javadoc: error - invalid flag: -html4

I'd prefer that the build uses java8 for java8 libraries because cross-compiling does not catch changes in the runtime libraries.

afs:
@rvesse  - The current plugin in the build is 3.0.0. `<option>` didn't change the situation. Can you given an example or point to one? (the online plugin doc does not talk about `<option>`

Profiles: that would be great. Again, a sample to start from would help. While all the suggestions here are great, it's going to be slower if I have to carry out each one from scratch.

jena-jdbc/jena-jdbc-core/pom.xml
afs: I'm unclear here. I tried with JDK10 and had to add the log4j1 dependency but not the log4j2 dependencies.

In the code I can only find imports of `org.apache.log4j`, which is log4j1, not `org.apache.logging.log4j` which is log4j2.

Elasticsearch does use log4j2 and Jena adds adapters theer for that.

rvesse:
@afs https://github.com/rvesse/airline/blob/master/pom.xml#L495-L497 which I discovered via JIRA https://issues.apache.org/jira/browse/MJAVADOC-475

jena-jdbc/jena-jdbc-core/pom.xml
afs: More:

Confirmed I have only have to add a single log4j1 dependency.
The error is otherwise from the maven-javadoc-plugin, not the compiler:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.0.0:jar (attach-javadocs) on project jena-jdbc-driver-bundle: MavenReportException: Error while generating Javadoc: 
[ERROR] Exit code: 1 - /home/afs/Jena/jena-jdbc/jena-jdbc-core/src/main/java/org/apache/jena/jdbc/JenaDriver.java:42: error: package org.apache.log4j does not exist
[ERROR] import org.apache.log4j.PropertyConfigurator;
[ERROR]                        ^
[ERROR] 
[ERROR] Command line was: /usr/lib/jvm/default-java/bin/javadoc @options @packages
[ERROR] 
[ERROR] Refer to the generated Javadoc files in '/home/afs/Jena/jena-jdbc/jena-jdbc-driver-bundle/target/apidocs' dir.
[ERROR] 
```

lewismc:
@afs 
> The problem is not html4/html5 (we can clean up the javadoc) but compatible with java8 javadoc tool. Java8 build causes:

Got it.

> I'd prefer that the build uses java8 for java8 libraries because cross-compiling does not catch changes in the runtime libraries.

So do you want to abandon this PR and state that Jena should be used with JDK 1.8? If so I can close out this issue. Unless of course if you can propose an alternative.

jena-jdbc/jena-jdbc-core/pom.xml
lewismc: Agreed. This is the behavior I get as well. 
My thinking was that the dependency needed to be available at compile time as oppose to only runtime... but maybe this thinking is wrong. 
The generated Javadoc look fine but the error prevents a stable build.

lewismc: From what I understand, the offending class [org.apache.log4j.PropertyConfigurator](https://logging.apache.org/log4j/2.0/log4j-1.2-api/apidocs/index.html?org/apache/log4j/PropertyConfigurator.html) resides in log4j 2.0... the runtime dependencies are detailed [as follows](https://logging.apache.org/log4j/2.0/runtime-dependencies.html) with the [Maven Artifacts detailed here](https://logging.apache.org/log4j/2.0/maven-artifacts.html).

rvesse: This code predates Log4j 2.x so shouldn't have any dependency upon it

afs: Wrong PropertyConfigurator.java  - there are two, one in log4j1 and one in log4j2 for compatibility code `log4j-1.2-api/src/main/java/org/apache/log4j/PropertyConfigurator.java`.

Jena is picking up the one from log4j1. So the log4j2 dependencies aren't need but the log4j1 is. Either will do to compile but log4j1 is intended.

rvesse:
I think there is a simpler solution to this, I have Jena JDBC building locally under JDK10 and will push a new PR shortly

lewismc:
@rvesse grand

afs:
"should be used with JDK 1.8" has several facets.
There are different levels of Java11 usage: the project discussions back when it was Java9:

[dev@ message](https://lists.apache.org/thread.html/880632ba6b6df6f2b26e4357462d94426ea4b14d83f08e9edac0b684@%3Cdev.jena.apache.org%3E) and sent as part of [users@ message](https://lists.apache.org/thread.html/5b31a3b6794781f6a02658608444043b5535957a6c94f3728efd8304@%3Cusers.jena.apache.org%3E).

Ultimately, it is a community decision. 

My personal opinion is that Jena should be able to run on Java8 runtime. Research projects may upgrade quickly but enterprise deployments are another matter.

To ensure no accidents, I prefer the release build used Java8 because it protects against runtime library changes (we have one currently about tools.jar).

I'd like the build to work with Java11 - hence the commented line in the parent `pom.xml` and the Jenkins job, both of which came out of the project discussions from Feb. Currently, "-Pdev" works, the full release build does not.

I have `~/.mavenrc` setup to make sure I gets builds right. I use JDK10 as the default java on my system (I am resisting installing Java11 manually and waiting for Ubuntu ... at least for now!), I make sure maven runs with Java8.

     # ~/.mavenrc
    JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64

Java11 will change the ecosystem and we don't know how that plays out or what may come out of [adoptjava](https://adoptopenjdk.net/) and the Jan2019 deadline from Oracle. I think it will rapidly become clear.

rvesse:
@afs Agreed, many folks are going to migrate Java 8 -> Java 11 because of the LTS release status are are skipping interim versions

afs:
`10 < 11`.  All these issues are in Java11 because Java 9 & 10 were the big changes and Java 11 the consolidation.

lewismc:
Closing in favor of #476

PR #478, opened 2018-10-06
JENA-1612: A lightly parallel loader plan.

afs:


afs:
Includes [JENA-1616](https://issues.apache.org/jira/browse/JENA-1616). Working on the JENA-1612 "light" loader, I noticed the logging output was not right for multiple file loads.

PR #479, opened 2018-10-07
1515/fix

Claudenw:
closes JENA-1515


jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/rewriters/ElementRewriter.java
```diff
@@ -103,17 +103,13 @@ public void visit(ElementBind el) {
...
-				bindings.next(); // skip the binding
-			} else {
-				retval.add(v);
-				retval.add(rewrite(bindings.next()));
-			}
+		for (Var v : el.getVars()) {
+			retval.add(v);
 		}
+		for (Binding binding : el.getRows())
+		{
```
afs: Minor: different style. There is more of this later.

Claudenw: not sure what this comment means.  The style of the looping is different and has a different result.

afs: Code layout. Some of same-line { , some next line {

PR #490, opened 2018-11-16
JENA-1635 Fix invalid Automatic-Module-Name values

acoburn:
The jena-fuseki-access, jena-fuseki-core and jena-fuseki-main modules
have invalid data for Automatic-Module-Name.

This aligns the data from those modules with the existing fuseki2 module.

### DIFF
--- a/jena-fuseki2/jena-fuseki-access/pom.xml
+++ b/jena-fuseki2/jena-fuseki-access/pom.xml
@@ -31,7 +31,7 @@
   <packaging>jar</packaging>
   
   <properties>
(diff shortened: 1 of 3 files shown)


afs:
PR #473 touches on these modules. Do you mind if I merge this PR into PR #473?

And "fuseki", not "fuseki2" as Fuseki1 is now well and truly legacy only?

acoburn:
@afs feel free to pull this into #486. I used `fuseki2` because that's what was used [here](https://github.com/apache/jena/blob/master/jena-fuseki2/pom.xml#L57), but I'd also prefer just `fuseki`, as you suggest.

PR #491, opened 2018-11-16
merge of IndexErrorOption

xristy:
PR for JENA #1636

### DIFF
--- a/jena-text/src/main/java/org/apache/jena/query/text/TextIndexConfig.java
+++ b/jena-text/src/main/java/org/apache/jena/query/text/TextIndexConfig.java
@@ -28,6 +28,7 @@ public class TextIndexConfig {
     String queryParser;
     boolean multilingualSupport;
     boolean valueStored;
+    boolean ignoreIndexErrors;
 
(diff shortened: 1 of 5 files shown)


xristy:
I think I see why I was getting unmerged commits. I was needed to do a commit after the ``merge --squash``

afs:
Looking at other recent PRs, it looks like they are pulled from your BDRC repo then close the PR. This is why they are "closed", not "merged".

The documented workflow is to merge from the github PR - it is not proscriptive and if your way works for you, then fine. It's the same commits (same SHA1), different route.

Normally I do:

git pull github pull/NNN/head --no-ff
git push

as in your modified "11.". Then when the GH repo sync's up with Apache master (a few minutes at most), the PR here goes purple/merged and has a "delete branch" option to delete the branch in your repo that you created the PR from.

The "This closes \#NNN" text does not have to be in the first line of the commit message. 

If convenient, could you put "JENA-nnnn:" on the front of important commits or the merge commit message? Don't matter if you don't, just makes things a but more findable from just git logs in my experience.

PR #498, opened 2018-11-25
JENA-1639: Model.getList and Statement.getList

afs:
Attempted to use the local code style at each addition point.

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/rdf/model/ModelCon.java
+++ b/jena-core/src/main/java/org/apache/jena/rdf/model/ModelCon.java
@@ -111,16 +111,17 @@ public interface ModelCon {
  *     It is in effect an unsafe downcast.</p>
  *
  * <p>Subsequent operations on the returned Alt may modify this model.</p>
- * <p>The bag is assumed to already exist in the model.  If it does not,
+ * <p>The Alt is assumed to already exist in the model.  If it does not,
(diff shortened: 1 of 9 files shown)


ajs6f:
I'm probably confused, but where is `Model::getList` being added? I just see a change to `createList` and `getList` is only added to `ModelCon`.

jena-core/src/main/java/org/apache/jena/rdf/model/ModelCon.java
```diff
@@ -111,38 +111,62 @@
...
+ * <p> This method enables an application to treat any resource as a list.
+ *     It is in effect an unsafe downcast.</p>
+ *
+ * <p>Subsequent operations on the returned list may modify this model.</p>
+ * <p>The list is assumed to already exist in the model.  If it does not,
+ * <CODE>createList</CODE> should be used instead.</p>
+ * @return a list instance
+ * @param r the resource of the list
+ */ 
+   RDFList getList(Resource r) ;
```
afs: Here.

afs:
`interface Model extends ModelCon`

(I think to put many of the convenience wrappings in a different place i.e. `ModelCon`)

ajs6f:
Ah, I _always_  get `ModelCon` and `ModelCom` (impl) confused. Cool!

afs:
Ditto!

PR #500, opened 2018-11-27
JENA-1642: Check has value for header instead of is true

jsarjeant:
When using schemagen from the maven plugin the `header` option does not work. `header` is intended to be a template string and when `isTrue` is checked the string is cast to a boolean and the `BadBooleanException` is thrown. I believe we should check for existence here instead of "truthiness"

Also, it appears that the maven tools are no being built as part of the main project. Is that the case?

### DIFF
--- a/jena-cmds/src/main/java/jena/schemagen.java
+++ b/jena-cmds/src/main/java/jena/schemagen.java
@@ -1815,7 +1815,7 @@ public class schemagen {
         @Override
         public String getOutputOption() { return getStringValue( OPT.OUTPUT ); }
         @Override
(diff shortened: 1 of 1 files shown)


afs:
> it appears that the maven tools are no being built as part of the main project. Is that the case?

Yes, I'm afraid it is. Some tests fail when used with Apache parent >19 (I just checked with v21 - this is still the case).


```
[ERROR] The following builds failed:
[ERROR] *  schemagen-integration-1/pom.xml
[ERROR] *  schemagen-integration-4/pom.xml
[ERROR] *  schemagen-integration-2/pom.xml
[ERROR] *  schemagen-integration-0/pom.xml
[ERROR] *  schemagen-integration-3/pom.xml

```

PR #501, opened 2018-11-30
JENA-1643: Use transactional dataset in clear().

afs:
The fix is the one line change in method `clear()`.

### DIFF
--- a/jena-tdb/src/main/java/org/apache/jena/tdb/store/GraphTDB.java
+++ b/jena-tdb/src/main/java/org/apache/jena/tdb/store/GraphTDB.java
@@ -53,15 +53,15 @@ public abstract class GraphTDB extends GraphView implements Closeable, Sync {
     }
 
     private String graphNameStr() {
-        if ( getGraphName() == null ) return DatasetPrefixesTDB.unnamedGraphURI;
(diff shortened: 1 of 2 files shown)


jena-tdb/src/main/java/org/apache/jena/tdb/store/GraphTDB.java
```diff
@@ -171,7 +171,10 @@ protected final int graphBaseSize() {
 
     @Override
     public void clear() {
-        getDatasetGraphTDB().deleteAny(getGraphName(), Node.ANY, Node.ANY, Node.ANY) ;
+        // Logically, this is "super.clear()" except the default implementation
+        // is a loop that materializes nodes. We want to call the dataset directly 
+        // so that nodes don't get materialized, just deleted from indexes.
```
ajs6f: @afs Just for my education, what do you mean by "materialized" here?

rvesse: Loaded from the node table into nodes in the cache I think?

afs: Yes.

afs:
And review approvals?

PR #510, opened 2018-12-15
JENA-1649: Retire jena-fuseki1

afs:
Code removal, leaving only a `jena-fuseki1/README.md`.


afs:
Yes - there are various documentation places to clearup.

PR #514, opened 2018-12-19
JENA-1647: Query string, with restrictions, allowed by RFC 8141.

afs:


afs:
My bad!  I misread the spec then promptly implemented and tested my misunderstanding.

Two new tests added - the two examples in the spec.

Forced push to the PR as a replacement.

PR #519, opened 2019-01-04
JENA-1656: Make DatasetGraphMonitor transaction-aware.

pipcet:
This adds transaction information to the DatasetChanges interface (and uses default methods, is that okay?)

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/core/DatasetChanges.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/core/DatasetChanges.java
@@ -39,4 +39,15 @@ public interface DatasetChanges
     /** Release any resources */
     public void reset() ;
 
+    /** A transaction has begun. */
+    default void begin() {}
+
+    /** The current transaction has been committed. */
+    default void commit() {}
+
+    /** The current transaction has been aborted. */
+    default void abort() {}
+
+    /** The current transaction has been ended. */
+    default void end() {}
 }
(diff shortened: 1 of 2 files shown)


pipcet:
There are a few places in the tree that use `DatasetGraphMonitor` (or `DatasetChanges`), so it seemed like the best fit for my situation.

One problem I see with `RDFChanges` is that `txnCommit` is called before `super.commit`, so the change would not yet be visible to a thread that's woken up by `txnCommit`.

My current code also uses a new interface that monitors reads rather than writes, to determine which part of a graph has been looked at by a query (so we know when to repeat it to simulate a continuous query).

So there are probably three use cases:
- be informed that the graph has potentially changed and the changes, if any, are visible to other threads
- be informed that the graph is about to change, allowing you to veto transactions
- be informed that the graph is being looked at

I think combining the first two in `DatasetGraphMonitor`/`DatasetGraphChanges` makes sense, at least if we allow the relatively new use of default interface methods. I suspect different applications have different requirements for false positives, though.

pipcet:
I think it would indeed be best to have both handlers for the "inside" and "outside" case; as you said, your application needs the inside case; mine needs the outside case (or a custom DatasetWrapper).

I don't think the distinction is particularly useful for read transactions, though: those will probably have to deal with false positives anyway.

afs:
I'm not sure what you mean by a "false positive" here.

afs:
Sorry this has gone on for so long.

I went back to this and investigated the relationship to jena-text. It's hard to untangle because jena-text is involved in the transaction and coordinates with Lucene.

The contract for `DatasetChanges` is rather unclear - it predates full transaction support for all datasets.

I've raised [JENA-2071](https://issues.apache.org/jira/browse/JENA-2071) to deprecate `DatasetChanges` so it can be removed from general use, freeing up space to do an "outside" transaction monitor with new, definite contract.

If this is still relevant, or you have further thoughts, please add to the ticket [JENA-1656](https://issues.apache.org/jira/browse/JENA-1656).

PR #520, opened 2019-01-10
correct closing of http connections

vgordievskiy:
This fix for correct closing of http connections for the "describe" and "json" queries.


vgordievskiy:
@afs check please my changes (done with the try-with-resources)

> Was the effect to cause problems when the result data has a parse error or is truncated in someway?
No, I caught this problem as some freezing  of applications after 3 correct requests.

PR #522, opened 2019-01-18
Improve error message trying to use TDB2 location with TDB (JENA-1658)

rvesse:
Since the lock files have slightly different formats TDB can detect when a lock file appears to be TDB2 and issue an appropriate error message to suggest users try using TDB2 instead.

This should avoid user confusion as seen in JENA-1658

### DIFF
--- a/jena-tdb/src/test/java/org/apache/jena/tdb/base/file/TestLocationLock.java
+++ b/jena-tdb/src/test/java/org/apache/jena/tdb/base/file/TestLocationLock.java
@@ -22,10 +22,10 @@ import java.io.BufferedWriter;
 import java.io.FileWriter;
 import java.io.IOException;
 
-import org.apache.jena.tdb.TDBException ;
-import org.apache.jena.tdb.base.file.Location ;
-import org.apache.jena.tdb.base.file.LocationLock ;
-import org.apache.jena.tdb.sys.ProcessUtils ;
+import org.apache.jena.tdb.TDBException;
+import org.apache.jena.tdb.base.file.Location;
+import org.apache.jena.tdb.base.file.LocationLock;
+import org.apache.jena.tdb.sys.ProcessUtils;
 import org.junit.Assert;
 import org.junit.Assume;
 import org.junit.BeforeClass;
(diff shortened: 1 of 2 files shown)


jena-tdb/src/test/java/org/apache/jena/tdb/base/file/TestLocationLock.java
```diff
@@ -146,4 +146,34 @@ public void location_lock_dir_error_02() throws IOException {
...
+        }
+
+        // Trying to get the owner should error accordingly
+        try {
+            lock.canObtain();
+            Assert.fail("Expected the lock file to be considered invalid");
+        } catch (FileException e) {
+            String errMsg = e.getMessage();
+            Assert.assertNotNull(errMsg);
+            System.out.println(errMsg);
```
ajs6f: Is this `System.out.println` leftover from debugging?

rvesse: Guilty!

jena-tdb/src/test/java/org/apache/jena/tdb/base/file/TestLocationLock.java
```diff
@@ -90,8 +90,9 @@ public void location_lock_dir_02() throws IOException {
         Assert.assertTrue(lock.canObtain());
 
         // Write a fake PID to the lock file
-        try(BufferedWriter writer = new BufferedWriter(new FileWriter(dir.getPath("tdb.lock")))) {
-            writer.write(Integer.toString(-1234)); // Fake PID that would never be valid
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(dir.getPath("tdb.lock")))) {
+            writer.write(Integer.toString(-1234)); // Fake PID that would never
```
afs: Minor - line wrapped

afs:
The lock file is the only thing that stops TDB1 and TBD2 having the same directory! It wasn't "by design" either way round. But it is better to keep them separate as it get confusing. I can't see any value in both in the same directory.

PR #528, opened 2019-01-31
Allow direct loading of tuples without temporary tables

grahamtriggs:
Add a new interface to the TupleLoader that allows database engines with appropriate query constructs to bypass the use of a temporary table when adding data to the store.

Currently, this has only been implemented for MySQL using a layout2/hash, although in theory it could be possible to do it for other databases - e.g. Postgres 9.5+.

For a MySQL database, this reduces writing times by about 30%, (based on a local 5.7 instance running on an SSD)

### DIFF
--- a/jena-sdb/src/main/java/org/apache/jena/sdb/layout2/TupleLoaderBase.java
+++ b/jena-sdb/src/main/java/org/apache/jena/sdb/layout2/TupleLoaderBase.java
@@ -30,6 +30,9 @@ import org.apache.jena.sdb.sql.TableUtils ;
 import org.apache.jena.sdb.store.TableDesc ;
 import org.apache.jena.sparql.util.NodeUtils ;
 
+import static java.sql.Connection.TRANSACTION_NONE;
+import static java.sql.Connection.TRANSACTION_READ_UNCOMMITTED;
+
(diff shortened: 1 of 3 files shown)


afs:
See [JENA-1665](https://issues.apache.org/jira/browse/JENA-1665).

grahamtriggs:
I should add to this, that there are two different circumstances that should be considered:

1) If you are frequently doing small inserts of triples - using multiple DB sessions - then this will have the most benefit, as the creation of the temporary table was a significant overhead.

2) For large batch imports (e.g. sdbload), then it is important to include "rewriteBatchedStatements=true" as part of the MySQL connection string.

Using the rewriteBatchedStatements option takes batch triple loading from 3,500 - 6,000 triples a second to 12,000 to 20,000 triples per second, as the driver uses a more efficient operation for executeBatch().

(This is on a SSD-based laptop, which can ingest at up to 35,000 triples per second for TDB2).

As you would expect, tuning MySQL buffer settings is important - once the buffer reaches capacity, the ingest will slow dramatically.

afs:
The project position on SDB is currently that it is discouraged for new applications and that problems with the module SDB would not block a release; it would get dropped for the release.

On that basis, I'm inclined to be quite permissive about changes to SDB. If someone can make something better that works for them, then merge it.

I'm not sure how many users there are and what databases they use. Historicially MySQL has been the most commonly used backend.

afs:
Some tests maybe?

PR #533, opened 2019-02-15
Feature/sdborder

grahamtriggs:
This patch combines a few enhancements to the way SQL queries are formed:

1) The order of variables is retained in the ScopeBase (a HashSet is replaced with LinkedHashSet). When select a subject / property / object from the Triples/Quads table, the order in which the columns are selected can make a difference as to whether an SQL engine uses an index or not, and this at least makes it predictable (and in practice, the order will be subject - predicate - order, and allow the SubjPredObj index to be used - previously, with the HashSet losing the order in which the scopes were added, you might get predicate - subject - object, which isn't in an existing index).

2) Replace some DISTINCT clauses with GROUP BY statements, and fix the order in which columns are grouped to match an index. SQL Engines typically perform GROUP BY faster than a DISTINCT (e.g. in MySQL the optimizer explicitly rewrites DISTINCT to GROUP BY when it knows that it can). Again, fixing the order of the columns in the GROUP BY improves the likelihood that an index will be used for that part of the query.

3) Allow simple ORDER BY clauses to be pushed into the SQL - by simple, I mean ordering by a bound variable and without the use of a function.

This is DISABLED by default, because the order will be different to the order that the ordering iterator will return, and can be enabled with the "optimizeOrderClause" option. Whilst the ordering does not include the comparisons of the iterator, the order generated by the database should still be consistent with the SPARQL definition of ordering.

If you are returning the entire set of results from a query, there is generally negligible performance difference between the Java iterator and the SQL clause - it shifts a some of the execution time from the JVM to SQL.

However, passing the ORDER BY into SQL means that any LIMIT / OFFSET can also be passed into SQL. So in cases where you are LIMITing the amount of rows returned by SPARQL, the SQL version will be substantially faster. e.g.

SELECT ?s ?p ?o WHERE { ?s ?p ?o } ORDER BY ?s ?p ?o LIMIT 20

on a 500,000 triple store, this takes 12 seconds using the Java iterator, and only 1.5 seconds using the SQL ordering.

NB. If a SPARQL query does not consist of "simple" clauses only that can be rewritten, then it will use the Java iterator and not pass any ordering into the SQL.

### DIFF
--- a/jena-sdb/src/main/java/org/apache/jena/sdb/SDB.java
+++ b/jena-sdb/src/main/java/org/apache/jena/sdb/SDB.java
@@ -70,6 +70,8 @@ public class SDB
     public static final Symbol annotateGeneratedSQL     = SDBConstants.allocSymbol("annotateGeneratedSQL") ;
 
     public static final Symbol optimizeSubqueryFragments = SDBConstants.allocSymbol("optimizeSubqueryFragments") ;
+
+    public static final Symbol optimizeOrderClause = SDBConstants.allocSymbol("optimizeOrderClause") ;
     // ----------------------------------
     
     // Global context is the ARQ context.
--- a/jena-sdb/src/main/java/org/apache/jena/sdb/compiler/OpSQL.java
+++ b/jena-sdb/src/main/java/org/apache/jena/sdb/compiler/OpSQL.java
@@ -19,7 +19,6 @@
 package org.apache.jena.sdb.compiler;
 
 import org.apache.jena.atlas.io.IndentedWriter ;
-import org.apache.jena.query.SortCondition;
 import org.apache.jena.sdb.core.SDBRequest ;
 import org.apache.jena.sdb.core.sqlnode.SqlNode ;
 import org.apache.jena.sdb.store.SQLBridge ;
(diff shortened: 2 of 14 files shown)


afs:
Recorded as [JENA-1673](https://issues.apache.org/jira/browse/JENA-1673).

(If the JIRA is opened before the PR and the PR has the JENA-NNNN in the title, the automation will update the JIRA ticket with activity on the PR.)

afs:
Some test maybe?

grahamtriggs:
Functionally, the LinkedHastSet (order of Scope), and DISTINCT / GROUP BY translation are tested multiple times during the existing tests - the code is exercised and affects the SQL generated without breaking any of the tests.

The ORDER BY translation currently isn't because it's disabled by default, but it could be turned on for the tests (and again it doesn't break any existing tests).

I'll have a think about what tests could be included to ensure that the SQL translations occur as expected, beyond just testing the execution of the query.

afs:
I was thinking of tests that check this functionality is happening. The existing test suite should test SPARQL execution functionality but doesn't check how that happened. A small number tests per feature (e.g. 1!) that check expected transformations happen makes it more stable in to the future.

This isn't a blocker for me. I'm happy to commit to the codebase as is and leave any testing for a later followup.

grahamtriggs:
@afs Hi - I've added some basic tests:

1) There is a test to check that scopes added to the ScopeBase are returned in the correct order

2) There is a simple test of the DISTINCT optimization to see if the SQL generates a GROUP BY

3) There is a simple test of the ORDER optimization to see if the SQL generates an ORDER BY

I've also moved where the ORDER optimization is applied to the post-processing (alongside the LIMIT/OFFSET handling), as it is more cleanly applied there.

grahamtriggs:
@afs Ooops! I've added the Apache header to both of those files.

PR #535, opened 2019-02-20
JENA-1674: Don't sign extend Float.floatToIntBits.

afs:
Add TestFloatNode for value-based testing of FloatNode.
Use canonical NaN for packing Double.NaN.


afs:
Sorry @rvesse - discovered I'd been running the wrong test suite and the original PR was not passing. Caught thanks to TravisCI. The only change is in `TestNodeIdInline.nodeId_float_6` which was 

Wrong (the reconstructed value isn't lexical `"1"`:
```
    @Test public void nodeId_float_6()
    { test("'-1'^^xsd:float") ; }
```
fixed:
```
    @Test public void nodeId_float_6()
    { test("'-1'^^xsd:float", "'-1.0'^^xsd:float") ; }
```

PR #537, opened 2019-02-28
JENA-1676 , JENA-1677: Make the commit step more robust against external factors

afs:
[JENA-1676](https://issues.apache.org/jira/browse/JENA-1676) : TDB1

[JENA-1677](https://issues.apache.org/jira/browse/JENA-1677) : TDB2

### DIFF
--- a/jena-db/jena-dboe-transaction/src/main/java/org/apache/jena/dboe/transaction/txn/TransactionCoordinator.java
+++ b/jena-db/jena-dboe-transaction/src/main/java/org/apache/jena/dboe/transaction/txn/TransactionCoordinator.java
@@ -747,8 +747,9 @@ public class TransactionCoordinator {
                     throw new TransactionException("Exception during 'commit' - transaction rollback.", ex);
                 }
                 // Very bad. (This have been dealt with already and should get to here.)
-                SysErr.error("Transaction rollback failed. System unstable.");
-                throw new TransactionException("Exception during 'rollback' - System unstable.", ex);
+                SysErr.error("Transaction rollback failed. System unstable."+
+                    "\nPlease contact users@jena.apache.org, giving details of the environment and this incident.");
+                throw new Error("Exception during 'rollback' - System unstable.", ex);
             }
             catch (Throwable ex) {
                 SysErr.warn("Unexpected Throwable during 'commit' : transaction may have committed. Attempting rollback: ",ex);
--- a/jena-db/jena-dboe-transaction/src/main/java/org/apache/jena/dboe/transaction/txn/TransactionCoordinator.java
+++ b/jena-db/jena-dboe-transaction/src/main/java/org/apache/jena/dboe/transaction/txn/TransactionCoordinator.java
@@ -20,7 +20,9 @@ package org.apache.jena.dboe.transaction.txn;
 
 import static org.apache.jena.dboe.transaction.txn.journal.JournalEntryType.UNDO;
 
+import java.io.IOException;
 import java.nio.ByteBuffer ;
+import java.nio.channels.ClosedByInterruptException;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap ;
 import java.util.concurrent.Semaphore ;
@@ -30,6 +32,7 @@ import java.util.concurrent.locks.ReentrantReadWriteLock ;
 
 import org.apache.jena.atlas.logging.FmtLog;
 import org.apache.jena.atlas.logging.Log ;
+import org.apache.jena.dboe.base.file.FileException;
 import org.apache.jena.dboe.base.file.Location;
 import org.apache.jena.dboe.sys.Sys;
 import org.apache.jena.dboe.transaction.txn.journal.Journal;
@@ -64,7 +67,8 @@ import org.slf4j.Logger ;
  */
 final
 public class TransactionCoordinator {
-    private static Logger log = Sys.syslog ;
+    private static Logger SysLog = Sys.syslog ;
+    private static Logger SysErr = Sys.errlog;
     
     private final Journal journal ;
     // Lock on configuration changes. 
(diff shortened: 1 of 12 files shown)


afs:
The normal processing of transactions is not affected by this PR, except that there is code rewrite. What changes is that the error handling is expanded, 

TDB1:
Changes at commit point:
https://github.com/apache/jena/pull/537/files#diff-75d6cc4cd853b4f00f68431df7465d79R159
and its journal:
https://github.com/apache/jena/pull/537/files#diff-ba9400c61c593485f52fc26cbb831334R312

TDB2:
Changes at commit point:
https://github.com/apache/jena/pull/537/files#diff-dccb2f7577f47b71391afce72cdc0c7dR716
and its journal:
https://github.com/apache/jena/pull/537/files#diff-d9dbd4b5900d1a55896d8b038a8012deR257

jena-db/jena-dboe-transaction/src/main/java/org/apache/jena/dboe/transaction/txn/journal/JournalControl.java
```diff
@@ -18,280 +18,38 @@
...
+        Iterator<JournalEntry> iter = journal.entries();
+
+        for ( ; iter.hasNext() ; ) {
+            JournalEntry e = iter.next();
+            // System.out.println("Posn: "+journal.position()+" :
+            // ("+(journal.size()-journal.position())+")") ;
+            System.out.println(JournalEntry.format(e));
         }
     }
-
```
rvesse: I guess all this deleted stuff is irrelevant/outdated stuff that you cleaned up with these changes?

jena-tdb/src/main/java/org/apache/jena/tdb/transaction/Transaction.java
```diff
@@ -205,158 +296,162 @@ public void abort() {
...
-                        throw new TDBTransactionException("Exception during abort - transaction did abort", ex) ;
-                    }
-                    state = TxnState.ABORTED ;
-                    outcome = TxnOutcome.W_ABORTED ;
-                    // journal.truncate to last commit
-                    // Not need currently as the journal is only written in
-                    // prepare.
-                    break ;
+                        throw new TDBTransactionException("Transaction has already committed or aborted");
+                    // Application abort. Didn't startwriing to the journal.
```
rvesse: Typo, words run together

rvesse:
NB - For other reviewers ignoring whitespace changes in the diff view eliminating a bunch of non-changes

jena-tdb/src/main/java/org/apache/jena/tdb/transaction/Transaction.java
afs: Fixed

jena-db/jena-dboe-transaction/src/main/java/org/apache/jena/dboe/transaction/txn/TransactionCoordinator.java
```diff
@@ -683,28 +687,80 @@ private boolean promotionWaitForWriters() {
...
+                    SysErr.warn("IOException during 'commit' : transaction may have committed. Attempting rollback: "+ex.getMessage());
+                else
+                    SysErr.warn("Exception during 'commit' : transaction may have committed. Attempting rollback. Details:",ex);
+                if ( abandonTxn(transaction, sysabort) ) {
+                    SysErr.warn("Transaction rollback");
+                    throw new TransactionException("Exception during 'commit' - transaction rollback.", ex);
+                }
+                // Very bad. (This have been dealt with already and should get to here.)
+                SysErr.error("Transaction rollback failed. System unstable.");
+                throw new TransactionException("Exception during 'rollback' - System unstable.", ex);
```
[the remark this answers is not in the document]
afs: If it gets here, I don't know what happened but likely some external environment (I/O) issue. Rollback does not use more disk space; it does do I/O to reset the journal removing bytes at the end.

It is a log message so won't be seen until later.

"System unstable" was supposed to communicate that.

How about changing it to include "contact users@jena" and making it a `java.lang.Error`?

jena-db/jena-dboe-transaction/src/main/java/org/apache/jena/dboe/transaction/txn/journal/JournalControl.java
afs: Yes - I found this large block of stuff that looks like it started in TDB1.

jena-db/jena-dboe-transaction/src/main/java/org/apache/jena/dboe/transaction/txn/TransactionCoordinator.java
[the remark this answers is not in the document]
rvesse: Perhaps something asking users to report back to us so we get an idea of how prevalent (or not) this situation is in the wild.  If something might be wrong with the users IO setup (whether transient or not) may be worth making that clear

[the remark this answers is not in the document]
afs: There is a good reason why each exception message has different working - points to the exact line if they email.

See the [additional commit](https://github.com/apache/jena/pull/537/commits/98caf7d4a8792cc1ca4faca4a4cadebcbf25249a) that asking the user to contact users@apache.

It isn't prevalent. There would already have been warnings and exceptions reported. The main change now is an attempt to handle the situation. 

It happened in our development when Thread.interrupt was called on a long running thread that happened at the time to be in a transaction. We have stopped using Thread.interrupt - a better mechanism is the pattern used for query cancellation. It is the response of the Java I/O close the file asynchronously. That is covered further up; it should only impact the journal which as some recovery code.

PR #538, opened 2019-03-02
JENA-1678 : Update version of sjf4j

afs:


afs:
For CVE-2018-8088.

While Jena does not use this feature of slf4j, it helps if the jar dependencies can be kept CVE-clean to assist users in environments that have automatic scanning of software.

PR #546, opened 2019-03-21
JENA-1687: refine NodeFactory return types

jmkeil:


afs:
Given the other PRs and discussion on [JENA-1687](https://issues.apache.org/jira/browse/JENA-1687), I don't think we should make the change from the top class (in effect, the interface) and the specific concrete implementations.

Thank you to @jmkeil for suggesting it and making sure we do think about issues.

jmkeil:
Yes. The change is unnecessary.

afs:
Thanks!

PR #548, opened 2019-03-21
JENA-1689: change type of Polyadic#m_subGraphs from List to Set

jmkeil:
…ted addition of graphs causing performance issues and unexpected behavior in case of removal of graphs


afs:
What is the performance issue? Is itbecause a graph is added two or more times?

MultiUnion.addGraph suppresses duplicate graphs at the point of addition of the graph. Application-written subclasses of Polyadic could do the same.

(graphs are .equals by identity not by value (contents)).

    if (!m_subGraphs.contains( graph )) {
                m_subGraphs.add( graph );
            }

jmkeil:
You are right. Sorry.

PR #549, opened 2019-03-28
JENA-1696: Control of jackson versions. Update jsonld-java

afs:
Expicit control the version of jackson.core so we can pick ones with security fixes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,11 +57,15 @@
     <ver.libthrift>0.10.0</ver.libthrift>
 
     <!-- jsonld-java depends on depends on Jackson core
-         artifacts so the versions must align. Consult jsonld-java's 
-         POM for the correct dependency versions 
+         and Jackson databind. databind has had a few CVEs 
+         so we exclude from jsonldjava and select the version
+         ourselves.
+         When upgrading jsonld, consult jsonld-java's  
+         POM for the correct dependency versions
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -57,11 +57,15 @@
     <ver.libthrift>0.10.0</ver.libthrift>
 
     <!-- jsonld-java depends on depends on Jackson core
-         artifacts so the versions must align. Consult jsonld-java's 
-         POM for the correct dependency versions 
+         and Jackson databind. databind has had a few CVEs 
+         so we exclude from jsonldjava and select the version
```
rvesse: Is that exclusion pre-existing or was that missed from this PR?

afs: The dependency is picked up in jena-arq which explicitly pulls in jackson.core. This replaces the jsonld-java ones when they are compatibility (by maven defn).

Can add exclusions if you want. As it stands, if mistakenly jsonld-java is ahead of Jena's, resolution will be the later ones.

See dependency:tree for jena-arq.

ajs6f: I would prefer to add the exclusions for clarity... otherwise, if I understand the situation rightly, changing the order of dependencies in a downstream project could create a different dependency tree. That would surprise me!

afs: Exclusions for the two jackson artifacts added.

PR #552, opened 2019-04-08
JENA-1700: RDFParserBuilder.strict. Strict Fuseki SPARQLer data validator.

afs:


jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/validation/html/DataValidatorHTML.java
```diff
@@ -109,7 +110,7 @@ public static void executeHTML(HttpServletRequest httpRequest, HttpServletRespon
                     outStream.flush(); 
                     System.err.flush() ;
                 } catch (RiotException ex) {
-                    ex.printStackTrace(stderr); 
+                    //ex.printStackTrace(stderr);
```
kinow: Maybe delete the line?

afs: Done!

PR #553, opened 2019-04-08
JENA-1701: Print an error on bad URI string

afs:


jena-arq/src/main/java/org/apache/jena/sparql/expr/nodevalue/NodeFunctions.java
```diff
@@ -421,18 +421,30 @@ public static Node iri(Node n, String baseIRI) {
...
 
+    private static String getOneViolation(IRI iri, boolean includeWarnings) {
+        Iterator<Violation> iter = iri.violations(includeWarnings);
+        if ( ! iter.hasNext() )
+            return null; 
+        Violation viol = iter.next() ;
+        return viol.getShortMessage() ;
+    }
+
     // The Jena version can be slow to inityailise (but is pure java)
```
rvesse: Random unrelated typo here if you feel like fixing it

afs: Yes - I've removed the whole comment and commented code.  It is not true nowadays because Java got better at what was the slow point of initialization.

PR #555, opened 2019-04-12
JENA-1703: Combined artifacts NOTICE files

afs:
Sets up the NOTICE inclusions for shaded Fuseki jars.
Removes unnecessary files which helps us keep things current.


afs:
Extra commit to set `/META-INF/NOTICE` in the war file. After rereading various descriptions and also the Apache principles, I have concluded we should include a NOTICE file.

(I don't like the way this is inserted but it works and gets the intended war file layout.)

PR #557, opened 2019-04-22
JENA-664 GeoSPARQL Jena

galbiston:
GeoSPARQL implementation and Fuseki application. Using pom.xml provided by Andy Seabourne for both modules. Tidied up licences to same format as other Jena modules.

- jena-geosparql is failing the rat check due to the OGC schema file. 
- jena-fuseki-geosparql is using Logback Classic for logging and may need to be replaced with preferred logger.
- Apache SIS logging is not using SLF4J, so is not redirected via logging setup (mainly relating to coordinate reference system transformations if using default setup).


ajs6f:
Hey, @galbiston, this is awesome to see! Thank you so much for all this work!

On the specific question of SIS logging (and admitting in advance that I know nothing about SIS), could we use something like [this](https://sis.apache.org/apidocs/org/apache/sis/util/logging/LoggerFactory.html) to plug in our usual SLF4J API, or is there something more subtle going on here?

afs:
Is that what `org.slf4j:jul-to-slf4j` is for?

ajs6f:
@afs, but there are some [quirks](https://www.slf4j.org/legacy.html#jul-to-slf4j) that require a little more work. That having been said, I use `jul-to-slf4j` all the time without issue.

galbiston:
`org.slf4j:jul-to-slf4j` has worked once I'd found the programmatic initialisation step. I've applied it to the jena-fuseki-geosparql application but only suppressed the SIS logging when I tried the same in the jena-geosparql library. Will try to have another look at it.

afs:
Let's look at jena-geosparql then move on to jena-fuseki-geosparql.

There is java8/java11 for `java.xml.bind:jaxb-api` (not in Java11 - has to be a dependency) but other than that it compiles. Not urgent, know fix.

Running the tests I get at the start:

    WARNING: The “SIS_DATA” environment variable is not set.

then some test problems which I'm assuming come from the lack of `SIS_DATA`. What should this be set to?

afs:
I have found http://sis.apache.org/epsg.html which needs some thinking about. Setting up locally simple enough.

Tests pass!

(with noise - I'm not worried about logging noise. Setting for the tests can be done with `src/test/resources`. Elsewhere, I have used log4j2 and also used (separately!) JUL.)

afs:
I have found http://sis.apache.org/epsg.html and got the tests running locally. Sorry for the noise above.

Test pass!

Some noise to be tuned out from running the tests because of `System.out.println` but that's minor.

I get 3 warning messages:
```
Apr 23, 2019 6:33:20 PM org.apache.sis.referencing.operation.transform.DefaultMathTransformFactory createParameterizedTransform
WARNING: The “SIS_DATA” environment variable is not set.
Apr 23, 2019 6:33:20 PM org.apache.sis.referencing.operation.CoordinateOperationFinder createOperation
WARNING: Can not parse “OSTN15_NTv2_OSGBtoETRS.gsb” as a file in the NTv2 format.
Apr 23, 2019 6:33:20 PM org.apache.sis.referencing.operation.CoordinateOperationFinder createOperation
WARNING: Can not parse “OSTN02_NTv2.gsb” as a file in the NTv2 format.
```

galbiston:
I had the SIS_DATA dependency resolved in Gradle for testing both modules by "testImplementation "org.apache.sis.non-free:sis-embedded-data:$sisVersion". I don't think there is an equivalent for Maven for test only. Due to the licensing conflict between EPSG and AL2 I didn't include it as a non-test dependency.

The other two warnings are to do with precision in transformations from the UK's OSGB36 to WGS84. OSGB36 was just an arbritary choice to use a planar coordinate system and conversions from WGS84. I can change the relevant test data to one of the main coordinate reference systems and it should make the warnings go away. There will just be this underlying configuration of the Apache SIS framework.

I asked Martin Desruisseaux from Apache SIS about the warnings and his response is below.

"For the warning that you get, actually I should improve the error message. I presume that the “OSTN15_NTv2_OSGBtoETRS.gsb” datum shift file is not present on your system, so the error message should be "Can not find OSTN15_NTv2_OSGBtoETRS.gsb datum shift file". In such case Apache SIS will use a less accurate fallback (more on it later).

The SIS_DATA environment variable is used not only for the EPSG database, but also for all other kind of data that SIS may need, including datum shift files. The EPSG embedded database is more convenient, but does not provide any datum shift files yet. We may include some datum shift files in sis-embedded-data in the future, but I don't think we could include all of them for space (and sometime licensing) reasons. Note also that I noticed that the embedded Derby database in a compressed JAR file is at least 5 times slower than a Derby database as ordinary files in a directory.

If you want to give a try to SIS_DATA, you can create a DatumChanges sub-directory in the SIS_DATA directory and put the OSTN15_NTv2_OSGBtoETRS.gsb file there (we can download it from internet). You may also opportunistically replace sis-embedded-data module by sis-epsg for better performances (it will create the Derby database in that directory). If you prefer to keep sis-embedded-data for convenience, then you can ignore those warnings. SIS could be improved for making those warnings less disruptive to the users, but we have not determined yet how. In any case, I recommend to take a look at the output of System.out.println(theCoordinateOperation) and check in particular the value inside OperationAccuracy[...] element. The value will be different depending if the datum shift file has been found or not; it is up to the application to decide if the reported accuracy if sufficient for their needs."

ajs6f:
Hello, @galbiston-- can you explain a bit what you mean by "I don't think there is an equivalent for Maven for test only." Are we trying to put resources on the classpath only for the test phases, or have them available in some other way for just those phases, or am I misunderstanding what you mean?

afs:
I'm using this in the POM for the tests:
```
    <!-- Non-free -->
      <dependency>
        <groupId>org.apache.sis.non-free</groupId>
        <artifactId>sis-embedded-data</artifactId>
        <version>0.8</version>
        <scope>test</scope>
      </dependency>
```
and then 

> Set the SIS_DATA environment variable to the path of an initially empty directory (preferred choice). 

applies. 

Where does SIS_DATA setting happen in the gradle setup?  My gradle foo ran out and I could not find it.

I tried exporting it in the shell - it works when I use a Java8 JDK, and fails for Java11 (compiling for Java8).

@ajs6f Is there a way to have a few lines of java run before all the tests as it is current setup or do we need to do the 

    @RunWith(Suite.class)
    @SuiteClasses({Test1.class, Test2.class})

thing?

afs:
Next issue:  `geosparql_vocab_all_v1_0_1_updated.rdf`

It references the OGC license for documents http://www.opengeospatial.org/legal/ 
But is it "document" or "software"? 

An OGC "Document" has no right to modify and so is not open source.
OGC "Software" does not have that restriction (whether the erst of the license is OK, I haven't gone into in depth.)

I suspect that it is considered part of the "GeoSPARQL 1.0 is an OGC Standard.", so its a document.

@desruisseaux - is that correct?

---

I did find that if I removed it from `GeoSPARQLOperations`, then recursive all methods that got broken by that, the only the `DatasetOperations.setup` call to `GeoSPARQLOperations.applyInferencing` was not available.

ajs6f:
@afs This PR is so big I cannot even load the list of files in Github's browser. Since we are still using JUnit 4 we can write a [`Rule`](https://junit.org/junit4/javadoc/4.12/index.html?org/junit/Rule.html) or [`@ClassRule`](https://junit.org/junit4/javadoc/4.12/index.html?org/junit/ClassRule.html) (`static` version) to manage the timing and state of setup ourselves. A `@ClassRule` on the suite itself would be what we want to run a few lines of code before and after the complete lineup of tests. Is that what you are asking about? Or something more granular?

If you can give me the names of some tests, I could pull this branch locally and start poking around.

desruisseaux:
Hello @afs . The question about the exact license governing the RDF file is a good one. I confirm that OGC "Software" license is BSD-like and classified as category A for inclusion in Apache software. But I don't know if `geosparql_vocab_all_v1_0_1_updated.rdf` is under "Software" or "Document" license. I can ask to OGC if desired.

About logging mentioned earlier in this thread, is `org.slf4j:jul-to-slf4j` satisfying? Apache SIS has a mechanism for redirecting logging to SLF4J, but if JUL work well I could deprecate (and remove later) the SIS mechanism and replace by a documentation mentioning JUL.

afs:
@ajs6f It is currently working by surefire finding the tests so there is no place to add a suite-wide step.  If we add such as step with `RunWith` / `SuiteClasses` we can code with \@BeforeClass (or the other ways).

I don't think it needs clearing between tests but I'm not completely sure of that.

ajs6f:
@afs Can you point me at what we actually want to run? Are we talking about setting up some of these GIS reference databases like the stuff in `SIS_DATA`, or something else?

afs:
@desruisseaux, if you could ask OGC that would be great. It makes sense to restrict documents that are the definitive specifications (as do most standards organisations in some way of other). Reference data can be the same. This vocabulary / ontology file is more about defined terms for interoperability.

desruisseaux:
Actually, while preparing the email to OGC, I found answer to the question in their [legal FAQ](http://www.opengeospatial.org/ogc/legalfaq#DTD). Quoting them:

> **Is a schema or document definition (DTD) covered by the document or software terms?**
> Schemas (and DTDs) are frequently part of our specifications and seemingly fall under the document copyright terms. However, as long as you do not use the same formal namespace or public identifier to identify that modified OGC schema/DTD (which might confuse applications), you may treat the schema/DTD under the software terms. This means that you are permitted to make a derivative or modified OGC schema/DTD, but even under the software terms you are obligated to include/retain the OGC copyright notice. We further appreciate a couple sentences regarding who made the modifications, when, and what changes were made in the original DTD -- a common software documentation practice.

afs:
> @afs  Can you point me at what we actually want to run? 

The tests are in src/test as usual and run with surefire finding \@Test. It is not urgent at this point; it is just a warning (I hope!).

ajs6f:
@afs No, sorry, I meant to ask you to point me at what we want to run _before_ the tests: the setup step.

afs:
@desruisseaux - perfect! Thank you.

afs:
@ajs6f - create an empty test directory, point `SIS_DATA` to it.  However, it takes some fixing to get the module at the moment.

ajs6f:
@afs Can we not use some Maven for that? We could put an empty `SIS_DATA` directory in `src/test/resources` and then:
```
<build>
    <testResources>
      <testResource>
        <directory>src/test/resources/SIS_DATA</directory>
      </testResource>
</testResources>
…
<plugins>
      <plugin>
        <artifactId>maven-surefire-plugin</artifactId>
        <configuration>
          <systemPropertyVariables>
            <propertyName>SIS_DATA</propertyName>
            <buildDirectory>${project.build.directory}/SIS_DATA</buildDirectory>
```
?

desruisseaux:
`SIS_DATA` is an environment variable, while Maven Surefire `<systemPropertyVariables>` is (in my understanding) for Java properties. Another issue is that pointing to `${project.build.directory}/SIS_DATA` implies that the actual directory change for each Maven module, which may be a performance concern in database is recreated each time.

I have not found an ideal solution yet. The way I handle that in Apache SIS is to make tests that depends on `SIS_DATA` optional, with a JUnit `assumeTrue` statement at the beginning of those tests. That way, those tests are reported as skipped by Surefire or Jenkins.

ajs6f:
@desruisseaux We have I/O tests that take five minutes to run-- I'm not sure that making a few empty directories is going to change our build time very much, but if it does, we can build an `SIS_DATA` directory once in the main GeoSPARQL module and pass that location into the submodules via a Maven property.

As for the ENV var point, that's fairly trivial: 
```
<plugin>
        <artifactId>maven-surefire-plugin</artifactId>
        <configuration>
          <environmentVariables>
            <propertyName>SIS_DATA</propertyName>
            <buildDirectory>${project.build.directory}/SIS_DATA</buildDirectory>
```

desruisseaux:
The performance interrogation was because SIS may try to create the EPSG geodetic database in that directory (depending if `org.apache.sis.non-free:sis-epsg` is found on the classpath and the database does not already exist), which may take 10~30 seconds.

ajs6f:
Cool, then we can build it once and pass the location around via a Maven property.

afs:
"mvn clean test" takes 12s.

afs:
## Status (2019-04-27)

### Next

I've made some changes locally and got the build to work with tests.

All the changes are detailed below.  If the changes are OK, I can merge the PR and apply them.

There is still some possible clearing up but the only necessary step before release is the L&N for the shared Fuseki server.

### Details

The parent POM has some stray formatting.

#### Module: jena-geosparql

Some line ends need fixing - they are windows style (when using a diff, this shows up).

Surefire plugin configuration is for another way to run tests but for this module runs no tests. The automatic test determination works fine for jena-geosparql.

Delete output in tests such as "System.out.println" and companion "//System.out.println".

Add dependencies:
```
      <!-- Non-free; testing only -->
      <dependency>
        <groupId>org.apache.sis.non-free</groupId>
        <artifactId>sis-embedded-data</artifactId>
        <version>0.8</version>
        <scope>runtime</scope>
      </dependency>

      <dependency>
        <groupId>javax.xml.bind</groupId>
        <artifactId>jaxb-api</artifactId>
        <version>2.3.1</version>
      </dependency>
```

Then

    mvn8 clean test -Drat.skip=true

passes. (mvn8 is my script to run mvn with OpenJDK 8)

#### RAT Legal Check

Add to the parent POM:

    <exclude>**/geosparql_vocab_all_v1_0_1_updated.rdf</exclude>
    Delete logback*.xml

then 

    mvn8 clean test

passes.

#### Java11

maven with OpenJDK11 : has problems which look like a precison/length change somewhere in
the depths.

`BufferFFTest` does some comparison of quite large objects, too big to
describe here.

There are some tests on `LatLonPoint` like:

    GreatCirclePointDistanceTest.testGetPoint_3args_East:185
    expected: <LatLonPoint{lat=48.84533344929755, lon=-0.9821733277296272}> 
    but was:  <LatLonPoint{lat=48.845333449297556, lon=-0.982173327729628}>

The test uses `assertEquals(Object,Object)`, no precision factor
considered as is available with `assertEquals(double, double, delta)`.
Something for later.

#### Module jena-fuseki-geosparql

Worked first time!

There will be a bit of L&N work for `geosparql_vocab_all_v1_0_1_updated.rdf`.

I think it would be better placed as: 

   /jena-fuseki2/jena-fuseki-geosparql

but still run from the top POM (not in jena-fuseki2/pom.xml for now).

A better Java package name would be to put in the Fuseki family:

`package org.apache.jena.jena_fuseki_geosparql`
==>
`package org.apache.jena.fuseki.geosparql`

desruisseaux:
About precision, note that `Math.sin`, `cos` _etc._ are not fully deterministic even in Java. The specification said only that the result must be within 1 or 2 ULP from the real answer. There is variations depending on the platform or Java version. Furthermore many map projection formulas are only approximations (some have no exact mathematical solution), so the numerical output may vary in future Apache SIS versions depending on improvements or refactoring done on those calculations.

As a general rule it is safe to compare with a one centimetre precision (this is the precision reported in EPSG guidance notes for the published formulas). In practice the precision is often better, but not guaranteed.

afs:
@galbiston -- Sorry - I've missed something.

Some of the files have the Apache License for contributed software, (example `GeoSPARQLConfig.java`) and some have different wording. It is better practice for the contributor to change things to do with license and copyright.

Here is a perl script removes any existing headers and puts in the appropriate one.

https://gist.github.com/afs/c32177d0614553edee4f0e5ae319fb48

galbiston:
Apologies for the license issue. I thought I'd tidied them all up. I've also removed the lines of test name output, removed the commented out test results and renamed the Fuseki GeoSPARQL packages to the suggested form.

Thanks,

Greg

afs:
@galbiston - Thanks. The license text was passing the RAT checks but there was extra wording like "Copyright the authors". Not wrong, but a variation.  Copyright and IP is best done cleanly at the start rather than treat like software and "can fix later".

afs:
Merged to master!

PR #560, opened 2019-04-24
JENA-1706: Handle "QueryParser"; suppress logging WARN

afs:


afs:
This follows on from #536.

PR #580, opened 2019-06-30
JENA-1719: Work around for iterators being held across transactions

der:
Provides a workaround to #JENA-1719.

Underlying problem is that when inferences are not run to completion
then tabled goals and associated interpreters and agenda entries can
get retained and in some cases can hold on to uncompleted iterators.
Correct fix would be to clean these up but doing this safely, especially 
in the presence of concurrent queries is tricky.

Long term solution might need serious redesign of the backward engine with 
concurrency and transactions in mind. 

This workaround wraps the iterators so that if one is retained and 
and reused across a transaction boundary then it will attempt to restart
it. So doesn't fix the underlying problem but does workaround it for this test case.

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/reasoner/rulesys/impl/SafeTripleIterator.java
+++ b/jena-core/src/main/java/org/apache/jena/reasoner/rulesys/impl/SafeTripleIterator.java
@@ -1,0 +1,56 @@
+package org.apache.jena.reasoner.rulesys.impl;
+
+import org.apache.jena.graph.Triple;
+import org.apache.jena.reasoner.TriplePattern;
+import org.apache.jena.reasoner.rulesys.BackwardRuleInfGraphI;
+import org.apache.jena.util.iterator.ClosableIterator;
+import org.apache.jena.util.iterator.ExtendedIterator;
+
+/**
+ * Wrapper around triple searches using in the back chainer to attempt a restart
+ * in the case of a failure such as cross-transaction query.
+ */
+public class SafeTripleIterator implements ClosableIterator<Triple> {
+
+    protected ExtendedIterator<Triple> matchIterator;
+    protected long offset = 0;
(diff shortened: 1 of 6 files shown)


jena-core/src/main/java/org/apache/jena/reasoner/rulesys/impl/SafeTripleIterator.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public SafeTripleIterator(LPInterpreter interpreter, TriplePattern goal) {
+        this.infGraph = interpreter.getEngine().getInfGraph();
+        this.goal = goal;
+        restart();
+    }
+
+    protected void restart() {
+        matchIterator = infGraph.findDataMatches(goal);
+        for (int i = 0; i < offset; i++) matchIterator.next();
```
rvesse: Doesn't this needs to defend against the case where the data changes rendering the offset invalid i.e. it might no longer be possible to call `next()` the expected number of times resulting in a `NoSuchElementException`

rvesse: For the case of `restart()` under `next()` this is fine but for `restart()` under `hasNext()` you'd expect to get a `false` return not an exception

der: If the data changes then there should have been either an explicit `rebind()` or an implicit `prepare()` call which should clear the tabled goals anyway. If you make changes to the data bypassing both of those then iterators will blow up anyway, don't think the restarting makes this any worse.

afs:
@der - clearing up: Can I delete the branch in the Jena repo for this ( branch "JENA-1719")?

der:
@afs Yes, fine by me, thanks.

PR #594, opened 2019-08-22
JENA-1742: Fix OSGi imports

acoburn:
Add exclusion for google appengine and apphosting packages

Note: with this change, I am able to successfully deploy Jena in Karaf 4.x

### DIFF
--- a/apache-jena-osgi/jena-osgi/pom.xml
+++ b/apache-jena-osgi/jena-osgi/pom.xml
@@ -261,6 +261,8 @@
               !org.apache.jena.ext.*,
               !org.checkerframework.checker.*,
               !com.google.errorprone.annotations.*,
+              !com.google.appengine.api,
+              !com.google.apphosting.api,
               sun.misc;resolution:=optional,
               *
             </Import-Package>


ajs6f:
No objection to this fix, but it would be nice to know where this stuff is arising. The GeoSPARQL contribution was the largest chunk I remember going in recently, but I don't know why this would come out of that. @galbiston, anything come to mind?

acoburn:
FWIW, I spent a fair amount of time with `jdeps`, `mvn dependency:tree` and looking through the git history, but all to no avail. Plus, the geosparql contribution isn't even part of the jena-osgi module, so in principle, that shouldn't have any effect on this. The only dependency changes from the perspective of `jena-osgi` that I could find were the Jackson and commons-codec libraries, neither of which brings in these dependencies. I would also like to know where these package dependencies are coming from.

ajs6f:
Fair enough. It certainly shouldn't stop us from using this fix for now.

galbiston:
Nothing I can think for GeoSPARQL.
I've run Gradle dependency insight for the original GeoSPARQL project and neither of those dependencies are found. Don't recall any changes to depedencies since it was incorporated.

PR #597, opened 2019-08-29
JENA-1745: Update jackson-databind version

afs:


rvesse:
Went ahead and merged since this is security related

afs:
@rvesse - thanks.

All these databind issues don't directly affect jsonld-java or jena because it is the mapping JSON to custom classes (ORM-like) that is the problem area.

However, it is easier to just fix so the dependencies don't get flagged by github and similar makes customer/users life easier to have dependences that don't show up in automated software scans.

PR #598, opened 2019-08-30
JENA-1748: Tests and fix for TDB2 DatasetGraph listGraphNodes

afs:


jena-db/jena-dboe-storage/src/main/java/org/apache/jena/dboe/storage/system/DatasetGraphStorage.java
```diff
@@ -83,7 +83,7 @@ public DatasetGraphStorage(StorageRDF storage, StoragePrefixes prefixes, Transac
 
     @Override
     public Iterator<Node> listGraphNodes() {
-        Iterator<Quad> iter = find(null, null, null, null);
+        Iterator<Quad> iter = findNG(null, null, null, null);
```
ajs6f: Just a question for my learning: am I reading this correctly that TDB2 gets its list of graphs by iterating every quad in the store?

afs: Yes. While that is the default implementation, it is the one used by TDB2.

Consider whether the app wants the right answer during a transaction ...

PR #599, opened 2019-09-03
JENA-1751: Enable inline use of Element Subclasses

jmkeil:


afs:
The code is fine ("does no harm"). 

There is discussion on the JIRA ticket. https://issues.apache.org/jira/browse/JENA-1751

I am curious as to why varargs is useful in practice because the call site has to have a fixed number of arguments, or manage a separate Java array, and if it's an array, a (collections) list and "foreach" works, and that which leads me to wonder about the relationship with jena-querybuilder.

jmkeil:
The idea was to use it for inline `Expr` building during `Query` building:


```java
new SelectBuilder().addFilter(new E_NotExists(
    new ElementPathBlock(new Triple(nodeA, nodeB, nodeC),
                         new Triple(nodeD, nodeE, nodeF))
));
```
However, I was not aware of the `ExprFactory` also allowing to do this inline:

```java
ExprFactory e = new ExprFactory();
new SelectBuilder().addFilter(e.notexists(
    new SelectBuilder().addWhere(nodeA, nodeB, nodeC)
                       .addWhere(nodeD, nodeE, nodeF)
));
```
So, it is not needed.

PR #604, opened 2019-09-12
JENA-1757: Deprecate GraphStatisticsHandler

afs:
Illustration of deprecating `GraphStatisticsHandler`, not for merging in the current state.

This does the work for src/main/java.  It does not touch very much.

There are still warnings in the jena-core test code and the way forward is probably to remove the tests.


afs:
Second commit does the removal of the tests for graph statistics handler.  They are for graph-mem and union graph when the components are all graph-mem.

Claudenw:
I think there are deprecation annotations missing from permissions code.
But I am traveling and can not check.

afs:
No problem - I'll take care of it.

afs:
Done and added the second commit - just a couple of uses.

PR #609, opened 2019-09-21
Clean javadoc

afs:
Preparation for 3.13.0 release.

### DIFF
--- a/jena-base/src/main/java/org/apache/jena/atlas/iterator/Iter.java
+++ b/jena-base/src/main/java/org/apache/jena/atlas/iterator/Iter.java
@@ -381,7 +381,7 @@ public class Iter<T> implements Iterator<T> {
     }
 
     /** Take the first N elements of an iterator - stop early if too few 
-     * @See {@link #limit(Iterator, long)}
+     * @see #limit(Iterator, long)
      */
     public static <T> List<T> take(Iterator<T> iter, int N) {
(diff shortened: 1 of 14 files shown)


kinow:
Ps: Jackson 2.9.10 being released too... I think it is a dependency in jena? Just in case you'd like to look at changelog and see if there's any reason to update it too for next release https://twitter.com/fasterxml/status/1175466334852747264?s=20

afs:
Thanks for pointing that out. It's been bumpy for upgrades these last few months.

We have a upstream dependency for Jackson for jsonld-java. I don't think any of Jena code directly uses the Jackson code. Now, Jena and jsonld-java don't actually use the part of databind that has been under attack, but it is easy to upgrade so if jena uses use jackson directly, they get the fixes.

jsonld-java is currently depending on 2.9.9 (core) and 2.9.9.2 (databind). Jena takes control of the exact version because of the releases for CVE's, giving us fine grained control (2.9.9.x) without needing to wait for jsonld-java to release, assuming x.x.x.+1 is fix-only.

In my $job, some customers scan jars and match against the CVE database. I'm sure they aren't the only ones. It is easier to upgrade that explain why the CVE does not affect the code.

afs:
Jackson 2.9.10 (core and databind) has hit central and the automated GH security alerts.

JENA-1762.

PR #610, opened 2019-09-25
JENA-1763: Move SHACL test files

afs:


afs:
This is "git mv" and fixed up the test driver code.

PR #611, opened 2019-09-25
JENA-1761: General option setting for RDFParser and RDFWriter

afs:
Two halves or JENA-1761:

"General option setting for RDFParser and RDFWriter" -- the mechanism
"Option to write Turtle with PREFIX and BASE" -- used for Turtle.

The default remains the "@prefix" forms.


afs:
I'll hold back from merging this until after 3.13.1.  It is a functionality change.

PR #612, opened 2019-09-25
JENA-1764: Fix Automatic-Module-Name metadata

acoburn:
- This adds missing AMN metadata to `jena-dboe-storage` (an empty string is invalid)
- This fixes incorrect AMN for `jena-shacl` (it currently duplicates the tdb module name)
- This also fixes duplicate AMN metadata for the `jena-elephas-*` and `jena-jdbc-*` modules

### DIFF
--- a/jena-shacl/pom.xml
+++ b/jena-shacl/pom.xml
@@ -35,7 +35,7 @@
   <properties>
     <maven.build.timestamp.format>yyyy-MM-dd'T'HH:mm:ssZ</maven.build.timestamp.format>
     <build.time.xsd>${maven.build.timestamp}</build.time.xsd>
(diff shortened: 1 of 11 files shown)


jena-shacl/pom.xml
```diff
@@ -35,7 +35,7 @@
   <properties>
     <maven.build.timestamp.format>yyyy-MM-dd'T'HH:mm:ssZ</maven.build.timestamp.format>
     <build.time.xsd>${maven.build.timestamp}</build.time.xsd>
-    <automatic.module.name>org.apache.jena.tdb</automatic.module.name>
+    <automatic.module.name>org.apache.jena.shacl</automatic.module.name>
   </properties>
```
afs: Sorry about that. Carelessness on my part.

PR #622, opened 2019-10-24
JENA-1774: Reorder SELECT DISTINCT * ORDER BY

afs:


jena-arq/src/test/java/org/apache/jena/sparql/algebra/optimize/TestOptDistinctReduced.java
```diff
@@ -218,7 +218,7 @@
         testQuery(queryString, tOrderByDistinctApplication, opExpectedString) ;
     }
     
-    @Test public void distinct_order_by_application_03()
+    /*@Test*/ public void distinct_order_by_application_03_old()
```
rvesse: If this test case is no longer valid why not simply delete it?

afs: Done. I meant to but missed coming back to this while juggling the other pull requests.

Comment added to 02.

afs:
I don't remember either. There is a case when there's a projection, but I can't find a reason not to when there is no projection.  If it had a projection which writes out the vars, it becomes a success on the other if-branch.

PR #628, opened 2019-11-10
JENA-1776: Set Content-Length when sending a graph or dataset

afs:


jena-rdfconnection/src/main/java/org/apache/jena/rdfconnection/RDFConnectionRemote.java
```diff
@@ -492,8 +495,33 @@ protected HttpEntity graphToHttpEntity(Graph graph) {
...
+        return graphToHttpEntityWithLength(graph, syntax);
+    }
+    
+    /** 
+     * Create an HttpEntity for the graph. The HTTP entity will have the length but this
+     * requires serialising the graph at the point when this function is called.  
+     */
+    private HttpEntity graphToHttpEntityWithLength(Graph graph, RDFFormat syntax) {
+        String ct = syntax.getLang().getContentType().getContentType();
+        ByteArrayOutputStream out = new ByteArrayOutputStream(128*1024);
```
ajs6f: No huge difference, but just out of curiosity, why not `try-with-resource` here?

afs: Needs to pass the ByteArrayOutputStream out of the "try" scope and until Java9 you can't write `try (out) {}` so it's a bit yukky.

And I added the "IO.close" for neatness as it is not really needed.

rvesse:
Reading the wider context of the code it looks like the new code path replaces the old code path and the only way to use the old code path would be to extend the class and override the relevant method to call the old method.  Except that you can't do that because some/all of the methods are `private`, can we make them all `protected` to leave users the option to customise behaviour if they need to?

afs:
Yes - this can cause an OOME. The only option to work with connection caching I can see are chunking or the in-memory serialization. This code path is used when the graph/dataset is a Java object. 

`graphToHttpEntity` and `datasetToHttpEntity`, the points where a customisation should override are both protected. This is the right way to customize.

The path of loading a (large) file into a remote triplestore is different code path and it sets the length from the file using httpClient `FileEntity` which sets the Content-Length from the file.

I don't like the idea of protected on the old code because it is broken in almost all cases of use in more than one transfer. I left it "for the record" and to have the method comment that this does not work over cached connections. "protected" seems encourage usage. I don't mind changing to protected from private but leaving code (3 lines) that will normally go wrong does not feel like a good idea. At least it can be copied (open source!) YMMV. 

@rvesse - do you have a usage that is sending a large amount of data via this code path?

PR #634, opened 2019-11-19
JENA-1781 update Thrift version

ktreimann:


afs:
This needs a workaround to [THRIFT-5022](https://issues.apache.org/jira/browse/THRIFT-5022) (as per apache/thrift#1942).

It looks like `BinRDF.apply` can work without using isOpen.

afs:
PR #635 includes this commit and workarounds needed because of a Thrift regression [THRIFT-5022](https://issues.apache.org/jira/browse/THRIFT-5022).

Closing this - the change will get made via #635 -  thanks!

PR #637, opened 2019-11-26
JENA-1784 CacheSimple doesn't check keys for equality

strangepleasures:
Added an explicit check for keys' equality

### DIFF
--- a/jena-base/src/test/java/org/apache/jena/atlas/lib/cache/CacheSimpleTest.java
+++ b/jena-base/src/test/java/org/apache/jena/atlas/lib/cache/CacheSimpleTest.java
@@ -18,12 +18,12 @@
 
 package org.apache.jena.atlas.lib.cache;
 
+import org.apache.jena.atlas.lib.Cache;
+import org.junit.Test;
+
 import static java.util.stream.Collectors.toMap;
 import static java.util.stream.IntStream.rangeClosed;
 import static org.junit.Assert.*;
-
-import org.apache.jena.atlas.lib.Cache;
-import org.junit.Test;
 
 /**
  * Simple test to ensure that {@link CacheSimple} evidences the fixed-size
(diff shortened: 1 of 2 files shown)


afs:
Thank you - I agree with the fix.

Writing the test is harder! :-)

Rather than rely on the interesting fact that `hash("Aa") == hash("BB")`, which is implementation dependent (it's true in OpenJDK at the moment), how about creating objects with the explicit features needed?

        Object obj1 = new Object() {
            @Override public int hashCode() { return 1; }
        };
        Object obj2 = new Object() {
            @Override public int hashCode() { return 1; }
        };
        assertEquals(obj1.hashCode(), obj2.hashCode());
        assertNotEquals(obj1, obj2);
        Cache<Object, Integer> cache = new CacheSimple<>(10);
        cache.put(obj1, 1);
        assertTrue("Same key, expected in cache", cache.containsKey(obj1));
        assertFalse("Keys with same hash code should not be considered equal", cache.containsKey(obj2));

strangepleasures:
Agree. Thank you for your suggestion.

jena-base/src/test/java/org/apache/jena/atlas/lib/cache/CacheSimpleTest.java
```diff
@@ -40,4 +40,20 @@ public void testFixedSize() {
 				.forEach(testCache::put);
 		assertEquals("Test cache failed to maintain fixed size!", maxSize, testCache.size());
 	}
+
+	@Test
+	public void testSameHash() {
+		Object key1 = new Object() {
```
ajs6f: The eventual call to `.equals()` in the main code falls through to `==` (which certainly passes the test), and it might be nice to make this even a little more explicit by overriding `.equals()` in these two guys and making it clear that they are not `equal()` each other, but it's not really necessary.

strangepleasures: It's indeed a good practice to override `equals` when you change `hashCode` and vice versa.

afs:
Thank you!

PR #640, opened 2019-11-28
JENA-1787: Improve performance of ValidationProc.simpleValidatationNode

strangepleasures:


afs:
Good improvement.

I think it is better to put the new logic in `simpleValidatationInternal`, the top of the evaluation process, as being a slightly more robust. The new logic isn't bypassed, especially the assumption that only true focus nodes get passed into the processing.

Code below (+ align calls to the function); it does incorporate the `isFocusNode` and singleton path without getting all focus nodes.

If this is OK, I can the changes after merging this PR; no requirement to update the PR itself.

```
    /**
     * Validation process.
     * Either all focusNode for the shape (argument node == null)
     * or just for one node of the focusNodes of the shape.
     */
    private static void simpleValidatationInternal(ValidationContext vCxt, Graph data, Node node, Shape shape) {
        Collection<Node> focusNodes;
   
        if ( node != null ) {
            if (! isFocusNode(shape, node, data))
                return ;
            focusNodes = Collections.singleton(node);
        } else {
            focusNodes = getFocusNodes(data, shape);
        }
        . . . 
```

strangepleasures:
I'm fine with the proposed changes

afs:
All done - thank you.

PR #651, opened 2019-12-12
JENA-1796: Add missing DV for XSD DateTimeStamp data type

georgschmidtdumont:


afs:
Hi @georgschmidtdumont -- looks good.

Is this taken from the Xerces2-j code base, v2.12.? branch xml-schema-1.1-dev? If possible that would be good to try to align the source code - but they may be there is a reason why not?

georgschmidtdumont:
Hi @afs ,

I did not take this from the Xerces2-j code base. I had a look at the Xerces code base and there a one or two small differences in the Xerces implementation and this implementation. I tried to align the code and it is not entirely possible at the moment since the `isXMLSchema11` flag in the `parse` method is not available in Jena implementation.
Since this flag is added in the xml-schema-1.1-dev branch in Xerces and this branch is not merged or has a PR I would suggest using the implementation in this PR and update it once the xml-schema-1.1-dev branch has been merged.
I hope this is not a blocker for this PR.

Do you know why the Xerces classes are duplicated in Jena instead of pulling Xerces as a dependency?

afs:
Jena uses Xerces code for the datatypes and used to have dependency on Xerces. That does mean that the XML parser used is Xerces because it wires itself in if on the classpath. 

Sometimes people want to use a different parser when incorporated into a larger application, and a fixed dependency gets in the way with OSGi sometimes. So to reduce size and isolate Jena's specific needs, we pulled the code out JENA-1537.

The JDK has its own XML parser - a older xerces but one which has been significantly modified since.

afs:
Merged, thanks!

PR #652, opened 2019-12-12
JENA-1797: Shapes with class-based targets (sh:targetClass) are not applied to subclass instances

strangepleasures:


afs:
I've confirmed the problem and also checked it generates a violation running it at the [SHACL playground](https://shacl.org/playground/).

The fix works as well - there are two unused imports after the change. It would be good if those were cleaned up.

The test under "std" are a copy of the W3C test suite from the working group so I think it is better to leave them untouched and add a test under the "src/test/resources/local" area.

The tests are driven from a use of RDF defined by the SHACL WG using manifests and test cases; it has its own unique style. I've taken RDF from the test case the JIRA ticket and turned it into PR #653 . 

Please remove the test file and I'll add the other PR at the same time as this one. If you could confirm I've correctly extracted the test case, that would be great.

With the fix here, the test case passes; without the fix it fails.

strangepleasures:
Thanks for reviewing. I've removed the unused imports and the extra test case. Please also take a look at my comments on your PR.

PR #653, opened 2019-12-14
JENA-1797: Test for instance of subclass of implicit target

afs:
This is a test for PR #652 , fitted into the test suite.

### DIFF
--- a/jena-shacl/src/test/resources/local/additional/target-class-subclass-1.ttl
+++ b/jena-shacl/src/test/resources/local/additional/target-class-subclass-1.ttl
@@ -1,0 +1,51 @@
+PREFIX ex:       <http://example/>
+
+PREFIX rdf:     <http://www.w3.org/1999/02/22-rdf-syntax-ns#> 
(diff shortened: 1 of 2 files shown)


jena-shacl/src/test/resources/local/additional/implicit-subclass-2.ttl
```diff
@@ -0,0 +1,51 @@
...
+
+ex:A rdf:type rdfs:Class .
+
+ex:B #rdf:type rdfs:Class ;
+     rdfs:subClassOf ex:A .
+     
+ex:x rdf:type ex:B .
+
+ex:ShapeDef
+    sh:targetClass sh:Shape ;
```
strangepleasures: Did you mean `rdf:type sh:Shape ;`?

afs: Yes, it should be sh:NodShape.

afs:
Both done, using `target-class-subclass-1` then PR squashed to one commit.

PR #661, opened 2019-12-21
1755 documentation update

Claudenw:


afs:
@Claudenw Would it be possible to format "JENA-NNNN: ...." for the PR? Then the JIRA ticket is automatically updated.

PR #662, opened 2019-12-21
JENA-1806: Update examples

afs:


jena-tdb/src-examples/tdb/examples/ExTDB_Txn2.java
```diff
@@ -18,66 +18,56 @@
...
+ */
+public class ExTDB_Txn2 {
+    public static void main(String...argv) {
+        String directory = "MyDatabases/DB1";
+        Dataset dataset = TDBFactory.createDataset(directory);
 
-        // Start WRITE transaction. 
-        //   It's possible to read from the datet inside the write transaction.
+        // Start WRITE transaction.
+        // It's possible to read from the datet inside the write transaction.
```
kinow: s/datet/dataset ?

Claudenw:
These changes are in place.  I assume this is to correct issues before moving them to the jena-examples project.  Is that corect?

afs:
The changes are orthogonal to whether to move or not.

There were comments on email and on JENA-1806 so let's discuss the details there.

The idea of moving the examples is good - needs some details and refinement next.

jena-tdb/src-examples/tdb/examples/ExTDB_Txn2.java
afs: Done!

PR #666, opened 2020-01-07
JENA-1811: Dispatch on Content-Type. Accumulated code and comment cleanup.

afs:
Deprecation of DatasetAccessor etc (Use RDFConnection).

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/atlas/web/TypedInputStream.java
+++ b/jena-arq/src/main/java/org/apache/jena/atlas/web/TypedInputStream.java
@@ -34,10 +34,6 @@ public class TypedInputStream extends FilterInputStream
     private boolean isClosed = false ;
     
     public static TypedInputStream wrap(InputStream in) {
-        //Sometimes this is used to intentional loose the content type (in tests).
-//        if ( in instanceof TypedInputStream ) {
-//            return (TypedInputStream)in;
-//        }
         return new TypedInputStream(in);
     }
     
@@ -47,11 +43,11 @@ public class TypedInputStream extends FilterInputStream
     public TypedInputStream(InputStream in, String contentType)
     { this(in, ContentType.create(contentType), null) ; }
 
-    public TypedInputStream(InputStream in, String mediaType, String charset)
-    { this(in, mediaType, charset, null) ; }
-    
-    public TypedInputStream(InputStream in, String mediaType, String charset, String baseURI)
-    { this(in, ContentType.create(mediaType, charset), baseURI) ; }
+//    public TypedInputStream(InputStream in, String mediaType, String charset)
+//    { this(in, mediaType, charset, null) ; }
+//    
+//    public TypedInputStream(InputStream in, String mediaType, String charset, String baseURI)
+//    { this(in, ContentType.create(mediaType, charset), baseURI) ; }
     
     public TypedInputStream(InputStream in, ContentType ct)
     { this(in, ct, null) ; }
@@ -63,7 +59,7 @@ public class TypedInputStream extends FilterInputStream
         this.baseURI = baseURI ;
     }
     
-    public String getContentType()          { return mediaType == null ? null : mediaType.getContentType() ; }
+    public String getContentType()          { return mediaType == null ? null : mediaType.getContentTypeStr() ; }
     public String getCharset()              { return mediaType == null ? null : mediaType.getCharset() ; }
     public ContentType getMediaType()       { return mediaType ; }
     public String getBaseURI()              { return baseURI ; }
--- a/jena-arq/src/main/java/org/apache/jena/atlas/web/TypedInputStream.java
+++ b/jena-arq/src/main/java/org/apache/jena/atlas/web/TypedInputStream.java
@@ -43,12 +43,6 @@ public class TypedInputStream extends FilterInputStream
     public TypedInputStream(InputStream in, String contentType)
     { this(in, ContentType.create(contentType), null) ; }
 
-//    public TypedInputStream(InputStream in, String mediaType, String charset)
-//    { this(in, mediaType, charset, null) ; }
-//    
-//    public TypedInputStream(InputStream in, String mediaType, String charset, String baseURI)
-//    { this(in, ContentType.create(mediaType, charset), baseURI) ; }
-    
     public TypedInputStream(InputStream in, ContentType ct)
     { this(in, ct, null) ; }
     
@@ -62,7 +56,7 @@ public class TypedInputStream extends FilterInputStream
     public String getContentType()          { return mediaType == null ? null : mediaType.getContentTypeStr() ; }
     public String getCharset()              { return mediaType == null ? null : mediaType.getCharset() ; }
     public ContentType getMediaType()       { return mediaType ; }
-    public String getBaseURI()              { return baseURI ; }
+    public String getBaseURI()             { return baseURI ; }
     public InputStream getInputStream()     { return super.in ; }
     
     @Override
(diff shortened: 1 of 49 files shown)


jena-arq/src/main/java/org/apache/jena/atlas/web/TypedInputStream.java
```diff
@@ -47,11 +43,11 @@ private TypedInputStream(InputStream in)
...
-    public TypedInputStream(InputStream in, String mediaType, String charset)
-    { this(in, mediaType, charset, null) ; }
-    
-    public TypedInputStream(InputStream in, String mediaType, String charset, String baseURI)
-    { this(in, ContentType.create(mediaType, charset), baseURI) ; }
+//    public TypedInputStream(InputStream in, String mediaType, String charset)
+//    { this(in, mediaType, charset, null) ; }
+//    
+//    public TypedInputStream(InputStream in, String mediaType, String charset, String baseURI)
+//    { this(in, ContentType.create(mediaType, charset), baseURI) ; }
```
kinow: Maybe we can delete this?

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/auth/Auth.java
```diff
@@ -103,4 +104,20 @@ public static boolean allow(String user, AuthPolicy policy, Runnable notAllowed)
...
         return false;
     }
+
+    /**
+     * Calculate the value of the "Authentication" HTTP header for basic auth. Basic
+     * auth is not secure when used over HTTP (the password can be extracted). Use
+     * with HTTPS is better.
+     * <p>
+     * Unlike digest auth, basic auth can be setup without an extra round trip to the
+     * server, making it easier for scripts where teh body is not replayable.
```
kinow: s/teh/the (I think Java 8+ javadoc is stricter and requires the closing `</p>`, unless we are passing those args to reduce strictness during build)

jena-rdfconnection/src/main/java/org/apache/jena/rdfconnection/examples/RDFConnectionExample6.java
```diff
@@ -40,4 +40,3 @@ public static void main(String ...args) {
         }
     }
 }
-
```
kinow: I thought git asked for this extra newline at the end of the file.

afs: Git copes either way (I think). These example changes are removing an extra blank line. It was "}\n\n".

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/auth/Auth.java
afs: Done

jena-arq/src/main/java/org/apache/jena/atlas/web/TypedInputStream.java
afs: Agreed.

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/auth/Auth.java
afs: Thanks for raising this. I can't find a definitive answer by searching the web.

Not sure about the `<p>`. It is legal HTML (not xHTML). I don't get a warning/error in Eclipse. The POM has `<doclint>none</doclint>` and given the scale and age of Jena, other places will have no closing `</p>`.

Elsewhere, I have projects using Java11 and no special configuration of the javadoc plugin.

namedgraph:
@afs this is another change I do not agree with, just like #759 

`DatasetAccessor` is now surely used in a bunch of depending libraries.
But more importantly, it is conceptually a sound interface, representing GSP operations -- they are not going away nor are they changing any time soon. Now all kinds of RDF access is being lumped together into the `RDFConnection`, while modular and composable interfaces like the `DatasetAccessor` are being deprecated.

I mean, why can't `RDFConnection` simply extend `DatasetAccessor`, what is the problem with that? The end result is exactly the same.

Do I need to join the dev team to protest changes like this before it's too late?..

rvesse:
https://jena.apache.org/documentation/javadoc/rdfconnection/org/apache/jena/rdfconnection/RDFConnectionModular.html

`RDFConnection` is already composed of modular interfaces and there is even a concrete implementation that can be composed as such

namedgraph:
@rvesse OK. So `RDFDatasetConnection` replaces `DatasetAccessor`? If it does the same thing, why couldn't `DatasetAccessor` be used?

rvesse:
Well actually `RDFDatasetConnection` is a generalised RDF access abstraction whereas `DatasetAccessor` was a HTTP specific interface.  As the [documentation](https://jena.apache.org/documentation/rdfconnection/) says:

> RDFConnection provides a unified set of operations for working on RDF with SPARQL operations. It provides SPARQL Query, SPARQL Update and the SPARQL Graph Store operations. The interface is uniform - the same interface applies to local data and to remote data using HTTP and the SPARQL protocols ( SPARQL protocol and SPARQL Graph Store Protocol).

So `RdfDatasetConnection` is more capable than the API it replaces

namedgraph:
More capable as in having more methods?

@rvesse I am talking about conceptually sound, uniform interfaces that both the server and the client can share and that represent a certain specification (GSP, SPARQL query/update etc.).

For example, GSP spec supports 4 methods for the default graph and 4 (with `HEAD`) for the named graph. 9 methods total - that is what `DatasetAccessor` has.

Does GSP support adding quads i.e. `Dataset`s? No. Does it specify anything re. closing connections? No.

So why start adding implementation details to a clean interface? And if quad support is necessary, why can't it have its own interface? Like this:
https://github.com/AtomGraph/Core/blob/master/src/main/java/com/atomgraph/core/model/DatasetQuadAccessor.java

afs:
`DatasetAccessor` is missing some things. Learning happens.

For example, data has to be in-memory to upload it, there is no "send a file" - the application writer hase to read into memory, and then it is serializes which is a burden on application writer and the system. The task is load a file.

Triples vs quads is really confusing.  It's HTTP, not just GSP, that matters.

I don't know how much it is used - no data one way or the other. Some change to use `RDFConnection` - and to simplify, Jena because there is limited development resource but it's hardly a radically different interface. Deprecation is the way of signalling evolution.

`RDFConnection` is focusing on user/application tasks combining all the SPARQL things, and HTTP quads, that can be performed on a store (local or remote). Placing single steps in isolation is to ask the application writer to stitch things together - and is a barrier for transactions.

----

You could look at ideas-in-progress:

https://github.com/afs/jena-http

which:
* uses Java11 using `java.net.http` (which is, at heart, an asynchronous layer with websockets and HTTP/2.
* provdes a specific GSP client if you want to go direct.
* Provides graph and model levels of (`RDFConnection/`RDFLink`) and the modular interfaces.
* Preserves `RDFConnection`.

namedgraph:
I just would very much prefer keeping specification-level and implementation-level methods separate. So that the former could potentially have multiple implementations (even if there is none in Jena) and the latter could extend them. So that we would have a stable, reusable set of "canonical" interfaces for RDF specs. If Jena is not doing that, then we'll do it ourselves.

 I like the direction of jena-http but frankly this is very much what [Core](https://github.com/AtomGraph/Core) has been doing. For example the [GSP client](https://github.com/AtomGraph/Core/blob/master/src/main/java/com/atomgraph/core/client/GraphStoreClient.java) looks _very_ similar. There are other clients in that folder.

The advantage being that `GraphStoreClient` and friends are JAX-RS compatible, while jena-http is a singular API on its own.
Given Linked Data = RDF x HTTP, a Jena x JAX-RS implementation makes _more_ than total sense to me.

Can you take a look at Core - doesn't it do what you want jena-http to do?

PR #667, opened 2020-01-07
JENA-1812 Migrated blank node hashing from MD5 to SHA-256

NSeydoux:
Resolves issue [1812](https://issues.apache.org/jira/browse/JENA-1812)


afs:
See the [comment on the JENA_1812](https://issues.apache.org/jira/browse/JENA-1812). This changes the length of blank node output in N-Triples, going from 32 to 64 hex digits.. The suggestion is to use Murmur3_128, this is the same length as MD5.

afs:
Closed in favour of #673 (use murmur3 128 bit because it is the same length as MD5).

@NSeydoux - thanks for suggesting the change.

PR #668, opened 2020-01-09
 JENA-1814: Update to Apache parent POM v22.

afs:
Also some general tidying to prepare for the 3.14.0 release: 

* Remove unnecessary <groupId> and <version> (Eclipse warnings of duplicates)
* POM fix to jena-db (old style causes a warning with Apache POM v22).
* A few plugin version updates
* Remove some code warnings.
* A javadoc fix.


afs:
Nice! Nothing comes to time as to why it would be faster except external things - maven causes network and server usage.

PR #672, opened 2020-01-16
JENA-1817: Flush NodeTableCache in CommitFinish

afs:
The relevant change is:
https://github.com/apache/jena/compare/master...afs:jena1817_tdb2_node_cache?expand=1#diff-e27c3f5f1ae576b9ebb6655fb0585cb6L328

[The test](https://github.com/apache/jena/compare/master...afs:jena1817_tdb2_node_cache?expand=1#diff-d0b153bc6d2468facb2de0d2cf244206R104) in involves waiting points.

From experience, this isn't safe long term. On build systems under load, we have seen very long pauses (many seconds) in thread scheduling. Therefore, the test may need to be tweaked or removed. It has done the necessary proof - undo the fix and the test reliably fails.

### DIFF
--- a/jena-db/jena-dboe-transaction/src/main/java/org/apache/jena/dboe/transaction/txn/TransactionCoordinator.java
+++ b/jena-db/jena-dboe-transaction/src/main/java/org/apache/jena/dboe/transaction/txn/TransactionCoordinator.java
@@ -122,20 +122,15 @@ public class TransactionCoordinator {
         this(Journal.create(location));
     }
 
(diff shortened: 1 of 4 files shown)


afs:
Great. Thanks for the quick response.

Turns out it has happened before:
https://builds.apache.org/job/Jena_Development_Deploy/org.apache.jena$jena-tdb2/1792/testReport/junit/org.apache.jena.tdb2.store/TestTransactionLifecycleTDB/transaction_concurrency_writer/

(other failures around that time were unrelated Jenkins related matters)

PR #673, opened 2020-01-17
JENA-1812: Switch to Murmur3 for blank node allocation

afs:
Use Murmur3-128bit for blank node label allocation. This code is called when a blank node appears in RDF syntax.

This is an alternative to #667 as discussed at [JENA-1812](https://issues.apache.org/jira/browse/JENA-1812) (Security software is being to flag up the use of MD5)

Murmur3-128 retains the length of blank node labels at 32 hex digits while not triggering security scans.

The murmurs implementation is the one from Apache Commons Codec. The differences between it and the Guava implementation are, for this usage, slight. For reference, the Guava version is includes in the source but the compiler does not generate any bytecode.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/lang/BlankNodeAllocatorHash.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/lang/BlankNodeAllocatorHash.java
@@ -135,7 +135,7 @@ public class BlankNodeAllocatorHash implements BlankNodeAllocator {
 
         String hexString;
         if ( true ) {
-            long[] x = MurmurHash3.hash128(input);
+            long[] x = MurmurHash3.hash128x64(input);
             // dev: String xs = String.format("%016x%016x", Long.reverseBytes(x[0]), Long.reverseBytes(x[1]));
             char[] chars = new char[32];
             longAsHexLC(x[0], chars, 0);
@@ -143,8 +143,8 @@ public class BlankNodeAllocatorHash implements BlankNodeAllocator {
             hexString = new String(chars);
         } else {
             // Guava. Several objects created.
-            // Using 104729 makes it agree with Apache Commons Codec value.
-            byte[] bytes = Hashing.murmur3_128(104729).hashBytes(input).asBytes();
+            // Using MurmurHash3.DEFAULT_SEED (which is 104729) makes it agree with Apache Commons Codec value.
+            byte[] bytes = Hashing.murmur3_128(MurmurHash3.DEFAULT_SEED).hashBytes(input).asBytes();
             hexString = Bytes.asHexLC(bytes);
         }
         return NodeFactory.createBlankNode(hexString);
--- a/jena-arq/src/main/java/org/apache/jena/riot/lang/BlankNodeAllocatorHash.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/lang/BlankNodeAllocatorHash.java
@@ -18,15 +18,12 @@
 
 package org.apache.jena.riot.lang;
 
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
 import java.util.UUID;
 import java.util.concurrent.Callable;
 
-import org.apache.jena.atlas.lib.Bytes;
-import org.apache.jena.atlas.lib.Cache;
-import org.apache.jena.atlas.lib.CacheFactory;
-import org.apache.jena.atlas.lib.InternalErrorException;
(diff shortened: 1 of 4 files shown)


jena-arq/src/main/java/org/apache/jena/riot/lang/BlankNodeAllocatorHash.java
```diff
@@ -113,20 +104,67 @@ public Node call() {
...
+        // which is the reverse of %d-formatting for a long which is 
+        // "high byte first" (in a left-to-right writing system).
+        //
+        // For byte output compatibility with byte[] from Guava,
+        // need to reverse the bytes of the longs so that it prints "low to high"
+        // Java works in big-endian -- high bytes first.
+
+        String hexString;
+        if ( true ) {
+            long[] x = MurmurHash3.hash128(input);
```
Claudenw: The original commons-codec `MurmurHash3.hash128` (like the Guava code) has a sign extension issue with the seed value.  Since the seed is 104729 (for some reason) the hashes will work correctly.  However a fair amount of work has gone into commons-codec recently and v 1.14 fixes the issue.  Because of the change several of the `MurmurHash3.hash128(x)` methods have been deprecated.  It is probably cleaner to call `MurmurHash3.hash128x86( input, 0, input.length, MurmurHash3.DEFAULT_SEED)` or  `MurmurHash3.hash128x86( input, 0, input.length, 104729L)`

afs: Thanks for the heads-up. It is also good to see the Murmur3 no longer uses `String.getBytes()`. 

Changed to jena dependency to codec 1.14 and now calling `MurmurHash3.hash128x86(x)`

In fact, v1.14 has made a change unintentionally: it does apply the mask. I'll take this to the CODEC ticket.

afs:
Both the Apache Commons and Google Guava impls are translations of the C++ code. I think (maybe Claude can confirm) that the "x86" in the method names is because Java uses the same endian-ness as x86. But maybe the int->long conversions are different - one of the two fixes in codec 1.14 is about int to long which in Java (signed numbers) does a sign extension.

Claudenw:
There are 4 implementations of Murmur3:  There are 32 and 128 bit implementations both available in x86 and x64 variants.  The x86 and x64 variants are turned for different platforms but return different values.  The version available in commons-codec is murmur3-x86-128.

The recent fix in commons-codec was a sign extension bug in seeds wherein if the seed was negative the conversion from int to long was not masked allowing the sign to extend across the upper bytes and changing the result with respect to the "C" code.  The name was also changed to reflect the actual hashing method being implemented.

jena-arq/src/main/java/org/apache/jena/riot/lang/BlankNodeAllocatorHash.java
```diff
@@ -90,6 +80,12 @@ protected UUID freshSeed() {
     @Override
     public void reset() {
         UUID seed = this.freshSeed();
+        long mostSigBits = seed.getMostSignificantBits();
+        long leastSigBits = seed.getLeastSignificantBits();
+        // Stamp on version and variant. Makes it an illegal UUID (unless all the bits are zero!)
```
rvesse: I assume this is needed because otherwise the version and variant segments of the UUID would always be the same leading to some predictability in the hashing?

afs: TBH it is not needed, it's being a little tidy saying "this is not being used as a UUID". We are using a UUID but only as convenient way to get a quality random number so large it is unique for practical purposes. Version and variant are together 6 bits (fixed), leaving 122 bits of input. 122 bits is a very large number. IIRC (e.g. 3Store paper) a few bits more that 64 is enough. 58 bits more is a lot more.

> The “Birthday Paradox”1 gives the probability of an accidental hash collision
> occurring anywhere in a knowledge base of 10^9 distinct nodes to be around
> 1 : 10^−10 when using an effective 63 bit hash.

(more details in the paper)

The requirements are uniqueness for any bnodes that are ever going to meet (assumed "global"), speed and not requiring keeping the seen labels over the life of a parser run. Prior to using the hash allocator, Jena kept labels it had seen and very occasionally a user would run out of heap due to the footprint.

rvesse: Sure I'm well aware of the birthday paradox and the parsing concerns and overheads involved, just wanted to clarify the intent of the code

PR #674, opened 2020-01-20
Geosparql precision fixes

galbiston:
Precision reduced to 6 decimal places for coordinate transformation but can be configured.
Previously failing tests now pass.
pom.xml no longer skips tests.


galbiston:
Andy, 

Since this changes the pom.xml for the GeoSPARQL module I thought you might want to check over and test.

Thanks,

Greg

jena-geosparql/pom.xml
```diff
@@ -105,7 +105,7 @@
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-surefire-plugin</artifactId>
             <configuration>
-              <skip>true</skip>
+              <skip>false</skip>
```
rvesse: If we're no longer skipping we should remove this configuration entirely as otherwise it may prevent end users from opting in to skip tests

galbiston: I've removed the `surefire` plugin configuration block.

PR #676, opened 2020-01-22
 JENA-1827: Fix thread safety issue with HashMaps in TypeMapper

superfell:
As raised on the dev mailing list, TypeMapper is a shared instance, but can mutate state stored in HashMaps, leading to potential issues.

One potential issue with this change is that HashMap allows null keys and ConcurrentHashMap does not. I've preserved the nullness behavior of TypeMappers public functions under the assumption that registering a type map for a null URI does't make sense. If that assumption is not valid, then this will need more work.


afs:
The assumption that null is not registered is correct.

PR #677, opened 2020-01-23
Cleanup

afs:
Various small things seen in passing while investigating users@ emails.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/RDFParser.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/RDFParser.java
@@ -96,7 +96,8 @@ public class RDFParser {
     private final ErrorHandler      errorHandler;
     private final Context           context;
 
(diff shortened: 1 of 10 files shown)


jena-arq/src/main/java/org/apache/jena/riot/RDFParser.java
```diff
@@ -96,7 +96,8 @@
     private final ErrorHandler      errorHandler;
     private final Context           context;
 
-    private boolean                 canUse = true;
+    // Some cases the parser is reusable (read a file), some are not (input streams).
+    private boolean                 canUseThisPrser = true;
```
kinow: `s/Prser/Parser`... That's an internal variable, so no biggie.

afs: Still worth changing. Fixed in the forced push below.

PR #678, opened 2020-01-23
JENA-1826: use plain RDF/XML in Fuseki RDF responses

osma:
The pretty (aka abbreviated) RDF/XML serialization of Fuseki RDF responses (CONSTRUCT and DESCRIBE queries) sometimes takes excessive amounts of time and CPU to generate. This PR makes Fuseki use plain RDF/XML instead, which is faster to generate.

I had to pass a `RDFFormat` instead of `Lang` to `RDFDataMgr.write()` as it allows more fine-grained control over the serialization syntax. The logic was copied from the GSP_R class.

Jira link: https://issues.apache.org/jira/browse/JENA-1826

### DIFF
--- a/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ResponseDataset.java
+++ b/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ResponseDataset.java
@@ -37,7 +37,9 @@ import org.apache.jena.query.DatasetFactory;
 import org.apache.jena.rdf.model.Model;
 import org.apache.jena.riot.Lang;
 import org.apache.jena.riot.RDFDataMgr;
+import org.apache.jena.riot.RDFFormat;
 import org.apache.jena.riot.RDFLanguages;
+import org.apache.jena.riot.RDFWriterRegistry;
 import org.apache.jena.riot.WebContent;
 import org.apache.jena.shared.JenaException;
 import org.apache.jena.web.HttpSC;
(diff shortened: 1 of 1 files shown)


jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ResponseDataset.java
```diff
@@ -113,16 +115,19 @@ public static void doResponseDataset(HttpAction action, Dataset dataset) {
         Lang lang = RDFLanguages.contentTypeToLang(contentType);
         if ( lang == null )
             ServletOps.errorBadRequest("Can't determine output content type: "+contentType);
+        // Choose the serialization. For RDX/XML use the fast, plain variant.
```
afs: `s/RDX/RDF/`

osma: Good catch! Fixed

osma:
Thanks for the quick review and approval @afs. I'll wait a few more days before merging, in case others want to comment.

afs:
Would you like me to merge this?

osma:
@afs Sure, you can merge this, thanks! I could probably do that myself too but I haven't done this for awhile and I think the processes might have changed.

afs:
Yes, it's changed. Press the big green button, update the JITA ticket. 

Now that the GH and the ASF repos are both write mode, and mirrored, the git updating can be done on either. i.e. GH tools work.

PR #679, opened 2020-01-24
JENA-1828: CommonsRDF implementation

afs:


jena-extras/jena-commonsrdf/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+    <relativePath>../..</relativePath>
+  </parent>
+
+  <name>CommonsRDF for Jena</name>
+  <description>Apache CommonsRDF Java library for Apache Jena</description>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+    <ver.commonsrdf>0.5.0</ver.commonsrdf>
```
acoburn: Perhaps add a `<automatic.module.name>org.apache.jena.commonsrdf</automatic.module.name>` property for use with Java11?

jena-extras/jena-commonsrdf/README.md
```diff
@@ -0,0 +1,14 @@
+# commonsrdf-jena
+
+Code showing using [Apache CommonsRDF](http://commonsrdf.incubator.apache.org/)
```
acoburn: I believe that the correct URL is https://commons.apache.org/proper/commons-rdf/ (commons RDF is no longer in the incubator)

jena-extras/jena-commonsrdf/README.md
```diff
@@ -0,0 +1,14 @@
...
+with [Apache Jena](http://jena.apache.org/).
```
acoburn: minor: `http://` -> `https://`

jena-extras/jena-commonsrdf/pom.xml
afs: Yes, certainly

jena-extras/jena-commonsrdf/README.md
afs: That's a bit of a give away about the original timeline of the code!

(The code has also been de-java11'ed.)

jena-extras/jena-commonsrdf/README.md
afs: Yes

jena-extras/jena-commonsrdf/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+      <version>${ver.commonsrdf}</version>
+      <classifier>tests</classifier>
+      <scope>test</scope>
+    </dependency>
+
+    <!-- Used for testing and examples -->
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-rdf-simple</artifactId>
+      <version>${ver.commonsrdf}</version>
```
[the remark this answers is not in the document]
afs: Yes, if the examples move to `src/test`. That is fine by me - eventually, they will end up in jena-examples ([JENA-1806](https://issues.apache.org/jira/browse/JENA-1806)).

afs:
> I would absolutely use this once it becomes part of a release

Which way round Jena operating as the CommonsRDF provider, or Jena code calling through CommonsRDF?

acoburn:
>> I would absolutely use this once it becomes part of a release

> Which way round Jena operating as the CommonsRDF provider, or Jena code calling through CommonsRDF?

The first method: I have library code that uses the CommonsRDF API and at present, the provider (via ServiceLoader) is JenaRDF (commons-rdf-jena), but I would immediately replace that with what you have here. I would also use this code as a bridge between the Jena-specific RDF processing (e.g. RDFParser) and the CommonsRDF APIs.

acoburn:
It would also be nice if this code had a `./src/main/resources/META-INF/services/org.apache.commons.rdf.api.RDF` file containing this line:

    org.apache.jena.commonsrdf.JenaRDF

That way, one could use this with the service loader.

afs:
ServiceLoader resource file added. I have put in a test although it is test setup dependent (both "simple" and "jena" implementations are in-scope).

acoburn:
I took a medium-sized codebase that uses `commons-rdf-jena` pretty extensively, and converted it to the CommonsRDF implementation here. The conversion was pretty straight-forward, and everything seems to work nicely. I have three suggestions for additions to this PR.

1. The `JenaCommonsRDF` class has a `o.a.c.rdf.api.Dataset fromJena(o.a.j.sparql.core.Dataset)` converter but no converter in the other direction. I would suggest adding something like this:

```java
public static org.apache.jena.sparql.core.DatasetGraph toJena(Dataset dataset) {
    if ( dataset instanceof JenaDataset )
        return ((JenaDataset)dataset).getDataset();
    org.apache.jena.sparql.core.DatasetGraph d = DatasetGraphFactory.createGeneral();
    dataset.stream().forEach(q->d.add(toJena(q)));
    return d;
}
```

2. The `JenaRDF` class in `commons-rdf-jena` also has converters to/from `o.a.c.rdf.api.Syntax`/`o.a.j.riot.Lang`. Those would be great to add to the `JenaCommonsRDF` class:

```java
public static Optional<Lang> toJena(final RDFSyntax syntax) {
    return Optional.ofNullable(RDFLanguages.contentTypeToLang(syntax.mediaType()));
}

public static Optional<RDFSyntax> fromJena(final Lang lang) {
    return RDFSyntax.byMediaType(lang.getContentType().getContentType());
}
```

3. I also tried using the `jena-commonsrdf` module in a Java11 JPMS context, and I found that the `./jena-extras/jena-commonsrdf/pom.xml` will need the following addition:

```xml
  <build>                                                                                       
    <plugins>                                                                                   
      <plugin>                                                                                  
        <artifactId>maven-jar-plugin</artifactId>                                               
        <configuration>                                                                         
          <archive>                                                                             
            <manifestEntries>                                                                   
              <Automatic-Module-Name>${automatic.module.name}</Automatic-Module-Name>           
            </manifestEntries>                                                                  
          </archive>                                                                            
        </configuration>                                                                        
      </plugin>                                                                                 
    </plugins>                                                                                  
  </build>  
```

Once this is present, the JPMS-based tests all succeed.

I have not tested this in an OSGi context, but I quite certain that some additional metadata would be required (especially to get the ServiceLoader mechanism to work in OSGi), but that could also be added later. I would be happy to work on that.

acoburn:
FYI, OSGi support can be added by using this maven configuration (in place of the `maven-jar-plugin` above -- that would no longer be needed):

```xml
  <build>
    <plugins>
      <plugin>
        <groupId>org.apache.felix</groupId>
        <artifactId>maven-bundle-plugin</artifactId>
        <extensions>true</extensions>
        <configuration>
          <instructions>
            <Automatic-Module-Name>org.apache.jena.commonsrdf</Automatic-Module-Name>
            <Bundle-DocURL>https://jena.apache.org</Bundle-DocURL>
            <Bundle-SymbolicName>org.apache.jena.commonsrdf</Bundle-SymbolicName>
            <Export-Package>org.apache.jena.commonsrdf;version=${project.version}</Export-Package>
            <Import-Package>*</Import-Package>
            <Provide-Capability>osgi.serviceloader; osgi.serviceloader=org.apache.commons.rdf.api.RDF</Provide-Capability>
            <Require-Capability>osgi.extender; filter:="(osgi.extender=osgi.serviceloader.registrar)";resolution:=optional</Require-Capability>
          </instructions>
        </configuration>
      </plugin>
    </plugins>
  </build>
```

And also setting the packaging type to `bundle`:

     <packaging>bundle</packaging>

With this configuration, the `<automatic.module.name>org.apache.jena.commonsrdf</automatic.module.name>` property isn't used, but if you'd prefer, that property can be interpolated in the maven-bundle-plugin config. (You will notice that the Java11 Automatic-Module-Name will be included in the MANIFEST.MF)

afs:
1,2,3 - yes. And thanks for the PR to the PR branch.

(I had forgotten automatic-module-name isn't inherited from the parent POM, requiring configuring the jar-plugin in each module.)

Re:OSGi - wouldn't that be better as part of apache-jena-osgi?

acoburn:
Re: OSGi - either way is fine by me. I wasn't sure if you wanted to keep the commonsrdf module separate, since it's part of `-extras`, but putting it all into `jena-osgi` has an appealing simplicity (and it should be simple to add it to the `jena-osgi` module; just make sure the `Provide-Capability` and `Require-Capability` entries are present).

afs:
I wasn't expecting it to be used quite so quickly! 

"jena-extras" has morphed a bit; it is not a few bits and pieces of interest, it's "things one level deeper in the directory hierarchy".

Done all the Java changes, and added all the extra stuff it implies. At the moment the POM is "jar". 

For OSGi, my mild preference is to put it in apache-jena-osgi so that all OSGi related stuff is together.

Would it be OK to merge as "jar", and continue the discussion for later changes?

acoburn:
> For OSGi, my mild preference is to put it in apache-jena-osgi so that all OSGi related stuff is together.

Yes, that makes sense

> Would it be OK to merge as "jar", and continue the discussion for later changes?

Yes, sounds good.

PR #681, opened 2020-01-28
JENA-1831: Fix SDB transform issue with using UNION in restrincting a query

grahamtriggs:
Encountered a case using a UNION in a query to select e.g. objects where type is asserted as X, Y or Z (but not A, B, C, etc.)

Pull request prevents the transformer from copying restrictions out of a UNION so that the algebra is correct in this case.

Includes a test for a query that exhibits the problem. All previously existing transformer tests are unmodified and continue to pass, as do all the other unit tests in SDB.

https://issues.apache.org/jira/browse/JENA-1831

### DIFF
--- a/jena-sdb/src/main/java/org/apache/jena/sdb/compiler/TransformOptimizeSubqueryFragments.java
+++ b/jena-sdb/src/main/java/org/apache/jena/sdb/compiler/TransformOptimizeSubqueryFragments.java
@@ -169,7 +169,8 @@ public class TransformOptimizeSubqueryFragments extends TransformCopy {
         private Op addStatements(Op dest, Op src, BasicPattern potentialAdditions) {
             // Only needed if destination is a Basic Graph Pattern
             if (OpBGP.isBGP(dest)) {
-                if (src instanceof Op2) {
+                // Only add statements from an Op2 if it is not a Union
+                if (src instanceof Op2 && !(src instanceof OpUnion)) {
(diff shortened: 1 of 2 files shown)


afs:
I'm thinking that anyone looking back on this in the future may want to understand it and the test case is the illustration.

Is there a simpler test case? (it does not need to be CONSTRUCT for example). Are all the WHERE clause parts necessary?

grahamtriggs:
In theory, it might be possible to cut the WHERE clause down to the following:

WHERE {
    ?publication a ?type .
    OPTIONAL { ?publication <http://localhost/access> ?access . }
    {
       ?publication a <http://localhost/Report>  .
    }
    UNION
    {
        ?publication a <http://localhost/AcademicArticle>  .
    }
}

Without the OPTIONAL element, it makes it similar to the existing UNION queries, and it transforms correctly.

It's only with the combination of OPTIONAL and UNION in this form that generates incorrectly transformed algebra without the patch applied.

afs:
I'll leave the decision to you. After all, the next person to read it may well be future-you!

grahamtriggs:
I've simplified the test slightly by removing a couple of unnecessary patterns, and cleaned it up by removing the escaped tabs in the query. Also added a comment to say why it is necessary.

Test has been checked to ensure that it fails with the previous version of the transformer, so is a valid test of the issue that is being fixed.

afs:
@grahamtriggs  - thank you!

PR #682, opened 2020-01-28
JENA-1830: Put LTS in default travis-ci setup

afs:
https://issues.apache.org/jira/browse/JENA-1830

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,6 @@ install: true
 script: mvn -B clean install
 jdk:
   - openjdk8
-  - openjdk9
-  - openjdk10
   - openjdk11
-  - openjdk12
 env:
   - JAVA_OPTS="-Xmx3072M -Xms512M -XX:+UseG1GC"


.travis.yml
```diff
@@ -4,9 +4,6 @@ install: true
 script: mvn -B clean install
 jdk:
   - openjdk8
-  - openjdk9
```
kinow: What about including the next version(s) too as allowed failures?

We use that in commons - https://github.com/apache/commons-lang/blob/master/.travis.yml#L27

afs: How do "allow_failures" get reported?  Does it still send out an email alert?

If we had travis tied to, for example, project github actions, having a (non-failing) step for openjdk-ea makes sense. So if there is a preference and migration to use GH actions/Travis more (e.g. consolidating on GH for the sake of simplicity) then obviously this all needs revisiting in a new light.

ATM The travis file is for contributors. For the project, we do have Jenkins builds that reports as well. I'm not sure whether "allow fail" makes a difference - not sure people would look!

kinow: Oh, that's a good question. As far as I can tell, they are not reported. At least I never got an e-mail for `openjdk-ea` (early-access) in Commons projects.

I think people wouldn't look. It would be more helpful for developers. But if that's handled in Jenkins, then I guess it's not necessary to have it in Travis.

PR #684, opened 2020-01-31
JENA-1833: Add jena-commonsrdf to jena-osgi

acoburn:
This adds the commonsrdf project to the jena-osgi bundle

Tested with Karaf 4.2.6

### DIFF
--- a/apache-jena-osgi/jena-osgi/pom.xml
+++ b/apache-jena-osgi/jena-osgi/pom.xml
@@ -242,6 +242,7 @@
     <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-rdf-api</artifactId>
+      <version>${ver.commonsrdf}</version>
     </dependency>
     <dependency>
       <groupId>org.osgi</groupId>
--- a/apache-jena-osgi/jena-osgi/pom.xml
+++ b/apache-jena-osgi/jena-osgi/pom.xml
@@ -242,7 +242,6 @@
     <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-rdf-api</artifactId>
-      <version>${ver.commonsrdf}</version>
     </dependency>
     <dependency>
       <groupId>org.osgi</groupId>
--- a/apache-jena-osgi/jena-osgi/pom.xml
+++ b/apache-jena-osgi/jena-osgi/pom.xml
@@ -123,7 +123,14 @@
       <version>3.15.0-SNAPSHOT</version>
       <scope>provided</scope>
     </dependency>
-    
+
+    <dependency>
+      <groupId>org.apache.jena</groupId>
+      <artifactId>jena-commonsrdf</artifactId>
+      <version>3.15.0-SNAPSHOT</version>
+      <scope>provided</scope>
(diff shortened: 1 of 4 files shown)


afs:
I am encountering a problem with this.

mvn: 3.6.3

The commons-rdf-api:tests artifact does not have a version and the commons-rdf-simple artifact has `${ver.commonsrdf}`. These seem to be needed in /pom.xml.

      <dependency>
        <groupId>org.apache.commons</groupId>
        <artifactId>commons-rdf-api</artifactId>
        <version>${ver.commonsrdf}</version>
        <classifier>tests</classifier>
      </dependency>

      <dependency>
        <groupId>org.apache.commons</groupId>
        <artifactId>commons-rdf-simple</artifactId>
        <version>${ver.commonsrdf}</version>
      </dependency>

acoburn:
Interesting. I thought the `dependencyManagement` would have taken care of this, but I've adjusted the PR to be more explicit about the commons rdf dependencies

afs:
I should have been clearer - the build problem I'm getting is in the jena-commonsrdf module. Didn't get to jena-osgi because jena-commonsrdf didn't build. The explicit versions shouldn't be necessary. 

The real problem is the commons-rdf-api:tests. The "simple" issue is a red herring.

It seems that defining in the dependencyManagement of the parent for the main artifact does not apply to `<classifier>test` -- this may be related to being in `<scope>test`.

In jena-commonsrdf/pom.xml:

    <dependency>
      <groupId>org.apache.commons</groupId>
      <artifactId>commons-rdf-api</artifactId>
      <version>${ver.commonsrdf}</version>
      <classifier>tests</classifier>
    </dependency>

works and, to me, looks tidier than in the root /pom.xml.

acoburn:
Does that [last commit](https://github.com/apache/jena/pull/684/commits/737a6e0800cf102ac087ee81df8123dcff5d9fb0) work for you? I upgraded to maven 3.6.3 locally, and it builds for me.

afs:
It is `<scope>test` items causing build problems. The default scope dependencyManagement works fine. 

This is the set of changes mentioned in the review:
[jena1833-afs.txt](https://github.com/apache/jena/files/4147238/jena1833-afs.txt)

apache-jena-osgi/jena-osgi/pom.xml
```diff
@@ -232,6 +239,11 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-rdf-api</artifactId>
+      <version>${ver.commonsrdf}</version>
```
afs: Not needed. Picking up from the parent POM worked for this one.

pom.xml
```diff
@@ -575,12 +575,6 @@
         <version>${ver.commonscli}</version>
       </dependency>
 
-      <dependency>
```
afs: This worked for the default scope artifact.

acoburn:
Thanks for the feedback. This PR should now be in line with your suggestions.

PR #687, opened 2020-02-05
JENA-1836: Update Apache Shiro to 1.5.0

afs:


afs:
Good to hear you saw it.

On another project (much smaller), I have enabled `dependabot` which sends PRs for upgrades.  It works for simple cases, including use of property `ver.shiro`.

The second Shiro-related alert looks like a secondary effect. `jena-permissions` uses `<scope>provided`.

PR #688, opened 2020-02-08
JENA-1838: Updates to PrefixMap

afs:
https://issues.apache.org/jira/browse/JENA-1838

Significant changes:
* Remove use of `IRI` in the `PrefixMap` interface
* Remove use of `IRI` in the `ParserProfile` interface
* Rename `PrefixMap.contains` => `PrefixMap.containsPrefix`

with deprecations for the originals.

The output optimizations from earlier work for Turtle family writers is still in place.


afs:
I have tried running `TestAbbreviationPerformance` with lots of different values for the performance test of `FastAbbreviatingPrefixMap`. In all cases, up to 100k prefixes, the usual prefix map, with reverse URI lookup is faster.

The DBOE prefix maps don't implement the caching reverse lookup (yet?). Managing the cache invalidation with the storage layering means the base level storage of prefixes has to do it.

The Turtle family of writers copy the prefixes from the graph to a `PrefixMapStd` (which does have the reverse lookup cache). Writing is not the general case - the data isn't being updated while a writer is active so copy works.  For small numbers of prefixes (10s-100), this startup-time copy isn't significant compared to the work of writing.

PR #691, opened 2020-02-15
 JENA-1840: Add --syntax to tdbloader (TDB1 and TDB2)

afs:


jena-cmds/src/main/java/tdb2/tdbloader.java
```diff
@@ -92,6 +97,17 @@ else if ( loadername.matches("light") )
...
+        
+        if ( super.contains(argSyntax) ) {
+            String syntax = super.getValue(argSyntax) ;
+            Lang lang$ = RDFLanguages.nameToLang(syntax) ;
+            if ( lang$ == null )
+                throw new CmdException("Can not detemine the syntax from '" + syntax + "'") ;
+            this.lang = lang$ ;
+        }
+        
+        if ( super.graphName != null )
```
rvesse: So if I specify `--syntax=LANG` but also specify `--graph=URL` then my syntax spec is ignored?

afs: Thanks for spotting that - default syntax should be set before `--syntax` is processed.

PR #698, opened 2020-02-22
JENA-1846: Add basic CORS support

afs:
This PR provides server-wide CORS support (not per-dataset).


afs:
When it's in a dev build, I'll ask users@ if that is what is needed. It does not have a setup mechanism and only uses the natural defaults, based on Barry's request.

PR #699, opened 2020-02-27
JENA-1850: Update Jetty version; make use of Shiro consistent

afs:


afs:
@Claudenw - I tried to use Shiro 1.5.0 in Fuseki but got shading errors as well as warnings running Fuseki's tests.

To make all versions consistent, I've set the ver.shiro in /pom.xml to 1.2.6 (as it has been for Fuseki for a while). jena-permissions was using 1.5.0. As a usecase for jena-permissions is in a Fuseki, same version seems important.

Is 1.2.6 for jena-permissions?
(It can be overridden in jena-permissions/pom.xml but that seems to be delaying the potential problems to downstream).

afs:
Shiro 1.5.0 is needed (JENA-1836) so fixes needed first.
Don't understand how the use of 1.2.6 isn't flagged.

PR #700, opened 2020-02-28
JENA-1850, JENA-1851: Upgrades

afs:
Various version upgrades.


afs:
shiro: This includes shiro 1.5.0 for Fuseki Full. Pre Shiro 1.4.2 has a CVE against it CVE-2019-12422.

But the build o is made messy:

- The jars are packed wrongly and shading jena-fuseki-fulljar generates a large block of warnings about overlapping classes. Looks to me like the same bytecode, copied into the shiro-core jar. (They modify the published shiro-core for some OSGi reason).
- The fuseki-webapp tests generate warnings about configuration. Shiro uses reflection to do some initialization and with duplicate code in jars, initialization is called twice on `IniRealm`.

The warnings look safe as far as I can tell.

So the choice is messy build or explore excluding dependencies of shiro-core. In theory, this is due to be fixed but it has been around for a few versions now.

This PR leaves it as "messy build".

PR #702, opened 2020-03-02
JENA-1852: Create sources-jar; add automatic module name

afs:


jena-extras/jena-querybuilder/pom.xml
```diff
@@ -29,6 +29,10 @@
     <version>3.15.0-SNAPSHOT</version>
   </parent>
 
+  <properties>
+    <automatic.module.name>org.apache.jena.querybuild</automatic.module.name>
```
acoburn: Since the artifactId and package namespace both use `querybuilder`, I think it would make sense to add `er` to the end of this module name: hence `org.apache.jena.querybuilder`

afs: Done (by saving the editor buffer).

PR #705, opened 2020-03-09
JENA-1854: Fuseki to consume the whole request body when errors occur.

afs:
This PR catches various places where incomplete processing of the request body may happen and calls a function to consume any remaining bytes in the body input stream before starting sending the HTTP error response.

### DIFF
--- a/jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
+++ b/jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
@@ -437,4 +437,22 @@ public class IO
         }
         catch (IOException ex) { IO.exception(ex); return; }
     }
+
+    // Do nothing buffer.  Never read from this, it may be corrupt because it is shared.
+    private static int SKIP_BUFFER_LEN = 64*1024;
+    private static byte[] SKIP_BUFFER = null;
+    /** Skip to the end of the InputStream, discarding input. */
+    public static void skipToEnd(InputStream input) {
+        if ( SKIP_BUFFER == null )
+            // No harm in concurrent assignment.
+            SKIP_BUFFER = new byte[SKIP_BUFFER_LEN];
+        try {
+            for(;;) {
+                // Skip does not guarantee to go to end of file.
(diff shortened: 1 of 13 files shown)


jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
```diff
@@ -437,4 +437,22 @@ public FileVisitResult postVisitDirectory(Path dir, IOException e) throws IOExce
...
     }
+
+    // Do nothing buffer.  Never read from this, it may be corrupt because it is shared.
+    private static int SKIP_BUFFER_LEN = 64*1024;
+    private static byte[] SKIP_BUFFER = null;
+    /** Skip to the end of the InputStream, discarding input. */
+    public static void skipToEnd(InputStream input) {
+        if ( SKIP_BUFFER == null )
+            // No harm in concurrent assignment.
+            SKIP_BUFFER = new byte[SKIP_BUFFER_LEN];
```
kinow: The buffer here appears to start null/uninitialized. Then on the first call to this method it is created, then populated with whatever we have in the stream. But it's never used again I think.

Couldn't we define it as a local variable in the `skipToEnd` method, and just leave it to be collected when the method is done, or `= null` it?

rvesse: I think @afs's goal here is to minimise extra memory usage by having a shared buffer that we can just dump the unread data out into potentially by many threads at once.  We're never going to read the data back out so it doesn't matter if many threads are dumping unusable data into it at once.

If you had a local variable then you'd have a buffer by thread and a potential DoS vector because a malicious client could send a load of malformed request bodies in order to get Fuseki to run up its memory usage and OOM.

afs: Yes, as @rvesse says, the buffer is a "/dev/null".

I copied the pattern from CommonsIO:`IOUtils.skip` with a bigger buffer. This also explains why `InputStream.skip` can't be used.

We could allocate a local and null it. Allocation does initialize each time. May be a bit of over engineering!

PR #707, opened 2020-03-11
JENA-1857: DatasetGraphInMemory.size.

afs:


afs:
@ajs6f - Good news! 

I think this is the only case of an "access" operation missing an in-transaction test.

jena-arq/src/main/java/org/apache/jena/sparql/core/mem/DatasetGraphInMemory.java
```diff
@@ -400,7 +400,9 @@ public DatasetPrefixStorage prefixes() {
 
     @Override
     public long size() {
-        return quadsIndex().listGraphNodes().count() ;
+        return isInTransaction() 
+                ? quadsIndex().listGraphNodes().count()
+                : calculateRead(this, ()->quadsIndex().listGraphNodes().count());
```
ajs6f:
```suggestion
                : calculateRead(this, quadsIndex().listGraphNodes()::count);
```

I just have a thing for point-free! :man_shrugging:

afs: `access` returns `Iterator<T>`.

I had tried `quadsIndex().listGraphNodes()::count` -- it does not work.

It is different because is `quadsIndex().listGraphNodes()` is called as a plain argument (i.e. early), the method pointer calculated with `::count` and that passed in so only `count()` is delayed to inside the R transaction.

`()->quadsIndex().listGraphNodes().count()` delays calling `quadsIndex().listGraphNodes()` to inside the R transaction and `listGraphNodes()` is the critical step.

ajs6f: Ah, I see. Let's merge this and I will see if anything further is worth doing when I look at 1859.

ajs6f: Actually it turned out that there was no reason to restrict `access` to `Iterator<T>`, so I sent you a [PR](https://github.com/afs/jena/pull/5) to this branch to do that.

PR #711, opened 2020-03-18
Code cleanup

afs:
This is a PR based on the code line of #710.


afs:
Yes,the logging commit is PR#710.

What this adds over that is the clean-up. It is an entry in the "most boring PR" competition.

PR #712, opened 2020-03-19
JENA-1862: Query.clone is slow

Aklakan:
This PR changes the implementation of Query.clone to make use of the syntax transform machinery.
A parameterized test case runner with 1230 queries is provided in TestQueryCloningEssentials
A test case for inital corner cases is provided in TestQueryCloningCornerCases - the main purpose is to check that modification of elements of a clone do not affect the original query. Note, that  Query.elementDataBlock is not cloned, I'd said this is an issue in QueryTransformOps.
A little commented out by means of "// @ Test" benchmark runner is provided in TestQueryCloningCornerCases as well.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/query/Query.java
+++ b/jena-arq/src/main/java/org/apache/jena/query/Query.java
@@ -32,6 +32,7 @@ import org.apache.jena.sparql.core.* ;
 import org.apache.jena.sparql.engine.binding.Binding ;
 import org.apache.jena.sparql.expr.Expr ;
 import org.apache.jena.sparql.expr.ExprAggregator ;
+import org.apache.jena.sparql.expr.ExprTransform ;
 import org.apache.jena.sparql.expr.ExprVar ;
 import org.apache.jena.sparql.expr.aggregate.Aggregator ;
 import org.apache.jena.sparql.serializer.QuerySerializerFactory ;
@@ -39,6 +40,7 @@ import org.apache.jena.sparql.serializer.SerializerRegistry ;
 import org.apache.jena.sparql.syntax.Element ;
 import org.apache.jena.sparql.syntax.PatternVars ;
 import org.apache.jena.sparql.syntax.Template ;
+import org.apache.jena.sparql.syntax.syntaxtransform.* ;
 import org.apache.jena.sparql.util.FmtUtils ;
 import org.apache.jena.sys.JenaSystem ;
 
@@ -765,14 +767,15 @@ public class Query extends Prologue implements Cloneable, Printable
     public Object clone() { return cloneQuery() ; }
 
     /**
-     * Makes a copy of this query.  Copies by parsing a query from the serialized form of this query
+     * Makes a copy of this query using the syntax transform machinery.
      * @return Copy of this query
      */
     public Query cloneQuery() {
-        // A little crude.
-        // Must use toString() rather than serialize() because we may not know how to serialize extended syntaxes
-        String qs = this.toString();
-        return QueryFactory.create(qs, getSyntax()) ;
+        ElementTransform eltTransform = new ElementTransformCopyBase(true);
+        ExprTransform exprTransform = new ExprTransformApplyElementTransform(eltTransform, true);
+
+        Query result = QueryTransformOps.transform(this, eltTransform, exprTransform);
+        return result;
     }
 
     // ---- Query canonical syntax
--- a/jena-arq/src/main/java/org/apache/jena/sparql/syntax/ElementData.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/syntax/ElementData.java
@@ -33,11 +33,18 @@ import org.apache.jena.sparql.util.NodeUtils ;
 
 public class ElementData extends Element
 {
-    private List<Var> vars = new ArrayList<>() ;
-    private List<Binding> rows = new ArrayList<>() ;
-    
+    private List<Var> vars ;
+    private List<Binding> rows ;
+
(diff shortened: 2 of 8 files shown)


afs:
I'll review ASAP especially the skipped tests.

There is a significant code update as we switch to log4j2 - this should not have any impact here but with so many files touched in some minor way, the PR may not be still merge-clean.

Also, as @rvesse is aware, the project has some other matter it needs to attend to.

afs:
From [JENA-1862](https://issues.apache.org/jira/browse/JENA-1862):

> Query.valueDataBlock is not copied by QueryTransformOps - probably
> this should be done; maybe wrapping it as an ElementData passing it
> through the given ElementTransformer?

Yes, it should be transformed and that way seems better than any that adds a public method to `Query`. I had thought of a change `Query` from a `TableData` to using an `ElementData`, then call the transform on `ElementData`, `QueryTransformOp.transformat(Query,...)` which already processes projects etc. But seems to need `Query` to have new methods to get/set it.

> QueryTransformOps has a statement that says that the top level
> element is always a ElementGroup - however the 4 excluded test
> case files parse into a Query with a ElementSubQuery as the top
> level object. What behavior is preferred in this case?

The comment isn't quite right - it can be an `ElementSubQuery`. There is a special syntax case because it does not need to be nested in { } -- it already prints them as it does when it's part of a group.

There is a bug somewhere:

```
PREFIX : <http://example/> 
SELECT * {
   ?x ?y ?z 
   { SELECT ?s { ?s ?p ?o } }
 }
```

when new-cloned and printed becomes 

```
PREFIX  :     <http://example/>

SELECT  *
WHERE
  { ?x  ?y  ?z
    { PREFIX  :     <http://example/>
      
      SELECT  ?s
      WHERE
        { ?s  ?p  ?o }
    }
  }

```
The prefixes have leaked into the sub-query.

Aklakan:
Oh, I haven't noticed the prefix leaking in testing this code, but I have seen it happen before. IIRC it may be related to a PrefixMapping2 (the one with local / global maps) getting [copied into a PrefixMappingImpl](https://github.com/apache/jena/blob/master/jena-arq/src/main/java/org/apache/jena/sparql/syntax/syntaxtransform/QueryTransformOps.java#L180). Maybe PrefixMapping needs a clone method too?

Aklakan:
Hi Andy, is there anything more I can do to support on this PR?

* The `ElementSubQuery` case can be handled by just adding it to the if statement below the incorrect comment `Top level is always a group.`
* As for wrapping the value with an element: This seems reasonable, however an element transform may yield another element instead. While this is valid for in-pattern ElementData blocks, this is not the case for a valueDataBlock. On the other hand, if one really wanted to transform a valueDataBlock to another element, one could just use a pre-transform to move the ElementData into an ElementGroup pattern.
* As for the prefixes I could have another look.

afs:
ElementSubQuery: Agreed

ElementData: Seems reason to assume ElementData to ElementData at least for the trailing part of  a query. That can't be anything else. Bundling using a temporary ElementData is fine as well.

Prefixes: I now think the best solution is to clear the prefixes during query parsing. The subquery does carry round the prefixes of the outer query. Subquery code predates SPARQL 1.1 - mayb the intent was subquery scoped prefixes.  Far too long ago to know.

The parser can clear them after finishing the sub-query element, or maybe it does not need them at all. I'll investigate.

afs:
Hi Claus,

You OK for doing ElementSubQuery and ElementData? If those are done, I'll add the prefixes fixes.

Prefix change:
https://github.com/afs/jena/commit/5e16f6717e293dd21e9221d57b76002017454cc0
and a small cleanup
https://github.com/afs/jena/commit/6dd354b697855dbf4ab7f9556815d1133101b78a

Aklakan:
Hi Andy, I will add ElementSubQuery and ElementData over the weekend

afs:
I've run into the fact that GH knows where repos are cloned from originally and haven't worked out how to have two+ clones of Jena on my account.

Here is a diff to not have prefixes when parsing a sub-select:

```
diff --git a/jena-arq/src/main/java/org/apache/jena/sparql/lang/SPARQLParserBase.java b/jena-arq/src/main/java/org/apache/jena/sparql/lang/SPARQLParserBase.java
index d3fcd7b55f..274342f90d 100644
--- a/jena-arq/src/main/java/org/apache/jena/sparql/lang/SPARQLParserBase.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/lang/SPARQLParserBase.java
@@ -221,12 +221,12 @@ public class SPARQLParserBase extends ParserBase
     protected void startSubSelect(int line, int col)
     {
         pushQuery();
-        query = newSubQuery(getPrologue()) ;
+        query = newSubQuery() ;
     }
     
-    protected Query newSubQuery(Prologue progloue)
+    protected Query newSubQuery()
     {
-        return new Query(getPrologue());
+        return new Query();
     }
     
     protected void popQuery()
```

Aklakan:
Whoa, already 2 week passed - but the ElementSubQuery and ElementData implementation is ready for review. Note, that binding objects are not cloned in the process. Probably a clone of bindings is not desired anyway, as for all practical purposes, once they are constructed, they should be considered immutable anyway (the Binding interface does not provide modification methods).

afs:
Time flies :-)

I ran the code, together with prefixes patch, and there is one test failure: `TestSyntaxTransform.subst_query_10`. This is good! It expected a subquery nested with `{{ }}` and now `{ }` is used.

Also, I chekced the prinint of queries with my previous investigation code.

All looks good!

PR #716, opened 2020-03-25
cleanup: jena-shacl 

afs:


jena-cmds/src/main/java/shacl/shacl.java
```diff
@@ -70,7 +70,7 @@ public static void main(String...args) {
             case "val": case "validate": case "v":
                 cmdExec = "validate";
                 break;
-            case "parse": case "p":
+            case "parse": case "p": case "print":
```
kinow: Oh, I didn't know we could combine `case` statements like this in Java!

afs: It's regular switch case fallthrough. The formatting makes it looks special.
Presentation is everything.

PR #718, opened 2020-03-27
JENA-1871: Check and limit the interval for ActionSleep.

afs:
Fixes for:

[JENA-1871](https://issues.apache.org/jira/browse/JENA-1871) -- ActionSleep

[JENA-1872](https://issues.apache.org/jira/browse/JENA-1872) -- async task pool

### DIFF
--- a/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/async/AsyncPool.java
+++ b/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/async/AsyncPool.java
@@ -25,6 +25,7 @@ import java.util.concurrent.*;
 
 import org.apache.jena.fuseki.Fuseki;
 import org.apache.jena.fuseki.server.DataService;
+import org.apache.jena.fuseki.servlets.ServletOps;
 
 /** The set of currently active and recently completed tasks. */
 public class AsyncPool
@@ -34,10 +35,16 @@ public class AsyncPool
     // Number of finished tasks kept.
     private static int MAX_FINISHED = 20;
 
-    // See Executors.newCachedThreadPool and Executors.newFixedThreadPool
+    // A ThreadPoolExecutor with
+    // * 0 to nMaxThreads
+    // * no queue of waiting tasks 9tasks execute or are rejected)
+    // * dormant threads released after 120s.
+    //
+    // SynchronousQueue is a BlockingQueue that has zero length - it accepts and
+    // delivers an item or rejects immediately, no delay by queueing.
     private ExecutorService executor = new ThreadPoolExecutor(0, nMaxThreads,
                                                               120L, TimeUnit.SECONDS,
-                                                              new LinkedBlockingQueue<Runnable>());
+                                                              new SynchronousQueue<Runnable>(true));
     private final Object mutex = new Object();
     private long counter = 0;
     private Map<String, AsyncTask> runningTasks = new LinkedHashMap<>();
@@ -63,9 +70,14 @@ public class AsyncPool
                 return null;
             };
             AsyncTask asyncTask = new AsyncTask(c, this, taskId, displayName, dataService, requestId);
-            /* Future<Object> future = */ executor.submit(asyncTask);
-            runningTasks.put(taskId, asyncTask);
-            return asyncTask;
+            try { 
+                /* Future<Object> future = */ executor.submit(asyncTask);
+                runningTasks.put(taskId, asyncTask);
+                return asyncTask;
+            } catch (RejectedExecutionException ex) {
+                ServletOps.errorBadRequest("Async task request rejected");
+                return null;
+            }
         }
     }
 
--- a/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/async/AsyncPool.java
+++ b/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/async/AsyncPool.java
@@ -37,7 +37,7 @@ public class AsyncPool
 
     // A ThreadPoolExecutor with
     // * 0 to nMaxThreads
-    // * no queue of waiting tasks 9tasks execute or are rejected)
+    // * no queue of waiting tasks (tasks execute or are rejected)
     // * dormant threads released after 120s.
     //
     // SynchronousQueue is a BlockingQueue that has zero length - it accepts and
@@ -70,12 +70,12 @@ public class AsyncPool
                 return null;
             };
             AsyncTask asyncTask = new AsyncTask(c, this, taskId, displayName, dataService, requestId);
-            try { 
+            try {
                 /* Future<Object> future = */ executor.submit(asyncTask);
                 runningTasks.put(taskId, asyncTask);
                 return asyncTask;
             } catch (RejectedExecutionException ex) {
-                ServletOps.errorBadRequest("Async task request rejected");
+                ServletOps.errorBadRequest("Async task request rejected - exceeds the limit of "+nMaxThreads+" tasks");
                 return null;
             }
         }
--- a/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionSleep.java
+++ b/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionSleep.java
@@ -32,6 +32,8 @@ import org.slf4j.Logger;
 /** A task that kicks off a asynchronous operation that simply waits and exits.  For testing. */
 public class ActionSleep extends ActionCtl /* Not ActionAsyncTask - that is a container-item based. */
 {
+    private static int MaxSleepMillis = 20*1000;
+    
     public ActionSleep() { super(); }
 
     @Override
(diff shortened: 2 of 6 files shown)


jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/async/AsyncPool.java
```diff
@@ -34,10 +35,16 @@
     // Number of finished tasks kept.
     private static int MAX_FINISHED = 20;
 
-    // See Executors.newCachedThreadPool and Executors.newFixedThreadPool
+    // A ThreadPoolExecutor with
+    // * 0 to nMaxThreads
+    // * no queue of waiting tasks 9tasks execute or are rejected)
```
kinow: s/9tasks/(tasks

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionSleep.java
```diff
@@ -32,6 +32,8 @@
 /** A task that kicks off a asynchronous operation that simply waits and exits.  For testing. */
 public class ActionSleep extends ActionCtl /* Not ActionAsyncTask - that is a container-item based. */
 {
+    private static int MaxSleepMillis = 20*1000;
```
kinow: `maxSleepMillis`?

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionSleep.java
```diff
@@ -50,45 +52,57 @@ public void validate(HttpAction action) {}
...
+    protected SleepTask createRunnable(HttpAction action) {
         String interval = action.request.getParameter("interval");
         int sleepMilli = 5000;
-        if ( interval != null )
+        if ( interval != null ) {
             try {
                 sleepMilli = Integer.parseInt(interval);
             } catch (NumberFormatException ex) {
-                action.log.error(format("[%d] NumberFormatException: %s", action.id, interval));
+                ServletOps.errorBadRequest("Bad format for 'interval': integer required");
```
kinow: Maybe use the same prefix for all errors?

`"Bad format for 'interval': integer required"`

`"Bad format for 'interval': negative sleep interval"`

`"Bad format for 'interval': sleep internal greater than maximum allowed"`

?

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionSleep.java
```diff
@@ -50,45 +52,57 @@ public void validate(HttpAction action) {}
...
 
         @Override
         public void run() {
             try {
-                log.info(format("[%d] >> Sleep start", actionId));
-                Lib.sleep(sleepMilli);
-                log.info(format("[%d] << Sleep finish", actionId));
+                log.info(format("[Task %d] >> Sleep start", actionId));
+                for ( int i = 0 ; i < 10 ; i++ )
+                    Lib.sleep(sleepMilli/10);
```
kinow: Out of curiosity, why `/ 10`? Is it to yield control so other things get a chance to run after every 1/10th of the sleep time?

afs: Yes - that's the reason. It is dev code put in to nudge the JVM along when running all in in JVM (client and Fuseki embedded server). I find that it sometimes shows up concurrency issues better and I was trying multiple "sleeps" running.

It isn't supposed to be there for the codebase and I'll remove it - good thing we have code review.

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionSleep.java
afs: Good idea.

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionSleep.java
afs: Logically, it's a constant. I'll add `final`.  ALL_CAPS_IS_A_BIT_SHOUTY :-)

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/async/AsyncPool.java
afs: Will fix!

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionSleep.java
```diff
@@ -32,6 +32,8 @@
 /** A task that kicks off a asynchronous operation that simply waits and exits.  For testing. */
 public class ActionSleep extends ActionCtl /* Not ActionAsyncTask - that is a container-item based. */
 {
+    private static final int MaxSleepMillis = 20*1000;
```
rvesse: Could we make this public then the test case can refer to the constant rather than hard-coding it

jena-fuseki2/jena-fuseki-webapp/src/test/java/org/apache/jena/fuseki/TestAdmin.java
```diff
@@ -278,7 +278,29 @@
...
+            String x = execSleepTask(null, -1);
+            fail("Sleep call unexpectedly succeed");
+        } catch (HttpException ex) {
+            assertEquals(400, ex.getStatusCode());
+        }
+    }
+
+    @Test public void sleep_3() {
+        try {
+            String x = execSleepTask(null, 20*1000+1);
```
rvesse: Would be nicer if this just referred to the constant in `ActionSleep` so if we ever change that the test doesn't break

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/async/AsyncPool.java
```diff
@@ -34,10 +35,16 @@
...
     private static int MAX_FINISHED = 20;
 
-    // See Executors.newCachedThreadPool and Executors.newFixedThreadPool
+    // A ThreadPoolExecutor with
+    // * 0 to nMaxThreads
+    // * no queue of waiting tasks (tasks execute or are rejected)
+    // * dormant threads released after 120s.
+    //
+    // SynchronousQueue is a BlockingQueue that has zero length - it accepts and
+    // delivers an item or rejects immediately, no delay by queueing.
```
rvesse: +1 So provided there is at least one thread free then submitting a task will succeed, if all threads are in-use then it fails

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/async/AsyncPool.java
```diff
@@ -63,9 +70,14 @@ public AsyncTask submit(Runnable task, String displayName, DataService dataServi
...
             AsyncTask asyncTask = new AsyncTask(c, this, taskId, displayName, dataService, requestId);
-            /* Future<Object> future = */ executor.submit(asyncTask);
-            runningTasks.put(taskId, asyncTask);
-            return asyncTask;
+            try {
+                /* Future<Object> future = */ executor.submit(asyncTask);
+                runningTasks.put(taskId, asyncTask);
+                return asyncTask;
+            } catch (RejectedExecutionException ex) {
+                ServletOps.errorBadRequest("Async task request rejected");
```
rvesse: Should this be more specific so users understand why it was rejected e.g. add something like ` - already <maxTasks> in progress`

jena-fuseki2/jena-fuseki-webapp/src/test/java/org/apache/jena/fuseki/TestAdmin.java
```diff
@@ -278,7 +278,29 @@
...
+            fail("Sleep call unexpectedly succeed");
+        } catch (HttpException ex) {
+            assertEquals(400, ex.getStatusCode());
+        }
+    }
+
+    // Async task testing
```
rvesse: Do we want some additional test cases that validate that we can now have multiple simultaneous async tasks?  I see test cases that cover JENA-1871 but not any that cover JENA-1872

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionSleep.java
afs: OK

jena-fuseki2/jena-fuseki-webapp/src/test/java/org/apache/jena/fuseki/TestAdmin.java
afs: Tricky. Any ideas how to?

I can investigate reading the tasks list for active tasks but we have situations where loaded CI servers do weird things (pauses of 10s of seconds or longer - I guess the VM isn't get real CPU time) so "overlap " is hard to ensure in a reliable way, including with the client code in the `@Test` itself.

Having "works mostly" is bad -  it it then natural to ignore test failures because they become "expected".

I have tested locally, including hitting the limit.

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/async/AsyncPool.java
afs: OK

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/async/AsyncPool.java
afs: Yes - like that.  The actual mechanism is expands up to nMaxThreads, creating threads if needed.

jena-fuseki2/jena-fuseki-webapp/src/test/java/org/apache/jena/fuseki/TestAdmin.java
afs: I've pushed two tests.

If they become unreliable in practice, I'd prefer that we comment out the `@Test`.

rvesse: LGTM

PR #721, opened 2020-04-01
JENA-1878: switched API call to prefix.cc from http to https

DavidFichtmueller:
This caused a mixed-content error and the call was blocked by the browser (thereby rendering the entire prefix.cc lookup as useless) when Fuseki was run via https.


afs:
@DavidFichtmueller  - thanks!

PR #724, opened 2020-04-08
JENA-1880: ResultSetWriterJSON using Nodes

afs:


afs:
This will get revised again when RDF* support is introduced.

PR #729, opened 2020-04-18
JENA-1884: remove JenaUUID

afs:
And JENA-1885.


jena-base/src/main/java/org/apache/jena/atlas/lib/BitsInt.java
```diff
@@ -155,14 +155,27 @@ boolean test(int bits, int value, int start, int finish)
...
+     * Clear the bit specified.
+     *  @param bits
+     *  @param bitIndex
+     *  @return int
+     */
+    public static final
+    int clear(int bits, int bitIndex)
+    {
+        check(bitIndex);
+        return clear$(bits, bitIndex) ;
```
kinow, line 176: `clear$` already does the `check(bitIndex)`. Would it be better to have it only there, or is it better to be safe and perform that verification on both methods?

jena-base/src/main/java/org/apache/jena/atlas/lib/BitsLong.java
```diff
@@ -157,22 +157,35 @@ boolean test(long bits, long value, int start, int finish)
...
+    /**
+     * Clear the bit specified.
+     *  @param bits
+     *  @param bitIndex
+     *  @return long
+     */
+    public static final
+    long clear(long bits, int bitIndex)
+    {
+        check(bitIndex) ;
```
kinow, line 177: Ditto above about `check` in `clear$`

jena-base/src/main/java/org/apache/jena/atlas/lib/BitsInt.java
afs: Fixed.

jena-base/src/main/java/org/apache/jena/atlas/lib/BitsLong.java
afs: Can't find that one but what I have done is make the files exactly the same except s/int/long/. They now line up in diff and (easier to see!) line length.

PR #731, opened 2020-04-21
JENA-1888: Mark log4j2 optional in OSGi bundle

acoburn:
This excludes the log4j2 dependency from the OSGi bundle metadata

### DIFF
--- a/apache-jena-osgi/jena-osgi/pom.xml
+++ b/apache-jena-osgi/jena-osgi/pom.xml
@@ -269,6 +269,7 @@
               !com.ibm.uvm.tools,
               !com.sun.jdmk.comm,
               !org.apache.log,
+              !org.apache.logging.*,
               !org.apache.jena.ext.*,
               !org.checkerframework.checker.*,
               !com.google.errorprone.annotations.*,
--- a/apache-jena-osgi/jena-osgi/pom.xml
+++ b/apache-jena-osgi/jena-osgi/pom.xml
@@ -269,12 +269,12 @@
               !com.ibm.uvm.tools,
               !com.sun.jdmk.comm,
               !org.apache.log,
-              !org.apache.logging.*,
               !org.apache.jena.ext.*,
               !org.checkerframework.checker.*,
               !com.google.errorprone.annotations.*,
               !com.google.appengine.api,
               !com.google.apphosting.api,
+              org.apache.logging.*;resolution:=optional,
               sun.misc;resolution:=optional,
               *
             </Import-Package>


afs:
For my information: log4j and slf4j-log4j12 weren't excluded in the past so has something changed that log4j2 is being picked up? If so, then maybe the issue is earlier in the process.

acoburn:
I looked a little deeper into this. The issue is related to the `org.apache.logging.log4j.core` package, which isn't exported by the pax-logger implementation that is part of the newer Karaf platforms. (Running `package:exports | grep -i log4j` in the different versions of Karaf will give you this information).

The simple work-around is to just install the required bundle:

```
bundle:install -s mvn:org.apache.logging.log4j/log4j-core/2.13.1
```

(or the equivalent in a downstream features file).

Either way, it seems that this PR is _not_ the right approach, and this should not be merged. The open question is just whether this dependency should be added to the Jena-provided features.xml file or just leave things as they currently are, but documenting how this works.

kinow:
@acoburn just random question, would it be possible to use Jenkins/Travis/GH actions/etc to start a container with Karaf, build Jena from sources, deploy and test to see if any change broke OSGi? Or is it something that needs manually testing and reviewing the dependencies (only read about OSGi, and had very little contact while using jboss/wildfly).

acoburn:
@kinow yes, there are ways to automate the testing of OSGi. Typically, one uses [pax-exam](https://ops4j1.jira.com/wiki/spaces/PAXEXAM4/overview) for this. In fact, there is already [some code](https://github.com/apache/jena/blob/master/apache-jena-osgi/jena-osgi-test/src/test/java/org/apache/jena/osgi/test/JenaOSGITest.java) in the Jena source tree that uses pax-exam for this sort of test. That test code, however, is not currently enabled, and it is a bit out-of-date, but it would be quite possible to restructure that a bit to automate the OSGi testing.

kinow:
> @kinow yes, there are ways to automate the testing of OSGi. Typically, one uses [pax-exam](https://ops4j1.jira.com/wiki/spaces/PAXEXAM4/overview) for this. In fact, there is already [some code](https://github.com/apache/jena/blob/master/apache-jena-osgi/jena-osgi-test/src/test/java/org/apache/jena/osgi/test/JenaOSGITest.java) in the Jena source tree that uses pax-exam for this sort of test.

First time I heard about pax-exam, looks interesting, thanks @acoburn . 

>That test code, however, is not currently enabled, and it is a bit out-of-date, but it would be quite possible to restructure that a bit to automate the OSGi testing.

Should we create a JIRA ticket for that? Maybe one of us could work re-enabling it later, and make it easier to spot and fix issues with OSGi.

afs:
JENA-913 "mvn clean verify" did not work but I can't understand the thread and links to know what the problem really was. 

Ah - maybe - the preflight jenkins job "Jena_Development_Test" - and jena-osgi only works off installed, not built-in-job, artifacts which i think "mvn verify" normally does.

Another option would to run a separate job that checks OSGi and get it to run after a successful build.

acoburn:
I spent some time familiarizing myself with the log4j2 code dependency in jena-base. Because log4j-api and log4j-core are marked as optional in that module's pom.xml (and given how the code is structured in LogCtl.java), it seems more correct for those OSGi imports to be marked as optional. To me, this seems like a better solution than excluding the imports; plus, it mirrors the dependency structure of jena-base itself.

PR #732, opened 2020-04-21
JENA-1887: SplitIRI fix

afs:


jena-core/src/main/java/org/apache/jena/util/SplitIRI.java
```diff
@@ -136,45 +135,66 @@ private static boolean needsEscape(char ch, boolean finalChar) {
...
+        if ( idx1 >= 0 && idx2 < 0 )
+            // no "/" (or ":" if a urn)
+            limit = idx1;
+        else if ( idx1 < 0 && idx2 >= 0 )
+            // no fragment
+            limit = idx2;
+        else
+            // Fragment and path.
+            // If "/" is in the fragment, it is not the split.
+            if ( idx1 >= 0 && idx2 >= 0 )
```
kinow, line 180: Another interesting way to format code with comments! For one second thought we had two `else` statements. Nice!

PR #733, opened 2020-04-23
JENA-1861: Fixes for use of a single query object in parallel executions.

afs:
cc @Aklakan


afs:
@Claudenw - FYI - added some "setResultVars" to query builder as it finishes a query. These should make no difference.

I would do the same for this PR (i.e. call "setResultVars" as parsing finishes ) but it disturbs the syntax transform tests. It is the tranfsormation or tests (which use the parser) that are at fault and this is spun off into [JENA-1889](https://issues.apache.org/jira/browse/JENA-1889).

PR #738, opened 2020-04-30
JENA-1891: Correct StrMin and StrMax messages

afs:
[JENA-1891](https://issues.apache.org/jira/browse/JENA-1891)


jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/StrMinLengthConstraint.java
```diff
@@ -45,7 +45,7 @@ public ReportItem validate(ValidationContext vCxt, Node n) {
         String str = NodeFunctions.str(n);
         if ( str.length() >= minLength )
             return null;
-        String msg = toString()+": String too long: "+str;
+        String msg = toString()+": String not long enough: "+str;
```
lrncfly, line 48: `String too short` would be better in my opinion

afs:
Closed in favour of #739.

PR #741, opened 2020-05-08
JENA-1881: Add being able to read variations of RDF* JSON syntax.

afs:
The JSON result form changes are not defined in RDF* so there are variations.

This PR adds being about to read RDF4J variation of JSON results.

Also some cleanups that got noticed.


jena-arq/src/main/java/org/apache/jena/riot/resultset/rw/ResultSetReaderJSON.java
```diff
@@ -239,20 +239,30 @@ private static Node parseOneTerm(JsonObject term, LabelToNode labelMap) {
...
+            JsonObject oTerm = get(term, kObject, kObjectAlt);
             if ( sTerm == null || pTerm == null || oTerm == null )
                 throw new ResultSetException("Bad triple term: "+term);
             Node s = parseOneTerm(sTerm, labelMap);
             Node p = parseOneTerm(pTerm, labelMap);
             Node o = parseOneTerm(oTerm, labelMap);
             return NodeFactory.createTripleNode(s, p, o);
         }
+        
+        // Get a object from an object - use the first name inthe fields list
```
kinow, line 257: s/inthe/in the

PR #742, opened 2020-05-11
JENA-1881: Support RDF* triple terms in SPARQL XML results

afs:
In the same nested style as the JSON results: 

        {
          "type": "triple" , 
          "value": {
            "subject":   { "type": "bnode" , "value": "b0" } ,
            "property": { "type": "uri" , "value": "http://example/p" } ,
            "object":    { "type": "literal" , "datatype": "http://www.w3.org/2001/XMLSchema#integer" , "value": "123" }
          }

is in SPARQL XML results:

        <triple>
          <subject>
            <bnode>b0</bnode>
          </subject>
          <property>
            <uri>http://example/p</uri>
          </property>
          <object>
            <literal datatype="http://www.w3.org/2001/XMLSchema#integer">123</literal>
          </object>
        </triple>


jena-arq/src/main/java/org/apache/jena/riot/resultset/rw/ResultSetWriterXML.java
```diff
@@ -253,84 +254,138 @@ public void binding(String varName, RDFNode node) {
...
-                String datatype = literal.getDatatypeURI();
+                String datatype = literal.getLiteralDatatypeURI();
                 out.print(" ");
                 out.print(dfAttrDatatype);
                 out.print("=\"");
                 out.print(datatype);
                 out.print("\"");
             }
-
+
```
kinow, line 350: Extra white space

jena-arq/src/main/java/org/apache/jena/riot/resultset/rw/ResultSetWriterXML.java
```diff
@@ -253,84 +254,138 @@ public void binding(String varName, RDFNode node) {
...
+            out.println(">");
+        }            
+            
+        private void printBlankNode(Node node) {
+            String label = bNodeMap.get(null, node);
+            // Comes with leading "_:"
+            label = label.substring(2);
+            out.print("<");
+            out.print(dfBNode);
+            out.print(">");
```
kinow, line 323: The new methods `openTag`/`closeTag` cannot be used in `printBlankNode` too, as in `printTripleTerm`? (asking because I noticed that method also increments indentation, so maybe not supposed to be always used?)

jena-arq/src/main/java/org/apache/jena/riot/resultset/rw/ResultSetWriterXML.java
afs: I didn't remove all trailing WS because it is all over the place.

jena-arq/src/main/java/org/apache/jena/riot/resultset/rw/ResultSetWriterXML.java
afs: The XML write is very, very, _very_ old! I can't even track its history in git properly. 

The new `openTag`/`closeTag` made writing the nested structures for `printTripleTerm`.

I tried to limit the scope of changes and not do a whole-class clear-up, though I was sorely tempted to because the code is very un-DRY.

PR #743, opened 2020-05-14
JENA-1881: Model API changes for RDF*

afs:
This PR adds triple term to the mode API by making them a kind of "Resource", along side URI and blank node.

This is the only way to incorporate triple terms without a major change to the type-safe API because they can appear in the subject position.

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/rdf/model/ResourceFactory.java
+++ b/jena-core/src/main/java/org/apache/jena/rdf/model/ResourceFactory.java
@@ -53,14 +53,14 @@ public class ResourceFactory {
     private ResourceFactory() {
     }
 
-    /** get the current factory object.
+    /** Get the current factory object.
      * 
      *  @return the current factory object
      */
     public static Interface getInstance() {
         return instance;
     }
-    /** set the current factory object.
+    /** Set the current factory object.
      * 
      * @param newInstance the new factory object
      * @return the previous factory object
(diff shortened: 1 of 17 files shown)


jena-core/src/main/java/org/apache/jena/rdf/model/ResourceFactory.java
```diff
@@ -92,6 +92,15 @@ public static Resource createResource(String uriref) {
         return instance.createResource(uriref);
     }
     
+    /** 
+     * create a new resource representing an RDF* triple term.
```
kinow, line 96: s/create/Create

jena-core/src/main/java/org/apache/jena/rdf/model/ResourceFactory.java
```diff
@@ -212,6 +221,13 @@ public static Statement createStatement(
          */
         public Resource createResource(String uriref);
         
+        /** create a new resource representing an RDF* triple term.
```
kinow, line 224: Uppercase create? (Other methods seem to start with upper case...)

jena-core/src/main/java/org/apache/jena/rdf/model/ResourceFactory.java
afs: Many of the javadoc comments are like that.  ("Answer" but "create")

Redid the commit : I've gone through the javadoc on all methods and tidied up.

kinow: Sorry, I had looked only at a couple comments available in the GitHub interface.

afs: No problem - usual balance between focused changes for the ticket addressed and a desire to sort out things in passing.

PR #747, opened 2020-05-21
JENA-1900: Ensure only one Log4j2Plugins.dat file shaded

afs:


jena-fuseki2/apache-jena-fuseki/log4j2.properties
```diff
@@ -47,8 +47,8 @@ logger.jetty.name  = org.eclipse.jetty
 logger.jetty.level = WARN
 
 # Hide bug in Shiro 1.5.0
-logger.shiro.name = org.apache.shiro.realm.text.IniRealm
-logger.shiro.level = ERROR
```
afs, line 51: Fix logging config mistake - should set for "org.apache.shiro" (silences two log lines as the server starts).

jena-fuseki2/jena-fuseki-fulljar/pom.xml
```diff
@@ -38,6 +38,12 @@
       <groupId>org.apache.jena</groupId>
       <artifactId>jena-fuseki-webapp</artifactId>
       <version>${project.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.logging.log4j</groupId>
+          <artifactId>log4j-web</artifactId>
```
afs, line 44: Exclude `org.apache.logging.log4j:log4j-web` -- it isn't necessary running as a single app in Jetty because there isn't a stop.unload issue. But it does contain another Log4j2Plugins.dat and this was the problem - shaing picked one, not merged them.

The general fix - the 3rd party shade transformer - didn't work reliably so the simple fix is to exclude the artifact in `jena-fuseki-fulljar`. The transformer is also very version sensitive.

The emergence fix script put in the log4j-core Log4j2Plugins.dat and does not include the log4j-web additions -- so evidence that the log4j-web isn't needed.

jena-fuseki2/jena-fuseki-fulljar/pom.xml
```diff
@@ -73,15 +79,26 @@
...
+                 org.apache.logging.log4j:log4j-core and, unless excluded,
+                 one via jena-fuseki-webapp which has org.apache.logging.log4j:log4j-core-web
+                 If it becomes necessary, see: https://github.com/edwgiz/maven-shaded-log4j-transformer
+            -->
             <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
               <mainClass>org.apache.jena.fuseki.cmd.FusekiCmd</mainClass>
+              <!-- https://issues.apache.org/jira/browse/LOG4J2-2537  -->
+              <manifestEntries>
+                <Multi-Release>true</Multi-Release>
+              </manifestEntries>
```
afs, line 101: Logging setup when run on a java9+ JVM was outputting a warning. See LOG4J2-2537.

pom.xml
```diff
@@ -54,7 +54,8 @@
     <ver.junit>4.12</ver.junit>  
 
     <ver.slf4j>1.7.30</ver.slf4j>
-    <ver.log4j2>2.13.1</ver.log4j2>
+    <ver.log4j2>2.13.3</ver.log4j2>
+    <ver.shade-plugin>3.2.3</ver.shade-plugin>
```
afs, line 58: In trying to get the 3rd party shade transformer to work, upgrading the shade plugin and the log4j2 version were necessary so evenif not required after all, the x.x.1 updates are left in.

PR #749, opened 2020-05-28
JENA-1902: Set default logging to output to stderr

afs:


afs:
> Is this worth a bugfix release?

Yes, while leaving some time for any other problems to get reported.

While there is workaround to this issue, this isn't the only problem with 3.15.0 so if the PMC has the energy, I think we should do a fairly quick release cycle for 3.16.0.  

RDF* is working for TDB1 and TDB2 now (even if I can now see a better way to do SPARQL* for all situations ... but it is a deeper, intrusive change to query execution so should not be rushed).

PR #750, opened 2020-05-29
JENA-1904: The "defaultGraph" parameter is ignored here, so the sparqlService(…

dacasals:
…String service, Query query, String defaultGraph) method not filter the subGraph. This change fix the problem calling the other method.


afs:
@dacasals -- Thank you!

PR #753, opened 2020-06-06
JENA-1912: Fix LogCtl and non-Log4J2 provider. Logging cleanup.

afs:


afs:
Many of the file changes are due to a rename of `LogCtl.setCmdLogging` to `LogCtl.setLogging` (with deprecation of the old name).

PR #755, opened 2020-06-07
JENA-1913: Multiple --file on the Fuseki command line

afs:


jena-fuseki2/jena-fuseki-webapp/src/main/java/org/apache/jena/fuseki/cmd/FusekiCmd.java
```diff
@@ -246,21 +247,26 @@ protected void processModulesAndArgs() {
...
+                List<String> filenames = getValues(argFile);
                 // Directly populate the dataset.
                 cmdLineConfig.reset();
                 cmdLineConfig.dsg = DatasetGraphFactory.createTxnMem();
-                Lang language = RDFLanguages.filenameToLang(filename);
-                if ( language == null )
-                    throw new CmdException("Can't guess language for file: " + filename);
-                Txn.executeWrite(cmdLineConfig.dsg, ()->RDFDataMgr.read(cmdLineConfig.dsg, filename));
+                String desc = filenames.size() == 1
+                    ? "file: "+filenames.get(1)
```
kinow, line 255: .get(0)? (Spent the day looking at JS and Python, and on mobile, so I might be confusing langs... if so ignore please)

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/cmds/FusekiMain.java
```diff
@@ -282,21 +282,26 @@ protected void processModulesAndArgs() {
...
-            serverConfig.dsg = DatasetGraphFactory.createTxnMem();
-
-            // INITIAL DATA.
-            Lang language = RDFLanguages.filenameToLang(filename);
-            if ( language == null )
-                throw new CmdException("Can't guess language for file: " + filename);
-            Txn.executeWrite(serverConfig.dsg,  ()->RDFDataMgr.read(serverConfig.dsg, filename));
+            List<String> filenames = getValues(argFile);
+            String desc = filenames.size() == 1
+                ? "file: "+filenames.get(1)
```
kinow, line 287: Isn't the parameter the index of the element? Or is it the number of elements to pop/shift?

jena-fuseki2/jena-fuseki-webapp/src/main/java/org/apache/jena/fuseki/cmd/FusekiCmd.java
afs: Arrg! `.get(0)` both times (quick fix applied).

The messages need refining as well. Logging works different in the two different server startup sequences ("legacy" - sigh).

afs:
Better information, same style output in each server. Better handling of bad syntax.

PR #759, opened 2020-06-15
JENA-1917: FileManager as interface; Deprecate FileManager.get()

afs:


jena-core/src/main/java/org/apache/jena/ontology/OntDocumentManager.java
```diff
@@ -54,16 +54,14 @@
  * example:
  * </p>
  * <pre>OntDocumentManager dm = OntDocumentManager.getInstance();
- * dm.setFileManager( FileManager.get() );</pre>
- * <p>Note that in Jena 2.3, we have deprecated the capability of the document manager
- * to store a table of known prefixes, and a table mapping document URI's to ontology language
- * types. <strong>The intention is to remove both of these capabilities from
- * Jena 2.4 onwards</strong>. If this change would be problematic, please send email to the
- * <a href="http://groups.yahoo.com/group/jena-dev">Jena support
```
kinow, line 62: Oh, the list used to be on Yahoo! Groups? That was probably a long time ago?

jena-core/src/main/java/org/apache/jena/n3/N3JenaWriter.java
```diff
@@ -105,16 +80,20 @@ N3JenaWriterCommon chooseWriter()
...
+        if ( writerName.equals("N3") )
             return new N3JenaWriterPP() ;
         
-        if ( writerName.equalsIgnoreCase(n3WriterPlain) )
-            return new N3JenaWriterCommon() ;
-        
-        if ( writerName.equalsIgnoreCase(n3WriterTriples) ||
-             writerName.equalsIgnoreCase(n3WriterTriplesAlt) )
-            return new N3JenaWriterTriples() ;
+//        if ( writerName.equals(n3WriterPrettyPrinter) )
```
kinow, line 88: I think the only difference now is that `writerName.equals(n3WriterPrettyPrinter)` used to work, and now it won't? But probably no harm I guess?

jena-core/src/main/java/org/apache/jena/util/FileManagerImpl.java
```diff
@@ -0,0 +1,566 @@
...
+ * 
+ * A FileManager provides access to named file-like resources by opening
+ * InputStreams to things in the filing system, by URL (http: and file:) and
+ * found by the classloader.  It can also load RDF data from such a system
+ * resource into an existing model or create a new (Memory-based) model.
+ * There is a global FileManager which provide uniform access to system
+ * resources: applications may also create specialised FileManagers.
+ * 
+ * A FileManager contains a list of location functions to try: the global
+ * FileManger has one {@link LocatorFile}, one {@link LocatorClassLoader} and
```
kinow, line 44: s/`FileManger`/`FileManager`

jena-core/src/main/java/org/apache/jena/util/FileManagerImpl.java
```diff
@@ -0,0 +1,566 @@
...
+ * one {@link LocatorURL}
+ * 
+ * Main operations:
+ *  * <ul>
```
kinow, line 48: Extra `*`

jena-core/src/main/java/org/apache/jena/util/FileManagerImpl.java
```diff
@@ -0,0 +1,566 @@
...
+    
+    static { JenaSystem.init(); }
+
+    /** Get the global file manager.
+     * @return the global file manager
+     */
+    public static FileManager get()
+    {
+        // Singleton pattern adopted in case we later have several file managers.
+        if ( fmInstance == null )
```
kinow, line 92: Should this method be made `synchronized`, or should we tell users of this code that it is not thread-safe? (I think it's not, but my Java-fu is getting utterly confused due to writing more JS & Python at $work)

jena-core/src/main/java/org/apache/jena/util/FileManagerImpl.java
```diff
@@ -0,0 +1,566 @@
...
+        String scheme = FileUtils.getScheme(baseURI) ;
+        
+        if ( scheme != null )
+        {
+            if ( scheme.equals("file") )
+            {
+                if ( ! baseURI.startsWith("file:///") )
+                {
+                    try {
+                        // Fix up file URIs.  Yuk.
```
kinow, line 444: :smile:

jena-arq/src/main/java/org/apache/jena/riot/adapters/AdapterFileManager.java
```diff
@@ -331,4 +339,189 @@ public InputStream openNoMap(String filenameOrURI) {
...
+        String scheme = FileUtils.getScheme(baseURI) ;
+        
+        if ( scheme != null )
+        {
+            if ( scheme.equals("file") )
+            {
+                if ( ! baseURI.startsWith("file:///") )
+                {
+                    try {
+                        // Fix up file URIs.  Yuk.
```
kinow, line 480: This method looks a lot like another method I'd read on this PR. Could they be merged into one and re-used? Or are they doing something different? (Will have another look after submitting the review)

kinow: It's `FileManager#chooseBaseURI` and `AdapterFileManager#chooseBaseURI`. They look identical. Should they be moved somewhere else so that both classes can re-use it?

afs:
That's right.

The FileManager functionality of finding resources and remapping names is in StreamManager (in RIOT), and there is AdapterFileManager to turn a StreamManager into a FileManager. 

That makes the ontology document manager continue to work in jena-core along (tests!) and in normal use with all the Jena jars files on the classpath.

The deprecations on loadModel and readModel operations - I've just seen I can do the rename as "internal", then put back the old operation deprecated, to rename all Jena's internal uses of loadModel and readModel that don't have access to RDFDataMgr (mostly in jena-core tests).

I'm also checking the documentation and will put in a PR for that.

namedgraph:
Why does `FileManagerImpl()` have no public constructor anymore?

namedgraph:
So `RDFDataMgr` is now supposed to be used instead of `FileManager`, right? But `RDFDataMgr` does not support location mapping, correct? So how is that an alternative?

Moreover all the `RDFDataMgr` methods are static, which makes it impossible to override them.

rvesse:
@namedgraph [`StreamManager`](https://jena.apache.org/documentation/javadoc/arq/org/apache/jena/riot/system/stream/StreamManager.html) now provides the mapping pieces, so you'd do something like:

```
try (TypedInputStream input = RDFDataMgr.open(filenameOrUri, yourStreamManager)) {
  Model model = new YourDesiredModelImpl();
  Lang lang = RDFDataMgr.determineLang(filenameOrUri, input.getContentType(), fallbackLang);
  return RDFDataMgr.read(model, input, input.getBaseURI(), lang);
}
```

namedgraph:
@rvesse OK, good to have an example. I think this is missing in the documentation.

So if I want to plug in our own HTTP access (using Jersey Client), should this be done by extending  `LocatorHTTP` and replacing the default `LocatorHTTP` instance in `StreamManager`?

afs:
`RDFDataMgr` uses the global `StreamManager` if no other is specified. There is always a `StreamManger`

The most detailed configuration is to use `RDFParser` builder:

```
RDFParser.create()
    .streamManager
    .httpAccept
    .lang
    .source
    ...
    .parse(...)
```

namedgraph:
I still think that RDF parsing and HTTP access are orthogonal matters. And that there are higher-level HTTP client/server APIs that could be used. But I guess that ship has sailed.

But considering that:
* we want to use our own HTTP API across the board, including `OntDocumentManager`
* `OntDocumentManager` uses `FileManager` internally

What other options are there rather than extending `AdapterFileManager` or `FileManagerImpl`? Why should `FileManager` become deprecated if it's still in use, and internal if it's an extensibility point?

namedgraph:
@afs am I missing something?

afs:
> am I missing something?

`RDFParser` is not limited to HTTP.

RDF parsing is a process of fetching syntax bytes and converting to RDF. There are several steps in doing that.

The user/application task should be simple and natural. `RDFDataMgr` provides single function calls for the majority of use cases and provides convenience. The main set of functions are around "Get RDF from URI" - for files and HTTP.

It calls `RDFParser`, hiding the details.

If the application wants control it can:

* Use `RDFParser` directly - the builder pattern makes setting the many aspects of a request much easier that APIs for every variation.
* Provide some steps itself, for example, get the InputStream and determine the syntax then pass to a parser (bytes to triples)
* Do something else.

> we want to use our own HTTP API across the board, including `OntDocumentManager`

If you mean JAX-RS, it is a whole application approach and runtime. It is remote only. Not everyone will want to their applications done that way.

namedgraph:
I understand the utility of `RDFParser` and `RDFDataMgr`.

We don't want to force JAX-RS on Jena core, but we want Jena to be flexible enough to override all HTTP access.
BTW JAX-RS is not remote only (if you mean client API) - it is a server/client framework and supports a uniform HTTP request/response API on both the server _and_ the client. Jena could benefit from that as its HTTP APIs are very "ad hoc".

But more on the `FileManager` point: as long as it is used in `OntDocumentManager`, it should be a valid extension point. And as such it should not be deprecated nor can it really be internal. 3rd parties might need to supply their own implementations - as we do in order to override the HTTP access. Does it make sense?

afs:
Please enumerate the requirements for such flexibility. Pointing to a large amount of other code is opaque.

Jena's APIs are not HTTP specific - they include files.

namedgraph:
### Requirements
Provide (retain) Java interfaces for HTTP-based RDF specifications (Linked Data, SPARQL Protocol, Graph Store Protocol) that satisfy the REST [Uniform Interface constraint](https://www.ics.uci.edu/~fielding/pubs/dissertation/rest_arch_style.htm#sec_5_1_5), i.e. that the interface on both the client and server side should be the same.

That makes implementation of components such as proxies trivial:

    class ProxyServer
    {

        public Response get()
        {
            return getBackend().getClient().get();
        }

        public Response post(Model model)
        {
            return getBackend().getClient().post(model);
        }

        ...

    }

Server- or client-specific interfaces can naturally extend the uniform interfaces.
It should work with `StreamRDF` as well.

### Compatibility
IMO a great strategic advantage would be making these interfaces JAX-RS compatible. `javax.ws.rs-api` is used by almost 2500 artifacts according to Maven central.
Jena does not need to depend on Jersey or another JAX-RS implementation, it should be possible to depend on the [javax.ws.rs-api](https://mvnrepository.com/artifact/javax.ws.rs/javax.ws.rs-api) only.
A weaker compatibility would be reusing the JAX-RS API under Jena's namespace, making it a drop-in replacement.

### Prior work
This approach is implemented in [AtomGraph Core](https://github.com/AtomGraph/Core). We are willing to contribute its code - it is Apache-licensed already.

namedgraph:
Re. my other point - I can't seem to get an answer on why `FileManager` is being deprecated while it's still used in `OntDocumentManager`.

afs:
I asked what requirements it places on Jena for the flexibility for there to be a JAX RS API module.

As that exists in AtomGraph, it is clearly possible.

The rest seems to be able stability of specific interfaces, accentuated by jumping from Jena 3.0.1 to Jena current.

> Provide (retain) Java interfaces 

Nothing is being removed.

`DatasetAccessor` is still there.

There is a better IMO interface than `DatasetAccessor` in `RDFDatasetConnection extends RDFDatasetAccessConnection`. It reflects experience with `DatasetAccessor`. 

There is a split into "read" and "read-write" so that "read" can be provided on its own. Compile time robustness not "UnsupportedOperationException".

Better names to be task focused. "fetch" because "get" is more common as getter in Java and some IDE tools do rely on that.
Remove the unnecessary "Model".
`add` -> `load` -- hopefully clearer because it has been a source of questions. YMMV.

The conversion of `DatasetAccessor` to `RDFDatasetConnection`::
```
    @Override public default Model getModel() { return fetch();}
    @Override public default Model getModel(String graphName) { return fetch(graphName);}
    @Override public default boolean containsModel(String graphName) { return true; }
    @Override public default void putModel(Model model) { put(model); }
    @Override public default void putModel(String name, Model model) { put(name, model); }
    @Override public default void deleteDefault() { delete(); }
    @Override public default void deleteModel(String name) { delete(name); }
    @Override public default void add(Model model) { load(model); }
    @Override public default void add(String name, Model model) { load(name, model); }
```

> why FileManager is being deprecated

It's not. Some operations are deprecated.

It's still in `OntDocumentManager` because of the Jena module structure (i.e. legacy).

At runtime, the global `FileManager` is actually an `AdapterFileManager` and the machinery is a `StreamManager`.

> We are willing to contribute its code

Great but as I asked last time, what advantage is there in making it part of Jena? It has a cost - it has to be maintained and it slows Jena evolution down. Does it have it's own community around the code to support it? Is there a community of users to become contributors?

namedgraph:
Sorry I misread about the requirements...

I can't think of any requirements for Jena's flexibility except having modular interfaces that conceptually represent RDF specs and satisfy the UIC (can be shared by the server and the client).

What I don't like about `RDFDatasetConnection` is the "connection"-focus. I think "accessor" made more sense. And then `close()` and `isClosed()` got added which are implementation-specific and not conceptual, as [I was trying to explain to Rob](https://github.com/apache/jena/pull/666#issuecomment-651104758).

Core does not have a community around it. But there are other Linked Data frameworks that cross Jena with JAX-RS - e.g. [Apache Clerezza](https://clerezza.apache.org/) and I think there were more. Just like frameworks rely on Jena for well-established RDF API, they do so on JAX-RS for an HTTP API.

The motivation behind Core seems to be similar to https://github.com/afs/jena-http. But IMO depending on JAX-RS should not slow the development. Quite the opposite - having constraints such as an existing API lets us focus on features rather than API design.

If you don't agree, it's OK. We will continue as we have done. I don't want to use more of our time on this topic.

PR #760, opened 2020-06-18
JENA-1921: SHACL Compact Syntax parser

afs:


afs:
(rebased against latest master)

PR #763, opened 2020-06-20
Clean up use of "assert"

afs:
This PR is a precursor for removing the use of JUnit3 TestSuite feature for tests from manifest files and using Junit4 runners.

This PR tidies up the use of Assert.assert, to be static imports, not via inheriting statics. It is so large it is worth pulling up separately so the real changes of conversion are not swamped.

There are some other small clean up changes that got swept up.


afs:
This is a precursor to [JENA-1992](https://issues.apache.org/jira/browse/JENA-1922).
A separate PR will include the changes that are significant.

PR #765, opened 2020-06-25
Make FileManagerImpl constructors public

namedgraph:
We have an implementation of `LocationMapper` which reads its configuration from file using the `FileManager`. #759 for some reason changed the `FileManagerImpl` constructor visibility to `protected`, where it was `public` before. Therefore our code no longer compiles:
https://github.com/AtomGraph/Web-Client/blob/master/src/main/java/com/atomgraph/client/locator/PrefixMapper.java#L228

This PR restores the `public` visibility of the `FileManagerImpl` constructors.

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/util/FileManagerImpl.java
+++ b/jena-core/src/main/java/org/apache/jena/util/FileManagerImpl.java
@@ -105,7 +105,7 @@ public class FileManagerImpl implements FileManager
     
     /** Create an uninitialized FileManager */
 
-    protected FileManagerImpl() {}
+    public FileManagerImpl() {}
     
     protected static FileManager makeStd() {
         FileManagerImpl fm = new FileManagerImpl();
--- a/jena-core/src/main/java/org/apache/jena/util/FileManagerImpl.java
+++ b/jena-core/src/main/java/org/apache/jena/util/FileManagerImpl.java
@@ -114,7 +114,7 @@ public class FileManagerImpl implements FileManager
     }
     
     /** Create with the given location mapper */
-    protected FileManagerImpl(LocationMapper _mapper)    { setLocationMapper(_mapper) ; }
+    public FileManagerImpl(LocationMapper _mapper)    { setLocationMapper(_mapper) ; }
 
     @Override
     public FileManager clone() { return clone(this) ; } 


kinow:
I think that was intentional @namedgraph, to make that class more for internal use. Wouldn't this work for you?

```java
FileManager fm = FileManager.get();
```

And then later fix the deprecated warning and maybe use `RDFDataMgr` or `StreamManager` depending on what you used FileManager for I think?

namedgraph:
@kinow the `PrefixMapper.initFromPath()` code was copied (long time ago) verbatim from `org.apache.jena.util.LocationMapper`:
https://github.com/afs/jena/blob/master/jena-core/src/main/java/org/apache/jena/util/LocationMapper.java#L148

I can try `FileManager.get()`, and it probably will work. But in general I think the recent Jena's "internalization" has become too extreme.
E.g. `LocationMapper` is meant to be a pluggable class, with 3rd party implementations. So if Jena's `LocationMapper` implementation uses `FileManagerImpl`, why shouldn't the 3rd party implementations have access to it? The `LocationMapper` is only "lucky" to have `protected` visibility because it's in the same package as the `FileManagerImpl`.

afs:
There is `FileManager.create()` that provides a new file manager. It is deprecated and should not be considered long-term stable, but making `FileManagerImpl` public would not make it any more stable.

Does `FileManager.create()` work? Looking at the code: it does, but so also would switching to a `StreamManager`. 

There are `LocationMapper` for `StreamManager` in `org.apache.jena.riot.system.stream` including one for classpath.

afs:
Ping.

namedgraph:
`FileManager.create()` compiles after I've changed Java source version to 1.8.
Can't tell right now if it works because of the `ConcurrentModificationException` thing, but will check later.

afs:
Compiles with which version? We compile with 8,11,12,13,14,15 targetting java8 as set in the POM.

namedgraph:
Sorry, with 3.16.0-SNAPSHOT.

namedgraph:
`PrefixMapper` seems to work with `FileManager.create()`.

afs:
As I under is, `FileManager.create()` meets the requirement to make new `FileManger` using the legacy. If so,is this PR relevant? If it is then it needs updating because GH says there is a branch conflict.

(Note: `AdapterFileManager` for a `StreamManager`  gets injected so in normal operation the global FileManager is the global `StreamManager`).

namedgraph:
This is rather minor, so we can close this. The more major argument I have is in #759.

PR #766, opened 2020-06-26
[JENA-1927] Link to TDB FAQs in error message

rvesse:
Where relevant include a link to the TDB FAQs in error messages


rvesse:
There's probably more places this could be added but this covers the key errors currently called out on the TDB FAQs page

See also related PR apache/jena-site#13

PR #772, opened 2020-07-02
JENA-1934: Better error handling for bad HTML forms

afs:


jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/server/Dispatcher.java
```diff
@@ -373,11 +382,23 @@ public static Operation chooseOperation(HttpAction action) {
...
         // Place for an extension point.
         boolean hasParams = request.getParameterMap().size() > 0;
         if ( hasParams ) {
+            // One nasty case:
+            // Bad HTML form (content-type  application/x-www-form-urlencoded), but body is not an HTML form.
+            //  map is one entry, and the key is all of the body,
+            if ( WebContent.contentTypeHTMLForm.equals(request.getContentType()) ) {
+                ServletOps.errorBadRequest("Malformed request: unrecognized HTML form request");
+                return null;
+            }
```
kinow, line 395: I think I got hit by this when I was writing the Vue.js client. The default value was sending JSON, or some other value, not HTML form. I couldn't figure out what was wrong, so had to look at the browser requests in the old UI to understand what the server was expecting.

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/server/EndpointSet.java
```diff
@@ -104,7 +104,7 @@ private void resetSingle() {
     public Collection<Operation> operations() { return endpoints.keySet(); }
     
     /** Get the Endpoint for a singleton EndpointSet */
-    public Endpoint getOnly() {
+    public Endpoint getExactlyOne() {
```
kinow, line 107: Much more descriptive! Thanks!

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/server/Dispatcher.java
afs: Ah. I came across it with `curl` but from what you say, code defaults POST to a HTML form if there is no content-type given. I guess the common case other that forms is "just send this file, the other end knows what to do" (i.e. ignore the content-type).

Fuseki dispatch is quite complicated and HTML forms are a possibility for SPARQL Update and SPARQL Query so it's not just simply ignoring `application/x-www-form-urlencoded`.

PR #776, opened 2020-07-22
JENA-1937: SHACL Compact Syntax writer and jena-shacl clean-up.

afs:
More (and hopefully final) work on the SHALC Compact Syntax writer together with cleanup of jena-shacls, including "shacl" command (in jena-cmds).

This PR adds a a RIOT writer for `Lang.SHACLC` -- this can only write what is expressible in SHACLC and does not write general RDF that is unconnected so as a writer it writes only part of a graph. It is therefore debatable general RDF APIS should work for SHACLC output. However, it does make working with SHACLC convenient.

If an attempt is made to write SHCAL that can't be expressed on compact syntax, the code throws an exception but, because it is a streaming writer, partial output occurs.

-  Remove "verbose" validation flag. Whatever is the right way to do this, a specific API call for a specific validator implementation is not the way to do it.
- Reorganise the SHACLC tests.
- Rename "simple" validation as "plain". This is the implementation of the algorithm as given in the SHACL spec.
- Various small fixes to code after PR #775 based on using it.
- Refactoring and code tidying.

Also included:
- Changed checked exception `IOException` and make a runtime exception in various `IO.readWholeFile*` operations. This makes it more consistent and work with lambda's better.
- Whitespace fixing.


jena-shacl/src/main/java/org/apache/jena/shacl/compact/writer/WriterRIOTShaclc.java
```diff
@@ -0,0 +1,63 @@
...
+    public void write(OutputStream out, Graph graph, PrefixMap prefixMap, String baseURI, Context context) {
+        IndentedWriter iOut = new IndentedWriter(out);
+        iOut.setUnitIndent(4);
+        write(iOut, graph, prefixMap, baseURI, context);
+    }
+
+    private void write(IndentedWriter out, Graph graph, PrefixMap prefixMap, String baseURI, Context context) {
+        Shapes shapes = Shapes.parse(graph);
+        ShaclcWriter.print(out, shapes);
+        //out.flush();
```
kinow, line 60: Commented test/debug statement?

jena-shacl/src/main/java/org/apache/jena/shacl/Shapes.java
```diff
@@ -189,7 +196,7 @@ public int numRootShapes() {
 
     /** Iterator over the shapes with targets and with explicit type NodeShape or PropertyShape. */
     public Iterator<Shape> iteratorAll() {
+        // rootsShapes and declShaes are disjoint so no dupicates in the iterator.
```
kinow, line 199: s/dupicates/duplicates

afs: Done

jena-shacl/src/main/java/org/apache/jena/shacl/compact/writer/WriterRIOTShaclc.java
afs: Good catch. It should not be there but the routines that create an `IndentedWriter` should have a "flush". `IndentedWriter` has it's own buffering for chars to bytes. The builtin `IndentedWriter.stdout` or `IndentedWriter.stderr` have `setFlushOnNewline` (c.f. PrintStream autoFlush).

PR #779, opened 2020-08-05
JENA-1939: fix getUpdateCount() in the JDBC driver

rvesse:
The method JenaStatement.getUpdateCount() returns 0. But java.sql.Statement documentation says:

    Returns: the current result as an update count; -1 if the current result is a ResultSet object or there are no more results

Returning correct value is important because a single statement may have multiple result sets and cause mutiple updates (not only multiple updated records, but multiple sets of updated records).

Applications (e.g. SQL-DK) iterate over these multiple results and need to know when finish. The getMoreResults() documentation says:

    There are no more results when the following is true: ((stmt.getMoreResults() == false) && (stmt.getUpdateCount() == -1))

But if 0 is returned, the application enters infinite loop.


rvesse:
This LGTM as far as I can tell, been a long time since I looked at the JDBC stuff

PR #781, opened 2020-08-14
[JENA-1911] Migrate Backbone.js app to Vue.js

kinow:
For JENA-1911. Replacing the old Backbone.js app by a Vue.js app. The new app uses webpack/babel, and NPM/Yarn to maintain the packages.

After installing with `yarn install`, it should be possible to start the new app with `yarn run serve`. It runs a development server, redirecting queries to `localhost:3030`. Only missing features now are:

- [x] Query editor
  - NOTE: using YASQE and YASR directly, couldn't find qonsole or an easy way to use YASGUI
  - The full screen mode was removed in recent releases of YASQE (we can add it back later too, though it requires some CSS/HTML and then use the fullscreen addon of CodeMirror)
  - 99% done with prefixes, just need to allow users to add new prefixes (although, it looks like they are not persisted at the moment)
- [x] Dataset edit
  - [x] Add routers, update menu, display links
  - [x] Add UI elements
  - [x] Call backend (implementing same behaviour as before, limiting the execution on large graphs)
- [ ] Sort out Maven build
- [ ] Tests
- [ ] Documentation


afs:
Got it running locally! :+1:

kinow:
@afs whew! Finally got all the features "working" ™. It needs some further testing to see what is working as before, what needs changing, etc.

Could you take a look later, whenever you have time (no hurry :+1:).

Once we are OK with the new app, I'll start writing tests, as normally a few bugs pop when you shake the JavaScript code too hard :grin: 

And after that the final step would then be to look at the Maven build (where I would need more help).

Cheers
Bruno

afs:
I'll be delighted to try it out.  How do I build it?

kinow:
When you check out this branch, start FusekiCmd (fuseki2). I start it in Eclipse with no extra params.

You should have Fuseki Backbone.js UI available on localhost:3030. I used this to compare with new UI.

Now in the new module folder (jena-fuseki-ui) run `yarn install` which will install the dependencies from npmjs.com (js equivalent to maven central).

Now `yarn run serve` should start a dev server on localhost:8080. The dev server has also a proxy to send requests to localhost:3030 (that's why you need to initialize that one too).

And now just compare the two versions side by side.

https://yarnpkg.com/
https://docs.npmjs.com

Thanks!
Bruno

kinow:
Rebased. Reading e-mail threads about this issue, and also about "Future Fuseki Admin API Thought", to see if I can come up with a solution for the Maven build here.

kinow:
I think this may work... the current `jena-fuseki-webapp` stays as-is for now. Later we may split it into "view" and "admin" modules.

The `jena-fuseki-ui` will produce HTML, JS, CSS, image, etc., assets into `jena-fuseki-ui/target/dist/`.

And—somehow, still thinking how—`jena-fuseki-war` will include `jena-fuseki-webapp` as dependency, bringing the servlets, and using its `web.xml`; while also copying the assets of `jena-fuseki-ui` into its WAR webapp folder (currently it's using `jena-fuseki-webapp`'s `webapp` contents, including the old BackboneJS application).

jena-drawing

kinow:
Last commit has a WIP of the suggested design above. But he log4j and other configuration files are missing from the generated war, and not sure if the parent module that uses the assembly plugin to produce the ZIP and tar.gz files is working correctly.

kinow:
Alright, last commit of night. Tried to update the assembly file to use the files from `jena-fuseki-ui`. It generated the correct webapp files, and includes the log4j configuration file. Just need to find time to test it, and see if it's deployed correctly to Tomcat. :sleeping_bed:

kinow:
View on a mobile/smaller viewport.

GIFrecord_2021-03-23_144052

kinow:
Rebased, and updated JS dependencies (found no security issues). Had to pin sass-loader due to https://stackoverflow.com/questions/66082397/typeerror-this-getoptions-is-not-a-function (I have the same in a project at $work). Vue 3 is almost fully production ready, with a few dependencies needing to add support to this new version. That should improve performance (which doesn't seem to be a problem honestly, it's quite fast already), fixes other bugs, but may require some changes in the code (which I'll have to do at $work anyway). But dependency issues like this might disappear when we update :+1:

jena-fuseki2/apache-jena-fuseki/assembly-dist.xml
```diff
@@ -75,7 +75,7 @@
     <fileSet>
       <outputDirectory></outputDirectory>
       <includes>
-        <include>templates/*</include>
+        <include>static/*</include>
```
rvesse, line 78:
```suggestion
        <include>static/*</include>
```

kinow:
Rebased (accidentally rebased against `master` :rofl: saved by `git reflog` :pray: )

kinow:
@afs rebased, and squashed the commits (so instead of 96 commits, we have a single one now :sweat_smile: ).

I'll look into the Maven build next. I guess I can just make `jena-fuseki2` webapp to build the `war` and copy all the files produced by `jena-fuseki-ui` inside the generated `war`. That should — in theory — replace the existing Backbone.js app, and re-build it whenever `jena-fuseki-ui` Vue.js app is modified.

kinow:
@afs in my last commit I tried to modify the Jetty application start up in `FusekiCmd` to load the resources from a `target/webapp` folder that contains the original `jena-fuseki-webapp` files, and also the `jena-fuseki-ui` Vue.js app files copied. See the pom.xml change for more.

I don't understand completely how Jena & Fuseki & Jetty work, so there might be better ways to structure how we get the files from `jena-fuseki-ui` into `jena-fuseki-webapp` (which from what I understood, goes into `jena-fuseki-war` later on)

kinow:
Argh, was going to rebase to quickly update deps, but for the third time in this PR I've rebased onto `master` instead of `main` :cry:

kinow:
Was going to use `reflog` to undo the mess I did here, but looking at the list of things I still need to implement

- [ ] figure out how to get Fuseki and the Maven build with the new UI
- [ ] write unit / e2e tests
- [ ] update dependencies (JS libs are released quite rapidly, since the PR was created I updated it a few times, but today I was going to try some major updates)

Soon would also have to

- [ ] update to Vue3 and TS
- [ ] keep updating our JS libs

I think I'd be the only dev maintaining it, and moving it forward would need a few hours month, which I have for the next 1-2 years. But would still be better to have `busFactor > 1`. I think for now we can keep using the old Backbone.js app. Fix any JS bugs if necessary, and update the JS libs manually when/if necessary.

PR #783, opened 2020-08-17
JENA-1944: RDF Thrift tidy-up

afs:
and update notes about a test that fails only on certain ASF Jenkins nodes.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/RDFParserRegistry.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/RDFParserRegistry.java
@@ -31,9 +31,11 @@ import java.util.Set ;
 import org.apache.jena.atlas.lib.InternalErrorException ;
 import org.apache.jena.atlas.web.ContentType ;
 import org.apache.jena.riot.lang.* ;
+import org.apache.jena.riot.system.ErrorHandlerFactory;
 import org.apache.jena.riot.system.ParserProfile;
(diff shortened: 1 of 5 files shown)


jena-arq/src/main/java/org/apache/jena/riot/RDFParserRegistry.java
```diff
@@ -206,14 +206,24 @@ public ReaderRIOT create(Lang language, ParserProfile profile) {
...
         @Override
         public void read(InputStream in, String baseURI, ContentType ct, StreamRDF output, Context context) {
-            BinRDF.inputStreamToStream(in, output) ;
+            try {
+                BinRDF.inputStreamToStream(in, output);
+            } catch (RiotException ex) {
+                if ( profile != null && profile.getErrorHandler() != null )
+                    profile.getErrorHandler().error(ex.getMessage(), -1, -1);
+                else
+                    System.err.println(ex.getMessage());
```
kinow, line 225: Should it go to a `Logger` instead? Or is it expected to be written to stderr?

afs: Thanks for seeing that. `Logger` in some way would be better

RDFThrift is a bit different because the parser does not signal errors or warnings like the other text-syntax parsers. Such raw binary is supposed to be "right" and errors are actually low level encoding problems that Thrift will catch and complain about. But each case must throw an exception regardless of the errorhandler.

PR #785, opened 2020-08-23
JENA-1936: Value-equality indexing for binary datatypes

afs:


jena-core/src/main/java/org/apache/jena/graph/impl/LiteralLabelImpl.java
```diff
@@ -303,20 +304,59 @@ public String getLexicalForm() {
...
+        return getLexicalForm();
     }
 
+    /**
+     * {@code byte[]} wrapper that provides {@code hashCode} and {@code equals} based
+     * on the value of the array.
+     */
+    static class ByteArray {
+        private final byte[] bytes;
+        /*package*/ ByteArray(byte[] bytes) {
```
strangepleasures, line 335: Consider caching the hash

```
private int hash;

/*package*/ ByteArray(byte[] bytes) {
  this.bytes = bytes;
}

@Override
        public int hashCode() {
             if (hash == 0) {
               hash = Arrays..hashCode(bytes);
             }
             return hash;
        }
```

afs: OK - done. We are assuming the `byte[]` isn't changing.

kinow (approved):

PR #789, opened 2020-08-29
JENA-1957: targetClass=, missing propertyParams, NodeShape sh:node.

afs:
Adds some things missing from SHACL Compact Syntax spec.


jena-base/src/main/java/org/apache/jena/atlas/lib/CollectionUtils.java
```diff
@@ -31,6 +32,10 @@
 
     /** Return an element from a collection. */ 
     public static <T> T oneElt(Collection<T> collection) {
+        if ( collection.isEmpty() )
```
rvesse, line 35: Should this method have a `null` check for `collection`?

rvesse (approved):
LGTM, one minor comment

jena-base/src/main/java/org/apache/jena/atlas/lib/CollectionUtils.java
afs: Done (the calls were safe anyway).

PR #790, opened 2020-08-29
JENA-1958: Support SPARQL* result format variants

afs:
Also contains some other clean up code.


kinow (approved):

afs:
Added a variant of a variant.

PR #792, opened 2020-09-03
JENA-1959 Added JsonLdReadContext

umbreak:
Fixes https://issues.apache.org/jira/browse/JENA-1959

- Adds the analogous to `JsonLdWriteContext` when converting JSON to RDF (toRdf on the JSON-LD high level API)

- Updates Json-LD Java from 0.12.5 to 0.13.0

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/JsonLDReadContext.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/JsonLDReadContext.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
--- a/jena-arq/src/main/java/org/apache/jena/riot/JsonLDReadContext.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/JsonLDReadContext.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * <p>
- * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/jena-arq/src/main/java/org/apache/jena/riot/JsonLDReadContext.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/JsonLDReadContext.java
@@ -1,0 +1,74 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.jena.riot;
+
+import com.github.jsonldjava.core.JsonLdOptions;
+import org.apache.jena.atlas.web.ContentType;
+import org.apache.jena.riot.lang.JsonLDReader;
+import org.apache.jena.riot.system.StreamRDF;
+import org.apache.jena.sparql.util.Context;
+
+import java.io.InputStream;
+
+/**
+ * Set of parameters that can be used to control the reading of JSON-LD.
+ *
+ * This class provides setters to define a "Context" suitable to be passed as 
+ * last argument to  {@link ReaderRIOT#read(InputStream, String, ContentType, StreamRDF, Context)}
+ * when the ReaderRIOT has been created with one of the JSON-LD RDFFormat variants (that is, when it is
+ * an instance of {@link JsonLDReader})
+ *
+ * Parameters that are actually useful are ''documentLoader'' and ''produceGeneralizedRdf''.
+ *
+ */
+public class JsonLDReadContext extends Context {
+    /**
+     * Set the value of the JSON-LD "@context" node, used when reading a jsonld (overriding the actual @context in the jsonld). "Compact" and "Flattened" JSON-LD outputs.
+     *
+     * @param jsonLdContext the value of the "@context" node (a JSON value). Note that the use of an URI to pass an external context is not supported (as of JSONLD-java API 0.8.3)
+     *
+     * @see #setJsonLDContext(Object)
+     */
+    public void setJsonLDContext(String jsonLdContext) {
+        set(JsonLDReader.JSONLD_CONTEXT, jsonLdContext);
+    }
+
+    /**
+     * Set the value of the JSON-LD "@context" node, used when reading a jsonld (overriding the actual @context in the jsonld). "Compact" and "Flattened" JSON-LD outputs.
+     *
+     *
+     * @param jsonLdContext the context as expected by JSON-LD java API. As of JSON-LD java 0.8.3, a Map
+     * defining the properties and the prefixes is OK. Note that the use an URI to pass an external context is not supported (JSONLD-java RDF 0.8.3)
+     *
+     * @see #setJsonLDContext(String)
+     */
+    public void setJsonLDContext(Object jsonLdContext) {
+        set(JsonLDReader.JSONLD_CONTEXT, jsonLdContext);
+    }
+    /**
+     * Set the JSON-LD java API's options
+     *
+     * If not set, a default value is used.
+     *
+     * @param opts the options as defined by the JSON-LD java API
+     */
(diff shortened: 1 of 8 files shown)


jena-arq/src/main/java/org/apache/jena/riot/JsonLDReadContext.java
```diff
@@ -0,0 +1,74 @@
+/**
```
afs, line 1: Should be `/*`. (Yes - some other files have `/**` and need fixing)

jena-arq/src/main/java/org/apache/jena/riot/JsonLDReadContext.java
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * This class provides setters to define a "Context" suitable to be passed as 
+ * last argument to  {@link ReaderRIOT#read(InputStream, String, ContentType, StreamRDF, Context)}
+ * when the ReaderRIOT has been created with one of the JSON-LD RDFFormat variants (that is, when it is
+ * an instance of {@link JsonLDReader})
+ *
+ * Parameters that are actually useful are ''documentLoader'' and ''produceGeneralizedRdf''.
+ *
+ */
+public class JsonLDReadContext extends Context {
```
afs, line 40: OK - I see the intention here but subclassing isn't going to work. This is because `Context` is also used in places where it is a container of settings from different subsystems - e.g. `RIOT.getContext()` is a global system place to have context settings. A subclass doesn't because it can't be shared.

Instead, have a library class `JsonLD` with static functions.
```
public static void setOptions(Context cxt, JsonLdOptions opts) {
        cxt.set(JsonLDReader.JSONLD_OPTIONS, opts);
}
```

jena-arq/src/main/java/org/apache/jena/riot/RIOT.java
```diff
@@ -109,13 +109,6 @@ public static String getBuildDate() {
 
     // ---- Symbols
 
-    /**
-     * Symbol to use to pass (in a Context object) the "@context" to be used when reading jsonld
-     * (overriding the actual @context in the jsonld)
-     * Expected value: the value of the "@context",
-     * as expected by the JSONLD-java API (a Map) */
-    public static final Symbol JSONLD_CONTEXT = Symbol.create("http://jena.apache.org/riot/jsonld#JSONLD_CONTEXT");
-
```
afs, line 118: RIOT holds the public constants so an app does not need to import class `JsonLDReader` (which is in a subpackage).

It is possibly used by existing Jena applications so there is a compatibility issue here.

While not a perfect way to achieve a tidy separate of public and implementation I think we ought to keep the current style (and for `JSONLD_OPTONS`) unless there is a specific reason for this change.

jena-arq/src/test/java/org/apache/jena/riot/TestJsonLDReader.java
```diff
@@ -1,4 +1,4 @@
-/*
+/**
```
afs, line 1: `/*`

jena-arq/src/main/java/org/apache/jena/riot/lang/JsonLDReader.java
```diff
@@ -112,8 +130,21 @@ public void read(InputStream in, String baseURI, ContentType ct, StreamRDF outpu
...
     }
-    
-    private void read$(Object jsonObject, String baseURI, ContentType ct, final StreamRDF output, Context context) {
+
+    private void readWithJsonLDCtxOptions(Object jsonObject, String baseURI, final StreamRDF output, Context context)  throws JsonParseException, IOException {
+        JsonLdOptions options = getJsonLdOptions(baseURI, context) ;
+        Object jsonldCtx = getJsonLdContext(context);
+        if (jsonldCtx != null) {
+            if (jsonObject instanceof Map) {
+                ((Map) jsonObject).put("@context", jsonldCtx);
```
afs, line 139: I'm getting warnings on lines 114, 139: these should be suppressed (it makes it easier to see when new warnings appear in the codebase).

```
Map is a raw type. References to generic type Map<K,V> should be parameterized
	JsonLDReader.java	/jena-arq/src/main/java/org/apache/jena/riot/lang	line 139
Type safety: The method put(Object, Object) belongs to the raw type Map. References to generic type Map<K,V> should be parameterized
	JsonLDReader.java	/jena-arq/src/main/java/org/apache/jena/riot/lang	line 139
Unnecessary @SuppressWarnings("rawtypes")
	JsonLDReader.java	/jena-arq/src/main/java/org/apache/jena/riot/lang	line 114
Unnecessary @SuppressWarnings("unchecked")
	JsonLDReader.java	/jena-arq/src/main/java/org/apache/jena/riot/lang	line 114
```

jena-arq/src/test/java/org/apache/jena/riot/TestJsonLDReader.java
```diff
@@ -18,127 +18,123 @@
...
     //
     //
     //
 
-    /**
-     * Reading some jsonld String, using a Context
-     * @return a new Dataset
-     * @throws IOException
-     */
-    private Dataset jsonld2dataset(String jsonld, Context jenaCtx) throws IOException {
```
afs, line 118: JSON-LD is a dataset format and  can encode datasets (quads) as well as a single graph.

afs:
Generally - looks good and I think we do want to continue improving the java-jsonld usage. 

While [JENA-1948](https://issues.apache.org/jira/browse/JENA-1948) proposes using Titanium, I, for one, don't know if Titanium is fast/slow/good/bad. If they have different pros and cons, having both is possible.

afs:
Ping @fpservant

jena-arq/src/main/java/org/apache/jena/riot/JsonLDReadContext.java
umbreak: I was doing the same as it is being done in [JsonLdWriteContext](https://github.com/apache/jena/blob/master/jena-arq/src/main/java/org/apache/jena/riot/JsonLDWriteContext.java#L43). Why would it work for the writer and not for the reader? 

The only thing this class is doing is to forward the method calls to the super.context.

jena-arq/src/main/java/org/apache/jena/riot/RIOT.java
umbreak: Same argument goes here: I was trying to build it consistent to the [JsonLdWriter](https://github.com/apache/jena/blob/master/jena-arq/src/main/java/org/apache/jena/riot/writer/JsonLDWriter.java#L99).

You can keep this one as well in RIOT. It is exactly the same Symbol. We could flag it as `@deprecated` if we want...

jena-arq/src/test/java/org/apache/jena/riot/TestJsonLDReader.java
umbreak: In the tests I don't see what the dataset is doing that the model cannot do. But I can keep it if you find it better.

umbreak:
@afs in relation to Titanium and performance, here you have some current comparisons: https://github.com/umbreak/jsonld-benchmarks

But contributors of Titanium are well aware of it and are going to tackle the issue.

jena-arq/src/main/java/org/apache/jena/riot/lang/JsonLDReader.java
umbreak: I did not change this part of the code. It gives warning [as it gave before](https://github.com/apache/jena/blob/bfce1741cb12f9cf544235d32fba6598bc7341b5/jena-arq/src/main/java/org/apache/jena/riot/lang/JsonLDReader.java#L94).

What do you suggest?

jena-arq/src/main/java/org/apache/jena/riot/JsonLDReadContext.java
afs: OK - I see now. I hadn't remembered `JsonLdWriteContext`; I'd also jumped to the conclusion that `JsonLdReadContext` would be used where general `Context` is.

Same style as existing is fine; if is to be changed, that's a separate work item.

jena-arq/src/main/java/org/apache/jena/riot/RIOT.java
afs: As above, existing style is fine.

Leaving the constant, whether deprecated or not, helps migration.

(and I don't know why it is in RIOT in the first place.)

jena-arq/src/main/java/org/apache/jena/riot/JsonLDReadContext.java
```diff
@@ -0,0 +1,74 @@
...
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
afs, line 11: Please remove the <p> and indent the URL so that the text is exactly the same as other files.

Because this is a new file, it is better that the contributor does this and not the project. License/copyright notices are sacrosanct!

jena-arq/src/main/java/org/apache/jena/riot/JsonLDWriteContext.java
```diff
@@ -6,15 +6,16 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
afs, line 11: Ditto.

afs (changes requested):
Looks good - just the two license header issues. For the new file, it is much better if the contributor makes the changes, not the project.

Other than that, "approve".

afs (approved):

afs:
Thanks for the license header updates.

PR #799, opened 2020-09-17
JENA-1951 Add support for scripting languages other than JavaScript

strangepleasures:
First working version using the new polyglot scripting mechanism.
It should now support URIs of the form http://jena.apache.org/ARQ/pythonFunction#foo and language-specific configuration parameters.
I kept the old `org.apache.jena.sparql.function.js` classes and marked them as deprecated, because they're technically a part of public API, but they're no longer used anywhere.


jena-arq/src/main/java/org/apache/jena/sparql/function/scripting/NV.java
```diff
@@ -65,9 +65,9 @@ public static Object fromNodeValue(NodeValue nv) {
             return nv.getString();
         if ( nv.isNumber() ) {
             if ( nv.isInteger())
-                return nv.getInteger();
+                return nv.getInteger().longValue();
```
strangepleasures, line 68: Most script engines don't support BigInteger and BigDecimal :-(

afs:
I'm sorry this has taken so long. I was "out of the loop" for an extended period.

The conflict seems rather small yet causes some tools to get confused over it. I munged a patch locally and it seems fine.

Given this is Jena 4.0.0, would it be better to drop "sparql.function.js"?

Does any of the documentation change? https://jena.apache.org/documentation/query/javascript-functions.html

afs:
Possibly because of change of branch to 'main', the Github UI won't merge this.

PR #967 is a command line merge and resolution of this PR.

It then removes the "js" packages since for 4.0.0 we have the opportunity to make braking changes - because the common-case translation to native value types and back is unchanged, only the detailed access of `NodeValue`, the impact should be limited (I don't recall any mention of it in email or JITA tickets).

@strangepleasures - thank you.

PR #804, opened 2020-10-01
JENA-1974: G, a library of functions for working with Graph.

afs:
"G" is a set of functions for working with Graph at the detail triple level.

It packages up the checking needed around, for example, "getOne" when the code is expecting exactly one match, not zero, not several. It is taken from original use in jena-shacl but is itself generally useful when working in Java walking data in graphs.

There are some operations that are RDFS aware - these would be applicable for when RDFS reasoning is not enabled (in SHACL, certain parts of RDFS are required like transitive subclass but not full RDFS reasoning).


ajs6f:
This is useful stuff, and I have no objections by any means. Question: is this how you expect ideas for a new API to be presented? Via a "sidecar" library of functions like this? I'm inspired to ask because I see a number of things like "return a param or a default" which would make a lot of sense to impl via returning `Optional` from a hypothetical new graph abstraction.

jena-arq/src/main/java/org/apache/jena/riot/other/G.java
```diff
@@ -0,0 +1,593 @@
...
+    //---- get/list/iter
+
+    /** Does node {@code s} have property {@code p} in graph {@code g}? */
+    public static boolean hasProperty(Graph g, Node s, Node p) {
+        return g.contains(s, p, null);
+    }
+
+    /** Contains exactly one. */
+    public static boolean containsOne(Graph g, Node s, Node p, Node o) {
+        return g.contains(s, p, o) && Iter.count(g.find(s,p,null)) == 1;
```
strangepleasures, line 114: Could be more efficient without counting all triples, e.g. 
```
try (ExtendedIterator<Triple> it = g.find(s, p, null) {
  if (!it.hasNext()) {
     return false;
  }
  it.next();
  return !it.hasNext();
}
```

jena-arq/src/main/java/org/apache/jena/riot/other/G.java
```diff
@@ -0,0 +1,593 @@
...
+    public static Node object(Triple triple) {
+        return triple == null ? null : triple.getObject();
+    }
+
+    // ---- Node filter tests.
+    public static boolean isURI(Node n)         { return n != null && n.isURI(); }
+    public static boolean isBlank(Node n)       { return n != null && n.isBlank(); }
+    public static boolean isLiteral(Node n)     { return n != null && n.isLiteral(); }
+    public static boolean isResource(Node n)    { return n != null && (n.isURI()||n.isBlank()); }
+
```
strangepleasures, line 65: How about isNodeTriple?

afs:
@ajs6f 

> Question: is this how you expect ideas for a new API to be presented?

I don't have preference. This is a bit different as it is, mostly, moved from `jena-shacl`.  This library also "parked" in a not-API place for the moment so it can be morphed.

`Graph` is a bit different because it is the base of all RDF access. I would have expected a new API to build on top of `Graph` in the same way that `Model` does.

Actually changing `Graph` to add a whole bunch of operations and across different style of API does not seem like a good idea.

Specifically for `G`, it really is adding words to more accurately describe what it going on so even where the operations are one liners, I think it clearer to give the pattern a descriptive name. Other operations wrap up a bit of checking so moving visiual clutter out of the caller app. YMMV.

jena-arq/src/main/java/org/apache/jena/riot/other/G.java
afs: Agreed.

strangepleasures: Actually, try-with-resources wouldn't work here as ClosableIterator doesn't implement AutoCloseable. It would be quite convenient and shouldn't break anything. `public interface ClosableIterator<T> extends Iterator<T>,  AutoCloseable`

ajs6f: See https://issues.apache.org/jira/browse/JENA-1956...

jena-arq/src/main/java/org/apache/jena/riot/other/G.java
afs: Good point. You can see that the code is dated!
(and `isNodeGraph` even if it isn't current used)

jena-arq/src/main/java/org/apache/jena/riot/other/G.java
strangepleasures: Ah great!

afs:
Added

- comments for nearly all public functions.
- `Objects.requiresNonNull`
- rename variables e.g. `s` -> `subject`

jena-arq/src/main/java/org/apache/jena/riot/other/G.java
```diff
@@ -0,0 +1,669 @@
...
+        Objects.requireNonNull(object, "object");
+        ExtendedIterator<Triple> iter = find(graph, null, null, object);
+        if ( ! iter.hasNext() )
+            // Zero.
+            return false;
+        iter.next();
+        if ( iter.hasNext() )
+            // more than one
+            return false;
+        return true;
```
kinow, line 478: Do we have to close `iter`?

kinow (approved):
Don't know much about this API in Jena, so only had a brief look at the code :+1: looking good from what I could understand :+1:

jena-shacl/src/main/java/org/apache/jena/shacl/engine/SparqlConstraints.java
```diff
@@ -91,6 +94,27 @@ public static Constraint parseSparqlConstraint(Graph shapesGraph, Node shape, No
...
+    private static boolean absentOrOne(Graph g, Node s, Node p, Node o) {
+        ExtendedIterator<Triple> iter = G.find(g, s, p, null);
+        try {
+            if ( ! iter.hasNext() )
+                return false;
+            iter.next();
+            if ( ! iter.hasNext() )
+                return true;
+            long x = Iter.count(G.find(g, s, p, null));
+            throw new ShaclParseException("More then one (" + x + ") of " + String.format("(%s %s %s)", s, p, o));
```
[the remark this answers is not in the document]
afs: Done

jena-arq/src/main/java/org/apache/jena/riot/other/G.java
afs: Yes. (It is also most the same as findUnique)

Closing unfinished extended iterators is contract even if many (most?) apps don't care or attend to the detail. In fact, plain in-memory, TDB1, TDB2 and TIM don't rely on this. But as a library it ought to get it right!

PR #806, opened 2020-10-01
JENA-1975: Clean up logging

afs:


jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/JettyServer.java
```diff
@@ -352,7 +387,17 @@ protected void addFilter(ServletContextHandler context, String pathspec, Filter
         }
 
         /** Jetty server */
-        private static Server jettyServer(int port, boolean loopback) {
+        private static Server jettyServer(int port, boolean loopback, int minThreads, int maxThreads) {
+            ThreadPool threadPool = null;
+            // Jetty 9.4 : the Jetty default is200,8
```
kinow, line 392: 20? Or 2 perhaps?

kinow (approved):

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/JettyServer.java
afs: It really is 200,8 -- Jetty arguments are (max,min).

[QueuedThreadPool.java in Jetty 9.4](https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java)

One of the reasons for messing with this is that it becomes hard to debug with even the minimum 8 threads when you pause the whole JVM and want to find a thread in a particular state, especially if (for Delta testing) there are multiple Jetty servers in one JVM.

PR #809, opened 2020-10-09
JENA-1977: Parse rules header

afs:


jena-core/src/main/java/org/apache/jena/reasoner/rulesys/Rule.java
```diff
@@ -675,7 +673,9 @@ private static int nextSplit(int start, boolean white, String line) {
      * @throws ParserException if there is a problem
      */
     public static List<Rule> parseRules(String source,BuiltinRegistry registry) throws ParserException {
-        return parseRules(new Parser(source,registry));
+        BufferedReader reader = new BufferedReader(new StringReader(source));
+        Parser parser = rulesParserFromReader(reader, registry);
+        return parseRules(parser);
```
kinow, line 678: Could we close the reader before returning the created object?

afs: Done.

kinow (approved):

PR #815, opened 2020-10-19
JENA-1981 Unicode and SUBSTR

afs:
This fixed a problem identified from the W3C rdf-tests work. (JENA-1981)

It also includes improvements to JavaCC grammar for Turtle and SHACL Compact Syntax.
The Turtle grammar is kept for reference - it is not used to parse .ttl. The idea is to keep this as reusable input for other languages - SHACL Compact Syntax, for example. 

But JENA-1982 is not fixed for SPARQL yet. That needs care to ensure compatibility of corner cases.

As part of general tidying up in the area, rename the SPARQL template javacc file as main.jj.


jena-arq/src/test/java/org/apache/jena/sparql/expr/TestFunctions2.java
```diff
@@ -165,6 +167,23 @@
     public void substr_21()         { test("substr('', 'one', 3)",            "''") ; }
     @Test(expected=ExprEvalException.class)
     public void substr_22()         { test("substr('', 1, 'three')",            "''") ; }
+    
+    // Codepoint outside UTF-16. 
+    // These are  U+0001F46A 👪 - FAMILY
```
kinow, line 172: Emoticons being useful for testing :tada:

kinow (approved):

jena-arq/src/test/java/org/apache/jena/sparql/expr/TestFunctions2.java
afs: :-)

I didn't put them raw in the tests themselves because of portability concerns over binary/text.
I did test with 👪.
(ARQ is not handling \U in SPARQL properly yet.)

PR #817, opened 2020-10-21
[JENA-1985] Adding TriG and N-Quads content-type in Fuseki2.

marcelomachado:
Jena already work with N-Quads and Trig serialization. This is also useful for dataset page on Fuseki Webpage.

### DIFF
--- a/jena-fuseki2/jena-fuseki-webapp/src/main/webapp/dataset.html
+++ b/jena-fuseki2/jena-fuseki-webapp/src/main/webapp/dataset.html
@@ -138,6 +138,8 @@
                                       <option value="text/turtle">Turtle</option>
                                       <option value="application/ld+json">JSON-LD</option>
                                       <option value="application/n-triples">N-Triples</option>
+                                      <option value="application/n-quads">N-Quads</option>
+                                      <option value="application/trig">TriG</option>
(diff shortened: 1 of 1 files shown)


kinow (commented):
Looks good to me. Would need to test it, but change is really simple. @marcelomachado we would need a JIRA too (if there's not one already), in order to include it in the changelog. Could you create one and update the title of the pull request with the JIRA ID? 

https://issues.apache.org/jira/projects/JENA/issues

rvesse (approved):

afs (approved):

afs:
(Does not affect this PR) 
when checking the details, I found some Java code constants for TriG make it look like the correct MIME type is `text/trig`, not application/trig`. Both work through out the system so it's not observable - it's just the constant naming and `Lang` definition. I'll tidy this up.

marcelomachado:
> Looks good to me. Would need to test it, but change is really simple. @marcelomachado we would need a JIRA too (if there's not one already), in order to include it in the changelog. Could you create one and update the title of the pull request with the JIRA ID?
> 
> https://issues.apache.org/jira/projects/JENA/issues

Done!

kinow (approved):

PR #820, opened 2020-10-26
JENA-1982: SPARQL unicode escapes

afs:
Note:

This changes SPARQL - specifically Syntax.ARQ.
Syntax.SPARQL_11 remains untouched.

Syntax.ARQ ("with extensions") is the default in jena-arq and also used in Fuseki.

See discussion:
* https://github.com/w3c/rdf-tests/pull/67
* https://github.com/w3c/sparql-12/issues/77

SPARQL and Turtle differ in the way they handle \u and \U escapes.

In SPARQL, they can appear anywhere (even for special characters). ARQ does not support \U .

In Turtle, \u and \U can appear in strings and URIs, not in prefix names, and nowhere else. This is a better design.


afs:
This is quite easy to reverse if necessary.

PR #821, opened 2020-11-01
JENA-1949: Add https crypto modules from the JDK

afs:
Add modules to the docker build jlink step.


afs:
Merged so the OP for JENA-1949 can test it.

PR #831, opened 2020-11-11
Updates for httpclient/httpcore

afs:


afs:
This is PR #828 applied to both plain and OSGi httpClient.

PR #866, opened 2020-11-15
JENA-1998: Imports closure machinery

afs:


jena-shacl/src/main/java/org/apache/jena/shacl/Imports.java
```diff
@@ -0,0 +1,152 @@
...
+import org.apache.jena.riot.system.IRIResolver;
+import org.apache.jena.sparql.graph.GraphFactory;
+
+/**
+ * Import processing.
+ * <p>
+ * Imports are triggered by a base (a single triple "? rdf:type owl:Ontology")
+ * and imports (triples "base owl:Imports URI").
+ * <p>
+ * If there other "? owl:imports ?" triples, they are ignored.
```
kinow, line 47: If there are other?

jena-shacl/src/main/java/org/apache/jena/shacl/Imports.java
```diff
@@ -0,0 +1,152 @@
...
+ */
+public class Imports {
+    private Imports() {}
+
+    /**
+     * Load a graph and process it's owl:imports to create a single graph.
```
kinow, line 53: s/it's/its

kinow (approved):
Nice! I think I had seen `owl:imports` in some articles about Protege, but can't recall if I ever used it. Looks good to me :+1:

PR #867, opened 2020-11-17
JENA-1997: Remove old Turtle/N3 writer

afs:
Remaining Turtle parser supports jena-core tests.
Rename private TTL parser as org.apache.jena.ttl
Remove deprecated RDFFactoryF methods - includes changes to ModelCom.


jena-core/src/main/java/org/apache/jena/rdf/model/impl/RDFWriterFImpl.java
```diff
@@ -69,63 +72,15 @@ public RDFWriter getWriter(String lang) {
...
+    static { 
         reset();
     }
 
     private static void reset() {
         Class<? extends RDFWriter> rdfxmlWriter = org.apache.jena.rdfxml.xmloutput.impl.Basic.class;
         Class<? extends RDFWriter> rdfxmlAbbrevWriter = org.apache.jena.rdfxml.xmloutput.impl.Abbreviated.class;
         Class<? extends RDFWriter> ntWriter = org.apache.jena.rdf.model.impl.NTripleWriter.class;
-        Class<? extends RDFWriter> ttlWriter = org.apache.jena.n3.N3TurtleJenaWriter.class;
+//        Class<? extends RDFWriter> ttlWriter = org.apache.jena.n3.N3TurtleJenaWriter.class;
```
kinow, line 83: Can be removed too? And commented code below as well?

kinow (approved):

jena-core/src/main/java/org/apache/jena/rdf/model/impl/RDFWriterFImpl.java
afs: OK - my original attempt was merely comment out the wiring in of the writer leaving the code ... but turned into removal (we have git!) when I found all the "deprecated" that didn't actually do anything anymore.

Thanks.

afs:
The build failure is an artifact of Travis.

amatiushkin:
@afs Would you mind to share some context why this have been removed?

kinow:
@amatiushkin there's some explanation in the linked JIRA issue https://issues.apache.org/jira/browse/JENA-1997

afs:
Hi @amatiushkin - I'm curious as to what is behind your question. Does it affect you?

See [JENA-1997|https://issues.apache.org/jira/browse/JENA-1997] the ticket that gives some background.

To be clear: The proper writer is in jena/ARQ/RIOT.

The old (jena-core) is writer isn't compliant with the RDF 1.1 Turtle space. It is not accessible unless:
- the app uses jena-core only
- the app reconfigures the internal RIOT wiring in unsupported ways

And it isn't N3, and never was, despite its name.

If you are interested in developing N3 support, Jena does have th enecessary RDF term extension `Node_Graph` for representing N3 formulae.

amatiushkin:
@afs I was looking for API to parse input on older version 3.15-ish and then discovered this change which got me puzzled. 

Now I have more context, so it is all good.

afs:
`RDFDataMgr` for most tasks for read and write.

Then `RDFparser` and `RDFWriter` and their builders, for detailed control and setup.

PR #868, opened 2020-11-20
Add Dockerfile.snapshot for creating image from snapshot builds

SmallhillCZ:
For building Docker image of a snapshot version from https://repository.apache.org/content/groups/snapshots/org/apache/jena/apache-jena-fuseki/ use `Dockerfile.snapshot`.

Both JENA_VERSION and JENA_SNAPSHOT_VERSION must be provided.

    docker build -f Dockerfile.snapshot -t fuseki-snapshot --build-arg JENA_VERSION=3.17.0-SNAPSHOT --build-arg JENA_SNAPSHOT_VERSION=3.17.0-20201119.075025-46 .

### DIFF
--- a/jena-fuseki2/jena-fuseki-docker/Dockerfile.snapshot
+++ b/jena-fuseki2/jena-fuseki-docker/Dockerfile.snapshot
@@ -1,0 +1,125 @@
+## Licensed to the Apache Software Foundation (ASF) under one or more
+## contributor license agreements.  See the NOTICE file distributed with
+## this work for additional information regarding copyright ownership.
+## The ASF licenses this file to You under the Apache License, Version 2.0
+## (the "License"); you may not use this file except in compliance with
+## the License.  You may obtain a copy of the License at
+##
+##     http://www.apache.org/licenses/LICENSE-2.0
+## 
+## Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


afs:
This facility would be good to have.   Would it be possible to incorporate into the main Dockerfile? That way, as the fine tuning of the docker container operation are in one place and, given this module is new, there may well be suggestions for improvements.

Doe sit need to pull the tarball and not the jar file? The Dockerfile works with Fuseki Main (jena-fuseki-server).

There is combined jar for Fuseki/webapp (with UI) in jena-fuseki-fulljar but I don't think anyone has tested the file system layout with this form.

SmallhillCZ:
It could be implemented into the main Dockerfile, but would have to introduce conditionals for the default REPO arg and tarball extraction, which seemed messy, so I split it into separate one.

Is there a jar file alone generated for the snapshots? I know only of https://repository.apache.org/content/groups/snapshots/org/apache/jena/apache-jena-fuseki/, where only tarballs are present.

afs:
That was an automated close by github because the target branch has gone away.

I hope we can incorporate this - there seem to be two parts : building a docker image with SNAPSHOTS and building for the Fuseki/UI version of Fuseki.

PR #870, opened 2020-11-24
Add public getters to SHACL constraint implementations

fkleedorfer:
As documented in https://issues.apache.org/jira/browse/JENA-2002, this commit adds public getters for accessing internal data in SHACL constraints.

### DIFF
--- a/jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/CardinalityConstraint.java
+++ b/jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/CardinalityConstraint.java
@@ -39,14 +39,6 @@ public abstract class CardinalityConstraint extends ConstraintEntity {
         throw new ShaclParseException("Cardinality constraint on a node shape");
     }
 
-    public int getMinCount(){
-        return minCount;
-    }
-
-    public int getMaxCount(){
-        return maxCount;
-    }
-
     // -1 => no test
     protected CardinalityConstraint(int minCardinality, int maxCardinality) {
         this.minCount = minCardinality;
--- a/jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/CardinalityConstraint.java
+++ b/jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/CardinalityConstraint.java
@@ -38,7 +38,15 @@ public abstract class CardinalityConstraint extends ConstraintEntity {
         // Node shape with cardinality. Can this usefully be checked for in the parser?
         throw new ShaclParseException("Cardinality constraint on a node shape");
     }
-    
+
+    public int getMinCount(){
+        return minCount;
+    }
+
+    public int getMaxCount(){
+        return maxCount;
+    }
+
     // -1 => no test
     protected CardinalityConstraint(int minCardinality, int maxCardinality) {
         this.minCount = minCardinality;
--- a/jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/ConstraintComponentSPARQL.java
+++ b/jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/ConstraintComponentSPARQL.java
@@ -59,6 +59,18 @@ public class ConstraintComponentSPARQL implements Constraint {
         }
     }
 
+    public SparqlComponent getSparqlConstraintComponent() {
+        return sparqlConstraintComponent;
+    }
+
+    public Multimap<Parameter, Node> getParameterMap() {
+        return parameterMap;
+    }
+
+    public Query getQuery() {
+        return query;
+    }
+
     @Override
     public void validateNodeShape(ValidationContext vCxt, Graph data, Shape shape, Node focusNode) {
         SparqlValidation.validate(vCxt, data, shape, focusNode, null, focusNode, query, parameterMap,
--- a/jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/ConstraintComponentSPARQL.java
+++ b/jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/ConstraintComponentSPARQL.java
@@ -59,18 +59,6 @@ public class ConstraintComponentSPARQL implements Constraint {
         }
     }
 
-    public SparqlComponent getSparqlConstraintComponent() {
-        return sparqlConstraintComponent;
-    }
-
-    public Multimap<Parameter, Node> getParameterMap() {
-        return parameterMap;
-    }
-
-    public Query getQuery() {
-        return query;
-    }
-
     @Override
     public void validateNodeShape(ValidationContext vCxt, Graph data, Shape shape, Node focusNode) {
         SparqlValidation.validate(vCxt, data, shape, focusNode, null, focusNode, query, parameterMap,
--- a/jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/JViolationConstraint.java
+++ b/jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/JViolationConstraint.java
@@ -35,6 +35,10 @@ public class JViolationConstraint extends ConstraintTerm {
         this.generateViolation = generateViolation;
     }
 
+    public boolean isGenerateViolation() {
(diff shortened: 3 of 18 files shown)


jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/ConstraintComponentSPARQL.java
```diff
@@ -59,6 +59,18 @@ public ConstraintComponentSPARQL(SparqlComponent sparqlConstraintComponent,
         }
     }
 
+    public SparqlComponent getSparqlConstraintComponent() {
+        return sparqlConstraintComponent;
+    }
+
+    public Multimap<Parameter, Node> getParameterMap() {
+        return parameterMap;
```
kinow, line 67: Not sure if there's any risk here (or in the other getters that return objects and lists) to have concurrency issues. If so, we could copy the underlying object/collection.

kinow (commented):
Found nothing that called my attention in the changes, except the returned mutable objects, but if there's no issues with concurrency, then all good :tada: 

Tests are probably not necessary if these are only getters, I think?

jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/ConstraintComponentSPARQL.java
afs: Yes, it does matter. `.equals` relies on all parts of the shape and shapes go into datastructures so `hashCode` and `equals` must be stable. A `Collections.unmodifiable` (on "get" or in the constructor, not sure which is better) or a structure copy is needed. The current code probably isn't perfect.

Or documentation!

Another to look at is whether internal calculated members should be made public.

fkleedorfer: Sorry, good catch. I think wrapping with unmodifiable collections or cloning in the getter should do. I'll add another commit.

fkleedorfer: For the `Multimap` in `ConstraintComponentSPARQL` I am not sure how to proceed, I think it's best to revert the change on that one.

kinow: >For the Multimap in ConstraintComponentSPARQL I am not sure how to proceed, I think it's best to revert the change on that one.

If that's OK for you for now, then all good :-)

kinow (approved):

rvesse (approved):

jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/CardinalityConstraint.java
```diff
@@ -38,7 +38,15 @@ public void validateNodeShape(ValidationContext vCxt, Graph data, Shape shape, N
         // Node shape with cardinality. Can this usefully be checked for in the parser?
         throw new ShaclParseException("Cardinality constraint on a node shape");
     }
-    
+
+    public int getMinCount(){
```
afs, line 42: Not sure if this the best thing to do. The subclass for max and min already have accessors.

Being an abstract class, CardinalityConstraint can't be created directly.

"MaxCount" have a "getMinCount" function is a bit strange. If there were a general min and max at the same time constraint, then on the superclass would make more sense but there isn't.

Please remove theses and use overrides in the subclasses.

jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/JViolationConstraint.java
```diff
@@ -35,6 +35,10 @@ public JViolationConstraint(boolean generateViolation) {
         this.generateViolation = generateViolation;
     }
 
+    public boolean isGenerateViolation() {
```
afs, line 38: Minor: "generateViolation" read better.

jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/PatternConstraint.java
```diff
@@ -53,6 +53,18 @@ public PatternConstraint(String pattern, String flagsStr) {
         this.pattern = Pattern.compile(pattern, flags);
     }
 
+    public Pattern getPattern() {
```
afs, line 56: Minor: I'd prefer not the Pattern because it is an internal term, and then have the pattern string as `getPattern()`.

It is exposing the information once, not in two different ways.

jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/QualifiedValueShape.java
```diff
@@ -50,6 +50,22 @@ public QualifiedValueShape(Shape sub, int qMin, int qMax, boolean qDisjoint) {
         this.qDisjoint = qDisjoint;
     }
 
+    public Shape getSub() {
+        return sub;
+    }
+
+    public int getqMin() {
```
afs, line 57: AutoCase does not work out! And "Q" does work that well either! How about `public int qMin()`?

afs (changes requested):
Overall looks fine. Some fine tuning of names and exposed information.

afs:
Sorry for the delay - I was seeing the release through.

jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/CardinalityConstraint.java
fkleedorfer: ok

jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/JViolationConstraint.java
fkleedorfer: going with `generatesViolation()`?

jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/PatternConstraint.java
fkleedorfer: ok

jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/QualifiedValueShape.java
fkleedorfer: ok (`getQMin()` has its merits, but your call)

jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/JViolationConstraint.java
afs: Yes

jena-shacl/src/main/java/org/apache/jena/shacl/engine/constraint/QualifiedValueShape.java
afs: Ok

fkleedorfer:
If I am not mistaken, we're done. Please let me know if this is not the case. Cheers!

fkleedorfer:
Additionally, I've had to implement a kind of non-invasive visitor pattern to process constraint structures (such as sh:or/sh:node): 
https://github.com/researchstudio-sat/webofneeds/blob/feat-acl-impl/webofneeds/won-utils/won-utils-shacl2java/src/main/java/won/shacl2java/constraints/ConstraintVisitor.java

Of course, I would have preferred to use visitor methods offered by the Constraint classes. If you are interested, I can modify my implementation and add the visitor methods and visitor interface to the jena classes in a separate PR.

afs:
Seeing a PR for the visitor pattern would be interesting.

Sometimes it take time to get round to PRs when there are several in-progress. This week has been getting in the ones that are best done at the beginning of a release cycle.

afs:
ValueRangeConstraint and ConstraintPairwise - have two getters for the same thing. `Constraint` already has `getComponent`.

I've removed the stray.

PR #871, opened 2020-11-25
JENA-2003: Handle file URIs with URI scheme name

afs:
This fixes the problems with "file:".

Before passing to the OS, this has always been stripped and converted to an OS file name. There are other uses of filename handling code from Apache Commons IO, one of which `RDFLangauages.filenameToLang`, which takes a resource name (URI of filename) and uses the extension to guess the RDF syntax. It is used for file names and also URIs when there isn't content negotiation.

The fix, in IO.java, around line 100  (the rest of the PR is naming and cosmetic changes), is to extract the code from Apache Commons IO , remove the check for ":" as a trailing component of the filename ([NTFS ADS](https://en.wikipedia.org/wiki/NTFS#Alternate_data_streams_(ADS))) - windows uses ":" in two places, drive and ADS. This clashes with the use of URI scheme name, which also uses ":".

The fix is the safest approach - copying and restoring the old behaviour in a low risk fashion. Further factoring can be done after the 3.17.0 release to be tidier.

### DIFF
--- a/jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
+++ b/jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
@@ -30,7 +30,6 @@ import java.util.zip.GZIPOutputStream;
 import org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream;
 import org.apache.commons.compress.compressors.bzip2.BZip2CompressorOutputStream;
 import org.apache.commons.compress.compressors.snappy.SnappyCompressorInputStream;
-import org.apache.commons.io.FilenameUtils;
 import org.apache.jena.atlas.RuntimeIOException;
 import org.apache.jena.atlas.lib.IRILib;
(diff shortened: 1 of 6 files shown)


jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
```diff
@@ -85,24 +88,75 @@ static public InputStream openFileEx(String filename) throws IOException, FileNo
...
+        if (fileName == null) {
+            return NOT_FOUND;
+        }
+//        if (isSystemWindows()) {
+//            // Special handling for NTFS ADS: Don't accept colon in the fileName.
+//            final int offset = fileName.indexOf(':', getAdsCriticalOffset(fileName));
+//            if (offset != -1) {
+//                throw new IllegalArgumentException("NTFS ADS separator (':') in file name is forbidden.");
+//            }
+//        }
```
kinow, line 127: :+1:

kinow (approved):

PR #880, opened 2020-12-01
JENA-2006: DatasetGraph prefixes

afs:
Lots of repercussions and lots of checking and tidying while looking at the code areas affected.


rvesse (approved):
Lots of changes in here but don't see anything of concern

Does this constitute a breaking change (either at the API level or behavioural)?

afs:
> Does this constitute a breaking change (either at the API level or behavioural)?

No.

`Dataset.getPrefixMapping()` has a default method.

`DatasetGraph` does have a new interface method. We could add a default impl but at the moment, it signals "take note" to any downstream implementations I don't know about.

There is no intention to break or change any behaviour where code has not been unwrapping classes and directly getting at the TDB1/TDB2 prefix storage. But that's cheating anyway :-).

It is a big internal change. One reason for putting in right at the start of a version cycle is to give it maximum shakedown and to give downstream systems as much chance as possible to test. I know of very few cases of prefixes not being used as "parse in to set, write out to use".

jena-arq/src/main/java/org/apache/jena/riot/system/PrefixMap.java
```diff
@@ -35,6 +36,24 @@
  * </p>
  */
 public interface PrefixMap {
+    /**
+     * Return the URI for the prefix, or null if there is no entry for this prefix.
+     */
+    public String get(String prefix);
+
+    // Is this a good idea? Leave out until it is justified.
```
kinow, line 44: :+1:

jena-arq/src/main/java/org/apache/jena/riot/system/PrefixMapNull.java
```diff
@@ -18,18 +18,20 @@
...
 import java.util.function.BiConsumer ;
+import java.util.stream.Stream;
 
 import org.apache.jena.atlas.lib.Pair ;
 import org.apache.jena.shared.PrefixMapping ;
 
-/** Always empty prefix map */
-public class PrefixMapNull implements PrefixMap {
-    public static PrefixMap empty = new PrefixMapNull() ;
+/** Baes of always empty prefix maps {@link PrefixMapSink} and {@link PrefixMapZero}. */
```
kinow, line 31: Base?

kinow (approved):
Don't know much about the code changed, so just had a quick look at the docs for the changes, and looked over some new methods to see if anything appeared to be missing or wrong. Looks good to me! Left a few comments, no blockers :+1:

afs:
Thank you for the reviews. I've fixed the javadoc comments and squashed it down locally to a single commit which, if nothing else, aids reversing this should any problems arise.

afs:
The validation build has passed on my branch-build on travis-ci.org.

PR #882, opened 2020-12-07
JENA-2009: Implement Visitor Pattern for SHACL Constraints

fkleedorfer:
Notes:
* added `visit` method to `Constraint` interface. I experimented with providing a default implementation (calling `visitor.visit(Constraint)`) to avoid having to add an implementation to all implementing classes. This approach does not work, which I did not expect. Learned something about Java method invocation target selection there.
* added `visit` method non-abstract subclasses of `Constraint`. The pattern suggests they should be named `accept`, but jena uses `visit` wherever I checked. I guess it's better to stick with the existing flavor.
* Currently no unit tests as there is no logic apart from `visitor.visit(this)`. Please let me know if you feel differently. Using the PR in my client code, so I'm positive it works.

### DIFF
--- a/jena-shacl/src/main/java/org/apache/jena/shacl/parser/ConstraintVisitor.java
+++ b/jena-shacl/src/main/java/org/apache/jena/shacl/parser/ConstraintVisitor.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.jena.shacl.parser;
+
+import org.apache.jena.shacl.engine.constraint.*;
+
+public interface ConstraintVisitor {
+    default void visit(ReportConstraint constraint){}
+    default void visit(SparqlConstraint constraint){}
+    default void visit(ConstraintComponentSPARQL constraint){}
+    default void visit(UniqueLangConstraint constraint){}
+    default void visit(HasValueConstraint constraint){}
+    default void visit(MinCount constraint){}
+    default void visit(MaxCount constraint){}
+    default void visit(ShXone constraint){}
+    default void visit(ShAnd constraint){}
+    default void visit(ShOr constraint){}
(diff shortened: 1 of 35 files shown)


rvesse (approved):

afs:
Please could you put the vitro in a consistent place in the constraints? Just before printCompact would be fine as many of them are there already.

afs:
I am unsure that default methods on the visitor are the way to go. An abstract class "ConstraintBase" has been the style up until Java8 and now can have a mixin/trait to do the same.

Reason: robustness.

When writing complex visitors it is good to have a way for the programmer to know they have covered all cases and be notified when a new constraint is added. Otherwise it is possible to miss a case and compile time and and it becomes a runtime problem.

jena-shacl/src/main/java/org/apache/jena/shacl/parser/ConstraintVisitor.java
```diff
@@ -0,0 +1,56 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.jena.shacl.parser;
+
+import org.apache.jena.shacl.engine.constraint.*;
+
+public interface ConstraintVisitor {
+    default void visit(ReportConstraint constraint){}
```
afs, line 24: This one can be skipped.  It's a dummy that only exists to carry a message. It can't be an executable constraint.

jena-shacl/src/main/java/org/apache/jena/shacl/parser/ConstraintVisitor.java
```diff
@@ -0,0 +1,56 @@
...
+    default void visit(DisjointConstraint constraint){}
+    default void visit(EqualsConstraint constraint){}
+    default void visit(LessThanConstraint constraint){}
+    default void visit(ClosedConstraint constraint){}
+    default void visit(ClassConstraint constraint){}
+    default void visit(StrMaxLengthConstraint constraint){}
+    default void visit(StrLanguageIn constraint){}
+    default void visit(StrMinLengthConstraint constraint){}
+    default void visit(JViolationConstraint constraint){}
+    default void visit(DatatypeConstraint constraint){}
```
afs, line 47: Move to be with ClassConstaint.  (The ideal order is as in section 4: https://www.w3.org/TR/shacl/#core-components).

The J* then go separately.

We'll need a "ConstraitExt" for extensions (not a strong point of the visitor pattern!) but that can wait for now.

afs (changes requested):
Overall, looks good. Some refinement for the long term in comments.

jena-shacl/src/main/java/org/apache/jena/shacl/parser/ConstraintVisitor.java
fkleedorfer: We must implement `ReportConstraint.visit(ConstraintVisitor)`, though. Empty implementation for minimal friction?

afs: That'll do fine. The validation operations are `throw new UnsupportedOperationException()` but no-op works just as well for the visitor. (I suspect that ReportConstraint shouldn't exist at all! But from memory, passing the information it carries means adding other parameters in many places which even if purer, is verbose.)

kinow (approved):

afs (approved):

PR #883, opened 2020-12-07
JENA-2010: Unwrap to find a TDB2 dataset

afs:
With care, dig down to find a TDB2 dataset.

### DIFF
--- a/jena-fuseki2/jena-fuseki-webapp/src/main/java/org/apache/jena/fuseki/mgt/ActionCompact.java
+++ b/jena-fuseki2/jena-fuseki-webapp/src/main/java/org/apache/jena/fuseki/mgt/ActionCompact.java
@@ -20,11 +20,15 @@ package org.apache.jena.fuseki.mgt;
 
 import static java.lang.String.format;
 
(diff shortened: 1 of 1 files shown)


afs:
Leaving this as draft for the moment ... to think about whether any other cases will cause damage to the system.

It does apply in all settings (e.g. plain RDFDataset for an inference model over TDB2) but the text dataset is a wrapper layer on the base storage - much simpler.

rvesse (approved):
LGTM

After the discussion on list the other day I did think this might be possible but hadn't had chance to look yet.  Thanks for taking care of this 👍

rvesse:
Other cases that might prove difficult are where some sort of view/union is used across multiple distinct dataset.  However I think it's fine that the answer for that case is to provide another service that directly exposes the lower level dataset

afs:
Yes! This isn't all cases.

If the dataset does not fit the unwrapping algorithm, it does not find a TDB2.

It is only where one dataset wraps a single other dataset, not across multiple distinct datasets. An example of the lkatter not covered is inference and there, I agree, exposing the base dataset makes sense.

PR #884, opened 2020-12-08
JENA-2012: Test for UTF-16 high-low surrogates pairs

afs:
Add `checkString` into the parsing process that makes sure a high-surrogate is followed by a low-surrogate.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/lang/ParserBase.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/lang/ParserBase.java
@@ -153,7 +153,7 @@ public class ParserBase
     protected void startAggregate()   { aggregateDepth++; }
     protected int getAggregateDepth() { return aggregateDepth; }
     protected void finishAggregate()  { aggregateDepth--; }
-    
+
     protected Element compressGroupOfOneGroup(ElementGroup elg) {
         // remove group of one group.
         if ( elg.size() == 1 ) {
@@ -188,6 +188,25 @@ public class ParserBase
 
         lex = lex.substring(1) ;
         return NodeFactory.createLiteral(lex, lang, dt) ;
+    }
+
+    protected void checkString(String string, int line, int column) {
+        for ( int i = 0 ; i < string.length() ; i++ ) {
+            // Not "codePointAt" which does surrogate processing.
+            char ch = string.charAt(i);
+            // Check surrogate pairs are pairs.
+            if ( Character.isHighSurrogate(ch) ) {
+                i++;
+                if ( i == string.length() )
+                    throw new QueryParseException("Bad surrogate pair (end of string)", line, column);
+                char ch1 = string.charAt(i);
+                if ( ! Character.isLowSurrogate(ch1) ) {
+                    throw new QueryParseException("Bad surrogate pair (high surrogate not followed by low surrogate)", line, column);
+                }
+            } else if ( Character.isLowSurrogate(ch) ) {
+                throw new QueryParseException("Bad surrogate pair (low surrogate without high surrogate)", line, column);
+            }
+        }
     }
 
     protected Node createLiteral(String lexicalForm, String langTag, String datatypeURI) {
@@ -480,7 +499,7 @@ public class ParserBase
     public static String unescapeUnicode(String s, int line, int column) {
         return unescape(s, '\\', true, line, column);
     }
-    
+
     public static String unescapePName(String s, int line, int column) {
         char escape = '\\' ;
         int idx = s.indexOf(escape) ;
--- a/jena-base/src/main/java/org/apache/jena/atlas/io/InStreamUTF8.java
+++ b/jena-base/src/main/java/org/apache/jena/atlas/io/InStreamUTF8.java
@@ -28,11 +28,6 @@ import org.apache.jena.atlas.AtlasException ;
 /** Fast and streaming UTF-8 */
 public final class InStreamUTF8 extends Reader implements CharStream
 {
-    // TODO Add line and col counts.
-    // See arq.utf8.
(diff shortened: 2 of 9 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/lang/ParserBase.java
```diff
@@ -190,6 +190,24 @@ protected Node stripSign(Node node) {
...
+                if ( i == string.length() )
+                    throw new QueryParseException("Bad surrogate pair (end of string)", line, column);
+                char ch1 = string.charAt(i);
+                if ( ! Character.isLowSurrogate(ch1) ) {
+                    throw new QueryParseException("Bad surrogate pair (high surrogate not followed by low surrogate)", line, column);
+                }
+            } else if ( Character.isLowSurrogate(ch) )
+                throw new QueryParseException("Bad surrogate pair (low surrogate without high surrogate)", line, column);
+        }
+    }
```
kinow, line 209: :+1: 

I checked for a method that does something like this on Apache Commons Text, and Apache Commons Lang, but there's none. I thought some user would have requested that, interesting.

jena-base/src/main/java/org/apache/jena/atlas/io/InStreamUTF8.java
```diff
@@ -144,7 +144,18 @@ public final int advance()
...
         if ( x == -1 ) return -1 ;
-        return advance(input, x) ;
+
+        int codepoint = advance(input, x) ;
+        //System.out.printf("U+%04X\n", codepoint);
+
+        if ( false ) {
+//          if ( codepoint > Character.MAX_VALUE ) {
+//              throw new AtlasException("Out of range character (must use a surrogate pair)");
+//          }
```
kinow, line 154: Delete? :point_up:

jena-cmds/src/main/java/riotcmd/utf8.java
```diff
@@ -60,7 +60,8 @@ public static void main(String... args) {
                     } else
                         colNum++ ;
                     if ( !Character.isDefined(ch) )
-                        throw new AtlasException(String.format("No such codepoint: 0x%04X", ch)) ;
+                        //throw new AtlasException(String.format("No such codepoint: 0x%04X", ch)) ;
```
kinow, line 63: Delete? :point_up:

kinow (approved):

jena-cmds/src/main/java/riotcmd/utf8.java
afs: Done (this gives the test of being within the unicode version supported, not just any UTF-8 sequence).

jena-base/src/main/java/org/apache/jena/atlas/io/InStreamUTF8.java
afs: OK - removed the commented text, rearrange and "if(false)" and "Character.isDefined" (that is a switch between being a pure UTF-8 decoder and one that only accepts characters in the Unicode version that Java supports).

(I'd forgotten that this class would get into the PR!)

jena-arq/src/main/java/org/apache/jena/sparql/lang/ParserBase.java
```diff
@@ -190,6 +190,24 @@ protected Node stripSign(Node node) {
...
+                i++;
+                if ( i == string.length() )
+                    throw new QueryParseException("Bad surrogate pair (end of string)", line, column);
+                char ch1 = string.charAt(i);
+                if ( ! Character.isLowSurrogate(ch1) ) {
+                    throw new QueryParseException("Bad surrogate pair (high surrogate not followed by low surrogate)", line, column);
+                }
+            } else if ( Character.isLowSurrogate(ch) )
+                throw new QueryParseException("Bad surrogate pair (low surrogate without high surrogate)", line, column);
+        }
```
rvesse, line 208: Nit, the bracket style is pretty inconsistent here, I don't have a strong preference for using brackets when the `if` branch only has a single line but it would be nice to be consistent throughout the method 

For example the outer `if` uses `{}` because it has a multi-line block but its corresponding `else if` doesn't

Similarly the inner `if` are both single line blocks but one uses `{}` and one doesn't

jena-cmds/src/main/java/riotcmd/utf8.java
```diff
@@ -60,7 +60,7 @@ public static void main(String... args) {
                     } else
                         colNum++ ;
                     if ( !Character.isDefined(ch) )
-                        throw new AtlasException(String.format("No such codepoint: 0x%04X", ch)) ;
+                        System.out.printf("No such codepoint: 0x%04X\n", ch);
```
rvesse, line 63: Any reason why this is now just a log message rather than an error as it was previously?

jena-arq/src/main/java/org/apache/jena/sparql/lang/ParserBase.java
afs: Point taken. Changed.

jena-cmds/src/main/java/riotcmd/utf8.java
afs: Previous there was also a Character.isDefined in InStreamUTF8 and also a bad test that excluded surrogate use.

It is like this so first error does not terminate the command. Of course, it may flip to many errors!

When trying it out (on the rdf-tests data which is not large), stopping on the first error was unhelpful because the errors were in several bytes (surrogate pairs by the time it got to here).

rvesse (approved):

jena-cmds/src/main/java/riotcmd/utf8.java
rvesse: Makes perfect sense

PR #886, opened 2020-12-14
JENA-2015: RDF* annotation syntax

afs:
[JENA-2015](https://issues.apache.org/jira/projects/JENA/issues/JENA-2015)

includes as noted the Javacc turtle parser cleanup.

Does not include companion SPARQL* annotation syntax. (RDF* term syntax is already implemented.)


jena-arq/src/main/java/org/apache/jena/riot/RDFParser.java
```diff
@@ -334,8 +334,11 @@ private void parseNotUri(StreamRDF destination) {
             throw new RiotException("Failed to determine the RDF syntax (.lang or .base required)");
 
         ReaderRIOT readerRiot = createReader(ct);
-        if ( readerRiot == null )
-            throw new RiotException("No parser registered for content type: " + ct.getContentTypeStr());
+        if ( readerRiot == null ) {
+//            readerRiot = createReader(lang);
+//            if ( readerRiot == null )
+                throw new RiotException("No parser registered for content type: " + ct.getContentTypeStr());
```
kinow, line 340: Format/delete?

jena-arq/src/main/java/org/apache/jena/riot/lang/extra/LangParserBase.java
```diff
@@ -0,0 +1,201 @@
...
+    protected final Node nRDFobject     = RDF.Nodes.object ;
+
+    protected StreamRDF stream;
+    protected ParserProfile profile;
+
+    public LangParserBase() { }
+    
+    // These are essential calls unless the parser takes over the functions. 
+    // They can't easily be in the constructor because this class is inherited
+    // by the JaavCC generated parser.
```
kinow, line 66: s/JaavCC/JavaCC

jena-arq/src/main/java/org/apache/jena/riot/lang/extra/LangParserBase.java
```diff
@@ -0,0 +1,201 @@
...
+                    throw new QueryParseException("Bad surrogate pair (end of string)", line, column);
+                char ch1 = string.charAt(i);
+                if ( ! Character.isLowSurrogate(ch1) ) {
+                    throw new QueryParseException("Bad surrogate pair (high surrogate not followed by low surrogate)", line, column);
+                }
+            } else if ( Character.isLowSurrogate(ch) ) {
+                throw new QueryParseException("Bad surrogate pair (low surrogate without high surrogate)", line, column);
+            }
+        }
+    }
```
kinow, line 116: Was this code elsewhere and it was then moved here? Or is it duplicated? I remember seeing some similar code a few days ago I think.

kinow (approved):
:+1: looks good to me. Today-I-Learned about that `{| ... |}` syntax. Will try to remember that one, although I prefer to be more explicit and verbose I guess, with the syntax currently tested in this PR.

jena-arq/src/main/java/org/apache/jena/riot/lang/extra/LangParserBase.java
afs: Duplicated; there is some other methods copied as well.

The tiny difference is that the exceptions are RiotParseException ... except I missed this method!  Changed.

afs:
There are some functions copied - the exceptions are different (QueryParseException vs RiotParseException) - except I missed this function so updated to RiotParseException.

PR #887, opened 2020-12-16
JENA-2001: Print rdf:type if RDF prefix is available

afs:
Explaining "a" seems to come up when introducing RDF.

It is useful when "PREFIX rdf:" is not available. But when "PREFIX rdf:" is declared, explaining "a" is a (small) distraction when `rdf:type` is sufficiently concise.


jena-arq/src/main/java/org/apache/jena/riot/RDFWriterRegistry.java
```diff
@@ -22,14 +22,23 @@
...
 import org.apache.jena.riot.writer.* ;
 import org.apache.jena.sys.JenaSystem ;
 
+/**
+ * Writer registry. This is for writers presenting the functionality to write graphs and datasets, not streams.
+ * See {@link StreamRDFWriter} for registration of streaming writers.
+ *
+ * To register the language: see {@link RDFLanguages}.
+ *
+ * @see StreamRDFWriter for streamign writers.
```
kinow, line 37: s/streamign/streaming

kinow (approved):

PR #888, opened 2020-12-19
Fixes

afs:
JENA-2018
JENA-2019


kinow (commented):
Code looks good! Do we need unit tests for the cases detailed in the JIRA issues? Or is it already covered by other tests?

afs:
Good point

I've grok'ed the ruletest suite and added a test for that and also one for imports-as-shapes.
Without the fixes, they fail; with fixes, they pass.

kinow (approved):

PR #896, opened 2021-01-03
Update mockito(testing) to 3.6.28

afs:
Not an automatic dependabot upgrade. 

The artifact name has changed from `mockito-all` (stuck on v1) to `mockito-core` (v2, v3).

mockito-all seems to now have problem when used on Java16 (https://ci-builds.apache.org/job/Jena/job/Jena_JDK16/org.apache.jena$jena-core/18/testReport/ 03-Jan-2021 13:38:33)

This change seems to work (java8, java11) so hoping it works for Java16 as well.

### DIFF
--- a/jena-core/pom.xml
+++ b/jena-core/pom.xml
@@ -78,7 +78,7 @@
 
     <dependency>
       <groupId>org.mockito</groupId>
-      <artifactId>mockito-all</artifactId>
+      <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
 
--- a/jena-elephas/jena-elephas-mapreduce/pom.xml
+++ b/jena-elephas/jena-elephas-mapreduce/pom.xml
@@ -28,6 +28,7 @@
 
   <properties>
     <automatic.module.name>org.apache.jena.elephas.mapreduce</automatic.module.name>
+    <ver.mockito>1.9.5</ver.mockito>
   </properties>
 
   <dependencies>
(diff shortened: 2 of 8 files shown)


kinow (commented):
Good to know about this -all & -core issue. Travis CI failed for Java 8, but worked for Java 11.

I thought maybe they up'ed their minimum Java version, but looks like 3.x is Java 8 compatible 

>Mockito 3 does not introduce any breaking API changes, but now requires Java 8 over Java 6 for Mockito 2.

The errors are in the Elephas module

>[ERROR] Tests run: 111, Failures: 0, Errors: 111, Skipped: 0

I thought the Elephas module was perhaps broken on `master`, but Travis CI shows the last `master` build green. Perhaps something in this version of Mockito broke the tests of Elephas? :cry:

afs:
What is more mockito isn't a dependency of jena-elephas.

It comes (no PR):
```
[INFO] \- org.apache.mrunit:mrunit:jar:hadoop2:1.1.0:test
[INFO]    +- org.mockito:mockito-core:jar:1.9.5:test
```
(mrunit only runs on Java8, not later anyway)

but with the dependency mechanism seems to drag in as 3.2.28 which I can't explain.
Simple answer is to specify it directly in jena-elephas just to keep Java8 alive.

afs:
Seems to work (Java8) with the latest push. Forces the version of mockito to agree with mrunit, not jena elsewhere.

(It all might work anyway because there is also a initialization/start-up issue that showed up when dependencies changed.)

kinow (approved):

PR #901, opened 2021-01-25
JENA-2028: Jena cleanup

afs:
- Accumulation of code cleanup in core/arq
- removal of code marked deprecated in those and other modules.
- Sort out setup of SPARQL Update 

Other than removal of existing deprecated methods, this should not have an impact on user APIs.

The commit commits are indicators of the work done.


afs:
There will be a separate PR branched on this one for changes to abstract the IRI and establish the contract between IRI implement provider and the rest of Jena.

kinow (approved):

PR #910, opened 2021-02-01
JENA-2031: IRI provider/functional abstraction

afs:
This includes PR #901.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/system/IRIResolver.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/system/IRIResolver.java
@@ -19,23 +19,26 @@
 package org.apache.jena.riot.system;
 
 import java.io.PrintStream;
-import java.util.concurrent.Callable;
 
-import org.apache.jena.atlas.lib.Cache;
-import org.apache.jena.atlas.lib.CacheFactory;
-import org.apache.jena.atlas.lib.IRILib;
-import org.apache.jena.atlas.logging.Log;
-import org.apache.jena.iri.IRI;
-import org.apache.jena.iri.IRIException;
 import org.apache.jena.iri.IRIFactory;
 import org.apache.jena.iri.ViolationCodes;
 import org.apache.jena.iri.impl.PatternCompiler;
-import org.apache.jena.riot.RiotException;
-import org.apache.jena.riot.SysRIOT;
 
 /** IRI handling */
 public abstract class IRIResolver
 {
+    // ---- System-wide IRI Factory.
+
+    /** The IRI checker setup, focused on parsing and languages.
+     *  This is a clean version of jena-iri {@link IRIFactory#iriImplementation()}
+     *  modified to allow unregistered schemes and allow {@code <file:relative>} IRIs.
+     *
+     *  @see IRIFactory
+     */
+    public static IRIFactory iriFactory() {
+        return iriFactoryInst;
+    }
+
     private static boolean         showExceptions    = true;
 
     private static final boolean   ShowResolverSetup = false;
@@ -57,6 +60,7 @@ public abstract class IRIResolver
             printSetting(iriFactoryInst);
         }
 
+        // Jena 3.17.0 setup.
         // Accept any scheme.
         setErrorWarning(iriFactoryInst, ViolationCodes.UNREGISTERED_IANA_SCHEME, false, false);
 
@@ -65,12 +69,10 @@ public abstract class IRIResolver
         // The IRI is valid and does correct resolve when relative.
         setErrorWarning(iriFactoryInst, ViolationCodes.NON_INITIAL_DOT_SEGMENT, false, false);
 
-        // Turn off?? (ignored in CheckerIRI.iriViolations anyway).
-        // setErrorWarning(iriFactory, ViolationCodes.SCHEME_PATTERN_MATCH_FAILED, false, false);
-
-        // Choices
-        setErrorWarning(iriFactoryInst, ViolationCodes.LOWERCASE_PREFERRED, false, true);
+        // Choices: setting here does not seem to have any effect. See CheckerIRI and IRIProviderJena.
+        //setErrorWarning(iriFactoryInst, ViolationCodes.LOWERCASE_PREFERRED, false, true);
         //setErrorWarning(iriFactoryInst, ViolationCodes.PERCENT_ENCODING_SHOULD_BE_UPPERCASE, false, true);
+        //setErrorWarning(iriFactoryInst, ViolationCodes.SCHEME_PATTERN_MATCH_FAILED, false, true);
 
         // NFC tests are not well understood by general developers and these cause confusion.
         // See JENA-864
@@ -88,14 +90,12 @@ public abstract class IRIResolver
         // The set of legal characters depends on the Java version.
         // If not set, this causes test failures in Turtle and Trig eval tests.
         setErrorWarning(iriFactoryInst, ViolationCodes.UNASSIGNED_UNICODE_CHARACTER, false, false);
-       
 
         if ( ShowResolverSetup ) {
             System.out.println("---- After initialization ----");
             printSetting(iriFactoryInst);
         }
     }
-
 
     /** Set the error/warning state of a violation code.
      * @param factory   IRIFactory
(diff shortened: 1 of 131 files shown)


afs:
This is an unfinished draft.

See the package "irix" (this is a temporary location).

rvesse:
The basic interface looks fine as far as I'm able to pick it out amongst the swathe of related changes

afs:
This is now complete. There is further checking that can be done which is why it'll be "draft" for awhile.

`jena-iri` is still used directly in the XML I/O, `jena-jdbc-core` and cmds for IRI validation. Those can wait if they need changing at all.

Almost inevitably, from going through all the code, reviewing the parser tests, and generally going back to look at old code,



**Change 1**

Bad percent encoding e.g. `http://example/abc%XZdef` is strictly a syntax error in RFC 3986, not merely bad practice. In the changes here, it becomes an error.

There are no warnings for lowercase (upper case is recommended, but not required).

**Change 2**

URN handling: RFC 8141 tightens up the rules on URNs to be urn:NID:NSS, NID is at least 2 alphanumerics, and NSS at least one. This is more restrictive that the previous RFC 2141 which allowed NID to be "X-"

A consequence is that using `urn:NID:` as a prefix is strictly an illegal URN but Jena accepted it.

`urn:x:` was never legal but Jena accepted it and I would not be surprised if it is used in Jena users test data. It was in Jena although now that's all cleaned up and the Jena test suite passes in strict URI mode.

Currently, the default is to run with only light checking of URN specific rules.

afs:
Here is a list of links:

URI: https://tools.ietf.org/html/rfc3986
HTTP:     https://tools.ietf.org/html/rfc7230
URN:      https://tools.ietf.org/html/rfc8141
UUID:     https://tools.ietf.org/html/rfc4122 (as urn:uuid:)
DID:      https://www.w3.org/TR/did-core/
file:     https://tools.ietf.org/html/rfc8089#section-2

URI scheme registry: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
URN namespace registry:  https://www.iana.org/assignments/urn-namespaces/urn-namespaces.xhtml

afs:
Performance: pure parsing costs. (light test)

Test: Running "riot" with 4 BSBM 25 million .nt.gz in one run of the program (the argument being the first time warms the JVM; in fact the numbers don't show any effect). File has been used previously so in the file system cache.

Jena 3.17.0 and Jena+thisPR are within 0.5% of each other which is less than the timing variance.

Jena+PR with the [new IRI parser](https://github.com/afs/iri4ld/) is at least 12% faster. I'm surprised it's measurable.

afs:
JENA-2033 here to avoid a potential merge conflict because the fix is to a class modified by this PR.

jena-core/src/main/java/org/apache/jena/irix/IRIx.java
```diff
@@ -0,0 +1,194 @@
...
+ */
+
+package org.apache.jena.irix;
+
+import java.util.Objects;
+
+/**
+ * Support for RFC3986 IRIs.
+ * <p>
+ * The class {@code IRIx} is an abstraction layer; a provider is need to give an
```
kinow, line 26: s/is need/is needed

jena-core/src/main/java/org/apache/jena/irix/IRIx.java
```diff
@@ -0,0 +1,194 @@
...
+ * implementation. A provider is selected at start-up and is not expected to changes
```
kinow, line 27: s/to changes/to change

jena-core/src/main/java/org/apache/jena/irix/IRIx.java
```diff
@@ -0,0 +1,194 @@
...
+     * {@link IRIException}.
+     * <p>
+     * See {@link IRIs#resolve(String)} to create an absolute IRI, resolving
+     * against the system base if necessary.
+     * <p>
+     * See {@link IRIs#check(String)} to check a string is an absolute URI and is suitable for use in RDF.
+     * <p>
+     * See {@link IRIs#reference(String)} when the string is an absolute URI and
+     * should not be resolved against local system base (e.g. it was passed in from
+     * outside)to create an {@link IRIx} that is suitable for use in RDF.
```
kinow, line 50: `outside)to create` to `outside) to create`

jena-core/src/main/java/org/apache/jena/irix/IRIx.java
```diff
@@ -0,0 +1,194 @@
...
+     * See {@linkplain #isReference()} for testing whether a URI is suitable for use in RDF.
+     */
+    public abstract boolean isAbsolute();
+
+    /**
+     * A <a href="https://tools.ietf.org/html/rfc3986#section-4.2"><em>relative
+     * URI</em></a> one without a scheme, and maybe without some of the other parts.
+     * <p>
+     * Often it is just the path part.
+     * <p>
```
kinow, line 85: Unnecessary `<p>`? Unless it produces some extra line space I think.

jena-core/src/main/java/org/apache/jena/irix/IRIx.java
```diff
@@ -0,0 +1,194 @@
...
+     * See {@linkplain #isReference()} for testing whether a URI is suitable for use in RDF.
+     * <p>
+     * Note that a URI can be both "not absolute" and "not relative", e.g. {@code http://example/path#fragment}.
+     */
+    public abstract boolean isRelative();
+
+    /**
+     * An <em>RDF Reference</em> is an URI which has scheme.
+     * If it is hierarchical, it should a non-empty host- authority,
```
kinow, line 95: s/it should a non-empty host- authority/it should be a non-empty host-authority? (although I think the `-` is not used elsewhere, so just `host authority`?)

jena-core/src/main/java/org/apache/jena/irix/IRIx.java
```diff
@@ -0,0 +1,194 @@
...
+     * and may have a query component and may have a fragment component.
+     * This not a term in
+     * <a href="https://tools.ietf.org/html/rfc3986">RFC 3986</a>
+     * and it is not the same as "absolute URI".
+     * <p>
+     * In RDF data it is a
+     * <a href="https://www.w3.org/TR/rdf11-concepts/#section-IRIs">useful concept</a>.
+     * It is either an absolute URI, but if it is hierarchical, it must have a host.
```
kinow, line 103: Drop the `either`?

jena-core/src/main/java/org/apache/jena/irix/IRIx.java
```diff
@@ -0,0 +1,194 @@
...
+     * <p>
+     * Note that a URI can be both "not absolute" and "not relative", e.g. {@code http://example/path#fragment}.
+     */
+    public abstract boolean isRelative();
+
+    /**
+     * An <em>RDF Reference</em> is an URI which has scheme.
+     * If it is hierarchical, it should a non-empty host- authority,
+     * and may have a query component and may have a fragment component.
+     * This not a term in
```
kinow, line 97: Include an `is` after `This`?

jena-core/src/main/java/org/apache/jena/irix/package-info.java
```diff
@@ -0,0 +1,44 @@
...
+ * <p>
+ * The class {@link #IRIx} provides the application abstraction layer. There is an
+ * {@link #IRIProvider} that is the implementation of IRIs providing parsing and
+ * resolution. A provider is selected at start-up and is not expected to changes
+ * while the system is running.
+ * <p>
+ * {@link #IRIxResolver} is the main API. It provides {@link #IRIx} resolution with a
+ * base URI, and policy choices.
+ * <p>
+ * The class {@link #IRIs} providers functions related to IRIs including
```
kinow, line 30: s/providers/provides ?

jena-core/src/main/java/org/apache/jena/irix/package-info.java
```diff
@@ -0,0 +1,44 @@
...
+ * {@link IRIs#reference} to check that a string is suitable for us in RDF.
```
kinow, line 31: s/for us/for use/

jena-core/src/main/java/org/apache/jena/irix/IRIProvider.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Provider: an implementation of a factory for IRIs.
+ * This is not an application interface - is the plugin for the provider to jena.
+ *
+ * @see IRIxResolver
+ */
+public interface IRIProvider {
+    /**
+     * Create an IRI, throw {@link IRIException} is the string does not conform to the grammar.
```
kinow, line 29: s/is/if

jena-core/src/main/java/org/apache/jena/irix/IRIProvider.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    public IRIx create(String iri) throws IRIException;
+
+    /**
+     * Create an IRI, throw {@link IRIException} is the string does not conform to the grammar
```
kinow, line 34: s/is/if

jena-arq/src/main/java/org/apache/jena/riot/system/IRIResolver.java
```diff
@@ -129,391 +129,381 @@ private static void printErrorWarning(PrintStream ps, IRIFactory factory, int co
...
+//        public IRI resolve(String uriStr) {
+//            return other.resolve(uriStr);
+//        }
+//
+//        @Override
+//        synchronized
+//        public IRI resolveSilent(String uriStr) {
+//            return other.resolveSilent(uriStr);
+//        }
+//    }
```
kinow, line 508: Has it been moved to the irix package/module? Could it be deleted?

kinow (approved):
Minor comments in javadocs. But no issues with the code. Focused more on the IRIx and IRIs to understand what was the difference. Thanks for the great comments! :clap:

jena-arq/src/main/java/org/apache/jena/riot/system/IRIResolver.java
afs: This is the old code largely not used but there are still a few places using jena-iri IRI directly based on the IRIFactory defined in this class. These need tracking down and individually checking - jena-iri is better at warnings about IRIs than the future replacement.

The commented methods are a temporary way someone is (unexpected) using IRIREsolver and can restore their code easily. Hopefully, that won't happen! 

There is a balance - to give the new code as much time in the default branch as possible (or pushing out the release date), the task of chasing down every detail isn't completed but is a manageable length now.

PR #912, opened 2021-02-07
 JENA-2023: jena text cmds :: JENA-2038: rename legacy I/O interfaces.

afs:
JENA-2023: Fix package overlap of "jena" for jena-text commands.

JENA-2038: Rename interfaces (legacy APIs, but in the RDFWriter case, a name clash nowadays).


jena-cmds/src/main/java/org/apache/jena/cmd/Cmds.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.jena.sys.JenaSystem;
+
+/**
+ * Command execute by registering a name and a way to call "main".
+ * This is used by modules that are not a dependency of jena-cmds.
+ * <p>
+ * Example usage:
+ * <pre>
+ *
+ *
```
kinow, line 34: WIP or missing example? (also not sure if it was "command execution" or commands executed", etc)

jena-core/src/main/java/org/apache/jena/rdf/model/RDFWriterI.java
```diff
@@ -19,13 +19,13 @@
 package org.apache.jena.rdf.model;
 import java.io.Writer;
 import java.io.OutputStream;
-/** RDFWriter is an interface to RDF serializers.
+/** RDFWriterI is an interface to RDF serializers.
  *
  * <p>An <code>RDFWriter</code> is a class which serializes an RDF model
  * to some RDF serialization language.  RDF/XML, n-triple and n3 are
  * examples of serialization languages.</p>
```
kinow, line 26: The only thing that was a bit odd in this PR was the `I` added to the interface. Reading the JIRA issue, looks like it's to free the name `RDFWriter` and also because this is legacy code. If so, should we also `@deprecated` it?

kinow (approved):

jena-cmds/src/main/java/org/apache/jena/cmd/Cmds.java
afs: Done

jena-core/src/main/java/org/apache/jena/rdf/model/RDFWriterI.java
afs: I had quick look at deprecating it. It is still indirectly in the Model API with `Model` extends `RDFWriterF` and `RDFWriterF` has `getWriter`.

It is one of things where if you pull the thread, a lot more unravels!

It is the only way to do I/O in jena-core, without jena-arq/RIOT. RDF/XML input/output is in jena-core and together with basic NT and Turtle there to have the tests for RDF/XML and assemblers work in jena-core. Some or all of that could be moved to integration tests.

Can we treat that as a separate task for investigation?

kinow: >It is one of things where if you pull the thread, a lot more unravels!

:rofl: true! I bet you are finding a lot of these in all this pre Jena 4 work.

>Can we treat that as a separate task for investigation?

Sure thing, I've already approved it as I don't have a better suggestion for it now (we can rename it again in 4.1, 4.2, etc, if necessary anyway :man_shrugging: or deprecate it later too)

afs: It's not too bad overall - this (RDFReader etc) would quite a step though if it includes moving tests into the integration test area.

I prefer to have a period of stability before the actual release (do all the tidying things that got forgotten!).

PR #914, opened 2021-02-08
JENA-2042 FusekiServer::getPort() returns actual port

jmkeil:
`FusekiServer::getPort()` returns actual port, if it was randomly selected from the free ports. However, this does not guarantee to return the HTTPS port as described in the Java Doc.

### DIFF
--- a/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
+++ b/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
@@ -149,7 +149,7 @@ public class FusekiServer {
      */
     public int getPort() {
         //TODO garantee to return the HTTPS port, if HTTPS is enabled
-        return ((ServerConnector) fuseki.getJettyServer().getConnectors()[0]).getLocalPort();
+        return ((ServerConnector) getJettyServer().getConnectors()[0]).getLocalPort();
     }
 
     /** Get the underlying Jetty server which has also been set up. */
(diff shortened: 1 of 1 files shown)


afs:
Looking into the Jetty configuration is a bit uck. It's probably quite safe because that has been the way for many versions making it de facto the API.

Exploring alternatives:

A possibility is not support a port of zero and require the app choose one. The tests use `WebLib.choosePort` and explicitly set the port from outside.  Obviously, there is a timing issue that needs recovering by catching the bind exception and retrying. Probably no better than building in the game of "hunt the port" but I thought I'd mention it.

There is `fusekiServer.getJettyServer().getURI.getPort()` though all that does is to iterate to find the first connector that is a `NetworkConnector` in order, i.e. connector zero. and call `getLocalPort()`.

`org.eclipse.jetty.server.Server.getURI` shows how to dig information out of the connectors e.g. https.

jmkeil:
Further options:
* Remove promise to return the HTTPS port
* Make use of the way Fuseki initializes Jetty (see JettyHttps::server):
  * there are at most two connectors
  * second one is the HTTPS connector
  but that is error prone and not nice regarding maintainability
* iterate through all connectors

afs:
Bullet two looks good.

Code can look to see which is protocol a connector provides - returning some "http" and some "https" (and there will be at most one of a protocol in normal use).

https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java#L739

(that's 10.x but the 9.x code is the same - sometime Jena will move to Jetty 10 or 11, both require Java11 but we thought we'd let Jena moving to Java11 roll out first)

afs:
Hi - if you undraft this, I'll merge it and combine with work [JENA-2068](https://issues.apache.org/jira/browse/JENA-2068).

jmkeil:
Sure. Unfortunately, I didn't have time to apply the further discussed changes.

afs (approved):

PR #920, opened 2021-02-10
JENA-2032: Include https in prefix list

afs:
Misc cleanup:

Remove comments in POM files on how to incorporate ES.
Use IRIx in function iri()
Deprecate StandardFilter property that is no longer used.

### DIFF
--- a/jena-text/src/main/java/org/apache/jena/query/text/assembler/TextVocab.java
+++ b/jena-text/src/main/java/org/apache/jena/query/text/assembler/TextVocab.java
@@ -82,13 +82,6 @@ public class TextVocab
     public static final Resource lowerCaseFilter    = Vocab.resource(NS, "LowerCaseFilter");
     public static final Resource asciiFoldingFilter = Vocab.resource(NS, "ASCIIFoldingFilter");
 
-    // ElasticSearch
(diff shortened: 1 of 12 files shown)


jena-text/src/main/java/org/apache/jena/query/text/assembler/TextVocab.java
```diff
@@ -78,17 +78,16 @@
     public static final Resource whitespaceTokenizer = Vocab.resource(NS, "WhitespaceTokenizer");
     
     // Filters
+	/** 
+	 * "Standard Filter" is a no-op in Lucene 7.x and removed in Lucene 8.x.
+	 * @deprecated Do not use. To be removed.
+	 */
```
kinow, line 84: tabs? or maybe github UI is not showing spaces correctly.

kinow (approved):

jena-text/src/main/java/org/apache/jena/query/text/assembler/TextVocab.java
afs: Fixed - must have been some tabs in the file. Eclipse tries to copy the local indent style rather than apply the user settings regardless. It gets confused, not unreasonably, if the original context is then removed.

PR #921, opened 2021-02-13
JENA-2031: Fix handling of base URI for Windows file systems.

afs:
This is fixing the build failures on Windows where full file names including the drive letter, in `\` form.
This is a bug in the JENA-2031 changes causing an invalid URI for the base, which in turn lead to an exception.


jena-arq/src/main/java/org/apache/jena/sparql/expr/aggregate/AggMode.java
```diff
@@ -1,139 +1,139 @@
...
+    	        }
+    	        else {
+    	            amode.put(arrDouble[i],1);
+    	            if(mode==0) {
+    	            	mode = arrDouble[i];
+    	            }
+    	        }
+    	    }
+
+            return NodeValue.makeDecimal(mode);
```
kinow, line 132: Extra spaces

jena-arq/src/main/java/org/apache/jena/sparql/expr/aggregate/AggModeDistinct.java
```diff
@@ -1,138 +1,138 @@
...
+    	        }
+    	        else {
+    	            amode.put(arrDouble[i],1);
+    	            if(mode==0) {
+    	            	mode = arrDouble[i];
+    	            }
+    	        }
+    	    }
+
+            return NodeValue.makeDecimal(mode);
```
kinow, line 131: Using GH mobile UI. This file has a few lines with extra spaces (or tabs?) I think.

kinow (approved):

kinow:
Would it be worth adding a unit test (if not yet covered) for the single letter windows partition name? To prevent a regression..

jena-arq/src/main/java/org/apache/jena/sparql/expr/aggregate/AggModeDistinct.java
afs: Yes - tabs crept in, not at the start of the line - the files  in the PR were converted from windows line mode.

afs:
There is a test in `TestJenaReaderRIOT` - it tries all the ways to write file names.

It was `read_url_3` and `read_url_3x` that were failing on the Jenkins windows build.

PR #922, opened 2021-02-13
JENA-2019: Parser fix; printing fix; more tests

afs:
Fixes for reopened JENA-2019.

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/reasoner/rulesys/FunctorDatatype.java
+++ b/jena-core/src/main/java/org/apache/jena/reasoner/rulesys/FunctorDatatype.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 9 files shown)


jena-core/src/main/java/org/apache/jena/reasoner/rulesys/FunctorDatatype.java
```diff
@@ -0,0 +1,50 @@
...
+    public boolean isEqual(LiteralLabel value1, LiteralLabel value2) {
+        return isEqualByTerm(value1, value2) ;
+    }
+
+    @Override
+    public Object parse(String lexicalForm) { return lexicalForm ; }
+
+    @Override
+    public String unparse(Object value) { return value.toString(); }
+}
```
kinow, line 50: Missing newline for git

jena-core/src/main/java/org/apache/jena/reasoner/rulesys/Rule.java
```diff
@@ -1080,6 +1092,7 @@ private Rule doParseRule(boolean retainVarMap) {
                 r.isBackward = backwardRule;
                 return r;
             } catch (NoSuchElementException e) {
+                e.printStackTrace();
```
kinow, line 1095: Debug?

kinow (approved):

jena-core/src/main/java/org/apache/jena/reasoner/rulesys/Rule.java
afs: Arg! I've put that in, then taken it out, then put it in ...
Thanks for spotting that.

PR #925, opened 2021-02-15
JENA-2047: JENA-2048: More deprecations - jena-core, jena-arq.

afs:
A fairly mechanical processing of jena-core and jena-arq to search for \@Deprecated and remove class and methods where these have been deprecated for several years.


kinow (approved):
Looked at half of the changes, appears to be mainly removing code or formatting. If CI passes, then +1 :+1:

afs:
If you go "Files Changed" -> cog -> "Hide Whitespace Changes" it's better. It's still big and not very interesting; more deletions than additions.

We have several downstream projects, some do track dev builds, so we can see if it's OK that way as well.  I've already checked with RDF Delta.

The build does pass.

afs:
If you go "Files Changed" -> cog -> "Hide Whitespace Changes" it's better. It's still big and not very interesting; more deletions than additions.

We have several downstream projects, some do track dev builds, so we can see if it's OK that way as well.  I've already checked with RDF Delta.

The build does pass.

kinow:
>If you go "Files Changed" -> cog -> "Hide Whitespace Changes" it's better.

Ah! Today-I-Learned! Thanks @afs !

PR #926, opened 2021-02-17
JENA-2049: Binding improvements

afs:
While this is a big PR, most of the files changed are because all uses of binding in Jena have been converted to the new code. The old way (`BindingMap`) still works.

While making these changes, the process was to change some files to new style, then run the tests suite so that if behaviour changed, it was simpler to track down whether it was the translation of the usage code or the binding sub-system implementation.

The old style is still there in `BindingFactory` as deprecated static methods to smooth migration. The original classes in case for code using the classes directly and not via the factory.

The key new class is `BindingBuilder`.

The main change in code is `BindingMap BindingFactory.create()` becomes `BindingBuilder Binding.builder()` and the effects of that.


jena-arq/src/main/java/org/apache/jena/sparql/engine/binding/Binding4.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.Iterator;
+import java.util.Objects;
+import java.util.function.BiConsumer;
+
+import org.apache.jena.graph.Node;
+import org.apache.jena.sparql.core.Var;
+import org.apache.jena.sparql.engine.binding.itr.Itr;
+
+/**
+ * A binding implementation that supports 3 binding pairs.
```
kinow, line 29: s/3/4 ?

jena-arq/src/main/java/org/apache/jena/sparql/engine/binding/BindingBuilder.java
```diff
@@ -0,0 +1,486 @@
...
+import org.apache.jena.graph.Node;
+import org.apache.jena.sparql.core.Var;
+import org.apache.jena.sparql.engine.binding.itr.Itr;
+import org.apache.jena.sparql.util.FmtUtils;
+
+/**
+ * Build Bindings.
+ * <p>
+ * {@linkplain Binding} are immutable and can not be changed once constructed. A
+ * {@code BindingBuilder} accumulates some var/node pairs and at he build step,
```
kinow, line 35: "and at he build step" appears to be missing something?

jena-arq/src/main/java/org/apache/jena/sparql/engine/binding/BindingBuilder.java
```diff
@@ -0,0 +1,486 @@
...
+ *
+ * See also {@link BindingFactory}. When the number of var/node pairs is fixed and
+ * known, operations in {@link BindingFactory} directly create a binding without
+ * needing a builder.
+ */
+public class BindingBuilder
+{
+    // Do not implement .hashCode or .equals as value-based operations.
+    // Do not implement Binding or else it makes the coding accident of returning a
+    // BindingBuilder as a Binding result.
```
kinow, line 83: :+1:

jena-arq/src/main/java/org/apache/jena/sparql/engine/binding/BindingBuilder.java
```diff
@@ -0,0 +1,486 @@
...
+        map.put(var4, node4);
+        var1 = null;
+        node1 = null;
+        var2 = null;
+        node2 = null;
+        var3 = null;
+        node3 = null;
+        var4 = null;
+        node4 = null;
+        map.put(var, node);
```
kinow, line 248: Took me a few seconds re-reading it to see that `$add` and `$set` are not identical :smile: need more :coffee:

jena-arq/src/main/java/org/apache/jena/sparql/graph/NodeTransformLib.java
```diff
@@ -18,195 +18,228 @@
...
+import org.apache.jena.sparql.expr.Expr;
+import org.apache.jena.sparql.expr.ExprList;
 import org.apache.jena.sparql.expr.ExprTransform;
 
-public class NodeTransformLib
-{
-    /** Do a node{@literal ->}node conversion of an Op - return original BGP for "no change" */
+public class NodeTransformLib {
+    /**
+     * Do a node{@literal ->}node conversion of an Op - return original BGP for "no
```
kinow, line 43: Ah, I've been using `{@code ->}` when I needed something like this in Javadocs. Learned something new reviewing this PR. Thanks!

jena-extras/jena-querybuilder/src/test/java/org/apache/jena/arq/querybuilder/handlers/WhereHandlerTest.java
```diff
@@ -710,27 +688,26 @@ public void testAddValueRow_pfx_collection() {
...
-	
-		BindingHashMap binding = new BindingHashMap();
-		binding.add( v, NodeFactory.createURI( "one" ));
-		binding.add( x, NodeFactory.createLiteral("three"));
-		edat.add( binding );
-		binding = new BindingHashMap();
-		binding.add( v, NodeFactory.createURI( "two" ));
-		binding.add( x, NodeFactory.createLiteral("four"));
-		edat.add( binding );
+        setupBindings(edat, x, v);
```
kinow, line 697: Thought the space was wrong on this line, but looks like the rest of the file has tabs. Test file, no need to change it :+1:

kinow (approved):
Always learning something new from review these PR's. Thanks @afs!

jena-extras/jena-querybuilder/src/test/java/org/apache/jena/arq/querybuilder/handlers/WhereHandlerTest.java
afs: Old code had tabs.

This is why tabs are bad - display in GH is not as context sensitive as an IDE.

jena-arq/src/main/java/org/apache/jena/sparql/engine/binding/BindingBuilder.java
afs: ... or maybe less :coffee: 😄

afs:
Comments responded to and fixed in the force-pushed updates. 

Trying to make this a single commit.

PR #928, opened 2021-02-18
Jena-2052 run github action maven.yml also for windows server and macos

OyvindLGjesdal:
The  [commit](https://github.com/apache/jena/pull/928/commits/669ffcabc11ca85501e69cce383ded733b0d7d95)   contains [the last run](https://github.com/OyvindLGjesdal/jena/actions/runs/579527299) where the action behaves like I would expect, returning success for ubuntu, and failures (different failing tests) for macos and windows. There's a commit, removing branch `jena-2052` from branches, to only run the job on `main`.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -11,7 +11,12 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
 
     steps:
     - uses: actions/checkout@v2


afs:
Looks good.

Would you mind squashing this down to one commit please (it retains your name rather than us squash it which doesn't).

OyvindLGjesdal:
I've squashed this into a single commit, but when I'm trying to run this on my forked repo, I now get a disk full error on the windows build each time, during the tdb tests. I'll debug a bit and see if I can see what quotas the disk has.

afs:
TDB (1 and 2) tests can run into issues.  Memory mapped files on MS Windows do not get deleted from the file system until the JVM process exits This is a long standing Java bug.

The tests for Windows try to reuse database but it still leaves a significant over head.

Aslo, any crashes have before left dead directories (JUnit clean up not happening fully?).

If it causes problems, then just have Linux and MacOS - we have MS Windows testing on Jenkins . Adding MacOS to the GH action CI is still a step forward.

OyvindLGjesdal:
I'll remove the `windows-latest` from the matrix then. I didn't manage to change the directory or find a way around it.

It looks like there is 12 GB left for the run on the `D:` volume before running maven. I tried to change some options to use the `c:\ volume`, which has 65 GB free, but no luck. 

https://github.com/OyvindLGjesdal/jena/actions/runs/590467566 contains the latest job with disk info before the run. I see some errors in the build, but mostly "not found". Not sure what that implies.

I had one sucessful run on MacOS, but it also does timeout.

afs (approved):

afs:
Thanks - merged.

We can see what happens when run in the Apache Software Foundation account.

PR #929, opened 2021-02-18
JENA-2053: Use SslContextFactory.Server 

afs:
Other that his one change, which also works on 9.4.36, Fuseki runs with Jetty10.


kinow (approved):
Looks good to me :+1:

PR #930, opened 2021-02-20
JENA-2054: Allow ~ in file URIs

afs:
Review tests
Fix to run jena-iri tests in a build


jena-core/src/test/java/org/apache/jena/irix/TestIRIx.java
```diff
@@ -166,6 +170,8 @@ public TestIRIx(String name, IRIProvider provider) {
 
     @Test public void reference_07() { reference("http://example/", true); }
 
+    @Test public void reference_08() { reference("file:///a:/~jena/file", true); }
```
kinow, line 173: :+1:

kinow (approved):

PR #931, opened 2021-02-22
JENA-2031: More tests for IRIx

afs:
This PR adds a test suite for IRIx rather than relying on the provider to be correct.

It pins down the contract for IRIs in Jena.

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/irix/IRIProviderJenaIRI.java
+++ b/jena-core/src/main/java/org/apache/jena/irix/IRIProviderJenaIRI.java
@@ -20,6 +20,8 @@ package org.apache.jena.irix;
 
 import java.io.PrintStream;
 import java.util.Iterator;
+import java.util.Objects;
+import java.util.regex.Pattern;
 
 import org.apache.jena.iri.*;
 import org.apache.jena.iri.impl.PatternCompiler;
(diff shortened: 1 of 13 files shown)


afs:
There are mentions of "IRI3986". This is the replacement IRI code (https://github.com/afs/iri4ld) for after 4.0.0. Having the tests helps pin down the functionality. IRI3986 does pass the test suite so jena-iri and IRI3986 agree.

jena-core/src/main/java/org/apache/jena/irix/IRIProviderJenaIRI.java
```diff
@@ -161,33 +189,21 @@ private static IRI exceptions(IRI iri) {
...
-
                 case Violation.SCHEME_PATTERN_MATCH_FAILED:
                     if ( isURN(iri) && ! STRICT_URN )
                         continue;
                     if ( isFILE(iri) )
                         continue;
                     break;
                 case Violation.REQUIRED_COMPONENT_MISSING:
-                    if ( isFILE(iri) && ! STRICT_FILE )
+                    // jena-iri handling of file: URIs is only for (an interpreation of) RFC 1738.
```
kinow, line 201: s/interpreation/interpretation

jena-core/src/main/java/org/apache/jena/irix/IRIProviderJenaIRI.java
```diff
@@ -161,33 +189,21 @@ private static IRI exceptions(IRI iri) {
...
+                    // RFC8089 allows relative fil eURIs and a wider use of characters.
```
kinow, line 202: s/fil eURIs/file URIs

kinow (approved):

PR #940, opened 2021-03-01
 JENA-2031: Refine IRI settings

afs:
This is work inspired by parsing wikidata with Jena3, finding all the types of warnings and errors that arise then making the new code agree with Jena3, except where newer RFCs have changed the situation (e.g. percent encoding in DNS host names is now legal). The code also reflect jena's behaviour of warning about bad IRIs but not signalling a parse error - very large datasets have some less-than-perfect IRIs in them and aborting a load because of this is annoying.

Tests added to pin down expectations.

There is also a lot of clearing up and refactoring.

Performance checked: the Jena3 and Jena4 times to parse (BSBM data) are the same.

Hopefully, the last large IRI-related PR!


jena-core/src/main/java/org/apache/jena/irix/SystemIRIx.java
```diff
@@ -68,15 +89,24 @@ private static IRIx establishBaseURI() {
...
             // e.g. No filesystem.
-            return IRIx.create("urn:base:");
+            return IRIx.create(fallbackBaseURI);
         }
     }
 
-    private static IRIx setSystemBase(String baseStr) {
+    /**
+     * Create an {@link IRIx} suitable for a system base.
+     * This oepration always returns an {@link IRIx}
```
kinow, line 102: s/oepration/operation

kinow (approved):

PR #941, opened 2021-03-02
JENA-2060: Improve test robustness

afs:
And some information gathering code for `TestDatabaseOps.compact_prefixes_3` (which only fails with an intermittent NPE in deterministic code).

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -16,8 +16,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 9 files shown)


.github/workflows/maven.yml
```diff
@@ -16,8 +16,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # "windows" takes up a lot of diskspace due a longstanding JDK issue
```
kinow, line 19: :+1:

kinow (approved):
New code comments look good! Thanks!

PR #943, opened 2021-03-03
Code cleaning.

afs:
A few internal changes:

* Rename `dboe.Sys` as `SysDB` (this causes most of the file changes) to avoid two classes called `Sys`.
* Subcontract OS determination to Apache Commons lang3.
* Limit Fuseki to 8 threads when embedded. Also set the minimum to 2 which makes debugging the app easier.

### DIFF
--- a/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
+++ b/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
@@ -44,7 +44,6 @@ import org.apache.jena.fuseki.ctl.ActionMetrics;
 import org.apache.jena.fuseki.ctl.ActionPing;
 import org.apache.jena.fuseki.ctl.ActionStats;
 import org.apache.jena.fuseki.jetty.FusekiErrorHandler;
-import org.apache.jena.fuseki.jetty.JettyHttps;
 import org.apache.jena.fuseki.jetty.JettyLib;
 import org.apache.jena.fuseki.metrics.MetricsProviderRegistry;
 import org.apache.jena.fuseki.server.*;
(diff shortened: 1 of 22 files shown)


jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
```diff
@@ -736,6 +738,41 @@ private void serviceEndpointOperation(String datasetName, String endpointName, O
...
+         * <li>Use (-1,-1) for Jetty "default". The Jetty 9.4 defaults are (min=8,max=200).
+         * <li>If (min != -1, max is -1) then the default max is 20.
+         * <li>If (min is -1, max != -1) then the default min is 2.
+         * </ul>
+         */
+        public Builder numServerThreads(int minThreads, int maxThreads) {
+            if ( minThreads >= 0 && maxThreads > 0 ) {
+                if ( minThreads > maxThreads )
+                    throw new FusekiConfigException(String.format("Bad thread setting: (min=%d, max=%d)", minThreads, maxThreads));
+            }
```
kinow, line 756: Just wondering if this is set by a user external configuration (e.g. assembler file). In such case, couldn't the user pass `-1, -11` by accident? Does that create a useful error somewhere else?

I tried building Jena locally but it's failing for some reason. Upgraded to Java 11 and latest Maven, but still some annoying build error. Will figure out later what's wrong, but otherwise I would test to see what happens in this case, sorry.

kinow (approved):

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
afs: This isn't accessible from an assembler ATM.

Jetty has a test "if (maxThreads < minThreads)" but can change the use of min/max to make any negative numbers be "default".

PR #945, opened 2021-03-03
Update ModeSRS.java

chrdebru:
Changed hasSerialisation (UK English) to hasSerialization (US English) in the exception message. This error message is accidentally very misleading.


kinow (approved):

afs:
"serialization" is UK English as well.

The rule isn't as simple as "ise"=UK/"ize"=US.

In the UK, it depends on the root of the word (latin/greek/modern) and also the dictionary preference.

chrdebru:
Thanks for the clarification. The problem is that the predicate is geo:hasSerialization in the vocabulary, but the error message was mentioning geo:hasSerialisation. I thought the cause of the problem was related to UK English vs. US English.

afs (approved):

PR #948, opened 2021-03-05
jena-jdbc maintenance

afs:
Fix warnings:

* In tests, add `@SuppressWarnings("resource")` to classes.
* In main, rewrite to use try-with-resource (one case is not possible).
* Replace `c.newInstance()` with `c.getDeclaredConstructor().newInstance()`

Add `src/test/resources/log4j.properties` files (content of the same file in each case).

Remove `AbstractJenaDriverTests.river_connect_bad_06` which now does not pass and no longer seems to test a feature.


jena-core/src/main/java/org/apache/jena/irix/IRIProviderAny.java
```diff
@@ -21,7 +21,7 @@
 import java.util.Objects;
 
 /**
- * Non-resolving provider that accepts anything string for <...>
+ * Non-resolving provider that accepts anything string for &lt;...&gt;.
```
afs, line 24: Unrelated to jena-jdbc - another javadoc issue that checking the build showed up.

kinow (approved):

afs:
@rvesse - nice that the CVE warning has gone away.

OK to merge this post-fix cleanup now?

"CTR" still exists - review can be async.

rvesse:
@afs Thanks for catching that, weirdly that test was still passing locally

afs:
I don't doubt it - likewise, I've had cases where "old" code has been used. Seems to be SNAPSHOT related.
And not just in Eclipse.

PR #954, opened 2021-03-15
JENA-2004: Add application methods to Transactional

afs:
This PR adds operations to Transactional. They are default methods that call Txn.

See #872 - this is that discussion where Txn is not deprecated or code calling it changed. 

It means there complex code in the default methods.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/core/Transactional.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/core/Transactional.java
@@ -18,6 +18,8 @@
 
 package org.apache.jena.sparql.core;
 
+import java.util.function.Supplier;
+
 import org.apache.jena.query.ReadWrite ;
 import org.apache.jena.query.TxnType;
 import org.apache.jena.sparql.JenaTransactionException;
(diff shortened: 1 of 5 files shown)


afs:
cc @rvesse @strangepleasures 

I think this is the state of the discussions on #872.

jena-arq/src/main/java/org/apache/jena/sparql/core/Transactional.java
```diff
@@ -28,15 +30,17 @@
...
- * <p>{@code end()} is optional but preferred.
+ * <pre> begin(WRITE) ... abort() or commit() end()</pre>
+ * <p>{@code end()} is optional for "write" but is preferred.
  * <p>
- * Helper code is available {@link Txn} so, for example:
- * <pre>Txn.executeRead(dataset, {@literal ()->} { ... sparql query ... });</pre>
- * <pre>Txn.executeWrite(dataset, {@literal ()->} { ... sparql update ... });</pre>
- * or use one of <tt>Txn.calculateRead</tt> and <tt>Txn.executeWrite</tt>
+ * <h4>Application use</h4>
+ * Applications can conveniently execute the lifecycle with mthos to read or write:
```
rvesse, line 37: Typo: `mthos` -> `methods`

rvesse (approved):
I think as an incremental step this is fine

kinow (approved):

afs:
From #872:
An alternative design with TxnTrait (the default methods added separately) might be better. This leaves Transaction cleaner as an implementation API.

afs:
Sorry this has gone on for so long.

I went back to this and investigated the relationship to jena-text. It's hard to untangle because jena-text is involved in the transaction and coordinates with Lucene.

The contract for DatasetChanges is rather unclear - it predates full transaction support for all datasets.

I've raised [JENA-2071](https://issues.apache.org/jira/browse/JENA-2071) to deprecate `DatasetChanges` so it can be removed from general use, freeing up space to do an "outside" transaction monitor with new, definite contract.

If this is still relevant, or you have further thoughts, please add to the ticket [JENA-1656](https://issues.apache.org/jira/browse/JENA-1656).

afs:
(user error)

PR #955, opened 2021-03-15
JENA-2068: Fuseki main jetty and port configuration

afs:
Includes handling port 0 - allocate atomically when the server starts then finding what we got - in place of the "ask" then use the given port which has a timing hole albeit small.


kinow (approved):
Is the change described in the description text already included, and then we are just tidying up the workflows to make sure they run and verify the change is working? Or are you including more commits?

afs:
Sorry - wrong branch PR.

PR #956, opened 2021-03-15
JENA-2068: Fuseki main jetty and port configuration

afs:
Includes handling port 0 - allocate atomically when the server starts then finding what we got - in place of the "ask" then use the given port which has a timing hole albeit small.

### DIFF
--- a/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/cmds/FusekiMain.java
+++ b/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/cmds/FusekiMain.java
@@ -28,8 +28,6 @@ import arq.cmdline.CmdARQ;
 import arq.cmdline.ModAssembler;
 import arq.cmdline.ModDatasetAssembler;
 import org.apache.jena.assembler.exceptions.AssemblerException;
-import org.apache.jena.atlas.json.JSON;
-import org.apache.jena.atlas.json.JsonObject;
 import org.apache.jena.atlas.lib.FileOps;
 import org.apache.jena.atlas.logging.FmtLog;
 import org.apache.jena.atlas.web.AuthScheme;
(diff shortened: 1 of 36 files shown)


jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/cmds/FusekiMain.java
```diff
@@ -414,13 +433,14 @@ protected void processModulesAndArgs() {
 
     private int portNumber(ArgDecl arg) {
         String portStr = getValue(arg);
+        if ( portStr.isEmpty() )
```
kinow, line 436: Not sure if blanks could also be handled here with `StringUtils.isBlank` or .trim()?

jena-fuseki2/jena-fuseki-main/src/test/java/org/apache/jena/fuseki/main/TestFusekiServerBuild.java
```diff
@@ -0,0 +1,72 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.apache.jena.rdfconnection.RDFConnection;
+import org.apache.jena.rdfconnection.RDFConnectionFactory;
+import org.apache.jena.sparql.core.DatasetGraphFactory;
+import org.junit.Test;
+
+public class TestFusekiServerBuild {
+    // Most testing happens by use in the test suit.
```
kinow, line 31: s/suit/suite

kinow (approved):

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/cmds/FusekiMain.java
afs: This is from the command line which will have discarded (unquoted) whitespace.

`--port=""` or `--port=` is clearer whitespace.

There is a argument for applying it to numbers but the command hasn't up to now.

jena-fuseki2/jena-fuseki-main/src/test/java/org/apache/jena/fuseki/main/TestFusekiServerBuild.java
afs: Done - thanks.

PR #963, opened 2021-03-23
JENA-2074: JENA-2061: Buffer write RDF/XML

afs:
JENA-2076: Prune non-standard content types

And clean-up.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/atlas/web/ContentType.java
+++ b/jena-arq/src/main/java/org/apache/jena/atlas/web/ContentType.java
@@ -45,9 +45,9 @@ public class ContentType
      * Get the type/subtype as a string.
      * @see #toHeaderString toHeaderString for use in HTTP headers.
      */
-    
+
     public String getContentTypeStr() {
-        return mediaType.getContentType() ;
+        return mediaType.getContentTypeStr() ;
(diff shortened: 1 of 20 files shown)


rvesse (approved):
I think I'm fine with these changes around deprecating the old non-standard MIME types as part of Jena 4.x

*BUT* we probably want to explicitly mention this in the 3 -> 4 Migration guide because it could cause issues for users who are using Jena to interact with old/legacy servers that still use the non-standard MIME types.  I think in most cases users can likely workaround this by using existing Jena mechanisms to override the `Accept` header that gets sent and/or registering the non-standard MIME types explicitly.  However we should make users aware of the potential issue so we don't get a lot of "You broke our stuff" type emails once Jena 4 starts to get adopted.

afs:
Added to notes for release notes

The ones removed are "x-" and "text/nquads" (as ASCII, not UTF-8), none of which were ever formalised. I don't know of any client software that uses them. The RFC definition of "text/*" has morphed away from ASCII.  The yukky text/plain for N-triples is still there (RDF 1.0).

My guess is that Java11 with the removal of JDK-provided packages is the most likely user issue and its not even a Jena issue.

But expect the unexpected.

PR #977, opened 2021-04-01
Various small cleanup things.

afs:
Comment fixes, whitespace tidying, package access for RDF-star/TDB operations (public left over from development)

Iterator close support for Iter.flatMap and stream.close <=> iterator close.


.github/workflows/maven.yml
```diff
@@ -1,10 +1,11 @@
 ## Licensed under the terms of http://www.apache.org/licenses/LICENSE-2.0
 
 name: Apache Jena CI
+# Manual
+on: workflow_dispatch
 
-on:
-  push:
-    branches: [ main ]
```
kinow, line 7: Not useful to have it running automatically?

jena-core/src/main/java/org/apache/jena/util/iterator/WrappedIterator.java
```diff
@@ -31,78 +32,90 @@ a WrappedIterator is an ExtendedIterator wrapping around a plain (or
...
         Answer an ExtendedIterator wrapped round <code>it</code> which does not
         permit <code>.remove()</code> even if <code>it</code> does.
     */
     public static <T> WrappedIterator<T> createNoRemove( Iterator<T> it )
         { return new WrappedIterator<>( it, true ); }
-   
-    
+
+    /**
+     * Answer an ExtendedIterator wrapped round a {@link Stream} The extended
```
kinow, line 56: missing period before `The extended`?

kinow (approved):

.github/workflows/maven.yml
afs: GH actions serve multiple purposes and audiences but it is one mechanism. It was the same with travis.

One usage is for contributor-developers who have cloned jena and they are likely not using main but their development branch (looking at our PRs, they have all been developments branches in the cloned repo IIRC). And they should not accidently try to update the snapshots repo. What is more, for these people, it will be running on their github account and action-spend limits.

The other usage is for the project. But (ideally) without exporting the project setup to clones which as far as  I know isn't possible.

The project has Jenkins for various jobs. "main" gets updated when PRs are merged and the Jenkins jobs run. I think deploying to the ASF snapshots repo is easier from Jenkins (security credentials).

Building on "main" is duplicating Jenkins, so we have time to get refine this.

e.g. We could enable on PRs coming in.

jena-core/src/main/java/org/apache/jena/util/iterator/WrappedIterator.java
afs: Done! (Big forced push done!!)

PR #978, opened 2021-04-04
JENA-2080: TriX-star

afs:


jena-arq/testing/RIOT/Lang/TriX/trix-star-bad-emb-nested-3.trix
```diff
@@ -0,0 +1,19 @@
...
+        </triple>
+        <uri>http://example/P1</uri>
+        <uri>http://example/P2</uri>
+        <uri>http://example/o2</uri>
+      </triple>
+      <uri>http://example/p</uri>
+      <uri>http://example/o1</uri>
+    </triple>
+  </graph>
+</trix>
```
kinow, line 19: Reading the JIRA issue this was exactly what I thought, would a nested triple work as in RDF-star? :+1: Question answered!

kinow (approved):
Looks good to me!

afs:
Added a commit.

I made the mistake of looking and pulling on the loose thread of whether the TriX writer properly writes with prefixes safely. Writing `<qname>` is disabled but the code should work is enabled.

It is disabled because the regularity of no use of `<qname>` is more useful than space saving, which are small. The TriX writer is streaming and prefixes in-stream can't be handled anyway.

There wasn't a test for reading with xmlns and `<qname>` - test added.

kinow (approved):

PR #980, opened 2021-04-06
JENA-2081: add /$/compact/* endpoint to fuseki main

bsara:
Compaction can be a very important feature. But, if you're using something like rdf-delta to run your fuseki server, you don't have any way to run compaction while the server is still running. This PR moves the `/$/compact/*` endpoint from the webapp into main so that it can be supplied even when the non-full version of Fuseki is being used.

### DIFF
--- a/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
+++ b/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
@@ -46,6 +46,7 @@ import org.apache.jena.fuseki.access.DataAccessCtl;
 import org.apache.jena.fuseki.auth.Auth;
 import org.apache.jena.fuseki.auth.AuthPolicy;
 import org.apache.jena.fuseki.build.FusekiConfig;
+import org.apache.jena.fuseki.ctl.ActionCompact;
 import org.apache.jena.fuseki.ctl.ActionMetrics;
 import org.apache.jena.fuseki.ctl.ActionPing;
 import org.apache.jena.fuseki.ctl.ActionStats;
@@ -357,6 +358,7 @@ public class FusekiServer {
         private int                      maxThreads         = -1;
 
         private boolean                  verbose            = false;
+        private boolean                  withCompact        = false;
         private boolean                  withPing           = false;
         private boolean                  withMetrics        = false;
         private boolean                  withStats          = false;
@@ -501,6 +503,12 @@ public class FusekiServer {
             return this;
         }
 
+        /** Add the "/$/compact/*" servlet that triggers compaction for specified dataset. */
+        public Builder enableCompact(boolean withCompact) {
+            this.withCompact = withCompact;
+            return this;
+        }
+
         /**
          * Add the dataset with given name and a default set of services including update.
          * This is equivalent to {@code add(name, dataset, true)}.
@@ -627,6 +635,7 @@ public class FusekiServer {
             withPing  = argBoolean(server, FusekiVocab.pServerPing,  false);
             withStats = argBoolean(server, FusekiVocab.pServerStats, false);
             withMetrics = argBoolean(server, FusekiVocab.pServerMetrics, false);
+            withCompact = argBoolean(server, FusekiVocab.pServerCompact, false);
 
             // Extract settings - the server building is done in buildSecurityHandler,
             // buildAccessControl.  Dataset and graph level happen in assemblers.
@@ -1207,7 +1216,9 @@ public class FusekiServer {
                 addServlet(context, "/$/stats/*", new ActionStats());
             if ( withMetrics )
                 addServlet(context, "/$/metrics", new ActionMetrics());
-            // TODO Should we support registering other functionality e.g. /$/backup/* and /$/compact/*
+            if ( withCompact )
+                addServlet(context, "/$/compact/*", new ActionCompact());
+            // TODO Should we support registering other functionality e.g. /$/backup/*
 
             servlets.forEach(p-> addServlet(context, p.getLeft(), p.getRight()));
             filters.forEach (p-> addFilter(context, p.getLeft(), p.getRight()));
--- a/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
+++ b/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
@@ -50,6 +50,7 @@ import org.apache.jena.fuseki.ctl.ActionCompact;
 import org.apache.jena.fuseki.ctl.ActionMetrics;
 import org.apache.jena.fuseki.ctl.ActionPing;
 import org.apache.jena.fuseki.ctl.ActionStats;
+import org.apache.jena.fuseki.ctl.ActionTasks;
 import org.apache.jena.fuseki.jetty.FusekiErrorHandler;
 import org.apache.jena.fuseki.jetty.JettyLib;
 import org.apache.jena.fuseki.metrics.MetricsProviderRegistry;
@@ -362,6 +363,7 @@ public class FusekiServer {
         private boolean                  withPing           = false;
         private boolean                  withMetrics        = false;
         private boolean                  withStats          = false;
+        private boolean                  withTasks          = false;
 
         private String                   jettyServerConfig  = null;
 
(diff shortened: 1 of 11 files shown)


bsara:
found some issues in the tests...please wait to merge

bsara:
The tests have been fixed. Ok to merge if wanted.

kinow (approved):
Change sounds reasonable to me. Will it be called via a `curl` request, or is there a plan to update Fuseki interface to call it too? Perhaps display when it's running like a backup request? (not sure how long the compact action may take?)

Had a look at JIRA for issues related to compact but couldn't find one, so I think it'd need a new one.

rvesse:
I think this looks OK, I know we had a discussion on the dev lists about this change previously and can't remember exactly why we didn't make this change at the time

Probably needs @afs to review

afs:
Visual scan of the PR and it looks good.

> we didn't make this change at the time

IIRC it is the lack of "local only" operation of admin operations that change the database. stats/metrics don't do that. But this feature for compaction is "opt-in" by server startup. Anyway - from memory.

I'll try the PR out.

bsara:
@afs changes applied. Thank you for the quick response.

bsara:
I also figured that if you all were ok with this change. I might submit another for `/$/backups-list` and `/$/backup/*`, I just don't have an immediate need for it right now is all.

bsara:
> not sure how long the compact action may take?

@kinow in my situation, it takes about 18-20 minutes to complete because I have a very large dataset. From what I could tell, there is no way to determine if the process has completed unless you ping the server `/$/tasks/{taskId}` endpoint to check if is finished. I would imagine that continuously pinging that from the front-end wouldn't be very desirable. One could check once on load of some page component though.

bsara:
I've added a second commit to add the `/$/tasks` endpoint as well. I didn't put it in a separate PR because it seems to make sense to have it added when `/$/compact/*` is added since you can't actually ever check the status of your compact process without it. It is still has an opt-in separate from compact though.

jena-fuseki2/jena-fuseki-main/src/test/java/org/apache/jena/fuseki/main/TestFusekiMainCmd.java
```diff
@@ -102,4 +103,11 @@ private void server(String... cmdline) {
             assertNotEquals(0, x.readAllBytes().length);
         }
     }
+
+    @Test public void tasks_01() {
+        server("--mem", "--tasks", "/ds");
+        String x = HttpOp.execHttpGetString(serverURL+"/$/tasks");
+        assertNotNull(x);
+        assertNotEquals(0, x.length());
```
bsara, line 111: `JSON.parse` throws an error when calling it on `x` here because the root type of the JSON returned is a list, not an object. I opted to not update the parser to handle lists because it seemed silly to update it for this one test.

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
```diff
@@ -501,6 +505,18 @@ public Builder enableMetrics(boolean withMetrics) {
...
         }
 
+        /** Add the "/$/compact/*" servlet that triggers compaction for specified dataset. */
+        public Builder enableCompact(boolean withCompact) {
+            this.withCompact = withCompact;
+            return this;
+        }
+
+        /** Add the "/$/tasks" servlet that responds with info about tasks run on the server */
+        public Builder enableTasks(boolean withTasks) {
```
rvesse, line 515: Does it not make more sense to just automatically enable tasks if any background task feature is enabled?

jena-fuseki2/jena-fuseki-main/src/test/java/org/apache/jena/fuseki/main/TestFusekiMainCmd.java
afs: `JSON.parseAny` will read a list or array (or a atomic JSON type).

`jena-fuseki-weabapp TestAdmin` has some tests of tasks.

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
bsara: I considered this. But just because you're enabling a background task, that doesn't always mean that someone might want/need to track those tasks, so I decided to make it opt-in. But if automatically including it is what is wanted, then I can add it.

rvesse: Yeah I can see arguments for both approaches.

We recently had one user who was trying to `cron` a script that was requesting a compaction and then immediately deleting files out from under Fuseki/TDB ultimately causing it to crash.  Our response to them was to refer them to using the tasks API to only clean up files after compaction was finished.

So my strong preference is always to have task monitoring enabled if we have any of the potentially long running background task APIs enabled.

Be interested to see what @kinow and @afs's opinions are as they use Fuseki more heavily than me

jena-fuseki2/jena-fuseki-main/src/test/java/org/apache/jena/fuseki/main/TestFusekiMainCmd.java
bsara: fixed.

kinow:
> > not sure how long the compact action may take?
> 
> @kinow in my situation, it takes about 18-20 minutes to complete because I have a very large dataset. From what I could tell, there is no way to determine if the process has completed unless you ping the server `/$/tasks/{taskId}` endpoint to check if is finished. I would imagine that continuously pinging that from the front-end wouldn't be very desirable. One could check once on load of some page component though.

I think that's what happens with the backup task (I think it's around [this function](https://github.com/apache/jena/blob/31dc0d328c4858401e5d3fa99702c97eba0383a0/jena-fuseki2/jena-fuseki-webapp/src/main/webapp/js/app/views/dataset-management.js#L144-L153) that is [scheduled](https://github.com/apache/jena/blob/31dc0d328c4858401e5d3fa99702c97eba0383a0/jena-fuseki2/jena-fuseki-webapp/src/main/webapp/js/app/models/task.js#L79-L83) to run [every 500ms or 1s I think](https://github.com/apache/jena/blob/31dc0d328c4858401e5d3fa99702c97eba0383a0/jena-fuseki2/jena-fuseki-webapp/src/main/webapp/js/app/models/task.js#L91-L99).

But that happens only when you execute the backup action. And if you refresh the screen that's gone, you have no more access to the task progress.

In the branch where I am porting it to Vue, I am setting a permanent view for tasks that [updates every 10 seconds](https://github.com/kinow/jena/blob/7580da22cab35b6fbf76006c10544eeba851d84d/jena-fuseki2/jena-fuseki-ui/src/views/manage/Tasks.vue#L118-L126).

But my intention is to first have a replacement for Fuseki Backbone.js app, and then discuss the possibility of perhaps having a secure WebSockets that sends the data to the UI when available. This way we remove the polling, and can send information about any kind of server management tasks to the UI - so I think in the future we could show back up, compression, and other kinds of tasks if that would save users of having to log in to servers and run commands, or having to save HTTP requests to submit to the server.

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
kinow: I am not maintaining a production Jena instance at the moment. But at least for me I prefer to have the tasks endpoint accessible to that I can either check task statuses manually via command line `curl`, or preferably via Fuseki UI.

bsara: sounds like there is a consensus to add it by default. I guess if someone wants to have it disabled, an opt-out option could always be added in the future.

bsara: OK, I've updated the code to enable tasks by default if compact is enabled.

afs (approved):
Everything done!

afs:
> I might submit another for /$/backups-list and /$/backup/*

@bsara -- yes, please. That would be great.

bsara:
Is there a schedule for releases on this project? I.E. when can I expect to see a production jena with these changes?

afs:
The Jena project aims for a release every 3-4 months.

The project operates a "green main branch" - the development main branch is releasable unless there is special reason why it is not in a fit state. There is no special preparation needed other than set the version number.

You are welcome to clone the repo and build the code ("mvn clean install").

As an Apache project, any release by the project is subject to a vote by the PMC. At least 3 PMC members need to vote. This is important because a formal release is transfer from any development build (which is "personal" - it has no standing as a Apache release) to one released by the Apache Software Foundation. The Release Manager is not personally legally liable.

On this issue, we need to decide on access control, or not. These functions are open. As a principle, we need to take these security issues seriously. In some environments, the network/VM/container setup means that serves are not generally accessible but that isn't universal.

One way is restrict to localhost. Another is have a admin/password.

If you want an advance, non-release, build of RDF Delta with development Fuseki, then that is for discussion on RDF Delta. (It is a matter of setting the Jena version but SNAPSHOTS for dependencies can't be released to maven central.)

PR #981, opened 2021-04-07
Handle shortened IpV6 as resolved by chrome

sparkhi:
Chrome browser appears to resolve localhost as IpV6 in the form [::1] which does not work as part of the `LocalhostFilter` and we get 403 forbidden
OS: Ubuntu 20.04.2 LTS
Browser: Chrome 89.0.4389.90

### DIFF
--- a/jena-fuseki2/jena-fuseki-webapp/src/main/java/org/apache/jena/fuseki/authz/LocalhostFilter.java
+++ b/jena-fuseki2/jena-fuseki-webapp/src/main/java/org/apache/jena/fuseki/authz/LocalhostFilter.java
@@ -53,14 +53,14 @@ public class LocalhostFilter extends AuthorizationFilter403 {
 
     private static String LOCALHOST_IpV6_a = "[0:0:0:0:0:0:0:1]";
     private static String LOCALHOST_IpV6_b = "0:0:0:0:0:0:0:1";
+    private static String LOCALHOST_IpV6_c = "[::1]";
     private static String LOCALHOST_IpV4 =  "127.0.0.1";   // Strictly, 127.*.*.*
 
     private static final Collection<String> localhosts = new HashSet<>(
-        Arrays.asList(LOCALHOST_IpV4, LOCALHOST_IpV6_a, LOCALHOST_IpV6_b));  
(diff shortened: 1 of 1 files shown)


afs:
Hi there,

Which version of Jena is this about and is it the Fuseki packaging for the standalone server with UI?
What goes wrong?

There was a problem in Jena 3.17.0 (and only that version) : [JENA-2072](https://issues.apache.org/jira/browse/JENA-2072) (related [JENA-2011](https://issues.apache.org/jira/browse/JENA-2011)).

Upgrading components, (presumably Jetty, which is used for that form of Fuseki) made it go away but we didn't find the root cause.
`[::1]` should become `[0:0:0:0:0:0:0:1]` but I, for one, am not clear which component is responsible for converting to canonical form. We can add this to be robust anyway.

rvesse (approved):

sparkhi:
I am actually using an older version, but went looking into source code if it has been fixed in latest code. 
**Symptom**: Could not browse dataset in the UI because `http://localhost:3030/$/server` was returning 403. It worked well in Firefox but not in Chrome. When examining the remote address in chrome developer tools, the address appeared to be shortened to `[::1]`

Comparison of firefox / chrome in the attached image 
image

afs:
@sparkhi -- that is the same details as [JENA-2011](https://issues.apache.org/jira/browse/JENA-2011). That will have affected earlier versions than 3.17.0 and which added "[...]" form.

I have just tried in Chrome, with Jena 4.0.0 and when I see `[::1]:3030` in a request, and it is correctly being let through to localhost only operations by `[0:0:0:0:0:0:0:1]`

`[::1]` is just a way of writing `[0:0:0:0:0:0:0:1]` - it is not a different address -- `::` is "fill as many '0:' as needed to get the right length".

Would it be possible to try Jena 4.0.0? It would be good to pindown what is really happening even when applying the patch. If it is still broken we may need a x.x.1 release.

sparkhi:
Got it. Feel free to reject the PR atm, I might not be able to update on my current system, and I do not have another machine handy which would do the IpV6 request. 
Apologies!!

afs:
No problem. It is harmless to add it anyway.

Thanks for the details - all helps to nail down what has been happening.

kinow (approved):
Looks harmless :+1:

afs:
"mostly harmless"

42

kinow:
> mostly harmless

No spoilers, I'm still at the book 3 :laughing:

PR #983, opened 2021-04-12
JENA-2086: No need to reset the journal every transaction.

afs:


kinow (approved):

rvesse (commented):
Honestly I'm struggling to understand what's actually changed here.  AFAICT this just addresses an issue where `end()` could get called multiple times unnecessarily.  Not sure how the underlying issue is actually getting fixed

jena-db/jena-dboe-transaction/src/main/java/org/apache/jena/dboe/transaction/txn/TransactionCoordinator.java
```diff
@@ -705,12 +705,11 @@ private boolean promotionWaitForWriters() {
     /** Signal that the transaction has finished. */
     /*package*/ void completed(Transaction transaction) {
         finishActiveTransaction(transaction);
-        journal.reset();
         notifyEnd(transaction);
```
rvesse, line 708: I guess the above is the actual substantiative change in this PR. Any journal clean up actually happens elsewhere in abort/commit logic?

afs: Yes.

One line change!

The reported issue is that threads are piling up on locks. Internally in the JDK, the channel.sync() causes low level synchronization.

For a TDB2 reader, there is no journal handling needed. No `sync`.

For a writer (a transaction that finishes in write state - it may have been promoted), the journal handling is in `executeCommit `-`executeCommitWriter` which ends with `truncate`-`executeCommit` (which has an moderately unnecessary `sync` - it leaves things in a neat state but is strictly not needed).

The journal in TDB2 is never very big, especifically, less than a disk block. One transaction's work of root identifiers for the indexes and node table which is some 100's of bytes. It is not the transaction changes.

rvesse (approved):

PR #987, opened 2021-04-20
JENA-2094: Run tests in strict environment

afs:
PR #986 fixed the problem reported but it broke some test that run in "strict" mode ("strict" mode validates the IRIProvider, not the way Jena uses it). 

This PR runs the TS_IRIx tests in strict mode.

### DIFF
--- a/jena-core/src/test/java/org/apache/jena/irix/AbstractTestIRIx.java
+++ b/jena-core/src/test/java/org/apache/jena/irix/AbstractTestIRIx.java
@@ -27,7 +27,10 @@ import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.runners.Parameterized.Parameters;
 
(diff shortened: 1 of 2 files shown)


jena-core/src/test/java/org/apache/jena/irix/AbstractTestIRIx.java
```diff
@@ -27,7 +27,10 @@
 import org.junit.BeforeClass;
 import org.junit.runners.Parameterized.Parameters;
 
-/** Test suite driver for IRIx */
+/** Test suite driver for IRIx.
+ * The test execution environment is set to be "strict".
+ * Tests can  change this; it is reset each test.
```
kinow, line 32: missing "after" before each I think, but not important

kinow (approved):

PR #988, opened 2021-04-26
[JENA-2096] Added ability to delete old DB after compaction has completed successfully

bsara:
Adds two things:
1. adds `--delete` option to `tdbcompact`
2. adds `deleteOld` parameter to `/$/compact` endpoint.

These options enable the deletion of the previous DB after compaction has successfully completed. The default behavior remains the same as it was previous to this PR (I.E. deletion does not take place).

### DIFF
--- a/jena-cmds/src/main/java/tdb2/tdbcompact.java
+++ b/jena-cmds/src/main/java/tdb2/tdbcompact.java
@@ -18,18 +18,33 @@
 
 package tdb2;
 
+import org.apache.jena.cmd.ArgDecl;
 import org.apache.jena.tdb2.store.DatasetGraphSwitchable;
 import org.apache.jena.tdb2.sys.DatabaseOps;
 import tdb2.cmdline.CmdTDB;
 
 public class tdbcompact extends CmdTDB {
+    private static final ArgDecl argDeleteOld = new ArgDecl(ArgDecl.NoValue, "deleteOld");
+
+    private boolean shouldDeleteOld = false;
+
+
     static public void main(String... argv) {
         CmdTDB.init() ;
         new tdbcompact(argv).mainRun() ;
     }
 
     protected tdbcompact(String[] argv) {
-        super(argv) ;
+        super(argv);
+
+        super.add(argDeleteOld, "--deleteOld", "Delete old database after compaction");
+    }
+
+    @Override
+    protected void processModulesAndArgs() {
+        super.processModulesAndArgs();
+
+        shouldDeleteOld = contains(argDeleteOld);
     }
 
     @Override
@@ -41,7 +56,7 @@ public class tdbcompact extends CmdTDB {
     protected void exec() {
         DatasetGraphSwitchable dsg = getDatabaseContainer() ;
         long start = System.currentTimeMillis();
-        DatabaseOps.compact(dsg) ;
+        DatabaseOps.compact(dsg, shouldDeleteOld) ;
         long finish = System.currentTimeMillis();
         System.out.printf("Compacted in %.3fs\n", (finish-start)/1000.0);
     }
--- a/jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/DatabaseMgr.java
+++ b/jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/DatabaseMgr.java
@@ -33,7 +33,7 @@ import org.apache.jena.tdb2.sys.TDBInternal;
  * @see TDB2Factory
  */
 public class DatabaseMgr {
-    
+
     private DatabaseMgr() {}
 
     // All creation of DatasetGraph for TDB2 goes through this method.
@@ -55,11 +55,29 @@ public class DatabaseMgr {
      * Compact a datasets which must be a switchable TDB database.
      * This is the normal dataset type for on-disk TDB2 databases.
      *
+     * Deletes old database after successful compaction if `shouldDeleteOld` is `true`.
+     *
      * @param container
+     *
+     * @deprecated Use `compact(container, false)` instead.
      */
+    @Deprecated
     public static void compact(DatasetGraph container) {
+        compact(container, false);
+    }
+
+    /**
+     * Compact a datasets which must be a switchable TDB database.
+     * This is the normal dataset type for on-disk TDB2 databases.
+     *
+     * Deletes old database after successful compaction if `shouldDeleteOld` is `true`.
+     *
+     * @param container
+     * @param shouldDeleteOld
+     */
+    public static void compact(DatasetGraph container, boolean shouldDeleteOld) {
         DatasetGraphSwitchable dsg = requireSwitchable(container);
-        DatabaseOps.compact(dsg);
+        DatabaseOps.compact(dsg, shouldDeleteOld);
     }
 
     /**
--- a/jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/sys/DatabaseOps.java
+++ b/jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/sys/DatabaseOps.java
@@ -20,7 +20,9 @@ package org.apache.jena.tdb2.sys;
 
 import java.io.*;
 import java.nio.file.*;
+import java.util.Comparator;
 import java.util.List;
+import java.util.stream.Stream;
 import java.util.zip.GZIPOutputStream;
 
 import org.apache.jena.atlas.RuntimeIOException;
@@ -83,19 +85,19 @@ public class DatabaseOps {
             return new DatasetGraphSwitchable(null, location, dsg);
         }
         // Exists?
-       if ( ! location.exists() )
-           throw new TDBException("No such location: "+location);
-       Path path = IOX.asPath(location);
-       // Scan for DBs
-       Path db = findLocation(path, dbPrefix);
-       if ( db == null ) {
-           db = path.resolve(dbPrefix+SEP+startCount);
-           IOX.createDirectory(db);
-       }
-       Location loc2 = IOX.asLocation(db);
-       DatasetGraphTDB dsg = StoreConnection.connectCreate(loc2, params).getDatasetGraphTDB();
-       DatasetGraphSwitchable appDSG = new DatasetGraphSwitchable(path, location, dsg);
-       return appDSG;
+        if ( ! location.exists() )
+            throw new TDBException("No such location: "+location);
+        Path path = IOX.asPath(location);
+        // Scan for DBs
+        Path db = findLocation(path, dbPrefix);
+        if ( db == null ) {
+            db = path.resolve(dbPrefix+SEP+startCount);
+            IOX.createDirectory(db);
+        }
+        Location loc2 = IOX.asLocation(db);
+        DatasetGraphTDB dsg = StoreConnection.connectCreate(loc2, params).getDatasetGraphTDB();
+        DatasetGraphSwitchable appDSG = new DatasetGraphSwitchable(path, location, dsg);
+        return appDSG;
     }
 
     public static String backup(DatasetGraphSwitchable container) {
@@ -107,13 +109,13 @@ public class DatabaseOps {
 
         DatasetGraph dsg = container;
 
-//  // Per backup source lock.
-//  synchronized(activeBackups) {
-//      // Atomically check-and-set
-//      if ( activeBackups.contains(dsg) )
-//          Log.warn(Fuseki.serverLog, "Backup already in progress");
-//      activeBackups.add(dsg);
-//  }
+        //  // Per backup source lock.
+        //  synchronized(activeBackups) {
+        //      // Atomically check-and-set
+        //      if ( activeBackups.contains(dsg) )
+        //          Log.warn(Fuseki.serverLog, "Backup already in progress");
+        //      activeBackups.add(dsg);
+        //  }
 
         Pair<OutputStream, Path> x = openUniqueFileForWriting(backupDir, BACKUPS_FN, "nq.gz");
         try (OutputStream out2 = x.getLeft();
@@ -165,7 +167,16 @@ public class DatabaseOps {
     // JVM-wide :-(
     private static Object compactionLock = new Object();
 
+
+    /**
+     * @deprecated Use `compact(container, false)` instead.
+     */
+    @Deprecated
     public static void compact(DatasetGraphSwitchable container) {
+        compact(container, false);
+    }
+
+    public static void compact(DatasetGraphSwitchable container, boolean shouldDeleteOld) {
         checkSupportsAdmin(container);
         synchronized(compactionLock) {
             Path base = container.getContainerPath();
@@ -195,6 +206,19 @@ public class DatabaseOps {
             LOG.debug(String.format("Compact %s -> %s\n", db1.getFileName(), db2.getFileName()));
 
             compact(container, loc1, loc2);
+
+            if ( shouldDeleteOld ) {
+                Path loc1Path = IOX.asPath(loc1);
+                LOG.debug("Deleting old database after successful compaction (old db path='" + loc1Path + "')...");
+
+                try (Stream<Path> walk = Files.walk(loc1Path)){
+                    walk.sorted(Comparator.reverseOrder())
+                        .map(Path::toFile)
+                        .forEach(File::delete);
+                } catch (IOException ex) {
+                    throw IOX.exception(ex);
+                }
+            }
         }
     }
 
--- a/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionCompact.java
+++ b/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionCompact.java
@@ -88,10 +88,17 @@ public class ActionCompact extends ActionAsyncTask
     }
 
     static class CompactTask extends TaskBase {
-        static private Logger log = Fuseki.compactLog;
+        static private final Logger log = Fuseki.compactLog;
+
+        private final boolean shouldDeleteOld;
 
         public CompactTask(HttpAction action) {
             super(action);
+
+            String deleteOldParam = action.request.getParameter("deleteOld");
+
+            this.shouldDeleteOld = ( deleteOldParam != null
+                                     && ( deleteOldParam.isEmpty() || deleteOldParam.equalsIgnoreCase("true") ) );
         }
 
         @Override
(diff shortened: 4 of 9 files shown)


jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionCompact.java
```diff
@@ -88,18 +88,25 @@ private static DatasetGraph unwrap(DatasetGraph dsg, Predicate<DatasetGraph> pre
...
+
+        private final boolean shouldDeleteOld;
 
         public CompactTask(HttpAction action) {
             super(action);
+
+            String deleteOldParam = action.request.getParameter("deleteOld");
+
+            this.shouldDeleteOld = ( deleteOldParam != null
+                && ( deleteOldParam.isEmpty() || deleteOldParam.equalsIgnoreCase("true") ) );
```
kinow, line 101: If I understood it correctly (having first :coffee: of the day), in the command line the database is not deleted by default.

But here in Fuseki if you do not pass a parameter (`http//url/`?) then it is not deleted. If you pass an empty parameter (`http//url/?deleteOld=` ?) then it is deleted? Or if you pass `"true"` (`http//url/?deleteOld=true` ?) then it's also deleted.

If that's correct, then I think it's the old database is not deleted only if I do not specify the parameter or if I set its value to `"false"`? Shouldn't we ignore blank/empty strings too, and only delete if the param is `true`, having this way the same behaviour as in the cli?

```suggestion
            this.shouldDeleteOld = deleteOldParam != null && deleteOldParam.trim().equalsIgnoreCase("true");
```

kinow:
(We'll also need a JIRA for this)

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionCompact.java
bsara: The reason that I coded it this way is so that the following will both trigger a deletion: `http//url/?deleteOld` & `http//url/?deleteOld=true`. Since `deleteOld` is a boolean/flag parameter, its mere presence implies `true` if no explicit value is set. This is pretty common for boolean URL params in my experience.

bsara: I actually believe that this probably should be in the body of the POST, but I'm trying to follow already existing conventions in the project.

kinow: Ah, now that makes sense @bsara. I was wondering why you'd chose a different behavior here :+1: thanks!

kinow:
Created https://issues.apache.org/jira/browse/JENA-2096 for this issue

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/sys/DatabaseOps.java
```diff
@@ -198,6 +198,12 @@ public static void compact(DatasetGraphSwitchable container) {
             LOG.debug(String.format("Compact %s -> %s\n", db1.getFileName(), db2.getFileName()));
 
             compact(container, loc1, loc2);
+
+            if ( shouldDeleteOld ) {
+                Path loc1Path = IOX.asPath(loc1);
+                LOG.info("Deleting old database after successful compaction (old db path='" + loc1Path + "'...");
+                IOX.delete(loc1Path);
```
rvesse, line 205: Does this suppress errors?

Doing this on Windows has a good chance of failing due to the longstanding JVM bug with deleting memory mapped files

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionCompact.java
rvesse: Makes sense but we should ensure that this is explicit in the documentation

rvesse (commented):
Generally looks good, couple of minor comments

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/sys/DatabaseOps.java
afs: `Sys.isWindows`.

On windows, it will work ... but the files do not go away until the JVM exits.

afs, line 205: `IOX.delete` (which calls JDK `Files.delete`) only delete a file or an empty directory.

See documentation for `Files.delete` - it needs a file walk to delete recursively (Don't use `FileOps.clearAll` - old code, needs updating.)

Called on a directory with files, `IOX.delete` throws `DirectoryNotEmptyException`.

If there a test for this?

bsara: how do I handle this?

bsara: No test yet. Wanted to see how the change was going to be received first.

bsara: I've updated this to use `Files.walk()...`.

bsara: I've updated this to use `Files.walk()...`

afs: > No test yet. Wanted to see how the change was going to be received first.

Better to set this as "draft" if it is not ready for merging to be clear.

bsara: test added

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/sys/DatabaseOps.java
```diff
@@ -198,6 +199,20 @@ public static void compact(DatasetGraphSwitchable container) {
             LOG.debug(String.format("Compact %s -> %s\n", db1.getFileName(), db2.getFileName()));
 
             compact(container, loc1, loc2);
+
+            if ( shouldDeleteOld ) {
+                Path loc1Path = IOX.asPath(loc1);
+                LOG.info("Deleting old database after successful compaction (old db path='" + loc1Path + "')...");
+
+                try {
```
afs, line 207: Should use try-with-resources here. (See `Files.walk` `@apiImpl` javadoc.)
This closes the stream which in turn closes the internal `DirectoryStreams`.
```
        try (Stream<Path> walk = Files.walk(Paths.get(...)) ) {
            walk.sorted(Comparator.reverseOrder())
                .map(Path::toFile)
                .forEach(File::delete);
        } catch (IOException ex) {
            IO.exception(ex);
        }
```

rvesse (approved):

afs:
I don't see any test for compact+delete=true/false for Fuseki.

There are exceptions when running the test suite left over from the previosu PR. There are changes on main to fix some of them 
a4346e6493 : `jena-fuseki-main/testing/FusekiEmbedded/tdb2-config.ttl` and 927b137bcd test `compact_01`.
(so please rebase).

`TestConfigFile.serverTDB2` is still throwing an exception. The compact there can simply be removed and a test for compact in Fuseki provided separately, which run compact both with and without delete=true and tests the Fuseki disk state. (All in one test means only one new database created.) 

The compact/delete-true case can be skipped for Windows.

The problem is that the tests have a memory TDB2 database and that does not provide compact. Compact is a disk-based operation.

afs (changes requested):

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/DatabaseMgr.java
```diff
@@ -55,11 +55,14 @@ public static DatasetGraph connectDatasetGraph(String location) {
      * Compact a datasets which must be a switchable TDB database.
      * This is the normal dataset type for on-disk TDB2 databases.
      *
+     * Deletes old database after successful compaction if `shouldDeleteOld` is `true`.
+     *
      * @param container
+     * @param shouldDeleteOld
      */
-    public static void compact(DatasetGraph container) {
+    public static void compact(DatasetGraph container, boolean shouldDeleteOld) {
```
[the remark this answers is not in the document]
bsara: done

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/sys/DatabaseOps.java
bsara: done

bsara:
@afs whoops. seems that I missed that test. I'll add it now.

bsara:
@afs I've added a test for fuseki main

jena-cmds/src/main/java/tdb2/tdbcompact.java
```diff
@@ -18,18 +18,36 @@
...
     static public void main(String... argv) {
         CmdTDB.init() ;
         new tdbcompact(argv).mainRun() ;
     }
 
     protected tdbcompact(String[] argv) {
-        super(argv) ;
+        super(argv);
+
+        if (!Sys.isWindows) {
```
afs, line 41: It is ("should be") fine on Windows.  The compact works ... it is just the effect of delete does not happen until the JVM exits and in this command that's OK.

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/sys/DatabaseOps.java
```diff
@@ -195,6 +207,19 @@ public static void compact(DatasetGraphSwitchable container) {
             LOG.debug(String.format("Compact %s -> %s\n", db1.getFileName(), db2.getFileName()));
 
             compact(container, loc1, loc2);
+
+            if ( !Sys.isWindows && shouldDeleteOld ) {
```
afs, line 211: As above. Don't need to check for windows - it is only tests that are really affects by this.

If you think it is better to alter the user, a log message could be issued.

jena-cmds/src/main/java/tdb2/tdbcompact.java
```diff
@@ -18,18 +18,36 @@
...
+            super.add(argDeleteOld, "--delete", "Delete old database after compaction");
+        }
+    }
+
+    @Override
+    protected void processModulesAndArgs() {
+        super.processModulesAndArgs();
+
+        shouldDeleteOld = ( !Sys.isWindows && contains(argDeleteOld) );
```
afs, line 50: Again, the windows issue.

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionCompact.java
```diff
@@ -88,18 +89,26 @@ private static DatasetGraph unwrap(DatasetGraph dsg, Predicate<DatasetGraph> pre
...
+        static private final Logger log = Fuseki.compactLog;
+
+        private final boolean shouldDeleteOld;
 
         public CompactTask(HttpAction action) {
             super(action);
+
+            String deleteOldParam = action.request.getParameter("deleteOld");
+
+            this.shouldDeleteOld = ( !Sys.isWindows
```
afs, line 101: Windows!

jena-cmds/src/main/java/tdb2/tdbcompact.java
```diff
@@ -18,18 +18,36 @@
...
 package tdb2;
 
+import org.apache.jena.base.Sys;
+import org.apache.jena.cmd.ArgDecl;
 import org.apache.jena.tdb2.store.DatasetGraphSwitchable;
 import org.apache.jena.tdb2.sys.DatabaseOps;
 import tdb2.cmdline.CmdTDB;
 
 public class tdbcompact extends CmdTDB {
+    private static final ArgDecl argDeleteOld = new ArgDecl(ArgDecl.NoValue, "deleteOld");
```
afs, line 28: This calls it `--deleteOld` but the help is `--delete`.

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/sys/DatabaseOps.java
```diff
@@ -195,6 +207,19 @@ public static void compact(DatasetGraphSwitchable container) {
...
+                Path loc1Path = IOX.asPath(loc1);
+                LOG.info("Deleting old database after successful compaction (old db path='" + loc1Path + "')...");
```
afs, line 213: Make this `LOG.debug` because the other message level debug.

afs (changes requested):
I've run it and it worked for me. 

2 comments: 
1. wrong name in command help
1. skipping "windows" isn't needed and is really only for tests. As we leave the DBs around (async issue), there tests don't need to do anything at present.

The only addition that might be nice to check-after-compact but let's get this PR into the codebase after the two review comments are dealt with.

afs:
https://github.com/apache/jena/pull/988#issuecomment-828611307

Still getting exceptions from the background task: TS_FusekiMain
```
[INFO] Running org.apache.jena.fuseki.main.TS_FusekiMain
13:01:21 WARN  Compact    :: [66] **** Exception in compact
org.apache.jena.tdb2.TDBException: Dataset does not support admin operations
	at org.apache.jena.tdb2.sys.DatabaseOps.checkSupportsAdmin(DatabaseOps.java:134) ~[jena-tdb2-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.jena.tdb2.sys.DatabaseOps.compact(DatabaseOps.java:181) ~[jena-tdb2-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.jena.tdb2.DatabaseMgr.compact(DatabaseMgr.java:80) ~[jena-tdb2-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.jena.fuseki.ctl.ActionCompact$CompactTask.run(ActionCompact.java:111) ~[jena-fuseki-core-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.jena.fuseki.async.AsyncPool.lambda$submit$0(AsyncPool.java:66) ~[jena-fuseki-core-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.jena.fuseki.async.AsyncTask.call(AsyncTask.java:100) [jena-fuseki-core-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630) [?:?]
	at java.lang.Thread.run(Thread.java:832) [?:?]
13:01:21 ERROR Server     :: Exception in task 4 execution
org.apache.jena.tdb2.TDBException: Dataset does not support admin operations
	at org.apache.jena.tdb2.sys.DatabaseOps.checkSupportsAdmin(DatabaseOps.java:134) ~[jena-tdb2-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.jena.tdb2.sys.DatabaseOps.compact(DatabaseOps.java:181) ~[jena-tdb2-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.jena.tdb2.DatabaseMgr.compact(DatabaseMgr.java:80) ~[jena-tdb2-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.jena.fuseki.ctl.ActionCompact$CompactTask.run(ActionCompact.java:111) ~[jena-fuseki-core-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.jena.fuseki.async.AsyncPool.lambda$submit$0(AsyncPool.java:66) ~[jena-fuseki-core-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at org.apache.jena.fuseki.async.AsyncTask.call(AsyncTask.java:100) [jena-fuseki-core-4.1.0-SNAPSHOT.jar:4.1.0-SNAPSHOT]
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630) [?:?]
	at java.lang.Thread.run(Thread.java:832) [?:?]
```

afs (changes requested):
I'm getting two exceptions in the build from background compact tasks.

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/sys/DatabaseOps.java
bsara: updates

jena-cmds/src/main/java/tdb2/tdbcompact.java
bsara: whoops. good catch.
fixed.

jena-cmds/src/main/java/tdb2/tdbcompact.java
bsara: fixdd

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/sys/DatabaseOps.java
bsara: fixed

jena-cmds/src/main/java/tdb2/tdbcompact.java
bsara: fixed

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/ctl/ActionCompact.java
bsara: fixed

bsara:
@afs I believe that I've also fixed the exception thrown by the fuseki main tests.

afs (approved):
Looks good!

PR #1000, opened 2021-05-09
Escape HTML chars

afs:


kinow (approved):
Code looks good. Had a quick tour of Fuseki with the new code to check if it's working OK, I could create dataset, upload data, and query it. All the menu items are still working, appears to be harmless.

And congrats on PR/issue number 1000 on Jena's GitHUb @afs :tada: :partying_face:

PR #1001, opened 2021-05-14
JENA-2105: Builder pattern for DataService and Endpoint

afs:
Also "JENA-2104: Check for bnodes for using ()-list syntax" which is a small fix.


jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/server/DataService.java
```diff
@@ -64,22 +63,39 @@
...
         //goActive();
     }
 
-    /** Clone this {@code DataService} except use a different dataset graph */
-    public DataService alter(DatasetGraph dataset) {
-        DataService dSrv2 = new DataService(dataset);
-        this.forEachEndpoint(ep->dSrv2.addEndpoint$(ep));
-        return dSrv2;
-    }
+//    /** Clone this {@code DataService} except use a different dataset graph */
```
kinow, line 93: The commented out code might be useful later? If not we could remove it perhaps?

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
```diff
@@ -417,6 +418,15 @@ public void logServer() {
             OperationRegistry.copyConfig(operationRegistry, this.operationRegistry);
         }
 
+        /**
+         * This name already register in this builder?
```
kinow, line 422: s/register/registered

kinow (approved):

PR #1004, opened 2021-05-17
JENA-2107: Substitute in RDF-star triple pattern

LorenzBuehmann:
JENA-2107: Substitute in RDF-star triple pattern to improve performance


afs:
Small point: could the commit message say something like "JENA-2107: Substitute in RDF-star triple pattern"

LorenzBuehmann:
> Small point: could the commit message say something like "JENA-2107: Substitute in RDF-star triple pattern"

done

PR #1006, opened 2021-05-20
JENA-2109: Permissions v4 main

Claudenw:
Code to differentiate between hard and soft read errors.

Changes to migrate to V4.
SecurityEvaluator change adding isHardReadError() method and changes to test cases to test it.
Change from V3 functionality in that by default isHardReadError() is false so that 
 * when a method returns a list and `read` permission is not granted, an empty list is returned.
 * when a methods return boolean based on presence of triple and `read` permissions are not granted, `false` is returned.

Previously these states would throw a `ReadDeniedException`

Closes JENA-2109


afs (approved):

afs:
@Claudenw : Is there a JIRA for this?

For each version, there is a URL for all tickets labeled as fixed for a release for people who want details of all changes.
For next release: https://s.apache.org/jena-4.1.0-jira (it works per and post release state)

If you want it on that list, could you please add a JIRA and ref the PR  or main commit?

PR #1015, opened 2021-06-16
JENA-2112: ShEx

afs:


afs:
2100 of the new files are the ShEx community test suite.

This is available under the W3C Software License (as is SPARQL).

afs:
Added a fix for JENA-2120 to this PR because it builds on changes made for the ShEx test suite that also address JENA-2118.

PR #1016, opened 2021-06-17
JENA-2119: Restore read config from run/

afs:


Humbedooh:
test message to check linkage with JIRA, ignore me please :)

PR #1043, opened 2021-08-05
JENA-1948: Limited JSON-LD 1.1 Reader

afs:
This PR contains a JSON-LD reader based on Titanium. Treat as experimental.

The linkage is minimal - it parses the JSON-LD 1.1 document, prints quads to bytes and reads the N-Quads bytes.

Better would be direct translation from the Titanium RDF abstraction to Jena - contributions welcome!

To use:
```
   RDFparser.create.forceLang(Lang.JSONLD11)...
```
Unfortunately, the JSON-Processing API, jakarta.json-api, is licensed under EPL-2.0 (even just the API, let alone an implementation) so needs careful handling.

For now, add to the application dependencies:
```
     <dependency>
        <!-- Apache License -->
        <groupId>com.apicatalog</groupId>
        <artifactId>titanium-json-ld</artifactId>
        <version>1.1.0</version>
      </dependency>

      <!-- Eclipse Public License 2.0 -->
      <dependency>
        <groupId>org.glassfish</groupId>
        <artifactId>jakarta.json</artifactId>
        <version>2.0.0</version>
      </dependency>
```
Gradle:
```
compile group: 'com.apicatalog', name: 'titanium-json-ld', version: '1.1.0'
compile group: 'org.glassfish', name: 'jakarta.json', version: '2.0.0'
```

### DIFF
--- a/jena-arq/pom.xml
+++ b/jena-arq/pom.xml
@@ -69,6 +69,27 @@
       <artifactId>jsonld-java</artifactId>
     </dependency>
 
+    <!-- For Titanium JSON-LD 1.1 -->
+    <dependency>
+      <!-- Apache License -->
+      <groupId>com.apicatalog</groupId>
+      <artifactId>titanium-json-ld</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <!-- Eclipse Public License 2.0 (category-B) -->
+    <dependency>
+      <groupId>jakarta.json</groupId>
+      <artifactId>jakarta.json-api</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <!--  And an implementation : Eclipse Public License 2.0 (category-B) -->
+    <dependency>
+      <groupId>org.glassfish</groupId>
+      <artifactId>jakarta.json</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <!-- End Titanium JSON-LD 1.1 -->
+    
     <!-- See parent POM. Our choice of version here. -->
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
--- a/jena-arq/src/main/java/org/apache/jena/riot/lang/LangJSONLD11.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/lang/LangJSONLD11.java
@@ -1,0 +1,125 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.jena.riot.lang;
+
+import java.io.InputStream;
+import java.io.Reader;
+import java.io.StringWriter;
+
+import com.apicatalog.jsonld.JsonLd;
+import com.apicatalog.jsonld.document.Document;
+import com.apicatalog.jsonld.document.JsonDocument;
+import com.apicatalog.rdf.RdfDataset;
+import com.apicatalog.rdf.io.RdfWriter;
+import com.apicatalog.rdf.io.nquad.NQuadsWriter;
+
+import org.apache.jena.atlas.web.ContentType;
+import org.apache.jena.riot.*;
+import org.apache.jena.riot.system.StreamRDF;
+import org.apache.jena.sparql.util.Context;
+
+/**
+ * JSON-LD 1.1 {@link ReaderRIOT}.
+ * Placeholder implementation!
+ */
+public class LangJSONLD11 implements ReaderRIOT {
+    public LangJSONLD11() {}
+
+    @Override
+    public void read(InputStream in, String baseURI, ContentType ct, StreamRDF output, Context context) {
+        Sys.checkForTitanium();
+        try {
+            Document document = JsonDocument.of(in);
+            read(document, output, context);
+        } catch (Exception ex) {
+            throw new RiotException(ex);
+        }
+    }
+
+    @Override
+    public void read(Reader in, String baseURI, ContentType ct, StreamRDF output, Context context) {
+        Sys.checkForTitanium();
+        try {
+            Document document = JsonDocument.of(in);
+            read(document, output, context);
+        } catch (Exception ex) {
+            throw new RiotException(ex);
+        }
+    }
+
+    // Simple - read with Titanium, output N-Quads to a string, parser string.
+    // Minimal linkage!
+    private void read(Document document, StreamRDF output, Context context) throws Exception {
+        // JSON-LD to RDF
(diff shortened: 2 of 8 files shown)


afs:
@acoburn Is there this part of the Titanium API stable for javax, non-jakarta?

acoburn:
I took this branch and made a few minor modifications (to see if it would work with JakartaEE 8). I was able to get it to compile and the initialization steps in LangJSONLD11.Sys worked correctly. This was with Titanium 0.8.6 (the last version to support JakartaEE8) and org.glassfish:jakarta.json version 1.1.6

The challenge is that the Titanium API has changed in subtle ways between 0.8.6 and 1.x. For example, in both versions, the `com.apicatalog.jsonld.JsonLd.toRdf(com.apicatalog.jsonld.document.Document)` method returns a `ToRdfApi` object. But in 0.8.6, that class is located in the `com.apicatalog.jsonld.api.impl` package, and in 1.1.0, that class is in the `com.apicatalog.jsonld.api` package.

So when using this with a JakartaEE 8 platform, I end up with this error: 

```
java.lang.NoSuchMethodError: 'com.apicatalog.jsonld.api.ToRdfApi com.apicatalog.jsonld.JsonLd.toRdf(com.apicatalog.jsonld.document.Document)
```

One approach would be to only support JakartaEE 9 with this code. That might be a fine first step.

The way I was approaching this involved using two separate packages: `jena-jsonld` and `jena-jsonld-jee8` in which the first one used the JakartaEE 9 packages and Titanium version 1.1.0. The second package used the JakartaEE 8 packages and the earlier version of Titanium. Then the ServiceLoader did all the work in terms of bringing in the right version (provided that the user of the library included the correct dependency)

acoburn:
One thing to note is that this code (as currently written) continues to work in JakartaEE 8 environments, provided one doesn't use `Lang.JSONLD11` in the parser.

afs:
I don't understand the point about `Lang.JSONLD11` - that constant definition does not depend on Titanium.

Definition of Lang constants is not ("should not be") tied to implementation.

We can tune the guards so that the provided code only does anything if Titanium 1.1 is present, and not 0.8.6.
That is, test for `com.apicatalog.jsonld.api.impl.JsonLD`.

A separate module can add Titanium 0.8.6 (which can be done with the released Jena - that's what you do currently I presume?)(there is a example for adding a language).

It would be nice to remove the print-parse.

acoburn:
> I don't understand the point about Lang.JSONLD11 - that constant definition does not depend on Titanium.

Indeed, I meant `riot.lang.LangJSONLD11`

> A separate module can add Titanium 0.8.6

Yes, that is what I do currently

afs:
What is a good test class for the looking for Titanium 1.1.0 and not 0.8.6?

acoburn:
How about `com.apicatalog.jsonld.JsonLdVersion`?

That class was renamed from `com.apicatalog.jsonld.lang.Version` in the 0.8.6 -> 1.0.0 transition.

afs:
Fixup changes:

* Test class for Titanium changed to `com.apicatalog.jsonld.JsonLdVersion`
* Always register constant `Lang.JSONLD11` (not in `LangJSONLD11`) so the reader class does not have to.

afs:
If this Quarkus statement about JakartaEE9 is still current:

https://developers.redhat.com/blog/2020/09/11/quarkus-and-jakarta-ee-together-or-not

things are not hopeful.

acoburn:
w.r.t statement about Quarkus and JakartaEE, I think that might be a little dated. Quarkus makes extensive use of MicroProfile (as mentioned in that post), and MicroProfile is very much moving towards using [JakartaEE 9](https://github.com/eclipse/microprofile/issues/223). Once MP makes the jump to JakartaEE 9+, I would suspect that Quarkus will follow suit.

jena-arq/src/main/java/org/apache/jena/riot/lang/LangJSONLD11.java
```diff
@@ -0,0 +1,125 @@
...
+                // This class is in Titanium 1.1.0 (JakartaEE9 - jakarta.json) and not in
+                // 0.8.6 (which uses JakartaEE8 - javax.json) where it is "com.apicatalog.jsonld.lang.Version"g
+                Class.forName("com.apicatalog.jsonld.JsonLdVersion", false, Thread.currentThread().getContextClassLoader());
+                titaniumPresent = true;
+            } catch (ClassNotFoundException ex) {
+                return;
+            }
+            try {
+                Class.forName("jakarta.json.Json", false, Thread.currentThread().getContextClassLoader());
+                jakataJsonPresent = true;
```
acoburn, line 118:
```suggestion
                jakartaJsonPresent = true;
```

jena-arq/src/main/java/org/apache/jena/riot/lang/LangJSONLD11.java
```diff
@@ -0,0 +1,125 @@
...
+            } else {
+                factory = (language, profile) -> new LangJSONLD11();
+            }
+            RDFParserRegistry.registerLangTriples(Lang.JSONLD11, factory);
+        }
+
+        public static void checkForTitanium() {
+            if ( ! titaniumPresent )
+                throw new RiotException("Artifact com.apicatalog:titanium-json-ld not on classpath");
+            if ( ! jakataJsonPresent )
```
acoburn, line 103:
```suggestion
            if ( ! jakartaJsonPresent )
```

jena-arq/src/main/java/org/apache/jena/riot/lang/LangJSONLD11.java
```diff
@@ -0,0 +1,125 @@
...
+        w.write(dataset);
+
+        // Do something with the string.
+        RDFParser.fromString(writer.toString()).forceLang(Lang.NQUADS).parse(output);
+    }
+
+    public static class Sys {
+
+        private static boolean titaniumPresent = false;
+        private static boolean jakataJsonPresent = false;
```
acoburn, line 84:
```suggestion
        private static boolean jakartaJsonPresent = false;
```

jena-arq/pom.xml
```diff
@@ -69,6 +69,21 @@
...
+    <dependency>
+      <!-- Apache License -->
+      <groupId>com.apicatalog</groupId>
+      <artifactId>titanium-json-ld</artifactId>
+      <!-- For Titanium 0.8.6, last for javax.json -->
+      <scope>provided</scope>
+    </dependency>
+    <!-- Eclipse Public License 2.0 (category-B) -->
+    <dependency>
+      <groupId>org.glassfish</groupId>
```
acoburn, line 82: would it be better here to depend on `jakarta.json:jakarta.json-api`? That way Jena depends only on the API rather than on a concrete implementation of the JEE API. A user of Jena will still need to depend on the glassfish impl, but that's already a given. For any test code, the glassfish impl would also be needed.

jena-arq/src/main/java/dev/Run.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.jena.sparql.core.DatasetGraph;
+import org.apache.jena.sparql.core.DatasetGraphFactory;
+import org.apache.jena.sys.JenaSystem;
+
+public class Run {
+
+    public static void main(String[] args) {
+        JenaSystem.init();
+        DatasetGraph dsg = DatasetGraphFactory.createTxnMem();
+        RDFParser.create().forceLang(Lang.JSONLD11).source("/home/afs/tmp/D1.jsonld11").parse(dsg);
```
acoburn, line 33: Is this class expected here? (just noting the local path)

afs: Removed.

jena-arq/src/main/java/org/apache/jena/riot/lang/LangJSONLD11.java
afs: Done

acoburn (approved):

PR #1044, opened 2021-08-06
JENA-2142: Extend DatatypeFormatException

jmkeil:
This PR changes the class `DatatypeFormatException` in the following way:
* add constructors with cause parameter (I have a use case this would be helpful for)
* improve JavaDoc

All changes should be API and binary compatible.


afs (approved):
LGTM

ajs6f (approved):
+1

PR #1046, opened 2021-08-10
JENA-2147: Add DatasetGraph.stream

afs:


jena-arq/src/main/java/org/apache/jena/sparql/core/DatasetGraph.java
```diff
@@ -129,6 +133,22 @@ public default void addAll(DatasetGraph src) {
...
+    public default Stream<Quad> stream(Node g, Node s, Node p, Node o) {
+        return Iter.asStream(find(g, s, p, o));
+    }
+
+    /** Returns a {@link Stream} of {@link Quad Quads} in this dataset.
+     *
+     * @return a stream of quads in this dataset.
+     */
+    public default Stream<Quad> stream() {
+        return stream(Node.ANY, Node.ANY, Node.ANY, Node.ANY);
```
ajs6f, line 149: Would `Iter.asStream(find())` be better in case someone overrides the `default` above?

afs: Maybe - seems like it is one way, as is the other of providing stream/4 specially and all the find() are adapters to streams so making stream() having additional layers.

So just following `Graph` here which defaults like this.

ajs6f: Well, I guess ultimately we want people to actually reimpl the `stream` methods (when they have access to underlying streams or are willing to do the big work to create real streams (not just wrapped iters)). So I'm pretty agnostic, just happened to notice it.

ajs6f (approved):

jena-arq/src/main/java/org/apache/jena/sparql/core/DatasetGraph.java
afs: FWIW I found a patch file where I'd tried this set of changes out and did have `Iter.asStream(find())`.
Then copied from `Graph` for the real thing to pick up "the same layout".

afs: Thanks for noticing it!

PR #1063, opened 2021-09-02
JENA-2154: Custom SERVICE executors

Aklakan:
The implementation introduces a ServiceExecutorRegistry akin to the existing FunctionRegistry.
Motivation and design are outlined at https://issues.apache.org/jira/browse/JENA-2154.
Test cases are provided in the TestCustomServiceExecutor class.

Cheers,
Claus

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/query/ARQ.java
+++ b/jena-arq/src/main/java/org/apache/jena/query/ARQ.java
@@ -33,6 +33,7 @@ import org.apache.jena.sparql.mgt.Explain ;
 import org.apache.jena.sparql.mgt.Explain.InfoLevel ;
 import org.apache.jena.sparql.mgt.SystemInfo ;
 import org.apache.jena.sparql.pfunction.PropertyFunctionRegistry ;
+import org.apache.jena.sparql.service.ServiceExecutorRegistry;
 import org.apache.jena.sparql.util.Context ;
 import org.apache.jena.sparql.util.MappingRegistry ;
 import org.apache.jena.sparql.util.Symbol ;
@@ -611,6 +612,7 @@ public class ARQ
 
             // Initialise the standard library.
             FunctionRegistry.init() ;
+            ServiceExecutorRegistry.init();
             AggregateRegistry.init() ;
             PropertyFunctionRegistry.init() ;
 
--- a/jena-arq/src/main/java/org/apache/jena/sparql/ARQConstants.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/ARQConstants.java
@@ -290,6 +290,10 @@ public class ARQConstants
     public static final Symbol registryFunctions =
         SystemARQ.allocSymbol("registryFunctions") ;
 
+    /** The service executor library registry key */
+    public static final Symbol registryServiceExecutors =
+        SystemARQ.allocSymbol("registryServiceExecutors") ;
+
     /** The function library registry key */
     public static final Symbol registryProcedures =
         SystemARQ.allocSymbol("registryProcedures") ;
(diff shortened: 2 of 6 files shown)


Aklakan:
A couple of remarks which you probably want to comment on:

* The rationale for the supplier indirection is to defer starting the actual execution. This - in principle - allows for checking which of the executor factories could handle the arguments without immediately starting the execution.
```java
class ServiceExecutorFactory {
  Supplier<QueryIterator> createExecutor(OpService op, Binding binding, ExecutionContext execCxt);
}
```
* Probably it is indeed better to move the code from QueryIterService to Service - but add the 'Binding' argument to the function in Service

afs:
This is looking like good timing. While I, personally, think using SERVICE for extensions for stored procedures isn't perfect, it does stay within standard syntax which is valued by users and has some take-up.

Thanks for following existing design patterns (even if they are, with hindsight a bit overly complex in places!).

Starting with design review:

_Deferred start_

Delayed initialization can handled orthogonally. See
`IteratorDelayedInitialization` for example. `Supplier` is coping with an uncommon case
that I think can be done within an implementation anyway.

_Location_

> Probably it is indeed better to move the code from QueryIterService to Service - but add the 'Binding' argument to the function in Service

Possible, though I don't think it make much difference either way.  Leave it in `QueryIterService` for now?

_Default_

I'd like to handle the default case (in the PR, no factory found) as a regular `ServiceExecutorFactory` added from `ARQ.init` rather than code here. It is nice to have a "really no SERVICE support" done by clearing the registry even if we keep the old context approach.

The code will need to handle "no factory found" as `QueryExecException` or a join-identity
query iterator depending on `SILENT`.

_Substituted vs original op_

This isn't so easy to treat as separate.

* A dispatch may be `SERVICE ?service` with `?service` already bound but the
registration is to `<someURI>`.

* This code is in `nextStage` so iteration is happening further out. If the implementation wants bottom-up only, might have to have a different optimizer - i.e. needs to take a wider view. It can still do a single execution and keep that internally (`QueryIterator.close` will be called for clear-up).

I don't think that is common so the API can focus on the usual case because the other can still be done as long as the original is available.

How about providing both in the factory call so the implementation is informed?

With those points:

```java
public interface ServiceExecutorFactory {
    QueryIterator createExecutor(OpService op, OpService original, Binding binding, ExecutionContext context);
}
```

The switch to `ExecutionContext` is a good idea.

_Timing_

We are close to the 4.2.0 release point. As an extension to advertise, we should take the time to get it right. A whole cycle (3-4 months) is a bit long so my suggestion is include in 4.2.0 with the understanding that it is "Experimental - work-in-progress - subject to refinement".

afs:
Suggested changes made to https://github.com/afs/jena/tree/service (including moving tests to jena-integration as per an old comment). The branch is based off the PR commits.

Aklakan:
I agree with most points - but one remark:

The rationale for the deferred start is mainly to allow for showing a list of applicable service executors for a given query - e.g. using a debug function or an admin web interface - without actually starting the execution. Returning the QueryIterator directly does not guarantee that; same iterators may already be live when they are returned - others only initialize on .next / .hasNext. 
As a custom service clause is essentially a specification of some task, createExecutor should essentially take the role of validating the OpService (raise an exception if something is wrong, like illegal arguments) and if all is fine then return the object from which the execution can be started.

So with the supplier I wanted to avoid to introduce yet another interface, but I actually had that in mind:

```java
public interface ServiceExecutorFactory {
    ServiceExecution createExecutor(OpService op, OpService original, Binding binding, ExecutionContext context);
}

public interface ServiceExecution { // Similar to QueryExecution
    QueryIterator exec(); // Similar to execSelect
}
```

afs:
`ServiceExecutor` works for me.

Better than `Supplier` which does not say anything about calling `get()` multiple times (same object? new object each time?) . Jena has `Creator` for that role.

But a specifically named `ServiceExecutor` is good.

afs:
Included in PR #1070 .

PR #1064, opened 2021-09-03
JENA-2155: Add Web Access Control Ontology

ebremer:
https://www.w3.org/wiki/WebAccessControl

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/vocabulary/ACL.java
+++ b/jena-core/src/main/java/org/apache/jena/vocabulary/ACL.java
@@ -17,97 +17,35 @@
  */
 
 package org.apache.jena.vocabulary;
-
-import org.apache.jena.graph.Node ;
+import org.apache.jena.rdf.model.Model;
+import org.apache.jena.rdf.model.ModelFactory;
 import org.apache.jena.rdf.model.Property ;
 import org.apache.jena.rdf.model.Resource ;
-import org.apache.jena.rdf.model.ResourceFactory ;
 
 public class ACL {
-    
 /**
-*    Basic Access Control ontology.
-*    https://www.w3.org/wiki/WebAccessControl
-*/
+    Basic Access Control ontology.
+*/    private static final Model m = ModelFactory.createDefaultModel();
+    public static final String NS = "http://www.w3.org/ns/auth/acl#";
 
-    public static final String uri = "http://www.w3.org/ns/auth/acl#";
-
-    /** returns the URI for this schema
-        @return the URI for this schema
-    */
-    public static String getURI()
-        { return uri; }
-
-    protected static final Resource resource( String local )
-        { return ResourceFactory.createResource( uri + local ); }
-
-    protected static final Property property( String local )
-        { return ResourceFactory.createProperty( uri, local ); }
-
-    public static Resource APPEND = Init.Append();
-    public static Resource WRITE = Init.Write();
-    public static Resource CONTROL = Init.Control();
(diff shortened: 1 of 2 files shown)


afs:
Hi @ebremer,

What's the long term situation here? Is the work in Solid a different namespace? Is it compatible with this (up to 2016) vocabulary?

jena-core/src/main/java/org/apache/jena/vocabulary/ACL.java
```diff
@@ -0,0 +1,113 @@
...
+    public static final Property agentClass = Init.agentClass();
+    public static final Property origin = Init.origin();
+    public static final Property _default = Init._default();
+    public static final Property agent = Init.agent();
+    public static final Property agentGroup = Init.agentGroup();
+    public static final Property accessToClass = Init.accessToClass();
+    public static final Property defaultForNew = Init.defaultForNew();
+    public static final Property owner = Init.owner();
+
+    public static class Init {
```
afs, line 68: This shouldn't need the "Init" 

Use of "Init" is needed in RDF.java uniquely to stop class preliminary initialization problems within jena-core. It arises because Jena internally uses `RDF` constants. Once jena-core initialization (which is the earliest) then everything should be normal.

If you encountered a problem, we need to track it down.

jena-core/src/main/java/org/apache/jena/vocabulary/ACL.java
```diff
@@ -0,0 +1,113 @@
...
+        public static final Node agentClass = Init.agentClass().asNode();
+        public static final Node origin = Init.origin().asNode();
+        public static final Node _default = Init._default().asNode();
+        public static final Node agent = Init.agent().asNode();
+        public static final Node agentGroup = Init.agentGroup().asNode();
+        public static final Node accessToClass = Init.accessToClass().asNode();
+        public static final Node defaultForNew = Init.defaultForNew().asNode();
+        public static final Node owner = Init.owner().asNode();
+    }
+}
```
afs, line 113: Minor: missing final newline.

afs:
Maybe the class name should be `WebAccessControl`.

jena-core/src/main/java/org/apache/jena/vocabulary/ACL.java
ebremer: No problems.  Was following the code for RDF.  I removed the Init function.  Added the final new line.

ebremer:
The name space is the same as what is used in Solid: https://solid.github.io/web-access-control-spec/
The long term goal is to track those newer changes introduced by the Solid project.  I called the class "ACL" as it was short and convenient and followed their prefix abbreviation ACL.

jena-core/src/main/java/org/apache/jena/vocabulary/ACL.java
ebremer: added

afs:
Thanks for the updates.

How about WebACL? It's just "ACL" just feels a bit weak because this is specific to web operations. 
(It should be "WAC" from the wiki ... wonder why that hasn't taken off!)

Vocabularies don't have to be in jena itself - it does not convey any magic to them - but this is a W3C work. If Solid add something, presumably it will go in a different namespace or get community agreement.

ebremer:
WAC would fine with me.  Keeps it short and less ambiguous.  What say you?

afs:
OK

ebremer:
Done.

afs (approved):

PR #1068, opened 2021-09-09
JENA-2159: Add Schema.org vocabulary

ebremer:


jena-core/src/main/java/org/apache/jena/vocabulary/SO.java
```diff
@@ -0,0 +1,2366 @@
...
+
+/**
+ *  Schema.org is a collaborative, community activity with a mission to create, maintain, and promote schemas for structured data on the Internet, on web pages, in email messages, and beyond.
+ *  <p>
+ *	See <a href="https://schema.org/">Schema.org</a>.
+ *  <p>
+ *  <a href="https://schema.org/">Base URI and namepace</a>.
+ */
+    private static final Model m = ModelFactory.createDefaultModel();
+    public static final String NS = "https://schema.org/";
```
afs, line 36: FYI:
schema.org might be "http://" as well. There is no right answer here - it's at the whim of the JSON-LD context. So I asked and got a response that "https" is slightly better. We'll see.

afs (commented):
The file name "SO" is a bit terse :-)

SchemaDO.java
SchemaDotOrg.java

I would expect that it would be static import much of the time so the name length isn't so crticial. Maybe.

jena-core/src/main/java/org/apache/jena/vocabulary/SO.java
ebremer: I changed the name.

afs (approved):
Recorded as [JENA-2159.](https://issues.apache.org/jira/browse/JENA-2159).

PR #1071, opened 2021-09-11
Capitalized class names for WAC vocabulary

afs:
cc @ebremer


acoburn:
Thanks, this is a nice improvement

PR #1073, opened 2021-09-18
Work on quoting for multiline strings and on prefixes.

afs:


jena-base/src/main/java/org/apache/jena/atlas/lib/EscapeStr.java
```diff
@@ -80,7 +80,9 @@ public static void stringEsc(AWriter out, String s, char quoteChar, boolean sing
                 // Multiline string.
                 if ( c == quoteChar ) {
                     quotesInARow++ ;
-                    if ( quotesInARow == 3 ) {
+                    if ( (quotesInARow == 3) || (!singleLineString && (i == len - 1)) ) {
```
bvosburgh-tq, line 83: The check for `!singleLineString` here is not necessary, as it will always be true, given the earlier check a few lines above. This line should probably look like this:
```
                    if ( (quotesInARow == 3) || (i == len - 1) ) {
```

jena-arq/src/main/java/org/apache/jena/riot/out/NodeFormatterTTL_MultiLine.java
```diff
@@ -18,95 +18,120 @@
...
-        super(baseIRI, prefixMap) ;
+        super(baseIRI, prefixMap);
     }
 
     public NodeFormatterTTL_MultiLine(String baseIRI, PrefixMap prefixMap, NodeToLabel nodeToLabel) {
         super(baseIRI, prefixMap, nodeToLabel);
     }
 
-    private Runnable noop = ()->{} ;
+    private Runnable noop = () -> {};
```
bvosburgh-tq, line 52: This `Runnable` can be declared `private static final`.

jena-arq/src/main/java/org/apache/jena/riot/out/NodeFormatterTTL_MultiLine.java
```diff
@@ -18,95 +18,120 @@
...
-    {
-        if ( ! writeAsMultiLine(w, lex) ) {
+    protected void writeLiteralLongForm(AWriter w, String lex, String datatypeURI) {
+        Runnable runnable = () -> {
+            w.print("^^");
+            formatURI(w, datatypeURI);
+        };
+        if ( !writeAsMultiLine(w, lex) ) {
+            writeLexicalSingleLine(w, lex, runnable);
             super.writeLiteralOneLine(w, lex, datatypeURI);
```
bvosburgh-tq, line 101: This call to `super.writeLiteralOneLine(w, lex, datatypeURI);` should be removed; but the point is moot, as the entire method should be removed. See comment above.

jena-arq/src/main/java/org/apache/jena/riot/out/NodeFormatterTTL_MultiLine.java
```diff
@@ -18,95 +18,120 @@
...
         }
-        writeLexicalMultiLine(w, lex, ()->{ w.print('@') ; w.print(langTag) ; }) ;
+        writeLexicalMultiLine(w, lex, runnable);
     }
 
     @Override
-    protected void writeLiteralLongForm(AWriter w, String lex, String datatypeURI)
-    {
-        if ( ! writeAsMultiLine(w, lex) ) {
+    protected void writeLiteralLongForm(AWriter w, String lex, String datatypeURI) {
```
bvosburgh-tq, line 94: This method is no longer necessary. It is called only from `NodeFormatterTTL.formatLitDT(AWriter w, String lex, String datatypeURI)` (i.e. in this class's superclass). But, since this class now overrides `formatLitDT(AWriter w, String lex, String datatypeURI)` and implements it differently, this method  will never be called.

jena-arq/src/main/java/org/apache/jena/riot/out/NodeFormatterTTL_MultiLine.java
```diff
@@ -18,95 +18,120 @@
...
-            indent = iw.getAbsoluteIndent() ;
+            iw = (IndentedWriter)writer;
+            iw.pad();
+            indent = iw.getAbsoluteIndent();
             iw.setAbsoluteIndent(0);
         }
-        escapeProc.writeStrMultiLine(writer, str) ;
+        escapeProc.writeStrMultiLine(writer, str);
 
         if ( action != null )
```
bvosburgh-tq, line 136: This `null` check is unnecessary, as all callers of this `private` method supply a `Runnable`. And that is sorta the point of the above-defined `noop` `Runnable` - it obviates the need for a `null` check. : )

jena-arq/src/main/java/org/apache/jena/riot/out/NodeFormatterTTL_MultiLine.java
afs: Can do - the compiler will see it's not re-assigned and a constant anyway.

jena-arq/src/main/java/org/apache/jena/riot/out/NodeFormatterTTL_MultiLine.java
afs: Overriding `formatLitDT` was wrong. It didn't pick up abbreviated forms. Test added.

afs:
Thanks for the review. I've put some changes in.  

The tests are in `TestNodeFmt`.

PR #1076, opened 2021-09-22
JENA-2167: Protobuf based RDF binary format

afs:
Similar to the Thrift encoding but using Protobuf.

In parsing, it is slight slower than Thrift and both are considerably faster (x3-x4) than N-Triples/Turtle.


acoburn:
Just a note that this is very exciting. I have been using the Thrift binding with Jena for some time, and this option will be nice.

afs:
@acoburn  Interesting to hear!

PR #1078, opened 2021-09-24
JENA-2148 and JENA-2170

afs:


afs:
Adding other small cleanups.

PR #1080, opened 2021-09-30
JENA-2169: Dataset interface should support blank node labels for graphs

jmkeil:
Draft for blank node name support, but not tested yet, as I did not manage to find the tests for `DatasetImpl` to add some.

As I didn't wanted to go the way over a IRI String in `DatasetImpl::listModelNames`, I decided to directly use the `ResourceImpl(Node, ModelCom)` constructor with second parameter `null`. In the end, that should have the same result as `ResourceImpl(String)`. But I was not sure, if it would make sense to add the default graph (or the named graph) as second parameter.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/query/Dataset.java
+++ b/jena-arq/src/main/java/org/apache/jena/query/Dataset.java
@@ -21,6 +21,7 @@ package org.apache.jena.query;
 import java.util.Iterator ;
 
 import org.apache.jena.rdf.model.Model ;
+import org.apache.jena.rdf.model.Resource;
 import org.apache.jena.riot.system.Prefixes;
 import org.apache.jena.shared.Lock ;
 import org.apache.jena.shared.PrefixMapping;
@@ -51,8 +52,14 @@ public interface Dataset extends Transactional
     /** Get a graph by name as a Jena Model */
     public Model getNamedModel(String uri) ;
 
+    /** Get a graph by name as a Jena Model */
+    public Model getNamedModel(Resource uri) ;
+
     /** Does the dataset contain a model with the name supplied? */
     public boolean containsNamedModel(String uri) ;
+
+    /** Does the dataset contain a model with the name supplied? */
+    public boolean containsNamedModel(Resource uri) ;
 
     /**
      * Set a named graph.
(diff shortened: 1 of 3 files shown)


jmkeil:
Having the default graph as referenced `Model` in the returned `Resource`s would allow to add statements about the named graphs using the Resources methods, which feels right. But I don't have an overview about possible side effects, e.g. in case if there is no statement about the resource in the default graph yet.

jena-arq/src/main/java/org/apache/jena/query/Dataset.java
```diff
@@ -51,9 +52,15 @@
     /** Get a graph by name as a Jena Model */
     public Model getNamedModel(String uri) ;
 
+    /** Get a graph by name as a Jena Model */
```
afs, line 55: s/by name/by resource/

jena-arq/src/main/java/org/apache/jena/query/Dataset.java
```diff
@@ -51,9 +52,15 @@
...
+    public Model getNamedModel(Resource uri) ;
+
     /** Does the dataset contain a model with the name supplied? */
     public boolean containsNamedModel(String uri) ;
 
+    /** Does the dataset contain a model with the name supplied? */
```
afs, line 61: s/name/resource/

jena-arq/src/main/java/org/apache/jena/sparql/core/DatasetImpl.java
```diff
@@ -213,11 +248,30 @@ public boolean containsNamedModel(String uri) {
...
 
+    @Override
+    public boolean containsNamedModel(Resource uri) {
+        // Does not touch the cache.
+        checkGraphName(uri) ;
+        return dsg.containsGraph(uri.asNode()) ;
+    }
+
     @Override
     public Iterator<String> listNames() {
```
afs, line 259: NodeUtils.nodesToURIs does not meet its contract to skip non-URIs.

Here is an impl that does.
```
    public static Iterator<String> nodesToURIs(Iterator<Node> iter) {
        ExtendedIterator<Node> eIter = WrappedIterator.create(iter) ;
        Iterator<String> conv = eIter.filterKeep(Node::isURI).mapWith(Node::getURI);
        return conv ;
    }
```

afs (commented):
Some suggested changes - nothing major.

PR #1083, opened 2021-10-05
JENA-2178: Generalize service loading

afs:


rvesse:
Since we're generalising this mechanism a couple of things jump out to me:

- Do we want to force initialisers of the same level to run in a predictable order likely by breaking the comparison tie with a lexical comparison on class name.  That way if folks are using this mechanism and (un)intentionally use the same level as a system module the order of operations remains predictable.  This seems particularly important since this PR increases the range of levels considered to be Jena system so pre-existing initialisers might get bitten by this.
- A bare `sl.forEach()` without any exception handling is going to throw an exception should it encounter an unloadable class likely leading to a) a JVM exit with nasty stack trace or b) silently carrying on in a partially initialised state without any obvious indication to the user depending on how and when initialisation is getting called.  Some explicit error handling and bail out may be warranted.

afs:
It would be a good idea to handle bad initialization and log.

Initializers are supposed permanent extensions, not ad hoc extra functionality. It is not alt-OSGi! So better reporting but because there is no isolation guarantees, the system has to be considered unstable.

The debug helps with the overlap issue.

We could now have a reserved "jena service loader" then a separate "application service loader" if the interface names are different.

afs:
I am confused by talking about `sl.forEach()` - there is only one such call `sl.forEach(this::add)`and it is adding an existing object to a registry. It is not calling the loaded service in any way..

@rvesse -- what had you in mind here?

PR #1086, opened 2021-10-13
JENA-2169: Enable named Models with blank node name

afs:
This includes the main commit for this ticket : PR #1080 from @jmkeil .

This PR addresses the comments and adds tests to `AbstractTestDataset` which is used by all datasets implementation testing.


afs:
OK - I'll close the other one as this includes all the changes of #1080.

PR #1089, opened 2021-10-22
JENA-2181: Placeholder code for zst support

afs:
Not active - Apache Commons Compress uses an external JNI library for decompression. It means having to ship multi-architecture binaries for packaged binaries (Jena download and Fuseki).

Other clean up.


afs:
Closing - this PR isn't going anywhere.

If we are going to do anything, then let's do it comprehensively.

1. Have a registry of addition extension to decompressor.
2. In places that access a filename, consult the registry and add a wrapper.

There can be centralized in `IO`.

PR #1099, opened 2021-10-31
JENA-2173: Pull parsing based on AsyncParser

afs:
Use the async parser to provide a pull-parser.

Deprecate the Piped* classes.


rvesse (approved):
LGTM, will try it out later this week

afs:
Great - my main use case is the Async parser itself for the xloader. This requires all parser events to flow, not just triples or quads. It uses the StreamRDF form so isn't pull-parser.

jena-arq/src/main/java/org/apache/jena/riot/RDFDataMgr.java
```diff
@@ -1061,17 +1059,15 @@ public static void writeQuads(OutputStream out, Iterator<Quad> iterator) {
      * @param lang Language
      * @param baseIRI Base IRI
      * @return Iterator over the triples
+     * @deprecated Use {@link AsyncParser#asyncParseTriples} or for N-Triples, {@link RiotParsers#createIteratorNTriples}
      */
+    @Deprecated
```
rvesse, line 1064: Does this need to be deprecated, since it's updated to point at the new code AND it handles the NTriples special case it seems nice to keep this method.  Maybe just update the Javadoc to note that this potentially spawns a thread and provide a pointer to `AsyncParser`?

jena-arq/src/main/java/org/apache/jena/riot/RDFDataMgr.java
```diff
@@ -1080,18 +1076,14 @@ public static void writeQuads(OutputStream out, Iterator<Quad> iterator) {
      * @param lang Language
      * @param baseIRI Base IRI
      * @return Iterator over the quads
+     * @deprecated Use {@link AsyncParser#asyncParseQuads} or for N-Triples, {@link RiotParsers#createIteratorNQuads}
      */
+    @Deprecated
```
rvesse, line 1081: Same comment as above

jena-arq/src/main/java/org/apache/jena/riot/lang/PipedQuadsStream.java
```diff
@@ -23,13 +23,15 @@
 import org.apache.jena.sparql.core.Quad ;
 
 /**
- * Implementation of a producer class that sends Quads; must be connected to a {@code PipedRDFIterator<Quad>}. 
+ * Implementation of a producer class that sends Quads; must be connected to a {@code PipedRDFIterator<Quad>}.
+ * @deprecated To be removed
```
rvesse, line 27: Provide a pointer to `AsyncParser` as the replacement

jena-arq/src/main/java/org/apache/jena/riot/lang/PipedRDFIterator.java
```diff
@@ -56,7 +56,10 @@
  * @see PipedTriplesStream
  * @see PipedQuadsStream
  * @see PipedTuplesStream
+ *
+ * @deprecated To be removed
  */
+@Deprecated
```
rvesse, line 62: Again point to `AsyncParser` as replacement

jena-arq/src/main/java/org/apache/jena/riot/lang/PipedRDFStream.java
```diff
@@ -24,11 +24,14 @@
...
  * @param <T> Type corresponding to a supported RDF primitive
- * 
+ *
  * @see PipedTriplesStream
  * @see PipedQuadsStream
  * @see PipedTuplesStream
+ *
+ * @deprecated To be removed
  */
+@Deprecated
```
rvesse, line 34: Again point to `AsyncParser`

jena-arq/src/main/java/org/apache/jena/riot/lang/PipedTriplesStream.java
```diff
@@ -23,13 +23,16 @@
 import org.apache.jena.sparql.core.Quad ;
 
 /**
- * Implementation of a producer class that sends Triples; must be connected to a {@code PipedRDFIterator<Triple>}. 
+ * Implementation of a producer class that sends Triples; must be connected to a {@code PipedRDFIterator<Triple>}.
+ *
+ * @deprecated To be removed
```
rvesse, line 28: And again

jena-arq/src/main/java/org/apache/jena/riot/lang/PipedTuplesStream.java
```diff
@@ -25,13 +25,16 @@
 import org.apache.jena.sparql.core.Quad ;
 
 /**
- * Implementation of a producer class that sends @{code Tuple<Node>}; must be connected to a {@code PipedRDFIterator<Tuple<Node>}. 
+ * Implementation of a producer class that sends @{code Tuple<Node>}; must be connected to a {@code PipedRDFIterator<Tuple<Node>}.
+ *
+ * @deprecated To be removed
  */
+@Deprecated
```
rvesse, line 32: Yet again

rvesse (approved):
Some comments about the deprecations but otherwise works great.  Updating old code to use `AsyncParser` instead yields about a 1.7x performance improvement over `PipedRDFIterator`

One thing I don't see with this approach is any ability to get to the prefixes.  The underlying code passes them through the various layers up to the point of the `Iterator<EltStreamRDF> asyncParseIterator()` return value.  The callers of that function are applied a mapping from `EltStreamRDF` to `Triple`/`Quad` that discards anything other than exceptions and the `Triple`/`Quad` elements.

Might be useful to expose some ways of getting that information, whether via some additional interface or by making the `asyncParseIterator()` method public so advanced users who care can wrap that iterator in a way that lets them access that information?

afs:
About prefixes:

Internal class `EltStreamRDF` transfers one parser event (triple or quad or prefix or base or exception) so the call

 The main function `AsyncParser.asyncParse(String filesOrURL, StreamRDF output)` has a `StreamRDF` passes exactly the same as synchonous parsing to `output`.

To be used as a pull parser, using `Iterator<T>` - `asyncParseTriples` etc - and like Piped*, there has to be a specific `T`. As replacement for the existing functionality, `Iterator<Triple>`, there isn't the chance to see prefixes. But `AsyncParser.asyncParse(..., StreamRDF)` is the main operation.

jena-arq/src/main/java/org/apache/jena/riot/lang/PipedTuplesStream.java
afs: All Piped* updated to refer to `AsyncParser`.

jena-arq/src/main/java/org/apache/jena/riot/RDFDataMgr.java
afs: Added javadoc about a thread. 

I'd like the option to remove it; there is a fixed choice built-in. The app may not want NT/NQ done "same thread". Deprecation is the only way to let users/applications know that the async parser is now available. The javadoc does refer to `RiotParsers.createIteratorNQuads`.

rvesse: Agreed

rvesse:
Thanks for the thoughts on prefixes and more detail on how everything is fitting together 👍

PR #1104, opened 2021-11-05
Add jena-examples to modules in parent pom JENA-2195

obermeier:
JIRA number [JENA-2195](https://issues.apache.org/jira/browse/JENA-2195)


afs (approved):
Thank you!

PR #1108, opened 2021-11-10
TBS-4248: Jena 4.2.0-tq-1

bvosburgh-tq:
* Bump the Maven version to 4.2.0-tq-1
* Remove all checks for the Unicode Replacement Character (U+FFFD)
* Remove corresponding tests

### DIFF
--- a/apache-jena-libs/pom.xml
+++ b/apache-jena-libs/pom.xml
@@ -23,12 +23,12 @@
   <artifactId>apache-jena-libs</artifactId>
   <packaging>pom</packaging>
   <name>Apache Jena - Libraries POM</name>
-  <version>4.2.0</version>
+  <version>4.2.0-tq-1</version>
 
   <parent>
     <groupId>org.apache.jena</groupId>
     <artifactId>jena</artifactId>
-    <version>4.2.0</version>
+    <version>4.2.0-tq-1</version>
     <relativePath>..</relativePath>
   </parent>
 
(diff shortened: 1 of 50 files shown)


bvosburgh-tq:
Whoops - wrong repository - sorry about that.

afs:
Why not cherry pick the Jena codebase changes for FFFD handling? Then you'd know if 4.3 will work for you; otherwise you will need to revisit this issue.

Whether warning are reported or not is controlled by the `ErrorHandler` of the `TokenizerText`, or you can turn warnings off in the logger. For example, `NodecSSE`.

The Jena changes also include writer handling (JENA-2186) so that Jena does not write raw FFFD. Warnings only happen when data is first ingested.

(The original files mentioned in JENA-2179 were created by TQ tools in the first place.)

bvosburgh-tq:
Thanks for the pointers. I did look into some of what you mention:

I started with cherry picking "JENA-2179..."[1]; but it still logged warnings when the `TokenizerText.errorHandler` was built by `RDFParserBuilder` (using `ErrorHandlerFactory.getDefaultErrorHandler()`). We could change the factory default error handler, but that is used in a lot of places and is a bit risky at this point.

And elsewhere: although the warnings are ignored, there was no way to configure `NodecSSE` to pass a different `ErrorHandler` to `TokenizerText`. That is not critical to us; just an observation.

So: We are already in beta and I am trying to reduce the number and complexity of code changes. I would also like to sidestep the need to find all the places our code might trigger these warnings and verify the resultant behavior. So I took the most direct route of removing all the U+FFFD checks in `TokenizerText`. Of course, we will revisit all this post-release; worst case, when we move to Jena 4.3+.

Also, I do not understand the change to the writers to "escape" the U+FFFD. That just seems to be a workaround that stops the warnings in subsequent parsing. Which raises the question: Why were these warnings added to the Jena parsers? U+FFFD is a legal Unicode character; so why the warning? It seems like, by the time it is being read, the cat is out of the bag and there is nothing anyone can do to fix the implied problem (improper encoding/decoding). But maybe I am missing the use case(s)?

Anyway, thanks again for the suggestions.

[1] [394598d70ff4373b7fd335e79d100d8081af7fb0](https://github.com/apache/jena/commit/394598d70ff4373b7fd335e79d100d8081af7fb0)

afs:
`NodecSSE` does not print warnings, only errors and fatal.
What other configuration might be used?

[NodecSSE line 136](https://github.com/apache/jena/commit/394598d70ff4373b7fd335e79d100d8081af7fb0#diff-328894f6474e24eac980398f772f09938cbb2190b06a80c94977caf6d9fc8b2cR136)

TDB2 is not affected.

> seems to be a workaround that stops the warnings in subsequent parsing.

Exactly - though it is not a "workaround"; it is intentional. It means that IF
data in Jena has FFFD THEN it is silently round-tripped.

The reason for the change from other users experience working with corrupted
data. The raw FFFD is _not in the original input_, it is generated by the JDK.
The input file was not legal syntax. The FFFD is a translation artifact so that it is a
warning, not a fatal error. \uFFFD means two cases can be distinguished.

The syntax of Turtle allows various bad Unicode string cases (e.g. misused surrogates) and if the input isn't legal UTF-8 then it's not legal because Turtle is defined for UTF-8.  So, like IRIs, the Turtle (etc) grammar isn't the end of the story.

The files you are dealing with were written by TopQuadrant tools in the first place, and stored in TopQuadrant source control.
for Jena.

There aren't many places in TBS where the parser is called.

PR #1109, opened 2021-11-12
Align to JENA-2197 changes

afs:


afs:
In JENA-2197, the whitespace facet fix changed constructor usage to factory method `NodeValueDateTime.create`.

PR #1115, opened 2021-11-22
[JENA-1911]  vue

kinow:
:clapper: take 2 !

@afs finding myself now in a short sabbatical with extra spare time, how about we resuscitate the idea to revamp the UI #781? Doesn't need to be Vue, we can take another direction and do it incrementally, or use React, etc. WDYT?

Bruno

PR #1116, opened 2021-11-22
JENA-2201: Quick fix for validators CSS.

afs:
* Quick fix for JENA-2201.
* FusekiServer.Builder would report an error if the same pathspec is used twice
* Cleanup

Given #1115, putting in a quick fix now while we discuss the bigger picture for UI, seems the way forward.


kinow (commented):
I think it doesn't break anything. But I think in the future it could be useful if we had a variable pointing to the root directory. That variable then could be used in server or client side, to access assets like `$baseUrl/css/fuseki.css`, `$baseUrl/js/fuseki.js`, and so it goes.

afs:
These pages were a quick hack, for local use, that escaped into the wild. There shouldn't be any appearance-related HTML in the validators output!

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
```diff
@@ -411,8 +412,11 @@ else if ( httpPort <= 0 )
         private String                   httpsKeystore          = null;
         private String                   httpsKeystorePasswd    = null;
 
-        // Other servlets to add.
-        private List<Pair<String, HttpServlet>> servlets    = new ArrayList<>();
+        // Other servlets to add. Te pathspec for servlets must be unique.
+        // Order does not matter, the rules of patspec dispatch are "exact match"
```
rvesse, line 416: Typo `patspec` -> `pathspec`

jena-fuseki2/jena-fuseki-webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -114,48 +114,6 @@
...
-    <servlet-name>IRIValidator</servlet-name>
-    <url-pattern>/validate/iri</url-pattern>
-  </servlet-mapping>
-
-  <servlet-mapping>
-    <servlet-name>DataValidator</servlet-name>
-    <url-pattern>/validate/data</url-pattern>
-  </servlet-mapping>
-
   <!-- Server controls-->
```
rvesse, line 117: So the validators are no longer available if using the WAR distribution?

rvesse (approved):

jena-fuseki2/jena-fuseki-webapp/src/main/webapp/WEB-INF/web.xml
afs: Correct. This is misc cleaning up while diving back into the UI. 
The page in the webapp that was going to refer to them was never finished, and isn't linked to the navigation, so the user can not invoke them.

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
afs: Done!

PR #1121, opened 2021-12-01
JENA-2202: Skip if longitude not available

afs:
Basic fix.

### DIFF
--- a/jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/SpatialIndex.java
+++ b/jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/SpatialIndex.java
@@ -457,7 +457,11 @@ public class SpatialIndex {
             Resource feature = resIt.nextResource();
 
             Literal lat = feature.getRequiredProperty(SpatialExtension.GEO_LAT_PROP).getLiteral();
-            Literal lon = feature.getRequiredProperty(SpatialExtension.GEO_LON_PROP).getLiteral();
+            Literal lon = feature.getProperty(SpatialExtension.GEO_LON_PROP).getLiteral();
+            if ( lon == null ) {
+                LOGGER.warn("Geo predicates: latitude found but not longitude. " + feature);
+                continue;
+            }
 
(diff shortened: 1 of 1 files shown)


neumarcx:
or set to zero if null it might be intentional.

kinow (approved):

afs:
> or set to zero if null it might be intentional.

Null means "not present".

See report.

neumarcx:
> > or set to zero if null it might be intentional.
> 
> Null means "not present".
> 
> See report.

reasonable

rvesse:
> or set to zero if null it might be intentional.

Not sure we want [Null Island](https://en.wikipedia.org/wiki/Null_Island) being introduced here

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/SpatialIndex.java
```diff
@@ -457,7 +457,11 @@ public static final Dataset wrapModel(Model model) throws SpatialIndexException
             Resource feature = resIt.nextResource();
 
             Literal lat = feature.getRequiredProperty(SpatialExtension.GEO_LAT_PROP).getLiteral();
-            Literal lon = feature.getRequiredProperty(SpatialExtension.GEO_LON_PROP).getLiteral();
+            Literal lon = feature.getProperty(SpatialExtension.GEO_LON_PROP).getLiteral();
```
rvesse, line 460: I notice there's no `null` check for `lat` either

afs: The loop is over latitude properties:
```
ResIterator resIt = model.listResourcesWithProperty(SpatialExtension.GEO_LAT_PROP);
```

rvesse: Ah yes, so this is guaranteed to be non-null or we wouldn't be in the loop to start with

PR #1125, opened 2021-12-07
JENA-2207: Undo variable rename for scoping

afs:
[JENA-2207](https://issues.apache.org/jira/browse/JENA-2207)


kinow (approved):
Looks OK to me. There's a comment on JIRA from Rob :+1:

rvesse (approved):
LGTM

Agree with the point about needing to rename in the first place to guarantee correctness with other transforms and optimisations happening

PR #1126, opened 2021-12-09
fix documentation link for tdb-xloader documentation in cli-info

OyvindLGjesdal:
Hi, and thanks for the new version!

I'm looking forward to try out the new xloader, and noticed that I was unable to follow the url in the help text.


kinow (approved):
Thank you!

PR #1127, opened 2021-12-09
Xloader (more)

afs:
Reduces the intermediate space needed by 87%.

### DIFF
--- a/apache-jena/bin/tdb2.xloader
+++ b/apache-jena/bin/tdb2.xloader
@@ -16,13 +16,71 @@
 ## See the License for the specific language governing permissions and
 ## limitations under the License.
 
+## External environment variables.
+##   JENA_CP
+##   JENA_HOME
+##   JVM_ARGS
+
+## Programs used:
+## jq
+## sort
+## /usr/bin/gzip
+
+## ======= Check environment
+
+function testForProgram() {
+    local CMD="$1"
+    type -p "$CMD" &> /dev/null
+    if [[ $? -ne 0 ]]; then
+	echo "Command $CMD not found"
+	return 1
+    fi
+}
+
+## Test for "sort --parallel" (though without that, it's going to be slower!
+function testSortParallel() {
+    set +e
+    sort --parallel=3 < /dev/null &>/dev/null
+    if [[ $? -ne 0 ]]; then
+        echo "No --parallel support in sort(1)" 2&>1
+	exit 9
+    fi
+    set -e
+}
+
+JAVA="${JAVA:-java}"
+
+COMPLETE="yes"
+for F in jq sort /usr/bin/gzip java
+do
+    testForProgram "$F"
+    if [[ $? -ne 0 ]] ; then
+	COMPLETE="no"
+    fi
+done
+
+if [[ $COMPLETE -ne "yes" ]] ; then
+    echo "One or more programs missing" 2&>1
+    exit 9
+fi
+
+unset COMPLETE
+
+testSortParallel
+
+## ======== Setup
+
+## Environment variable TMPDIR is ignored.
+## it is often a small-ish area unsuitable for large temp files.
+## Use the --tmpdir flag
+
+TMPDIR=
+
+## Format used in logging with date(1).
 DATE="+%H:%M:%S"
-# JENA_CP=
-# JENA_HOME=
-# TMPDIR=
 
 ## Functions.
-# log "LEVEL" "MESSAGE"
+# Usage: log "LEVEL" "MESSAGE"
 function log() {
     local LEVEL="$1"
     local MSG="$2"
@@ -92,6 +150,8 @@ JVM_ARGS
     Do not set to all available RAM.
     Increasing it does not make the loader faster.
 
+The temporary directory defaults to the datbase directory.
+
 EOF
 }
 
@@ -160,7 +220,6 @@ if [ -L "${JENA_HOME}" ]; then
   #echo "Resolved symbolic links for JENA_HOME to $JENA_HOME"
 fi
 
-
 ## Classpath JENA_CP.
 if [ -z "$JENA_CP" ] ; then
    if [ -z "$JENA_HOME" ]; then
@@ -181,6 +240,9 @@ function exec_java() {
 while [ $# -gt 0 ]
 do
     ARG=$1
+    ## --tmpdir
+    ## --loc|--location
+    ## --help
     case $ARG in
 	-d|--debug)
 	    # Debug Mode
@@ -226,8 +288,7 @@ do
     esac
 done
 	       
-if [[ $# -eq 0 ]]
-then
+if [[ $# -eq 0 ]] ; then
     abort 1 "No files to load" 1>&2
 fi
 
@@ -239,9 +300,6 @@ fi
 
 [[ -z $TMPDIR ]] && TMPDIR=$LOC
 export TMPDIR
-## --tmpdir
-## --loc|--location
-## --help
 
 ## TDB1 / TDB2
 ## @@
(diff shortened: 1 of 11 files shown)


apache-jena/bin/tdb2.xloader
```diff
@@ -269,7 +278,7 @@ case "$SYSTEM" in
 	;;
 esac
 
-## Delete database!
+## Don't mess up an existsing database!
```
rvesse, line 281: Typo - `existsing` -> `existing`

apache-jena/bin/tdb2.xloader
```diff
@@ -278,65 +287,94 @@ fi
...
+QUADS_IDX="${QUADS_IDX:-$QUADS_DFT}"
 
-info
-info "Build OSP"
-T="$(now)"
-index OSP
-let TIME_IDX_OSP=$(($(now)-$T))
+if [ -e "$INFO" ] ; then
+    ## Skip a phase if there are no items to index.
+    TRIPLES="$(jq .triples < $INFO)"
```
rvesse, line 344: Do you want to verify that `jq` is actually available before trying to use it?

While it's an increasingly common and popular tool it often isn't installed by default and in $dayjob that's often bitten us

If it isn't available the subsequent `if` statements are liable to bomb out with Bash errors

jena-cmds/src/main/java/tdb2/xloader/AbstractCmdxLoad.java
```diff
@@ -51,12 +51,17 @@
     protected static ArgDecl argTmpdir     = new ArgDecl(true, "tmpdir", "tmp");
     protected static ArgDecl argIndex      = new ArgDecl(true, "index");
 
-    protected static ArgDecl argSortArgs   = new ArgDecl(true, "sortArgs", "sortargs");
+    // If this is put back, note there are two different sorts - one for the ndoe table and several for the indexes.
```
rvesse, line 54: Typo `ndoe` -> `node`

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/xloader/ProcIndexBuildX.java
```diff
@@ -143,31 +143,49 @@ private static long sort_build_index(Logger LOG, String datafile, DatasetGraph d
...
             List<String> sortCmdBasics = Arrays.asList(
                  "sort",
                     "--temporary-directory="+TMPDIR, "--buffer-size=50%",
                     "--parallel=2", "--unique"
                     //, "--compress-program=/usr/bin/gzip"
-                );
+            );
             List<String> sortCmd = new ArrayList<>(sortCmdBasics);
+            if ( BulkLoaderX.CompressSortIndexFiles )
+                sortCmd.add("--compress-program=/usr/bin/gzip");
```
rvesse, line 160: Should this be configurable in case this is used on some system with a weird installation layout?  Or better still can this just be `gzip`?

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/xloader/ProcIndexBuildX.java
```diff
@@ -143,31 +143,49 @@ private static long sort_build_index(Logger LOG, String datafile, DatasetGraph d
...
+
         try {
             //LOG.info("Step : external sort : "+indexName);
             //if ( sortArgs != null ) {}
+
+
             List<String> sortCmdBasics = Arrays.asList(
                  "sort",
                     "--temporary-directory="+TMPDIR, "--buffer-size=50%",
                     "--parallel=2", "--unique"
```
rvesse, line 155: Default `sort` on OS X doesn't have the `--parallel` option, should that be a configurable feature of the loader?

(I know OS X systems are most likely to have SSDs and not need this loader so much)

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/xloader/ProcNodeTableBuilderX.java
```diff
@@ -99,22 +104,27 @@
         Process proc2;
         try {
             //LOG.info("Step : external sort");
-            String[] sortCmd =
-                { "sort",
+            List<String> sortCmdBasics = Arrays.asList(
+                  "sort",
                     "--temporary-directory="+loaderFiles.TMPDIR, "--buffer-size=50%",
                     "--parallel=2", "--unique",
```
rvesse, line 110: Again `--parallel` not guaranteed to be available in all `sort`'s

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/xloader/ProcNodeTableBuilderX.java
```diff
@@ -99,22 +104,27 @@
...
-            if ( sortArgs != null ) {}
+                    );
+
+            List<String> sortCmd = new ArrayList<>(sortCmdBasics);
 
+            //if ( sortNodeTableArgs != null ) {}
+
+            // See javadoc for CompressSortNodeTableFiles - usually false
+            if ( BulkLoaderX.CompressSortNodeTableFiles )
+                sortCmd.add("--compress-program=/usr/bin/gzip");
```
rvesse, line 121: Again just use `gzip`?

Also this should mean that the commented out line 111 can be deleted

rvesse (commented):
Sounds like a great improvements, some minor suggestions/questions

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/xloader/ProcIndexBuildX.java
afs: Where do do programs live on macOS?

I'm not sure that the fork-exec in sort(1) will respect the PATH. Java ProcessBuilder does - so "sort" is OK.

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/xloader/ProcIndexBuildX.java
afs: There is enough work at the moment just to get it stable and effective on Linux!

While this tool is not too demanding on the machine, I can type and do email while it is running, are we really going to see usage on a Mac Book? It takes many hours to run so even from that POV, it's an offline/server side tool.

I'm inclined to do the variety refinement when it's working properly.  `jq` does need installing so a check in the shell script.

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/xloader/ProcIndexBuildX.java
rvesse: It lives in the same place, just having spent a lot of time in containers often using minimal Linux distro's over the last few years I've become very wary of assuming things will a) exist and b) always be where you expect

So it looks like they use `execlp` - https://github.com/wertarbyte/coreutils/blob/master/src/sort.c#L1084 - and the Linux manual pages say the following:

> # Special semantics for execlp() and execvp()

> The execlp(), execvp(), and execvpe() functions duplicate the actions of the shell in searching for an executable file if the specified filename does not contain a slash (/) character. The file is sought in the colon-separated list of directory pathnames specified in the PATH environment variable. If this variable isn't defined, the path list defaults to the current directory followed by the list of directories returned by confstr(_CS_PATH). (This confstr(3) call typically returns the value "/bin:/usr/bin".)
If the specified filename includes a slash character, then PATH is ignored, and the file at the specified pathname is executed.

So you should be able to pass in `gzip` and have it search the `PATH` just fine

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/xloader/ProcIndexBuildX.java
rvesse: No worries, maybe just something to mention in the documentation that a `sort` with `--parallel` support is assumed

apache-jena/bin/tdb2.xloader
kinow: Or tell users that somehow they must install it beforehand. In containers it's common to have a very limited number of utilities installed.

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/xloader/ProcIndexBuildX.java
kinow: Maybe we can just check the OS and warn if used in Mac OS, and add notes to the docs that this is a Linux tool. I think the `find` command in Mac OS is also different (remember being given a Mac OS and returning after I couldn't find anything :laughing: ).

afs:
I've added testing for programs and for --parallel.

Given the (unrelated!) desirability for 4.3.1, I've turned this into "ready for review" and will assume the comments so far apply. This is self contained - if it goes wrong, nothing else in Jena or Fuseki is impacted.

I ran this version:

1B Wikidata triples , 40kTPS / 6h 54m, database 81G needing an addition 11.6G for workspace for a total of 92G (+ the data).

on a 7 year old machine (quad core, 32G RAM) with the database and tmp files on spinning disk.

PR #1130, opened 2021-12-13
xloader: set LC_ALL=C for external sort command

OyvindLGjesdal:
`LC_ALL=C`  could maybe be added to the beginning of the sortCmd array instead of adding it in as environment var for the process builder?

### DIFF
--- a/jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/xloader/ProcIndexBuildX.java
+++ b/jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/xloader/ProcIndexBuildX.java
@@ -145,6 +145,7 @@ public class ProcIndexBuildX
         InputStream fromSortInputStream;
 
         try {
+            ProcessBuilder pb2;
             //LOG.info("Step : external sort : "+indexName);
             //if ( sortArgs != null ) {}
 
@@ -163,7 +164,9 @@ public class ProcIndexBuildX
                 sortCmd.add(datafile);
             // else this process will decompress and send the data.
 
-            proc2 = new ProcessBuilder(sortCmd).start();
+            pb2 = new ProcessBuilder(sortCmd);
+            pb2.environment().put("LC_ALL","C");
+            proc2 = pb2.start();
 
             // To process. Not used if uncompressed file.
             toSortOutputStream = proc2.getOutputStream();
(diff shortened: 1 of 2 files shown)


rvesse:
Don't believe you can add it to the start of the arguments because that is a Shell feature and a process invoked via `ProcessBuilder` is not being invoked via a Shell (unless you specifically build that)

Note that I believe @afs added this `export` already to the `xloader` scripts that invoke these commands so this PR may be unnecessary

OyvindLGjesdal:
Thank you @rvesse, I see it is set it is set in the xloader-index. I did see some speed improvements on a test-run, so I assumed the locale wasn't set. I'll double check  and see if it is already set and then close the PR.

afs:
The PR looks good to me -- it uses `ProcessBuilder.environment` to set the environment of the subprocess.

`xload-index` isn't part of `tdb2.xloader` - it's the TDB1 xloader (tdbloader2 revised) which I haven't been keeping in step.

With processes being setup and called from Java there is less reason to have the old multi-script setup. There is even a development-support java comman dline `CmdxLoader` which runs TDB2.xloader without shell. I did find that exiting JVMs after heavy use when swapping to a new step in the process helped probably because it reset OS predications (need to revisit sometime).

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/xloader/ProcIndexBuildX.java
```diff
@@ -141,6 +141,7 @@ private static long sort_build_index(Logger LOG, String datafile, DatasetGraph d
             return 0;
         // Sort task.
         Process proc2;
+        ProcessBuilder pb2;
```
afs, line 144: This can be inside the try loop - only proc2 needs to be passed out so the process can be stopped. Once built, the ProcessBuilder isn't needed.

```
ProcessBuilder procBuilder = new ProcessBuilder(sortCmd);
procBuilder.environment().put("LC_ALL","C");
proc2 = procBuilder.start();
```

afs (commented):
Looks good. One comment about code pattern.

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/xloader/ProcIndexBuildX.java
OyvindLGjesdal: Thanks for the comment @afs, moved inside the try block.

afs:
Thanks. (We might hold back merging for a few days so "main" stays in the same state as the 4.3.1 release.)

afs:
I've taken this commit and merged it into my working branch so it does not get lost while not touching `main` any more than necessary at the moment.

Thank you very much.

It will go into the release codebase.

PR #1131, opened 2021-12-14
Update log4j2. Clarify logging included artifacts in shading.

afs:


kinow (approved):

afs:
Thanks - still seems good to wait a day or so before merging to be sure the outside world has quietened down. 😸

jena-fuseki2/jena-fuseki-geosparql/pom.xml
```diff
@@ -98,6 +98,17 @@
...
 
+        <!-- Put in to ensure the dependency isn't optional. -->
+        <dependency>
+          <groupId>org.apache.logging.log4j</groupId>
+          <artifactId>log4j-api</artifactId>
+        </dependency>
+
+        <dependency>
+          <groupId>org.apache.logging.log4j</groupId>
+          <artifactId>log4j-core</artifactId>
```
berezovskyi, line 109: I think `<scope>runtime</scope>` can be added to this dependency and other `-impl` dependencies as logging code should only directly depend on the `api`.

afs: These are for the most-part packaging classes.

jena-fuseki-geosparql is both packing and code for the command line and Fuseki configuration.  Might be better to leave it as scope-compile but jena-fuseki-server should be scope-runtime as it's only shade-packaging.

afs: (pushed and change for scope-setting for jena-fuseki-server)

afs:
Merged because of [CVE-2021-45046](https://nvd.nist.gov/vuln/detail/CVE-2021-45046) which is rated "moderate".

PR #1132, opened 2021-12-14
JENA-2212: Common root DatasetAssembler

afs:
DatasetAssembler becomes an abstract class.

Deprecate `DatasetOne` (a dataset of only a default model), in favour of a wrapped `DatasetGraphOne`.

Rename apart TDB dataset assemblers.


kinow (approved):

afs:
(fixed the JIRA number)

rvesse (approved):

PR #1133, opened 2021-12-16
JENA-2220: Isolate Apache HTTP Client usage in jena-jdbc

afs:
This is part 1 of a 2 change. This PR copies the Apache HttpClient code used by jen-jdbc into that subsystem and puts in a compile-time dependency on Apache HttpClient into jena-jdbc-driver-remote.

The second part removes the legacy classes from jena-arq. See [JENA-2221](https://issues.apache.org/jira/browse/JENA-2221) and #1134.

### DIFF
--- a/jena-jdbc/jena-jdbc-driver-remote/src/main/java/org/apache/jena/jdbc/remote/http/QueryEngineHTTP.java
+++ b/jena-jdbc/jena-jdbc-driver-remote/src/main/java/org/apache/jena/jdbc/remote/http/QueryEngineHTTP.java
@@ -1,0 +1,870 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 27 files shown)


rvesse:
Okay, would it be better to transition these modules over to the new JDK support as well?

rvesse:
@afs Wasn't volunteering you, long time since I looked at this code and wondered if there was anything in particular that meant that it wasn't an option

rvesse (approved):

afs:
I'd be surprised if it couldn't be done - it's not like the protocols have changed.

It is two main/src/ classes, looks routine.  More changes in the tests which do client setup and auth stuff.

jena-jdbc/jena-jdbc-driver-remote/src/main/java/org/apache/jena/jdbc/remote/http/QueryEngineHTTP.java
```diff
@@ -0,0 +1,870 @@
...
+    // This may be null - if we were created form a query string,
+    // we don't guarantee to parse it so we let through non-SPARQL
+    // extensions to the far end.
+    @Override
+    public Query getQuery() {
+        if ( query != null )
+            return query;
+        if ( queryString != null ) {
+            // Object not created with a Query object, may be because there is forgein
+            // syntax in the query or may be because the queystrign was available and the app
```
kinow, line 605: s/queystrign/querystring

kinow (approved):

afs:
Noted. The companion PR is based off this one so change here affects that as well. (AKA needs rebase).

I'm running some of the bots to make sure the first batch of PRs today went in OK.

I'll put this in separately, if I may, in a bunch of "misc" - a collection of small things that don't change functionality that get noticed in passing.

(Or we might even get the jena-jdbc code updated and this AHC code removed by the release ... bit of a "maybe" on that)

PR #1137, opened 2021-12-19
JENA-2225: Handle large datasets size in TDB stats serializer

LorenzBuehmann:
This PR should fix the issue with serializing the size of a dataset in the TDB/TDB2 stats which failed for larger datasets beyond `int` value space.

### DIFF
--- a/jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/stats/Stats.java
+++ b/jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/stats/Stats.java
@@ -31,8 +31,10 @@ import java.util.Map.Entry;
 
 import org.apache.jena.atlas.lib.DateTimeUtils;
 import org.apache.jena.atlas.logging.Log;
+import org.apache.jena.datatypes.xsd.XSDDatatype;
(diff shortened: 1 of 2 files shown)


jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/stats/Stats.java
```diff
@@ -103,7 +105,7 @@ private static Item format(Map<Node, Integer> predicates, Map<Node, Integer> typ
         addPair(meta.getList(), "timestamp", NodeFactoryExtra.nowAsDateTime());
         addPair(meta.getList(), "run@",  DateTimeUtils.nowAsString());
         if ( count >= 0 )
-            addPair(meta.getList(), StatsMatcher.COUNT, NodeFactoryExtra.intToNode((int)count));
```
afs, line 106: The problem is the cast.

`NodeFactoryExtra.intToNode(count)` should work.  `intToNode(long integer)` exists.

They both return an xsd:integer which is a better choice than xsd:long because, if nothing else, it prints nicely.

afs (changes requested):

afs:
Thanks.

A fix is included in #1143.

PR #1142, opened 2021-12-27
JENA-2230: GeoSPARQL assembler

afs:


jena-geosparql/src/main/java/org/apache/jena/geosparql/assembler/GeoAssembler.java
```diff
@@ -0,0 +1,208 @@
...
+import org.apache.jena.sparql.util.graph.GraphUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class GeoAssembler extends DatasetAssembler {
+
+    private static Logger LOG = LoggerFactory.getLogger(GeoAssembler.class);
+
+    /** Helper for datasets that layer on top of other datasets. */
+    // Replace when createBaseDataset is is DatasetAssembler
```
kinow, line 57: Make it a TODO? Also duplicated "is"

jena-geosparql/src/main/java/org/apache/jena/geosparql/assembler/VocabGeoSPARQL.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+package org.apache.jena.geosparql.assembler;
+
+import org.apache.jena.rdf.model.Property;
+import org.apache.jena.rdf.model.Resource;
+import org.apache.jena.rdf.model.ResourceFactory;
+
+public class VocabGeoSPARQL {
+    static String NS = "http://jena.apache.org/geosparql#";
```
kinow, line 26: Should we use https? I think schema.org changed some time ago, don't remember what we use here in Jena.

afs: This is a namespaces/Turtle prefixes, not used as a URL. All our namespaces are http://.

Having some one way and some the other is probably confusing (to me!).

jena-geosparql/src/main/java/org/apache/jena/geosparql/assembler/GeoAssembler.java
afs: Done!

jena-geosparql/src/main/java/org/apache/jena/geosparql/assembler/VocabGeoSPARQL.java
kinow: Agreed!

jena-arq/src/main/java/org/apache/jena/sparql/util/graph/GraphUtils.java
```diff
@@ -132,6 +133,15 @@ public static boolean atmostOneProperty(Resource r, Property p) {
         return true ;
     }
 
+    public static boolean getBooleanValue(Resource r, Property p) {
+        if ( !GraphUtils.atmostOneProperty(r, p) )
```
kinow, line 137: Really nit-picking, and not related to this PR, but shouldn't that method be `atMost`? Like `atLeast`? Could be wrong, holidays minus :coffee: is not normally good for reviews.

kinow (approved):

afs:
Added code to control SIS log messages. SIS uses java.util.logging.

PR #1144, opened 2021-12-28
JENA-2232: Fix javadoc production for java11

afs:


jena-geosparql/pom.xml
```diff
@@ -94,20 +94,6 @@
...
   </dependencies>
 
-  <profiles>
-    <!-- Tests: Java8 only
-         A change in the doubles is resulting in 
-         test failures on precision but the test 
-         is string based of WKT.
-    -->
-    <profile>
-      <id>java-after-8</id>
```
afs, line 104: No-op. Jena now requires java11.

pom.xml
```diff
@@ -258,6 +258,27 @@
...
+        <jdk>11</jdk>
+      </activation>      
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-javadoc-plugin</artifactId>
+            <configuration>
+              <!-- Java 11 only, removed at Java13 -->
+              <additionalOptions>--no-module-directories</additionalOptions>
```
kinow, line 275: Does it break the build if someone uses Java13?

kinow (approved):

pom.xml
afs: Java 13 is EOL. The non-LTS have a lifetime of 6 months. No one should be using it (but ...)

It is not a problem in Java17 (which is LTS).

What does matter is that the flag has been removed and is illegal in Java17 (the javadoc plugin exits with an error).

https://bugs.openjdk.java.net/browse/JDK-8215291 says fixed at 13 (and the flag `--no-module-directories` removed).
The fix has been backported to java 12.

There is an open task to backport to Java11 (LTS) but no progress. https://bugs.openjdk.java.net/browse/JDK-8244171

kinow: >It is not a problem in Java17 (which is LTS).

Ah, perfect. I thought it would break on 13+, not only on 13 :+1: 

Thanks! #TodayILearned :)

PR #1153, opened 2022-01-02
[ui] Upgrade deps and fix small issues

kinow:
There's a tool called `ncu` (npm check updates) that's similar to the Maven plug-in that checks for latest dependencies. You run it as `ncu` to get a picture of available changes. Then `ncu -u` to update the `package.json`, and follow that with `yarn install` to install the dependencies locally.

Doing so, with React, Angular, Vue, etc., there's always the possibility that it will break something, due to some part of the framework using older versions. In the case of Vue, it's really common to see ESLint, and Sass Loader failing. It happened today, so I just reverted those two :+1: 

Vue 2 is stable and maintained. But eventually we will bump to Vue 3, and then will be able to update Sass Loader and ESLint. But **there are development-only dependencies**, so not a big issue :+1: 

Also fixing smaller issues. The only other main issue here is that `ncu` upgraded Bootstrap from 4, to 5. And the icons are not looking quite right. Other than that, it's looking good :-) Should be done this week.

image

-Bruno


jena-fuseki2/jena-fuseki-ui/.eslintrc.js
```diff
@@ -30,7 +30,8 @@ module.exports = {
   rules: {
     'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
     'no-debugger': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
-    'vue/custom-event-name-casing': 'off'
+    'vue/custom-event-name-casing': 'off',
+    'vue/multi-word-component-names': 'off'
```
kinow, line 34: When we update the linter, and the Vue plug-ins for the eslinter, it's normal to see new rules that were incorporated by the plug-in, but that are just really annoying for us :-) in cases like this I simply disable it, unless it's a rule that appears to be useful.

jena-fuseki2/jena-fuseki-ui/package.json
```diff
@@ -13,38 +13,38 @@
...
-    "@triply/yasr": "^4.2.8",
-    "axios": "^0.21.1",
-    "bootstrap": "^4.6.0",
+    "@fortawesome/fontawesome-svg-core": "^1.2.36",
+    "@fortawesome/free-solid-svg-icons": "^5.15.4",
+    "@fortawesome/vue-fontawesome": "^2.0.6",
+    "@triply/yasqe": "^4.2.20",
+    "@triply/yasr": "^4.2.20",
+    "axios": "^0.24.0",
+    "bootstrap": "^5.1.3",
```
kinow, line 22: Sorry, when I finished the PR they were finishing the version 5. I upgraded Skosmos from 3 to 5, but forgot to upgrade the Fuseki PR. Will try to review what changed, and what needs to be fixed (luckily we don't use Bootstrap per se, but Bootstrap Vue, which abstracts the Bootstrap syntax for us)

jena-fuseki2/jena-fuseki-ui/src/styles/index.scss
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-@import 'node_modules/bootstrap/scss/bootstrap';
+@import 'node_modules/bootstrap-vue/node_modules/bootstrap/scss/bootstrap.scss';
```
kinow, line 18: Commit message has a link to the README of Bootstrap Vue, where they changed the recommended way to import it (I got a warning using the old way)

jena-fuseki2/jena-fuseki-ui/src/views/dataset/Query.vue
```diff
@@ -212,8 +212,8 @@ export default {
         this.onYasqeOptionsChange()
         this.syncYasqePrefixes()
         this.loading = false
-      })
-    }, 300)
+      }, 300)
+    })
```
kinow, line 216: :lady_beetle: it worked because JS accepts the extra parameter in the wrong function, and then appears to just use `0` (like a `this.wait()` in Java I think).

kinow:
Fixed the buttons. They were getting the color: black from one of the scss files imported in the wrong order, and twice (it had been imported by an earlier import). Removed it in the last commit, and now it should be ready for review.

@afs I noticed that YASGUI/YASR is not adding borders to the table of results. But can't tell whether it's supposed to be that way, or if we want to have borders?

Bruno

afs:
I see 4.3:
Fuseki-4 3
and 4.4:
Fuseki-4 4

Some kind of border, possibly 1px?, around the results table, not the control above. Maybe it disappears depending on resolution  (4K screen here).

kinow:
@afs huh, funny. I cannot get the YASGUI styles applied correctly. Trying a few things locally...

afs:
It doesn't look wrong or unexpected to me. Are you seeing something different?

Other items:

- The table does not seem to resize when the window size changes - not sure if it is intended to do that - the query box does.

- Might be nice to have "query" as an immediate option on new dataset. Hardly urgent, more a bit of consistency.

- I'm wondering if we should have only one persistence choice for new databases - TDB2. Remove the third choice, changes the second choice to set TDB2.  I can't think of a reason to create TDB1 databases when starting from new and I'd rather people prefer TDB2 over TDB1.

kinow:
>It doesn't look wrong or unexpected to me. Are you seeing something different?

For some weird reason I cannot see borders in the results table. :disappointed_relieved: 

>The table does not seem to resize when the window size changes - not sure if it is intended to do that - the query box does.

Confirmed that. Will take a look, but since it's handled by YASGUI it might be easier to move to a separate issue for later.

>Might be nice to have "query" as an immediate option on new dataset. Hardly urgent, more a bit of consistency.

Ah, you mean when you create a new dataset and click on "create dataset", to have another "query" button? Or is it in the `/#/manage` Manage datasets view, to have a link for "query" there? This should take under one minute to do :+1: will include it here once I understand where it should go.

>I'm wondering if we should have only one persistence choice for new databases - TDB2. Remove the third choice, changes the second choice to set TDB2. I can't think of a reason to create TDB1 databases when starting from new and I'd rather people prefer TDB2 over TDB1.

I also cannot think why people would use TDB1. I was reading some old papers the other day, on inference & reasoners... I **think** there was one that used TDB1. But I never used it.

And now that you mentioned, I always went automatically to the in-memory or TDB2 options. But for a new user, that does not want an in-memory dataset, I think they would choose "Persistent – dataset will persist across Fuseki restart". No reason for choosing TDB2, unless they know what it means.

So +1 to simplify it. If necessary (i.e. if users ask to add it back) maybe we could have an "advanced" section, hidden by default, with the type of persistent dataset, allowing TDB1 or TDB2.

kinow:
@afs maybe you are seeing the old table, with borders.  Did you run `yarn install` after checking out this PR?

Here's `main`:

Screenshot from 2022-01-05 00-21-41

And here's this branch:

Screenshot from 2022-01-05 00-43-44

Newer versions of YASR (part of YASGUI, this being the **r**esult component) have a compact mode, and this change appears to have altered the borders of the table: https://github.com/TriplyDB/Yasgui/commit/19521998f035e718d3f1d5cfa6073ce2e34242e7

I tried setting the default settings of YASR to use `compact=false`, but that didn't seem to change anything. Then I've applied the borders & striped table styles manually. WDYT?

Screenshot from 2022-01-05 00-43-44

kinow:
Last commit of the day: added "query" to the Manage Dataset view @afs. If that was not what you meant, it'll be easy to undo.

image

:sleeping_bed:

afs:
Newbie alert.

I pull the PR branch and ran `mvn clean install`. In jena-fuseki-ui, I get an error

```
[INFO] --- frontend-maven-plugin:1.10.4:yarn (yarn install) @ jena-fuseki-ui ---
[INFO] Running 'yarn install' in /home/afs/ASF/jena-project/jena-fuseki2/jena-fuseki-ui
[INFO] yarn install v0.16.1
[INFO] [1/4] Resolving packages...
[INFO] [2/4] Fetching packages...
[INFO] error eslint-plugin-promise@6.0.0: The engine "node" is incompatible with this module. Expected version "^12.22.0 || ^14.17.0 || >=16.0.0".
[INFO] error Found incompatible module
[INFO] info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```
Is there something I need to do prior running maven?

kinow:
> Newbie alert.
> 
> I pull the PR branch and ran `mvn clean install`. In jena-fuseki-ui, I get an error
> 
> ```
> [INFO] --- frontend-maven-plugin:1.10.4:yarn (yarn install) @ jena-fuseki-ui ---
> [INFO] Running 'yarn install' in /home/afs/ASF/jena-project/jena-fuseki2/jena-fuseki-ui
> [INFO] yarn install v0.16.1
> [INFO] [1/4] Resolving packages...
> [INFO] [2/4] Fetching packages...
> [INFO] error eslint-plugin-promise@6.0.0: The engine "node" is incompatible with this module. Expected version "^12.22.0 || ^14.17.0 || >=16.0.0".
> [INFO] error Found incompatible module
> [INFO] info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
> ```
> 
> Is there something I need to do prior running maven?

Are you using node v10 or another old release? It looks like the min requirement for the eslint version is node 12+.

I'm on my phone, but quite sure mine is v14. https://nodejs.org/en/about/releases/

If you cannot upgrade node we can revert eslint and update it later.

afs:
I have node v16.13.1 installed.

But.

The POM has:

    <properties>
        <node.version>v12.8.0</node.version>
        <yarn.version>v0.16.1</yarn.version>
    </properties>

and the build creates a directory with `node/node` and it is v12.8.0.
It also contains a directory `node/yarn/`.

What are the right values? Do they have to coordinate in anyway?

kinow:
Ah, sorry, yes, running with Maven will use the downloaded node, you are right. I'm trying the `mvn clean install -Pdev` on my laptop now to see what's wrong.

kinow:
D'oh moment to myself. As you pointed, the `pom.xml` itself was getting us node 12.8, which is below the minimum for eslint requirements of 12.22 :man_facepalming: 

I changed that to v16.3.1 (stable LTS), and am running a build locally now to confirm it's building OK.

afs:
Looking good and the build works.

Aside:

I see that uploading data is using GSP, not HTML File Upload. 
That solves the problem of Common FileUpload version - we drop the upload operation!

1/
On startup: the display has "No datasets created" -- it would be nice to have a link to create one.

The old UI had "No datasets - add one"

When going into "Manage datasets" is "existing" - could it be "new dataset" if there are none?

2/ 
No "home" button.

Old UI had the name  "Apache Jena Fuseki" in the top bar

3/
On "Manage", could we have a "info" link per dataset?

4/
Creating a dataset:
Could we have two choices - memory and persistent, where persistent is TDB2.

(Alternatively, have the third choice as TDB1)

5/
Info/stats page.

"Available Service" has one URL per service. It is a random choice of the two available. There are two. Creating say /ds/sparql implicitly creates /ds for query.

In the Statistics it shows both.

This needs changing (the templates need updating to new configuration style but this release isn't the time to do it.

New:
stats-new-page
Old:
stats-old-page

kinow:
@afs on the last item, the two screenshots New & Old; I'll add back the two links (thought it wasn't necessary). Also, I intentionally changed the old layout, but forgot to point it out, sorry.

The old layout had a smaller section at the top, followed by a lot of blank space (on monitor) to the right. And then a table at the bottom, with many columns.

I found it easier to look at that information by filling the spaces in the screen, and then showing the statistics of each endpoint. Did you find the old layout better to visualize the Available Services and Statistics?

I think I should be able to have everything done in the next days and ready for review :+1: 

Thanks for the great feedback!
Bruno

afs:
The new layout for Available Services and Statistics is fine.

Checking the UI puts a whole new light on it!

kinow:
> On startup: the display has "No datasets created" -- it would be nice to have a link to create one.

image

image

kinow:
>Old UI had the name "Apache Jena Fuseki" in the top bar

@afs ah! I know why I removed it. The Home button links to the same view as datasets. The two links, home and datasets point to the same location, one next to the other, which looked strange.

The main reason for removing it was, actually, to show the current view. With an icon I'd leave it greyed out when not the current view. But we would end up with two active menu entries, both the home icon and the datasets link.

Do you think that's really needed, or would users adapt their navigation and recognize "datasets" as home? (really easy to change it, just asking so I'll see what to do about the active link...)

kinow:
>On "Manage", could we have a "info" link per dataset?

Done!

image

kinow:
>    Creating a dataset:
>    Could we have two choices - memory and persistent, where persistent is TDB2.
>
>(Alternatively, have the third choice as TDB1)

Super easy, done! I commented out TDB1, in case we decide to add it back. This way we should be able to take a look at the form with only two options, and think if that looks OK or if we need to have TDB1 too :+1: (or let me know if you'd like me to add it back)

image

kinow:
>    nfo/stats page.
>
>"Available Service" has one URL per service. It is a random choice of the two available. There are two. Creating say /ds/sparql implicitly creates /ds for query.
>
>In the "Statistics", area it shows both.

Aaah! We had two operations returned from the server. I remember not understanding why, now I realize they were not the same and it would make sense to display both. Sorry. Fixed!

image

kinow:
@afs most of the feedback addressed, with one question about the menu navigation.

I separated in commits, so we can revert or modify any of the changes (or if you'd like to take a look at what was required to change each one; I find it easier than with the old Bootstrap JS code, although it does require some understanding of Vue/JS.)

Oh, also found a nasty bug. I was using the array index as `:key` in Vue. Vue has the concept of `data` in a component. The `Menu` component will have a `data` section, which belongs to the component (you can pass variables as `props` that are accessible in `this.prop` or `this.data`, anyhoo).

When you want to display multiple items, you do something like

`<div v-for"item in someList" :key="item.id">`.

But I used the array index, which is a bad practice. In the end, deleting datasets too quickly could accidentally delete the wrong dataset :disappointed_relieved: :open_mouth: 

But it's fixed, and I couldn't find any other instance of that. So one possible bug fixed. I thought it would take longer, but since I finished everything in one day, I might still have time this week to start writing some tests to prevent any other possible bugs (inherited from the old code, or introduced by yours truly :)

Bruno

afs:
"Statistics" - stray "_5". The other "_" are unique keys for endpoints on the dataset URL itself and they don't get shown (correct). But for SPARQL query, it does.

There are 3 SPARQL Query services 

`fuseki-server --verbose` prints the endpoints:
```
21:56:02 INFO  Server          ::   Operation = query   Endpoints = [ "", "query", "sparql" ]
21:56:02 INFO  Server          ::   Operation = update  Endpoints = [ "", "update" ]
21:56:02 INFO  Server          ::   Operation = gsp-rw  Endpoints = [ "", "data" ]
21:56:02 INFO  Server          ::   Operation = gsp-r   Endpoints = [ "", "get" ]
21:56:02 INFO  Server          ::   Operation = upload  Endpoints = [ "", "upload" ]
```

afs:
Re: commits. I think the new UI is all-in-one. No parts, no nothing!

So whatever works for you - whether to squash to one commit or leave or squash some and not others. With the git viewer (Eclipse, or GH tools),it is quite easy to navigate changes on one file even of part of  larger commit.

My personal test is "what would I want to see if looking back a year from now".

kinow:
I honestly have no idea why I added an `if` to check for the SPARQL Query and return the name differently :roll_eyes: I also didn't iterate the endpoints and operations, instead accessing array indexes directly.

Just changed that now to be truly dynamic. I only don't display the endpoint name when it starts with an underscore. Looks like it render the lists correctly?

image

The changes are, on the left hand side, SPARQL Query now has three (or as many as returned) entries instead of only two as before. And on the right hand side, we display the endpoint name now as long as it doesn't start with an underscore.

How does that look @afs?

> My personal test is "what would I want to see if looking back a year from now".

Yeah, good test! I left the commits for your feedback, but I will squash into a single commit once we are OK with the changes here.

But I left a few commits where I'm fixing JS bugs or issues with dependencies. There's one where I added the link to the bug in a dependency. I use commits like this normally when troubleshooting another update, looking at blame/log/bisect/etc :+1: 

Thanks!

afs:
> How does that look @afs?

Looks right!

This is probably the time to remove file upload from the default setup that the UI uses. Not to remove the functionality from the server, just to remove it from setup you get via the UI. We can put it back if need be.

afs:
Could we make the logo top left be a link to the front page? Optionally : with the "Apache Jena Fuseki" text next to it?

Our site (where IMO the logo is a bit small) style:
(this images makes the whole look huge - which it isn't)
website-jena

BTW We have an SVG for the logo so we can make any resolution image we need.

kinow:
Done @afs!

image

Used variables from the Bootstrap SCSS (IDE's like WebStorm provide auto-complete, useful to avoid cluttering CSS files). So now the icon is the same SVG that we have in `jena-site`.

Also added the text "Apache Jena Fuseki" next to it, but customized the style so that it's gray-600 by default, then when you hover your mouse over it, it changes to gray-800 but without the underline. It matches the behavior of other links in the top menu.

I did not customized it when it is the active link, since the `datasets` option will be claiming that, and that could be confusing.

Let me know how that looks now :+1: :grimacing:

afs:
Looks better than good to merge!

I've done the removal HTML file upload from the default settings (JENA-2248).

Then with all the changes are in, we have a complete development to test.

kinow:
Brilliant! Rebased, and squashed commits. Would you like to take a final look & approve @afs? I saw the other PR for removing the HTML file upload :+1: looks good to me. Thanks!

afs:
I had to add a license line as the second line of theSVG file to keep RAT happy.

```
<?xml ... >
<!-- Licensed under the terms of http://www.apache.org/licenses/LICENSE-2.0 -->
```

then the build worked and a quick use of the UI (with cntl-F5 to get the latest!!) was fine.

In the build, there are some deprecation warnings such as:
```
[INFO] More info and automated migrator: https://sass-lang.com/d/slash-div
[INFO] 
[INFO]     ╷
[INFO] 302 │ $headings-margin-bottom:      $spacer / 2 !default;
[INFO]     │                               ^^^^^^^^^^^
[INFO]     ╵
[INFO]     node_modules/bootstrap-vue/node_modules/bootstrap/scss/_variables.scss 302:31  @import
[INFO]     node_modules/bootstrap-vue/node_modules/bootstrap/scss/bootstrap.scss 9:9      @import
[INFO]     src/styles/index.scss 18:9                                                     @import
[INFO]     src/App.vue 41:9                                                               root stylesheet
[INFO] 
[INFO] DEPRECATION WARNING: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.
[INFO] 
```

and also this:

```
[INFO] Browserslist: caniuse-lite is outdated. Please run:
[INFO] npx browserslist@latest --update-db
```
I tried `npx browserslist@latest --update-db` but it failed:
```
node:child_process:903
    throw err;
    ^

Error: Command failed: yarn info caniuse-lite --json
```
(some kind of version thing?)

but I did get a useable `fuseki-server`.

kinow:
>I had to add a license line as the second line of the SVG file to keep RAT happy.

Thanks!

>In the build, there are some deprecation warnings such as:
>[INFO] DEPRECATION WARNING: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.

We can ignore that warning for now, I had it in a previous project too. It's a library that's using a deprecated syntax for division. Sometimes JS moves really really fast. But sometimes you find issues that take a few months/years to be fixed, like this one (first saw it about 1 year ago?).

>and also this:
>I tried npx browserslist@latest --update-db but it failed:

I have that at $work as well :smile: so far most devs ignore it, but I will try to fix it and see if the warning goes away. If I find a solution will send a PR or just share how that was solved :+1: Shouldn't cause any inconvenience to users.

>but I did get a useable fuseki-server.
>These can be fixed, or merely noted because they "just are", after merge.

The final version is the important part. The most important thing to check is if the app works OK, and if there are no browser console errors. There are other things we can optimize, like tree-shaking, remove warnings, improve load time (loading resources asynchronously), etc. But we should be good for now IMO.

:+1: there might be more warnings in the console, or in the browser console. I'll go through these over the next days when writing tests too :+1:

afs (approved):
Good to go!

afs:
OK - so those UI build step outputs are "they happen". With compiled code, warnings from dependencies don't show so I mistakenly assumed they were from Fuseki UI code.

We are good to merge this and do further testing with all the server side changes.

kinow:
Thanks @afs! Next: tests 👍

afs:
FYI: When swapping between branches, I have had to run "yarn" manually.

My guess is that `node_modules` gets out of step when going backwards and forwards in time over different versions of jena-fuseki-ui.

kinow:
Hmmm, good to know. I haven't seen that issue yet, probably because my workflow is a little different.

- switch branches
- cd into fuseki2/fuseki-ui
- `yarn install` and later `yarn run serve`
- Start fuseki server from Eclipse
- Eventually `mvn clean install -Pdev`

But this is a workflow for when a change in the UI is needed, I guess 99% of the time you and other Jena devs will be using the `mvn` command, so will try to find a solution so that that doesn't happen if devs sync branches with UI changes :+1: 

Thanks for the heads up Andy.
Bruno

afs:
Yes, it's unlikely and (I think - fixing destroyed the evidence) only happened because the change of branch went back in time for jena-fuseki-ui but the node_modules was latest.

PR #1167, opened 2022-01-19
JENA-2256: Upgrades

afs:
Of note:

Lucene: 8.10.1 -> 8.11.1
```
        <requireMavenVersion>
          <version>3.2.5</version>
        </requireMavenVersion>
```
This was noted by the maven-version-plugin. It is just making explicit the minimum maven version. Maven version 3.2.5 was released 2017-10-18. Current maven is 3.8.4.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,14 +53,14 @@
    <properties>
     <ver.junit>4.13.2</ver.junit>  
 
-    <ver.slf4j>1.7.32</ver.slf4j>
+    <ver.slf4j>1.7.33</ver.slf4j>
     <ver.log4j2>2.17.1</ver.log4j2>
     <ver.shade-plugin>3.2.4</ver.shade-plugin>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -840,6 +839,9 @@
                   <message>No SNAPSHOT dependencies are allowed!</message>
                   <onlyWhenRelease>true</onlyWhenRelease>
                 </requireReleaseDeps>
+                <requireMavenVersion>
+                  <version>3.2.5</version>
```
rvesse, line 843: If we're gonna enforce this at all might want to enforce 3.8.1 (https://maven.apache.org/docs/3.8.1/release-notes.html) because that has some supply chain CVE related fixes

afs: 3.2.5 was just the number the versions plugins referred to.

That's a good reason to go to 3.8.1 (released 2021-03-30).

Done.

kinow (approved):

PR #1170, opened 2022-01-21
JENA-2258 LiteralLabelImpl: forward Throwabl, not only the message

jmkeil:
`LiteralLabelImpl`: forward `Throwable` instead of only the message to preserve causes.

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/graph/impl/LiteralLabelImpl.java
+++ b/jena-core/src/main/java/org/apache/jena/graph/impl/LiteralLabelImpl.java
@@ -78,10 +78,10 @@ final /*public*/ class LiteralLabelImpl implements LiteralLabel {
 	private boolean wellformed = true;
 	
 	/**
-	 * keeps the message provided by the DatatypeFormatException
-	 * if parsing failed for delayed exception thrown in getValue()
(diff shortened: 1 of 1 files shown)


jena-core/src/main/java/org/apache/jena/graph/impl/LiteralLabelImpl.java
```diff
@@ -81,7 +81,7 @@
 	 * keeps the message provided by the DatatypeFormatException
```
ajs6f, line 81: Perhaps change this comment to match the new code?

jena-core/src/main/java/org/apache/jena/graph/impl/LiteralLabelImpl.java
```diff
@@ -81,7 +81,7 @@
...
 	 * if parsing failed for delayed exception thrown in getValue()
 	 */
-	private String exceptionMsg = null; // Suggested by Andreas Langegger
+	private Throwable exception = null; // Suggested by Andreas Langegger
```
ajs6f, line 84: Not sure the in-line comment should remain after this change?

jmkeil:
Good points. Updated the PR.

ajs6f (approved):

afs (approved):

PR #1174, opened 2022-01-24
JENA-2259: (qtriple ...) and code maintenance in SSE.

afs:


afs:
This area of the code has not been touched in a long while. The code was old and inconsistently formatted so I took the opportunity of reformatting most of the files.

PR #1175, opened 2022-01-25
JENA-2260: Upgrade yarn from 0.16.x to 1.22.x

kinow:
Would be better to go to v2.x (called Berry), but I couldn't figure out how to tell the Maven plug-in to use the 2.x version yet. 1.x is still maintained, so we should be fine.

Also adds `--frozen-lockfile`. It won't update `yarn.lock` anymore, and will fail in case the lock file gets updated/out of sync. Hopefully that should improve the UI build.

Tested with `mvn clean test install -Pdev`, no changes to `yarn.lock`, build passed OK.

### DIFF
--- a/jena-fuseki2/jena-fuseki-ui/pom.xml
+++ b/jena-fuseki2/jena-fuseki-ui/pom.xml
@@ -57,7 +57,7 @@
                             <goal>yarn</goal>
                         </goals>
                         <configuration>
-                            <arguments>install</arguments>
+                            <arguments>install --frozen-lockfile</arguments>
                         </configuration>
                     </execution>
                     <execution>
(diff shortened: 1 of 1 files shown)


kinow:
@afs I had a look at ASF's Jenkins, but the Test Development job appears to be polling the `main` branch periodically. Do we have a job somewhere that will run for this pull request? Or is it triggered only after things are merged? Thanks!

afs:
We don't have a jenkins run to run PR branches. We "merge then test" :-). The Test Development job is running `mvn clean verify` (`test` will happen). We don't have build steps outside maven.

I'll pull the branch:
```
ID=...number...
NEWBRANCH=...local name...
git fetch upstream pull/$ID/head:$NEWBRANCH
git co "$NEWBRANCH"
```

afs:
... ran twice, `yarn.lock` is stable.

afs (approved):

kinow:
Thanks!

PR #1177, opened 2022-01-27
JENA-2262: Stabilize tests: separate TDB2 databases

afs:


kinow (approved):

afs:
The GH actions for Linux is more stable.

kinow:
@afs wouldn't it be useful to change the GH action to run with every PR?

afs:
That has a side-effect - it applies to anyone cloning the repo!

So there are there for manual invocation - any branch. We have Jenkins and (I believe) could setup "on branch" there.

Any contributor can run the build `mvn` on their clone as well. The build is simple - it doesn't require special setup or access to additional files.

PR #1180, opened 2022-01-28
JENA-2262: Skip to the end of response streams

afs:


afs:
This might be the cause of TestAdmin test failures under load.

kinow (approved):
Sounds like a plausible explanation for the test failures. +1

PR #1182, opened 2022-01-30
JENA-2265: Use window.location.pathname for the Fuseki URL, fix Vue p…

kinow:
…ublicPath

### DIFF
--- a/jena-fuseki2/jena-fuseki-ui/src/services/fuseki.service.js
+++ b/jena-fuseki2/jena-fuseki-ui/src/services/fuseki.service.js
@@ -27,19 +27,40 @@ const DATASET_COUNT_GRAPH_QUERY_1 = 'select (count(*) as ?count) {?s ?p ?o}'
 const DATASET_COUNT_GRAPH_QUERY_2 = 'select ?g (count(*) as ?count) {graph ?g {?s ?p ?o}} group by ?g'
 
 class FusekiService {
-  constructor () {
+  /**
+   * @param {Location} location
+   */
+  constructor (location) {
     this.isOffline = true
+    this.pathname = location.pathname
+  }
+
+  /**
+   * Gets the Fuseki URL, only the pathname onward. The protocol, server, port, etc,
+   * are left to the browser/JS engine & Vue to choose. Previously we were passing
+   * strings such as `/#/ping$`. But this did not work when the application was
+   * deployed on a Tomcat server, for example, where the base URL could be something
+   * like `http://localhost:8080/jena-fuseki/#/`.
+   *
+   * See RC1 vote thread for 4.4.0 for more: https://lists.apache.org/thread/z3gb5w95oc7c4v0g1jpk9jkxm0l4b7lh
+   *
+   * @param {string} url - Vue route URL
+   * @return {string} a new route URL that includes any pathname in the URL
+   */
+  getFusekiUrl (url) {
+    // modified version from: https://stackoverflow.com/a/24381515
+    return `${this.pathname}/${url}`.replace(/(\/)\/+/g, '$1')
   }
 
   async getServerData () {
-    const response = await axios.get('/$/server')
+    const response = await axios.get(this.getFusekiUrl('/$/server'))
     return response.data
   }
 
   async getServerStatus () {
     const startTime = new Date().getTime()
     try {
-      await axios.get('/$/ping')
+      await axios.get(this.getFusekiUrl('/$/ping'))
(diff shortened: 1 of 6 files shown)


jena-fuseki2/jena-fuseki-ui/src/services/fuseki.service.js
```diff
@@ -27,19 +27,40 @@ const DATASET_COUNT_GRAPH_QUERY_1 = 'select (count(*) as ?count) {?s ?p ?o}'
...
+   * like `http://localhost:8080/jena-fuseki/#/`.
+   *
+   * See RC1 vote thread for 4.4.0 for more: https://lists.apache.org/thread/z3gb5w95oc7c4v0g1jpk9jkxm0l4b7lh
+   *
+   * @param {string} url - Vue route URL
+   * @return {string} a new route URL that includes any pathname in the URL
+   */
+  getFusekiUrl (url) {
+    // modified version from: https://stackoverflow.com/a/24381515
+    return `${this.pathname}/${url}`.replace(/(\/)\/+/g, '$1')
```
kinow, line 52: We are not using the complete URL, i.e. no chance to have `http://` in the values, as we are only dealing with the `pathname` and the query string passed by Vue.

You can open some URL's in the browser like `localhost`, and `localhost/path`, and `localhost/path/#/` etc, to try, just type `window.location.pathname` in the browser console to see the value.

jena-fuseki2/jena-fuseki-ui/tests/unit/services/fuseki.service.spec.js
```diff
@@ -26,7 +26,9 @@ describe('FusekiService', () => {
   let fusekiService
   let clock
   beforeEach(async () => {
-    fusekiService = new FusekiService()
+    fusekiService = new FusekiService({
+      pathname: ''
+    })
```
kinow, line 31: Instead of accessing `window.location` directly, did some quick inversion of control, to make it easier to test too... later we can have better tests with and without a pathname, by just passing a JS object that we have full control how it behaves (it is hard/annoying to mock global objects like `window.location`)

jena-fuseki2/jena-fuseki-ui/vue.config.js
```diff
@@ -17,6 +17,7 @@
 const path = require('path')
 
 module.exports = {
+  publicPath: '',
```
kinow, line 20: Thanks for fixing this one @afs !

kinow (commented):
This should fix the issue found in the 4.4.0 RC1 vote, without breaking anything... :crossed_fingers:

namedgraph:
If this is a WAR to be deployed in a servlet container, I think you should be using [`ServletContext::getContextPath`](https://javaee.github.io/javaee-spec/javadocs/javax/servlet/ServletContext.html#getContextPath--) and/or [`ServletContext::getRealPath`](https://javaee.github.io/javaee-spec/javadocs/javax/servlet/ServletContext.html#getRealPath-java.lang.String-) for path resolution.

afs:
How? (this in in the browser)

afs (approved):
Tested for WAR file as ROOT.war, as "fuseki.war" and the standalone server.

namedgraph:
@afs inject the context path or base URL into HTML and build URls from it.

afs:
https://cli.vuejs.org/config/#publicpath

What advantage does doing any processing content have? At the moment, it is delivered without filtering from the static assets built by Vue.

It pairs with `getFusekiUrl` which does the callback URL building. The webapp container server may be behind a rewriting reverse proxy.

kinow:
> Tested for WAR file as ROOT.war, as "fuseki.war" and the standalone server.

Thanks Andy!

kinow:
> If this is a WAR to be deployed in a servlet container, I think you should be using [`ServletContext::getContextPath`](https://javaee.github.io/javaee-spec/javadocs/javax/servlet/ServletContext.html#getContextPath--) and/or [`ServletContext::getRealPath`](https://javaee.github.io/javaee-spec/javadocs/javax/servlet/ServletContext.html#getRealPath-java.lang.String-) for path resolution.

That could work too, I think. We would have to define a JSP page var, and load index.jsp instead of index.html.

Then set the var in an HTML element so the Vue browser can read it. I've seen it in another project I worked on, but that was later replaced by using the browser variables.

We can keep that in mind in case users with complex setups (proxies, crazy networks, ...) cannot use Fuseki 👍

Thanks!

PR #1184, opened 2022-02-03
JENA-2272: Introduce DSP. Refactor GSP. Common base StoreProtocol.

afs:


namedgraph:
Doesn't `StoreProtocol` duplicate the logic of `RDFDatasetConnection`, just on a lower level? If that is the intention, maybe their implementations should wrap/call each other?

afs:
`RDFDatasetConnection` is an interface, not an implementation.

`RDFConnection` is an interface implemented by an adapter to `RDFLink`.

`RDFLink` calls the engines `GSP` and `DSP` which are both "one shot" style whereas `RDFLink` is a connection (possibly many operations of different kinds, and various settings for all the operations).

`StoreProtocol` is the common HTTP setup for `GSP` and `DSP`.

PR #1188, opened 2022-02-05
JENA-2271: Do not remove slashes from graph name (with tests)

kinow:
Mea culpa. At least now we are able to add unit tests, so in theory this bug should not happen again (:crossed_fingers: ).

Bruno


afs (approved):
Works fine now!

PR #1200, opened 2022-02-18
JENA-2283: add back the SPARQL endpoint input field in the query page

kinow:


afs:
I tried and update and then a query to `/dataset` which provides both services and it worked.

The update response is HTML because yagsgui is sending a form. The "Accept" header is "text/plain,*/*;q=0.9" so we could respond with plain text.  ==> https://issues.apache.org/jira/browse/JENA-2287

afs (approved):

kinow:
Thanks Andy!

PR #1202, opened 2022-02-20
Explicit choice of versions (ajv, nanoid)

afs:
This PR puts the version numbers of into the "resolutions" section of package.json. This then satisfies the github security dependabot which does not seem to see that the versions actually used are via a different dependency route and are later (and fixed).

Downside: because the PR explicitly puts in the versions based checking node_modules/{ajv,nanoid}/package.json files, we are fixed to these even if the transitive dependencies move on.

An alternative course of action is that we instead simply mark the security alerts as not relevant - this then means we are not limited in the future by the named package.json/"resolutions" settings. Downside: all cloned repos get security alerts.

Which is best?

### DIFF
--- a/jena-fuseki2/jena-fuseki-ui/package.json
+++ b/jena-fuseki2/jena-fuseki-ui/package.json
@@ -56,6 +56,8 @@
   },
   "resolutions": {
     "node-forge": "^1.0.0",
-    "follow-redirects": "^1.14.7"
+    "follow-redirects": "^1.14.7",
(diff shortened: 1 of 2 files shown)


kinow (approved):
I think using resolutions for now should be OK. Before releases I normally check all the dependencies that can be upgraded (`ncu`), and also check if our resolutions need to be updated. So not a major issue IMO.

afs:
My downstream working copy of jena has cleared all the security alerts which is good.

PR #1204, opened 2022-02-24
JENA-2289: fix to setup the precomputed geospatial index properly

LorenzBuehmann:
Fixes a bug where the precomputed geospatial index isn't loaded from file and thus not being registered to the dataset context.

### DIFF
--- a/jena-geosparql/src/main/java/org/apache/jena/geosparql/assembler/GeoAssembler.java
+++ b/jena-geosparql/src/main/java/org/apache/jena/geosparql/assembler/GeoAssembler.java
@@ -176,14 +176,21 @@ public class GeoAssembler extends DatasetAssembler {
             return;
 
         try {
+            // no file given, i.e. in-memory index only
             if ( spatialIndex == null ) {
                 GeoSPARQLConfig.setupSpatialIndex(dataset);
                 return;
             }
 
+            // file given but empty -> compute and serialize index
             Path spatialIndexPath = Path.of(spatialIndex);
-            if ( ! Files.exists(spatialIndexPath) || Files.size(spatialIndexPath) == 0 )
+            if ( ! Files.exists(spatialIndexPath) || Files.size(spatialIndexPath) == 0 ) {
                 GeoSPARQLConfig.setupSpatialIndex(dataset, spatialIndexPath.toFile());
+                return;
+            }
+
+            // load and setup the precomputed index
+            GeoSPARQLConfig.setupPrecomputedSpatialIndex(dataset, spatialIndexPath.toFile());
         }
         catch (SrsException ex) {
             // Data but no spatial data.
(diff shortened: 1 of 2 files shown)


afs:
Thanks - how can this be tested? Run twice, no configuration changes, first time to build the index and second to use the previously built index?

LorenzBuehmann:
well, technically just providing an non-empty precomputed index file + querying would trigger the error.

But yes, reproduction via assembler would basically take two runs, the first would generate and save the spatial index, in the second run it would simply load the index and register it to the dataset context such that when querying the geospatial extension can make use of it

galbiston:
Is an additional function needed?

`GeoSPARQLConfig.setupSpatialIndex(dataset, spatialIndexPath.toFile());` ultimately goes to `SpatialIndex.buildSpatialIndex(Dataset dataset, String srsURI, File spatialIndexFile)`.
`buildSpatialIndex` loads the file and if it results in an empty SpatialIndex will then build a new one based on the `dataset`.

```
public static SpatialIndex buildSpatialIndex(Dataset dataset, String srsURI, File spatialIndexFile) throws SpatialIndexException {

        SpatialIndex spatialIndex = load(spatialIndexFile);

        if (spatialIndex.isEmpty()) {
            Collection<SpatialIndexItem> spatialIndexItems = findSpatialIndexItems(dataset, srsURI);
            save(spatialIndexFile, spatialIndexItems, srsURI);
            spatialIndex = new SpatialIndex(spatialIndexItems, srsURI);
            spatialIndex.build();
        }

        setSpatialIndex(dataset, spatialIndex);
        return spatialIndex;
    }
```

Is the fix not to remove the `if ( ! Files.exists(spatialIndexPath) || Files.size(spatialIndexPath) == 0 )` check?

LorenzBuehmann:
I'm not sure, but `setupSpatialIndex` does always scan for all SRS URIs first, I don't think this is efficient as it will increase the startup time

Aklakan:
Just a quick comment for robustness: You probably want to write the index to a temp file first and only if that completes successfully then copy it to the final file - otherwise if you hit ctrl+c during the build a restart would try to load the partially written file.
(If you are doing it like this anyway then ignore my remark :)

afs:
Mentioned on #1225.

I find it hard to jump from one area of Jena to another - 

Could someone please test this PR and see if it works? Or better, add a test to the test suite?

afs:
Is there consensus here? What is the next step here?

A name that describes exactly what happens looks better to me.

`GeoSPARQLConfig.setupPrecomputedSpatialIndex(dataset, spatialIndexPath.toFile());`

The question is what is its implementation. Is it `setupSpatialIndex(dataset, null, spatialIndexPath.toFile())` or 
```
    SpatialIndex si = SpatialIndex.load(spatialIndexFile);
    SpatialIndex.setSpatialIndex(dataset, si);
```

LorenzBuehmann:
> The question is what is its implementation. Is it `setupSpatialIndex(dataset, null, spatialIndexPath.toFile())`
well, this method would work indeed and is currently unused anyways - what we don't want is to call
`GeoSPARQLConfig.setupSpatialIndex(dataset, spatialIndexPath.toFile());` to avoid scanning for SRS URIs again.

Shall I remove my custom method and just use this method with `null` as SRS argument?

PR #1211, opened 2022-03-01
[JENA-2295] Fix YASGUI's share URL feature, add 404 page

kinow:


kinow:
Preview of the 404 page: 

image

kinow:
Done, I think. The last commit contains the comments & code that fix is. I tried to use some XSS payloads to trigger a popup/warning but couldn't pass through Vue + YASGUI, so I think we should be safe :crossed_fingers: 

To test it:

1. Choose or enter a SPARQL query
2. Click on the Share Button
3. Open that URL in a new tab. This validates the `created` component lifecycle method of the `Query` view component.
4. Now without reloading the page, modify the query in the browser (e.g. make it an invalid query, by replacing `SELECT` by `GET`, for instance). Hit enter, the UI does not reload (Vue will re-draw only the components that need to be updated), but the query must be updated. This validates the `beforeRouteUpdate` route navigation cycle/lifecycle method of the `Query` view component.

Then feel free to do any further tests. This is tricky to cover with a unit test, so I'm adding a note that this needs to be tested via an e2e test once we have those tests in the project :+1:

afs (approved):

PR #1218, opened 2022-03-08
Jena-2302: Streaming implementation of RowSetReaderJSON

Aklakan:
This PR provides a streaming implementation of RowSet for json result sets according to [JENA-2302](https://issues.apache.org/jira/projects/JENA/issues/JENA-2302).
In this initial commit the `RowSetReaderJSON.process` method has been modified to delegate to the new streaming approach.
A cleaner approach would be to use a separate RowSetReaderFactory and is subject to change as this PR progresses.

All existing tests in ARQ are green but there may yet be some uncaught corner cases.
In particular, a test case where 'head' comes after the 'results' key needs yet to be provided.


pom.xml
```diff
@@ -381,6 +381,13 @@
         <version>${ver.dexxcollection}</version>
       </dependency>
       
+      <!-- For RowSetJSONStreaming -->
```
afs, line 384: No need for this. By including it, it is reasonable to use it anywhere now.

jena-arq/src/main/java/org/apache/jena/riot/rowset/RowSetReaderRegistry.java
```diff
@@ -57,7 +57,8 @@ public static void init() {
         initialized = true;
 
         register(RS_XML,        RowSetReaderXML.factory);
-        register(RS_JSON,       RowSetReaderJSON.factory);
```
afs, line 60: Leave in-place and commented out. In time, after a release or two, we can drop that but it servers as a quick-fix marker for now.

jena-arq/src/main/java/org/apache/jena/riot/rowset/RowSetReaderRegistry.java
```diff
@@ -57,7 +57,8 @@ public static void init() {
...
+        // register(RS_JSON,       RowSetReaderJSON.factory);
+        register(RS_JSON,       RowSetReaderJSONStreaming.factory); // Experimental! JENA-2302
```
afs, line 61: Remove "experimental" comment.

jena-arq/src/main/java/org/apache/jena/riot/rowset/rw/JSONResultsKW.java
```diff
@@ -24,37 +24,37 @@
...
-    public static String kLink          = "link" ;
-    public static String kResults       = "results" ;
-    public static String kBindings      = "bindings" ;
-    public static String kType          = "type" ;
-    public static String kUri           = "uri"  ;
-
-    public static String kValue         = "value" ;
-    public static String kLiteral       = "literal" ;
-    public static String kUnbound       = "undef" ;
+    public static final String kHead          = "head" ;
```
afs, line 27: 👍🏻

jena-arq/src/main/java/org/apache/jena/riot/rowset/rw/RowSetReaderJSONStreaming.java
```diff
@@ -0,0 +1,50 @@
...
+        if (!Objects.equals(lang, ResultSetLang.RS_JSON ) )
+            throw new ResultSetException("RowSet for JSON asked for a "+lang);
+        return new RowSetReaderJSONStreaming();
+    };
+
+    @Override
+    public QueryExecResult readAny(InputStream in, Context context) {
+        return process(in, context);
+    }
+
```
afs, line 27: The codebase has one blank line between methods.

afs:
First comments:

#### Headers

Two files need the license header added. The build runs Apache RAT (Release Audit Tool) to check license headers are present and fails the build if they are not.

jena-arq/src/main/java/org/apache/jena/riot/rowset/rw/RowSetJSONStreamingBenchmark.java  
jena-arq/src/main/java/org/apache/jena/riot/rowset/rw/RowSetReaderJSONStreaming.java

after which the build passes. 

The build is `mvn clean install` -- there is a github action if you want to run it that way.

#### Warnings in the codebase

The code has warnings - could you deal with them please:
```
RowSetJSONStreaming.java
Javadoc: BufferedRowSet cannot be resolved to a type

RowSetJSONStreaming.java
The import com.google.gson.stream.MalformedJsonException is never used

RowSetJSONStreamingBenchmark.java
The import java.io.IOException is never used

RowSetJSONStreamingBenchmark.java
The import java.net.MalformedURLException is never used

RowSetReaderJSON.java
The import org.apache.jena.sparql.exec.RowSetBuffered is never used
```
#### History

When it is time to merge this, then the commits will need to be squashed down to one commits, or a few commits if a split makes sense, with "JENA-2302: " at the start of commit message. Just one "JENA-2302: Streaming JSON results" is fine. The details can be here and in JIRA.  

The aim is to make the commits meaningful to someone looking at the history is 1-5 years time. Detailed development commits don't stand the test of time.

Aklakan:
There is now a big update on the validation system of the streaming json system.
For a cleaner integration I have added a little extension to the riot system. Next to the `ErrorHandler` class there is now
* the class `Severity` with constants that match the methods of ErrorHandler
* the class `ErrorHandlerEvent` which captures the parameters of the ErrorHandler methods in one class
* the class `ErrorHandlers` with a `relay` method how te relay a `ErrorHandlerEvent` to an `ErrorHandler`

The `RowSetReaderJSONSteaming` validation system has now several settings with severity levels that relay violation events on that infrastructure.

As for further testing of the default settings which should capture the existing behavior, I still to plan to write a little script that randomly modifies a template json result set and compares the parsing with the non streaming approach.

Testing all possible combinations of validation settings would be an overkill though.

The comments above should have all been adressed.

Aklakan:
When I opened [JENA-2309](https://issues.apache.org/jira/projects/JENA/issues/JENA-2309) I realized that actually I should slightly refactor the method `RowSetJSONStreaming.moveToNext` to build upon an ClosableIterator<RsJsonElt> which would expose the head, binding and boolean elements - akin to AsyncParser.

afs:
Suggestion: prepare something that can go into the code base soon and then, as another PR, improve it.

Otherwise, a complicated thing will not have had time to stabilize for 4.5.0.

afs:
> For a cleaner integration I have added a little extension to the riot system

`ErrorHandler` is an interface. An error handler implementation can send warnings/errors to an event system by being an implementation. Do we need a separate concept "relays" and `ErrorHandlerEvent` for this one result set reader?

With `IGNORE` etc,. it feel like this is adding something more like logging. Error handlers are called only exception circumstances where things are wrong in some way.

Aklakan:
The aforementioned update is done. From my side the architecture is now complete.

> With `IGNORE` etc,. it feel like this is adding something more like logging. Error handlers are called only exception circumstances where things are wrong in some way.

The rationale is that this gives the possibility to decide on a per use-case basis on how severe certain violations should be considered. I think it was some legacy version of virtuoso that returned empty json for empty select results. The default severity in RowSetJSONStreaming is ERROR so such "ErrorEvents" will be relayed to an ErrorHandler's error method that will then raise an exception. With this update you can set the severity to WARNING which will log a warning or IGNORE which won't report anything.

As another example, what to do if there are multiple 'results' elements in the json? Is it an error? Issue a warning? Ignore because 'things on the Web happen' and the parser will just continue to emit bindings?

You have to actively change properties in the context though to get this behavior.

afs:
That can be done with an `ErrorHandler`. The `ErrorHandler` is the policy of what to do with notification from the application code. The resultset reader calls warning if it can recover and error if it can't or the results are not trusted, and fatal for catastrophic failure. 

Downgrading or ignoring "error" is not matter of filtering. "Error" is more significant.

Presumably, Virtuoso was fixed long ago - it hasn't been mentioned on stackoverflow or users@. Additional "maybe" features become technical debt. Unlike your big data processing platform, we don't know who uses what and in what way.

If you wish to propose an enhancement to error handlers then great - it should apply to all places error handler are used and have an agreed design. (It seems to me it is introducing error numbers and logging markers, for example.)

Aklakan:
So what is your suggestion w.r.t the ValidationSettings class?

At present you can configure the severity of violations individually, very much like shacl violations.

```java
    public static class ValidationSettings implements Serializable {
        /**
         * What to do if the JSON is effectively 'empty', i.e. if neither
         * the head nor the results key were present.
         * Unexpected elements are captured by onUnexpectedJsonElement.
         * e.g. returned older version of virtuoso open source
         * Mitigation is to assume an empty set of bindings.
         */
        protected Severity emptyJsonSeverity = Severity.ERROR;

        /** What to do if no head was encountered. We may have already
         * optimistically streamed all the bindings in anticipation of an
         * eventual head. */
        protected Severity missingHeadSeverity = Severity.ERROR;
        ...
}
```

The way it is used is like this: I check for the violation and based on its configured severity I forward some message to one of the ErrorHandler's methods. For example, the missing head violation is only revealed once the whole stream is consumed. However, because processing already took place (unless getResultVars was caled) so you may wish to only warn about that issue and move on. The default severity is ERROR so a message would get relayed to the single configured error handler's error() method which in turn would raise an exception.

```java
    public static void validateCompleted(RowSetJSONStreaming rs, ErrorHandler errorHandler, ValidationSettings settings) {        // Missing head
        if (rs.getKHeadCount() == 0) {
            ErrorHandlers.relay(errorHandler, settings.getMissingHeadSeverity(),
                new ErrorEvent(String.format("Mandory key '%s' not seen", kHead)));
        }
    }
```


 Without the relay mechanism one would either have to 
* specify one error handler per violation, such as `ARQ.context().copy().put(rsJsonSeverityMissingHead, someErrorHandler)`.
* or drop the ability to handle severity of violations and treat violations with a fixed severity. In this case there would only be one error handler with an all-or-nothing policy - accept or reject all violations because they cannot be discriminated.
* The relay method could also be seen as something RowSetJSONStreaming specific and be moved to that class as a private member.

Or what do you have a better suggestion?

Aklakan:
I added a few contrived test cases to test the validation and demonstrate the configuration.

The streaming json implementation causes tests of `org.apache.jena.jdbc.results.AbstractResultSetTests`run from `jena-jdbc-driver-remote` to raise warnings. This is because the test suite expects certain tests to fail, however `QueryExecHTTP` checks connections for consumptions; which obviously no longer occurs due to the streaming nature.
HttpClient's default behavior of consuming remaining data on close is a real pain: If an spo query fails midway then even with the streaming json implementation the http client might needlessly retrieve all remaining data. I once made a [PR to Apache VFS2](https://issues.apache.org/jira/browse/VFS-805#) for exactly issue for apache http client; not sure if the solution is applicable to java's native client but maybe it helps.

```java
if (retainedConnection.read() != -1)
    Log.warn(this, "HTTP response not fully consumed, if HTTP Client is reusing connections (its default behaviour) then it will consume the remaining response data which may take a long time and cause this application to become unresponsive");
```

afs:
The HTTP response must be finished. It is a feature of HTTP.  HTTP/1.1 is not a streaming protocol. HTTP/2 is much better but it must be used in the right way - overlaying HTTP/1.1 does not expose the two-flow of data in HTTP/2.

Sometimes, in HTTP/1.1, the client can close the TCP connection and open a new one. It messes up connection polling. TCP connections are expensive. There are client-wide and server-wide problems that arise. We know, it's happened. Jena's own test suite used to have these issues.

Closing the connection does not cause the client to clear-up at the point of close, nor does the server see a close to clear-up a that moment..  On unloaded machines, it can take 2 minutes; under load the Linux kernel delays reclaim close-waiting connections (it's a GC like task). 

The effect is either end can run out of file descriptors. 

In HTTP/2, the situation is both better and worse. Better because there is access to the chunking and flow of chunks but that is only if HTTP/2 specific code is written. Worse because there is only a single TCP connection per host pair which has multiple outstanding requests in-flight and can't be closed. It can lead to head-of-line blocking.

See javadoc for `BodySubscribers#ofInputStream()`

Jena needs reliability before performance. And it is the corner cases, not the main/normal usage.

afs:
> I started looking into this after all - so please don't (re)review the latest changes yet.

> What is your suggestion?

If this is ready now, I will review and make a proposal.

I'm sorry if discussions seems very cautious but this is a key feature of the Jena. A release is a significant commitment; a big difference to an SaaS platform. Adding APIs is a significant commitment. It much harder to remove or evolve features than add them. 

We have a lot of users, and that includes many who are data experts, not systems developers.

Aklakan:
Note that in the PR, the RowSetJSONStreaming implementation is independent of gson. Well, there are gson-specific static util functions (which might as well be made all private for the time being), but those methods are only used to implement `IteratorRsJSON` 'encoder' (maybe there is a better name for it) - i.e. the mapper of elements from the json reader to domain events.

afs:
GSON is small and has an Apache License 2.0. In terms of adding a new dependency, it is as easy as it gets.

titanium-json-ld uses jakarta.json - that's Eclipse Public License which is [category B](https://www.apache.org/legal/resolved.html#category-b). Odd that it isn't the Eclipse Distribution License which is BSD-ish. jena has already added the necessary text into the LICENSE in the distributions (zips). The shad plugin automatically includes all the licenses used by a combined jars.

Jackson is AL2. It's quite a bit bigger.

afs:
The implementation certainly streams. I did a test of reading 50 million results with Fuseki and client in one JVM.

The CPU was 130% so true parallelism, and the log messages from Fuseki came out at the expected points. 
The heap size was limited to 2Gbytes but the results are much larger (>8G) and there was heap issue.

afs:
> What is your suggestion?

Here is a diff for suggestion of what to do about the more sophisticated error handling. 

[Diff-1218-afs-1.txt](https://github.com/apache/jena/files/8325574/Diff-1218-afs-1.txt)

There is a new package `org.apache.jena.riot.rowset.rw.rs_json` with the additional machinery being introduced for JSON result set reading. This leaves `RowSetJSONStreaming` in the `rowset.rw` package.

It puts the additional error handler classes there, moving them out of the riot/system package. All the PR code is there without changing the system-wide API, which otherwise would, to me, be promising the facilities are available on other readers which is not the case. In Jena, with long version tail and the number of users, it is easier to add things than remove them.

The diff moves some inner classes where they are shared across classes into their own top-level files. There are some scope changes to package/private.

We can continue discussing changing the system error handling while still getting the reader capabilities into the codebase.

afs (commented):
These are incoporated into the diff.

afs:
Error handling discussion:

Error handling is trying to communicate the general consequences of some condition. Often, the application code is a long way away from the reader so details of the format are not understood by the application. In Fuseki, they are logged. Error handling and especially "warning", overlaps with logging. Even this "client" side code would do the same for federated queries.

In the parsers, errors generate exceptions so they are not recoverable; there are possibly occasionally exceptions to this, sometimes recovery is possible but the output should now be considered "wrong".

Unfortunately, this isn't a free choice. A warning is something the code notes but will continue to process, an error is serious enough that the results are suspect at best. Turing error into warning changes the contract.

In RowSetReaderStream, a situation like "Encountered bindings as well as boolean result" is an error. The incoming results JSON is illegal. The fault is at the server end. Any fixup is a decision as to which header to believe and is throwing the other choice away.

All the calls to `ErrorHandlers.relay` are serious errors, except, arguably, finding extra elements in the JSON which the reader doesn't understand. (e.g. JSON "comments"!!)

I tried to refactor to centralize error handling so the the processing code call "error(details)"  but it became too big a change.

I think allowing for fine-grained handling would benefit from an explicit error record of an enum "error kind", it's classification "error"/"warning" and the specific message. ValidationSettings enumerates the possible errors as statics.

The enum can be used to process the error which as looking up into a table of what is currently called Severity. The Severity class is how the caller wants to treat the issue arising. We need something to reflect the code raising the issue.

Related: logging system have "markers" to allow related log messages. Jena does not use markers - but the concept seems useful for a more detail error handling extension.

afs:
@Aklakan I'd like to get this PR into the codebase with the [diff](https://github.com/apache/jena/files/8325574/Diff-1218-afs-1.txt) applied. This does mean not no further changes.  With the diff, the subsystem is self-contained and so from my POV can go into Jena 4.5.0.

I can do that and sort out the commits where some are a record of work in progress e.g. like "Removed debug output". The downside is that it looses much or all of the recognition of the contribution.

What's the best way to proceed? One way is to squash done to one/a few commits on the PR, then I can pick it up, apply the diff for a new PR, then merge that.

Aklakan:
The last couple of weeks were so busy - but I finally managed to get educated in the art of rebasing and squashing and it should be done now :)
Its 3 commits now which should reflect the milestones of this PR.

afs:
Thanks for finding the time. And, yes, squash is an "education".

I've added the patch with the JIRA number in the commit message and all four commits will be a single merge.
(also - renamed the only one to `RowSetReaderJSON_V1` and fixed the javadoc about reverting.) I'm assuming we remove it fairly soon, or at least move it somewhere else.

PR #1253.

Some magic - this PR gets noted as merged as well.

PR #1219, opened 2022-03-10
JENA-2303: Don't allocated an RDFLink twice

afs:


rvesse (approved):
Approved

Seems like a bunch of unrelated changes snuck in here?

afs:
Yes - this has small things noticed while looking at the code - often because of use questions.

PR #1225, opened 2022-03-13
Geosparql transformation options

galbiston:
- Conversion between linear and non-linear units now throws exception due to risk of undisclosed inaccuracy, inconsistency and confusion. This is still compliant with the GeoSPARQL standard as invalid unit URIs for operations is an error. However, now the emphasis is on the user to undertake data preparation steps rather than hidden assumptions and potential inaccuracy being introduced.
- Option to disable transformation of mistmatching geometries. Enabled by default.
- Spatial Index now written to temp file before copying to target location (as per comment in JENA-2289).


kinow (commented):
I think git got confused with the diff for `GeometryWrapper.java`, but the rest of the code looks good. Haven't tested geosparql in Jena yet (before or after this change).

Aklakan:
@kinow Yes, silently using the temp directory would cause issues for us. With my comment I had a pattern like this in mind:

What I meant with my comment is a pattern like this:
```java
    Path dataFile = Path.of(finalTarget);
    Path dataFileTmp = dataFile.resolveSibling(dataFile.getFileName() + ".tmp");
    Callable<?> creationAction = () -> {
            try (OutputStream out = Files.newOutputStream(dataFileTmp)) {
                // Write the dataFileTmp file
                out.flush();
            }
    };
    safeCreate(dataFile, dataFileTmp, creationAction());

    // This method is pretty generic
    public static void safeCreate(Path dataFile, Path dataFileTmp, Callable<?> action) throws IOException {
        if (!Files.exists(dataFile)) {
            logger.info("Preparing " + dataFile);
            Thread deleteAction = new Thread(() -> Files.deleteIfExists(x));
            Runtime.getRuntime().addShutdownHook(deleteTmpFile);
            if (Files.exists(dataFileTmp)) {
                // Delete stale tmpFile or raise exception that it exists}
                // If ProcessFileLock in dboe supported checking if it's owning process is still is alive
                // we could even create a message such as
                // Temp file is already locked by still alive process 1234
            try {
                action.call();
                try {
                    Files.move(dataFileTmp, dataFile, StandardCopyOption.ATOMIC_MOVE);
                } catch (AtomicMoveNotSupportedException e) {
                    Files.move(dataFileTmp, dataFile);
                }
            } finally {
                deleteAction.run();
                Runtime.getRuntime().removeShutdownHook(deleteTmpFile);
            }
        }
```

afs:
`IOX.safeWrite`

By creating the replacement file in the same directory, the "move" to the right place is atomic.

Recovery is "if tempfile, not target file", then move tempfile.

ShutdownHooks aren't perfect - there has to be recovery code as well. JVMs crash-exit or are stopped abruptly. The fact that te recovery code is necessary means that shutdown hooks doesn't have to be added.

Aklakan:
> `IOX.safeWrite`
As a rule of thumb: Whenever you write boilerplate, you just haven't found its place in Jena :)

> ShutdownHooks aren't perfect - there has to be recovery code as well. JVMs crash-exit or are stopped abruptly. The fact that te recovery code is necessary means that shutdown hooks doesn't have to be added.
Yes, the shutdown hook is just there to try to free the temp-file space in case of an exception. Just overwriting the temp file may also cause problems if there is a concurrent write process - but usually in that case a some locking protocol (i.e. a lock file is created some time beforehand) should avoid that case from happening.
Note: IOX.safeWrite is strict about atomic move - if the FileSystem does not support it does not resort to a best-effort copy.

afs:
Yes, `IOX.safeWrite` does not copy - the correct recovery action will depend on the situation. Copy risks breaking in some cases. 

In the situation here, it's a non-cross volume operation of a real file system (`sync` must work for the database!).

afs:
Is this related to, or same as, https://github.com/apache/jena/pull/1204#issuecomment-1062209248?

afs:
Shall we merge this? Then make the use of the temp file safer and not dependent on the system tmp directory.

jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/GeometryWrapper.java
```diff
@@ -1,1248 +1,1240 @@
...
+            DirectPosition2D point = new DirectPosition2D(coord.getX(), coord.getY());
+
+            //Convert to WGS84. Use WGS84 and not CRS84 as assuming WGS8 is more prevalent.
+            CoordinateReferenceSystem wgs84CRS = SRSRegistry.getCRS(SRS_URI.WGS84_CRS);
+            MathTransform transform = MathTransformRegistry.getMathTransform(srsInfo.getCrs(), wgs84CRS);
+
+            DirectPosition wgs84Point = transform.transform(point, null);
+
+            //Find the UTM zone.
+            utmURI = SRSRegistry.findUTMZoneURIFromWGS84(wgs84Point.getOrdinate(0), wgs84Point.getOrdinate(1));
```
kinow, line 456: [`SRSRegistry.findUTMZoneURIFromWGS84`](https://github.com/apache/jena/blob/6e4837ee51f7c1a7bc3dfa18a21abd4c3f94d1bb/jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/registry/SRSRegistry.java#L124) is for `latitude` and `longitude`. But in the opengis [docs](https://docs.geotools.org/latest/javadocs/org/opengis/geometry/DirectPosition.html#getOrdinate-int-) all that I could find is that `getOrdinate` returns the ordinate at the dimension N :disappointed_relieved: 

Just checking if we are sure that `0` is latitude, and `1` is longitude?

jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/GeometryWrapper.java
```diff
@@ -1,1248 +1,1240 @@
...
+            transformedSourceGeometry = this.transform(SRS_URI.WGS84_CRS);
+        }
+
+        GeometryWrapper transformedTargetGeometry = transformedSourceGeometry.checkTransformSRS(targetGeometry);
+
+        //Find the nearest pair of coordinates from each Geometry using Euclidean distance (adjusting for wrap around).
+        //These are then used for Great Circle distance.
+        CoordinatePair coordinatePair = CoordinatePair.findNearestPair(transformedSourceGeometry, transformedTargetGeometry);
+
+        //Check whether the nearest pair are the same, i.e. the overlap or within each other.
```
kinow, line 601: s/the overlap/they overlap

kinow:
> Shall we merge this? Then make the use of the temp file safer and not dependent on the system tmp directory.

+1

galbiston:
> Is this related to, or same as, [#1204 (comment)](https://github.com/apache/jena/pull/1204#issuecomment-1062209248)?

Yes, this is what I was trying to address. Shall I just remove the temp file change?

afs:
As is OK - l'll convert it to a safer write based on [Claus's comment](https://github.com/apache/jena/pull/1225#issuecomment-1066316066) and having it like it is makes it clear what to change.

jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/GeometryWrapper.java
galbiston: In `WGS84` the first coordinate is the latitude and the second is the longitude. I've added a couple of tests using `WGS84` and `CRS84` (which reverses the coordinate order). The coordinates are both for the UK and fall within UTM zone 30N and the tests return this result.

jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/GeometryWrapper.java
galbiston: Fixed.

afs:
@galbiston - do you want to keep the commit history? Or I can squash+merge using the GH UI.

galbiston:
No problem having it squashed.

kinow (approved):

PR #1226, opened 2022-03-14
[JENA-2312] Use the service endpoint in the URL used by the UI to que…

kinow:
…ry the backend

### DIFF
--- a/jena-fuseki2/jena-fuseki-ui/src/views/dataset/Edit.vue
+++ b/jena-fuseki2/jena-fuseki-ui/src/views/dataset/Edit.vue
@@ -118,6 +118,8 @@ import {
 } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
 import { library } from '@fortawesome/fontawesome-svg-core'
+import currentDatasetMixin from '@/mixins/current-dataset'
+import { displayError } from '@/utils'
 
 library.add(faTimes, faCheck)
 
@@ -132,12 +134,9 @@ export default {
     FontAwesomeIcon
   },
 
-  props: {
-    datasetName: {
-      type: String,
-      required: true
-    }
-  },
+  mixins: [
+    currentDatasetMixin
+  ],
 
   data () {
     return {
@@ -218,13 +217,9 @@ export default {
       this.code = ''
       this.selectedGraph = ''
       try {
-        this.graphs = await this.$fusekiService.countGraphsTriples(this.datasetName)
+        this.graphs = await this.$fusekiService.countGraphsTriples(this.datasetName, this.services.query['srv.endpoints'][0])
       } catch (error) {
-        this.$bvToast.toast(`${error}`, {
-          title: 'Error',
-          noAutoHide: true,
-          appendToast: false
-        })
+        displayError(this, error)
       } finally {
         this.loadingGraphs = false
         this.loadingGraph = false
@@ -243,6 +238,7 @@ export default {
         const result = await this.$fusekiService.fetchGraph(this.datasetName, graphName)
         this.code = result.data
       } catch (error) {
+        console.error(error)
         this.$bvToast.toast(`${error}`, {
           title: 'Error',
           noAutoHide: true,
--- a/jena-fuseki2/jena-fuseki-ui/src/views/dataset/Edit.vue
+++ b/jena-fuseki2/jena-fuseki-ui/src/views/dataset/Edit.vue
@@ -24,16 +24,20 @@
           <b-card-header header-tag="nav">
             <Menu :dataset-name="datasetName" />
           </b-card-header>
-          <b-card-body>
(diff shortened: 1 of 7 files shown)


jena-fuseki2/jena-fuseki-webapp/pom.xml
```diff
@@ -57,6 +57,12 @@
       <version>4.5.0-SNAPSHOT</version>
     </dependency>
     
+    <dependency>
+      <groupId>org.apache.jena</groupId>
+      <artifactId>jena-geosparql</artifactId>
+      <version>4.5.0-SNAPSHOT</version>
+    </dependency>
```
kinow, line 64: Ops, I will remove this change. Only added it to test the GeoSparql pull request :+1:

kinow:
I think the `/` endpoint, used as a service for `query` cannot handle uploads? I'm extracting the upload URL from the `gsp-rw` service. Does that sound correct?

And will display a message to the user about the missing service. What others think?

image

kinow:
Sort of fixed. Pending:

- [ ] test other views and components
- [ ] test both cases, where we have multiple services configured for the dataset, and also a dataset with just the query service as in the geosparql example
- [ ] confirm that the `gsp-rw` is necessary to upload data
- [ ] confirm what is necessary for the edit graph
- [ ] check if unit tests are working after this change

Now :sleeping_bed:

kinow:
@afs , are you able to answer these two?

>confirm that the gsp-rw is necessary to upload data

The GeoSPARQL example from the documentation doesn't seem to allow uploading data? Or is it a bug in the UI?

```ttl
<#service> rdf:type fuseki:Service;
    fuseki:name "geo";
    fuseki:endpoint [ fuseki:operation fuseki:query; ] ;
    fuseki:dataset <#geo_ds> .

<#geo_ds> rdf:type geosparql:geosparqlDataset ;
    geosparql:spatialIndexFile     "DB/spatial.index";
    geosparql:dataset <#baseDataset> ;
    .

<#baseDataset> rdf:type tdb2:DatasetTDB2 ;
    tdb2:location "DB/" ;
    .
```

>confirm what is necessary for the edit graph

Similar to the previous question, is there a service that must be enabled so that the UI can edit the graph in the UI?

Thanks!

kinow:
Still not able to edit the data for the GeoSPARQL dataset.

image

I guess we also need the `gsp-rw` service enabled for that? If so I will disable the UI when the selected dataset is missing that option.

The rest of the UI is looking OK, so I think the change here will fix the issue of the service endpoint. Ready for an initial review, pending some questions/checks.

afs:
> confirm that the gsp-rw is necessary to upload data

Yes.

The "gsp-rw" includes the extension for loading quads.

There is another server "upload", which is no longer added in the default setup. "gsp-rw" calls the "upload" functionality when it receives a file without the GSP `?default` and `?graph=` parameters so it is a superset.

kinow:
Perfect, I'll update the Edit view to tell the user when it's not usable due to the missing service. Thanks @afs!

afs (approved):

jena-fuseki2/jena-fuseki-webapp/pom.xml
afs: Shall I merge then remove this? I'm clearing up prior to 4.5 so in the area.

kinow:
Using the GeoSPARQL docs dataset, and a normal TDB2 dataset to test. This PR changes `Upload` and `Edit` only. Here's a preview of these views for the two aforementioned datasets.

image

image

image

image

The views do not display the forms when there are no `gsp-rw` graphs configured. @afs all done now, if that's looking OK it should be ready to be merged. Thanks!

jena-fuseki2/jena-fuseki-webapp/pom.xml
kinow: Reverted the change :+1: But thanks for offering to handle this @afs :bow:

rvesse (approved):

jena-fuseki2/jena-fuseki-ui/src/views/dataset/Edit.vue
```diff
@@ -24,16 +24,20 @@
           <b-card-header header-tag="nav">
             <Menu :dataset-name="datasetName" />
           </b-card-header>
-          <b-card-body>
+          <b-card-body v-if="!this.services['gsp-rw'] || this.services['gsp-rw'].length === 0">
+            <b-alert show variant="warning">No service for Graph Store Protocol configured</b-alert>
```
afs, line 28: Uses may not associate GSP with thier task of adding data. Maybe working like:

```suggestion
            <b-alert show variant="warning">No service for adding data available. The Graph Store Protocol service should be configured to allow adding data.</b-alert>
```

afs (approved):

jena-fuseki2/jena-fuseki-ui/src/views/dataset/Edit.vue
kinow: Will copy the text in Upload.vue too. Thanks!

PR #1229, opened 2022-03-21
GH-1228: Process update flag for file datasets

### LINKED ISSUE #1228: --update ignored with file-provided data
Report: [users@ email](https://lists.apache.org/thread/gt3t8ntz6jf2n3s3rgbkqfm32rp1klvm).

`--file=someFile.ttl --update` still cause a read-only dataset to be configured.

This applies to both Fuseki binaries with a command line interface.

afs:
Resolves #1228.

The `--update` flag is not handled for `--file` datasets, which stil default to read-only.

Updates do not cause the file to be updated, only the in-memory database that the file was read into.

If you want persisted update, please use a TDB database.

### DIFF
--- a/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/server/FusekiInfo.java
+++ b/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/server/FusekiInfo.java
@@ -56,16 +56,15 @@ public class FusekiInfo {
     public static void logServerSetup(Logger log, boolean verbose,
                                       DataAccessPointRegistry dapRegistry,
                                       String datasetPath, String datasetDescription, String serverConfigFile, String staticFiles) {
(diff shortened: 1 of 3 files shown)


kinow (approved):

rvesse (commented):
I assume by setting this flag to `false` trying to use `--file` and `--update` produces a useful error?  Or does it silently ignore the `--update` flag

If it's the latter that's only going to provoke more user confusion

afs:
`--update` is largely legacy.

`--update` does not take a value. It is either present or not.

For `--loc` a and `--mem`, it isn't necessary and hasn't been for a awhile. These have update enabled. If the user wants detailed control, they write a configuration file.

`--file` defaults to "read-only" -- `--file` is "create in-memory dataset and read file into it", same as `ja:data` in a configuration file. i.e. it is publishing the file. With `--update`, update is allowed, but changes are not written back, hence the default is read-only.

PR #1235, opened 2022-03-25
[JENA-2311] query rewrite index does too expensive caching on geo literals

galbiston:
Replaced string concatenation for map key with test against separate string values to address issues highlighted in [JENA-2311](https://issues.apache.org/jira/browse/JENA-2311).

### DIFF
--- a/jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/index/GeometryLiteralIndex.java
+++ b/jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/index/GeometryLiteralIndex.java
@@ -60,17 +60,15 @@ public class GeometryLiteralIndex {
 
         if (INDEX_ACTIVE) {
 
-            if (index.containsKey(geometryLiteral)) {
-                geometryWrapper = index.get(geometryLiteral);
-            } else {
-                if (otherIndex.containsKey(geometryLiteral)) {
-                    geometryWrapper = otherIndex.get(geometryLiteral);
-                } else {
+            geometryWrapper = index.get(geometryLiteral);
+            if (geometryWrapper == null) {
+                geometryWrapper = otherIndex.get(geometryLiteral);
+                if (geometryWrapper == null) {
                     geometryWrapper = geometryDatatype.read(geometryLiteral);
                 }
                 index.put(geometryLiteral, geometryWrapper);
             }
-
+         
             return geometryWrapper;
         }
 
--- a/jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/index/QueryRewriteIndex.java
+++ b/jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/index/QueryRewriteIndex.java
@@ -20,19 +20,17 @@ package org.apache.jena.geosparql.implementation.index;
 import io.github.galbiston.expiring_map.ExpiringMap;
 import static io.github.galbiston.expiring_map.MapDefaultValues.MAP_EXPIRY_INTERVAL;
 import static io.github.galbiston.expiring_map.MapDefaultValues.UNLIMITED_MAP;
-import java.util.Map.Entry;
-import java.util.Objects;
 import org.apache.jena.geosparql.configuration.GeoSPARQLConfig;
 import org.apache.jena.geosparql.geo.topological.GenericPropertyFunction;
+import org.apache.jena.graph.Graph;
 import org.apache.jena.graph.Node;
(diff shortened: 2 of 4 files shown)


galbiston:
Commit history includes PR [#1234](https://github.com/apache/jena/pull/1234) because I'm tired and had forgotten I hadn't based it off `main`. Maybe just easiest to wait for that to go into `main` and then review this.

afs:
> easiest to wait for that to go into main and then review this.

I quite understand - too easy to do.

galbiston:
Cheers. It needs unit tests adding as well really. A bit of a gap in coverage.

LorenzBuehmann:
Hi, just want to give some feedback about the PR.

To keep it short, good job @galbiston - we merged the PR in our local Apache Jena Fuseki 4.6.0-SNAPSHOT deployment which provides a geospatial index on millions of geospatial geometries. Works as expected.

jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/index/GeometryLiteralIndex.java
```diff
@@ -60,17 +60,15 @@ private static GeometryWrapper retrieveMemoryIndex(String geometryLiteral, Geome
...
-                    geometryWrapper = otherIndex.get(geometryLiteral);
-                } else {
+            geometryWrapper = index.get(geometryLiteral);
+            if (geometryWrapper == null) {
+                geometryWrapper = otherIndex.get(geometryLiteral);
+                if (geometryWrapper == null) {
                     geometryWrapper = geometryDatatype.read(geometryLiteral);
                 }
                 index.put(geometryLiteral, geometryWrapper);
             }
```
LorenzBuehmann, line 70: a matter of taste, but in Java 11 we could also write
```
geometryWrapper = index.putIfAbsent(geometryLiteral, otherIndex.getOrDefault(geometryLiteral, geometryDatatype.read(geometryLiteral)));
```

LorenzBuehmann, line 70: a matter of taste, but in Java 11 we could also write
``` java
geometryWrapper = index.putIfAbsent(geometryLiteral, otherIndex.getOrDefault(geometryLiteral, geometryDatatype.read(geometryLiteral)));
```

jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/index/QueryRewriteIndex.java
```diff
@@ -140,19 +131,14 @@ public boolean isIndexActive() {
...
-                String[] parts = entry.getKey().split(KEY_SEPARATOR);
-                Resource subject = ResourceFactory.createResource(parts[0]);
-                Property property = ResourceFactory.createProperty(parts[1]);
-                Resource object = ResourceFactory.createResource(parts[2]);
-                model.add(subject, property, object);
+        Graph graph = GraphFactory.createDefaultGraph();
+        index.entrySet().forEach((entry) -> {            
+            if (entry.getValue()) {
+                graph.add(entry.getKey());
             }
```
LorenzBuehmann, line 138: Can be further simplified to either

``` java
index.forEach((key, value) -> {
            if (value) {
                graph.add(key);
            }
        });
```

afs:
Tests would be ideal but given it's been used so that will do.

So probably merge this in a few days and I'll apply the @LorenzBuehmann suggestions.

jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/index/GeometryLiteralIndex.java
afs: This isn't quite the same - it always evaluates the inner function calls `otherIndex.getOrDefault` and `geometryDatatype.read`.

LorenzBuehmann: True, my bad - should be `computeIfAbsent` to avoid unnecessary lookup of secondary index - so more like
``` java
geometryWrapper = index.computeIfAbsent(geometryLiteral, (lit) -> otherIndex.getOrDefault(lit, geometryDatatype.read(lit)));
```
would this be more readable? Don't know. Technically this is still more efficient from a Java internals point of view as it gathers the internal map bucket just once. But I don't think it matters in our case

PR #1262, opened 2022-04-17
Updates (Apache POM , UI components)

afs:
Apache POM to v26

In jena-fuseki-ui:
fortawesome/fontawesome-svg-core to ^6.1.1
fortawesome/free-solid-svg-icons to ^6.1.1
fortawesome/vue-fontawesome to ^2.0.6
triply/yasqe to ^4.2.23
triply/yasr to ^4.2.23
axios to ^0.26.1

### DIFF
--- a/jena-fuseki2/jena-fuseki-ui/package.json
+++ b/jena-fuseki2/jena-fuseki-ui/package.json
@@ -14,12 +14,12 @@
     "lint": "vue-cli-service lint"
   },
   "dependencies": {
-    "@fortawesome/fontawesome-svg-core": "^1.3.0",
-    "@fortawesome/free-solid-svg-icons": "^6.0.0",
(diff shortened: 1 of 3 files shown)


kinow (approved):

afs:
And a security alert on async goes away.

It arrived on the working copy of jena on my own account.
Forking on GH means that security alerts are not secret.

Attempting to update eslit didn't work so I left it alone.

kinow:
I think when you get a dependabot notification the issue is already published, so might be harder to fix it in secret.

I'm using [Snyk](https://snyk.io/) on another project, and that one appears to find issues that are not yet disclosed (actually quite a lot, and not JS, but Python + Docker). Maybe we can test that on Jena too some day.

afs:
You're right - the alerts are on published issues. It's only GH making the alerts private to the project ... when they are so simple to get anyway

PR #1268, opened 2022-04-21
fix for custom builtins not being used when parsing rules

mattmarshall:
`Rule.rulesFromURL(String, BuiltinRegistry)` does not pass the `BuiltinRegistry` argument to `Rule.rulesParserFromReader()`. This means that a user-specified builtin registry cannot be used with rules that are loaded using `rulesFromURL()`. This seems like a bug, so I can create a corresponding GitHub or JIRA issue, but wanted to first present this pull request. This is a simple one-line fix to pass `registry` into `rulesParserFromReader()`, along with some additional unit tests and a minor reorganization for some of the `.rules` files.

If you checkout just the `TestRuleLoader.java` unit test file from this PR, you can see that the `load_from_file_with_custom_builtins()` test fails. After adding the change to `Rule.java`, the test passes.

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/reasoner/rulesys/Rule.java
+++ b/jena-core/src/main/java/org/apache/jena/reasoner/rulesys/Rule.java
@@ -510,7 +510,7 @@ public class Rule implements ClauseEntry {
             InputStream in = FileManager.getInternal().open(uri);
             if (in == null) throw new RulesetNotFoundException( uri );
             br = FileUtils.asBufferedUTF8( in );
-            return parseRules( Rule.rulesParserFromReader( br ) );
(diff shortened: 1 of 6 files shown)


afs:
Hi @mattmarshall, thanks for the PR.

Yes, please do create a GH issue or a JIRA.

afs:
Looks good and the build is passing.

@der -- unless I heard otherwise, I'll merge this.

afs (approved):

der:
Yes, looks good to me. Thanks @mattmarshall

mattmarshall:
Thanks for merging this @afs and @der. I hadn't gotten around to creating an issue for this yet, so I appreciate it!

PR #1291, opened 2022-05-03
jena-fuseki-access - Propagate request/service context

vtermanis:
### What
Process request/service context in `AccessControlledDataset` the same way as is done in (parent) query processor, i.e. honour the server => dataset => endpoint context value priority order.

### Why
Standard SPARQL queries honour the context of the endpoint ([see docs](https://jena.apache.org/documentation/fuseki2/fuseki-config-endpoint.html)) as well as setting of reuqest-specific timeout via the `timeout` URL parameter. These two features allow one to:
- Override context values specified at server and/or dataset level (e.g. disable access to text indexing or change the default timeout for the endpoint)
- Specify per-request timeouts

(See bottom of summary of example. See also [mailing list thread](https://markmail.org/message/n7tcbi3ihfzlvbdv#query:+page:1+mid:2mvx45a2avki4tvc+state:results).)

### How
- Use `QueryExec` to create the `QueryExecution` instead of `QueryExecutionFactory` (the latter does not consider `HTTPActions`'s context)
- Use same `timeout` parameter logic for pre-request timeouts as in base SPARQL query processor

---

**Note**: I am not convinced that the way I've updated `AccessControlledDataset` (and promoted a helper method from `SPARQLQueryProcessor.java` to public) is the right way to go. But I can confirm that the following use-case works:

1. Define a dataset `DS1` with `jena:text` indexing enabled
2. Define a dataset `DS2`, with `AccessControlledDataset` wrapping `DS1`. (The access actual rules are irrelevant here.)
3. Define service `A` exposing a query endpoint for `DS1`, with extended context: `ja:context [ ja:cxtName "http://jena.apache.org/text#index" ; ja:cxtValue false ] ;`
4. Define service `B` the same as `A`, but for `DS2`

Current behaviour (pre-patch):
- `jena:text` is exposed only with query endpoint of `B`. SPARQL queries against `A` do not match text-indexed properties.

Expected behaviour (post patch):
- Neither `A` nor `B` match text-indexed properties

### DIFF
--- a/jena-fuseki2/jena-fuseki-access/src/main/java/org/apache/jena/fuseki/access/SecurityContext.java
+++ b/jena-fuseki2/jena-fuseki-access/src/main/java/org/apache/jena/fuseki/access/SecurityContext.java
@@ -24,6 +24,8 @@ import java.util.Collection;
 import java.util.function.Predicate;
 
 import org.apache.jena.atlas.iterator.Iter;
+import org.apache.jena.fuseki.servlets.HttpAction;
+import org.apache.jena.fuseki.servlets.SPARQLQueryProcessor;
 import org.apache.jena.graph.Node;
 import org.apache.jena.graph.NodeFactory;
 import org.apache.jena.query.Query;
@@ -31,6 +33,9 @@ import org.apache.jena.query.QueryExecution;
 import org.apache.jena.query.QueryFactory;
 import org.apache.jena.sparql.core.DatasetGraph;
 import org.apache.jena.sparql.core.Quad;
+import org.apache.jena.sparql.exec.QueryExec;
+import org.apache.jena.sparql.exec.QueryExecDatasetBuilder;
+import org.apache.jena.sparql.exec.QueryExecutionAdapter;
 import org.apache.jena.sparql.util.Context;
 
 /** A {@link SecurityContext} is the things actor (user, role) is allowed to do.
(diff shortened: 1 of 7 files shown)


jena-fuseki2/jena-fuseki-access/src/main/java/org/apache/jena/fuseki/access/SecurityContext.java
```diff
@@ -78,11 +83,22 @@ public default Collection<String> visibleGraphNames() {
...
     }
 
-    public QueryExecution createQueryExecution(Query query, DatasetGraph dsg);
+    public default QueryExecution createQueryExecution(HttpAction action, Query query, DatasetGraph dsg) {
+        QueryExecDatasetBuilder builder = QueryExec.newBuilder()
+            .dataset(dsg)
+            .query(query)
+            .context(action.getContext())
+            ;
+        // Note: Instead of this call (and the above build), the equivalent code from
```
vtermanis, line 96: Repeating the same step here for timeout seems not right (but I wasn't sure what a better / the right approach would be).

kinow (approved):
Use-case sounds good to me. Not sure about the implementation either, so probably needs @afs 's review here.

Thanks!

(also need to confirm all tests pass in Jenkins or GH Actions [manual])

afs:
From users@ thread https://lists.apache.org/thread/h0c81qjl8oc83yl2xf7xvt4l0pw4grrf

It looks like there is an issue as to whether the text dataset should push down the context setting for the index or not. Requiring endpoint configuration isn't so user-friendly.

This PR may contain a change to make anyway but this jena-fuseki-access uses context settings `DataAccessCtl.symAuthorizationService` itself so allowing the endpoints to modify context may be a security risk. Needs investigation - it's been a long time since I looked at the code!

An outstanding question [from email](https://lists.apache.org/thread/1t6cd3chn75twm1w6wrfwp5wrow6mk2g):

* Doesn't that [setting the context to an illegal value] cause warnings in the Fuseki log? 

It'll need some tests.

vtermanis:
Apologies - I updated the description since it had a mistake. The expected behaviour section has changed to:
> Current behaviour (pre-patch):
> 
> - jena:text is exposed only with query endpoint of B. SPARQL queries against A do not match text-indexed properties.
>
> Expected behaviour (post patch):
> - **Neither A nor B** match text-indexed properties

vtermanis:
(Changing to draft as suggested, since the best way to address this is still being debated.)

afs:
@vtermanis -- there has been a bug fixes for general context handling with queries and these are now on the main branch. I don't think it immediately relates to your report but I wanted to make sure you are working against the right state of the codebase.

vtermanis:
@afs thanks for the heads-up, do you mean:
- https://github.com/apache/jena/pull/1444

Is that the main one or other there others in particular? (Though it'll probably be clear when I rebase what's changed and look at the query servlet & processor code)

afs:
#1444 is about transactions.

#1375 and parts of some others around that time.

afs:
Closing due to age and the fact a lot has changed in Fuseki.

The context needs are possibly addressed by work that happened after this PR.

For the more general issues raised by [JENA-2339](https://issues.apache.org/jira/projects/JENA/issues/JENA-2339) / PR #1441 where control of security is made external to Fuseki, there is now a 3rd party system for Jena -- rdf-abac (https://github.com/telicent-oss/rdf-abac) which puts the policy control concern outside of the Fuseki engine that might be applicible to the use case.

It provides per-quad filtering of data coming out of a TDB2 database. A query operation only gets to see RDF data that it is permitted to see by policy. Each per gets a request specific dynamic view of the RDF data so query execution is only based on permitted data. The policy engine is accessed at request time.

For text query, either check for triples (and hence security permissions) after text index access or enhance the text search service to apply system-wide policies hared with the graph database.

Issue #1302, opened 2022-05-13
FusekiServer -- custom servlet filters added after FusekiFilter.


afs:
Application added servlet filter are added after the FusekiFilter despite the javadoc saying they are before.

This means they can't preprocess Fuseki data service requests.

PR #1303, opened 2022-05-13
GH-1302: Before and after servlet filters. Default to the "before" ones.

afs:
This closes GH-1302.

### DIFF
--- a/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
+++ b/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
@@ -417,8 +417,9 @@ public class FusekiServer {
         // Order does not matter, the rules of pathspec dispatch are "exact match"
         // before "prefix match".
         private Map<String, HttpServlet> servlets           = new HashMap<>();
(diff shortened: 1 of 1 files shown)


jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
```diff
@@ -1405,7 +1406,10 @@ private void servletsAndFilters(ServletContextHandler context) {
                 addFilterHolder(context, "/*", holder);
             }
 
-            // End of chain. May dispatch and not pass on requests.
+            beforeFilters.forEach(pair -> addFilter(context, pair.getLeft(), pair.getRight()));
+
+            // End of chain tghough there may be custom "afterFilters".
```
kinow, line 1411: s/tghough/though?

kinow (approved):

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
afs: Done!

Issue #1306, opened 2022-05-15
Upgrade to Vue 3


kinow:
Creating an issue so I don't lose track of this.

We are using Bootstrap Vue for the UI. It gives developers the same that vanilla Bootstrap offers, but instead of using CSS classes directly, it gives them Vue components, ready to use.

While it was great for rapid development for the migration of Backbone.JS to Vue.js, unfortunately now we are stuck with Vue 2 as Bootstrap Vue has not upgraded to Vue 3 yet - https://github.com/bootstrap-vue/bootstrap-vue/issues/5196

I **think** we might be able to move to vanilla Bootstrap and create simple components, assuming we are not using too many Bootstrap Vue components.

Alternatively, we can move to another UI library, with the disadvantage that unless it's a variation of Bootstrap, we would have a different UI look and feel.

Once this is solved, we might be ready to move to Vue 3 and Vite.

kinow:
List of components we are using:

```
kinow@ranma:~/Development/java/jena/jena/jena-fuseki2/jena-fuseki-ui$ grep -r -H -h -E "<b-.*[\s|>]" -o src | awk -F" |>" '{ print $1 }' | sort | uniq | cut -c2-
b-alert
b-badge
b-btn
b-button
b-card
b-card-body
b-card-header
b-card-title
b-col
b-collapse
b-container
b-form
b-form-s
b-input-group
b-input-group-append
b-lis
b-list-group
b-nav
b-navbar
b-navbar-brand
b-navbar-nav
b-navbar-toggle
b-nav-item
b-overlay
b-progress
b-row
b-skeleton
b-spinner
b-th
b-tr
```

afs:
:smile: just for getting the regex to work!

afs:
Same title as #1251. Is this a duplicate? continuation? sub-task?

kinow:
Bug in the regex, some `b-form-*` elements were swallowed by the regex. Here's the current list:

```
kinow@ranma:~/Development/java/jena/jena/jena-fuseki2/jena-fuseki-ui$ grep -r -h -E "<b\-[^ >]+" -o src | awk -F" " '{ print $1 }' | cut -c2- | sort | uniq -c | sort -h -r
     11 b-form-group
     10 b-nav-item
      9 b-card-body
      7 b-card-header
      7 b-card
      5 b-form-input
      4 b-table
      4 b-spinner
      3 b-skeleton
      3 b-popover
      2 b-navbar-nav
      2 b-nav
[...]

PR #1307, opened 2022-05-15
GH-1306, GH-1443, JENA-2307: Replace Bootstrap Vue by vanilla Bootstrap 5 CSS classes

### LINKED ISSUE #1443: Fuseki Query-UI sends request to wrong endpoint
### Version

4.5.0

### What happened?

# Current behaviour:
Queries in the Fuseki Query-UI are always sent to the query-endpoint. The in put in the input-field "SPARQL Endpoint" seems to be ignored.

fuseki_query_ui

# Desired behaviour:
Input field "SPARQL Edpoint" is considered within sparql query execution

_No response_

_No response_

kinow:
part of #1306 

closes #1443

- [x] b-alert
- [x] b-badge
- [x] b-btn
- [x] b-button
- [x] b-card
- [x] b-card-body
- [x] b-card-header
- [x] b-card-title
- [x] b-col
- [x] b-collapse
- [x] b-container
- [x] b-form
- [x] b-form-group
- [x] b-form-input
- [x] b-form-radio-group
- [x] b-form-select
- [x] b-input-group
- [x] b-input-group-append
- [x] b-list-group
- [x] b-list-group-item
- [x] b-nav
- [x] b-navbar
- [x] b-navbar-brand
- [x] b-navbar-nav
- [x] b-navbar-toggle
- [x] b-nav-item
- [x] b-overlay
- [x] b-pagination
- [x] b-popover
- [x] b-progress
- [x] b-row
- [x] b-skeleton
- [x] b-spinner
- [x] b-table
- [x] b-textarea
- [x] b-th
- [x] b-tr


afs:
Awesome - the simpler (less dependencies) the better.

kinow:
\- Me: _@afs do not worry, this update is going very well_
\- Also me: 
GIFrecord_2022-05-22_153237

:smile:

kinow:
Fixed a bug in the Edit, where a recent bug fix in the UI made it raise an error whenever we loaded the page. The issue was the order the components were loaded, so I just had to hook the reactivity another way (watch a property instead of doing when the component is created).

kinow:
Added the CSS of Bootstrap 5 directly after the Bootstrap Vue. It looks like Bootstrap Vue was either using an older Bootstrap 5, or customizing some classes. Now the layout has a few problems, but we can fix it as we update the components.

Once all components here have been migrated to Bootstrap, my plan is to remove the CSS styles and JS dependencies of Bootstrap Vue in this PR, and then adjust the layout to match what we have (it will be slightly different, font color may be a darker black, corners a little more or less rounded, and other cosmetic changes due to changes in Bootstrap, but nothing major, in theory).

kinow:
`form-group` was dropped in 5.2.
image

kinow:
Now pending just the table element. It's the most elaborate from the ones we used, so this one might take a bit longer.

kinow:
Finished with `b-table`, missing now just the Pagination component, and some testing.

I fixed one Vue reactivity issue in the query page. Will have to confirm, but it may be #1443, or related to it.

afs:
Sounds like this is in a good position.

kinow:
All of the BootstrapVue components have been converted. Now just missing

- [ ] remove the b-table from the upload view, use the new table component
- [ ] remove any left-over BootstrapVue code
- [ ] remove BootstrapVue
- [ ] manual tests

kinow:
This should be now working without Bootstrap Vue (blocker for Vue 3), and with the vanilla Bootstrap 5.2 CSS. I will keep working on and off to add more tests, especially since I had to write some custom components to replace the ones provided by Bootstrap Vue. But in case a release is necessary with this code (as it fixes one bug), it can be tested manually as well and released.

I have a separate issue to work exclusively on e2e tests, which should hopefully make Jena UI less error prone, and help with code refactoring (which unfortunately happens with some frequency in JS due to libraries, updated, security, etc).

-Bruno

kinow:
Managed to add e2e tests today too. They run similar to Selenium, but due to the way Cypress works, tests are less brittle.

Running the e2e tests will download some large dev dependencies, and video recording is disabled by default to reduce disk space, but that can be enabled to record the tests if needed.

```
yarn run coverage:e2e
yarn run test:e2e
yarn run test:e2e -- --headless --browser chrome
```

These are examples of how to run it. Without the `--headless`, Cypress will display a dialogue to choose the browser to run the tests with. You can go back in time and review what the UI looked like for each step of the test, which is quite handy when reproducing tests or user issues.

I've implemented a small JS mock server to store datasets and return some dummy data for the UI. This way we don't need to start a Docker container or something like that every time we run the tests (I found out that that helps speeding up the UI development, especially if a contributor is not familiar how to run the Jena backend.)

Screenshot from 2022-07-20 13-26-10

Test coverage is also quite good now with the e2e tests - the report below shows just the e2e tests coverage, but what's important is that it's covering the new code that I had to write to move away from Bootstrap Vue, and also that we can use these tests to write most of the new tests if needed (i.e. easy to copy-and-paste to create new tests).

Screenshot from 2022-07-20 13-57-08

Ready for review @afs, whew. After this one gets merged, my next task will be to upgrade all dependencies to the latest, and start moving things from Vue 2 to Vue 3. I will need to refresh my memory on what's new, and follow their migration guide. But it should be **a lot** easier than it was to remove the Bootstrap Vue dependency from our code base.

Thanks!
-Bruno

kinow:
@ramsel1508, my last commit above contains a e2e test for your issue #1443. So from now on, hopefully, this issue won't happen again :+1: Thanks for reporting and for the patience!

-Bruno

kinow:
Rebased. Then re-ran `yarn install`. Unit tests, and e2e tests both have passed successfully :-)

Closed the superseded issue for e2e tests, and linked this one in JIRA as well.

kinow:
Started migrating the app to Vue3 to save time (not committing anything here, will stash / create a diff).

Only blocker so far: https://github.com/vuejs/router/issues/454

Other than that, got the app up and running with no warnings after poking around with `yarn`, my IDE, and reading the great migration docs the Vue devs wrote :tada:

afs:
I downloaded the PR as a diff, applied it to current main (2022-07-28), and did a local build.

Only two things, one is quite minor:

The edit function: it is not loading the current contents of the graph when you click on the graph in the list of graph (neither default nor named). This means the edit are has no triples, and when put back the graph is empty.

When uploading data ("add data") there is a flash of orange in the progress bar then it goes green as it advances (guess: the default is orange). Minor.

Otherwise looking good and everything I tried worked.

kinow:
Shouldn't be hard to address these two issues. Will update it over the next days. Thanks @afs !

kinow:
>When uploading data ("add data") there is a flash of orange in the progress bar then it goes green as it advances (guess: the default is orange). Minor.

I think that's because either your data is really small, or your server is responding really fast. Here's what it looks like when I upload a ~3MB file (colors are a little off due to recording quality).

GIFrecord_2022-08-02_203732

It has no color by default. After you have selected a file and clicked upload, it will start the upload and change the bar color to orange and increase the progress status/number. Once it reaches 100% and the upload is complete it should turn green.

afs:
Yes - it was on the local machine and a very small test file. I assumed orange was an "error indicator", not "in-progress".

kinow:
> Yes - it was on the local machine and a very small test file. I assumed orange was an "error indicator", not "in-progress".

Good point! Follow-up to improve it: https://github.com/apache/jena/issues/1467

kinow:
>The edit function: it is not loading the current contents of the graph when you click on the graph in the list of graph (neither default nor named). This means the edit are has no triples, and when put back the graph is empty.

Excellent feedback.

There was another issue that when you clicked on add data or edit, you could see a yellow-ish box for a very quick moment, which says that No Service was found for the graphs. That's because the page is still loading the services from the service (i.e. an HTTP request and checking if we have the gsp-rw, and other services).

I changed it and now it should display that message only after it has finished loading the graphs and indeed couldn't locate the services. Before that we should see only a normal Loading/Spinner, or a Placeholder element until the page is fully loaded.

Now for the issue you found, there were two problems. One being a JavaScript error (thanks for catching that up!), and another left-over from the Bootstrap Vue migration :open_mouth: I've removed it, fixed the JS error, and wrote an e2e test to try to prevent it from happening again :+1: 

@afs in case you haven't seen the e2e tests yet, here's what it looks like when I re-run the e2e test for the Edit function. The HTTP calls are mocks that I copied/adapted from a running Fuseki instance. It creates 11 datasets (for testing pagination), and selects the first `a` dataset, then lists the current graphs, and clicks on the graph name. Finally, it confirms the code editor contains some expected text.

I'm hoping this will help us move with the UI without adding regressions in every release (No promises! JS is a really permissive language, and there are so many libs involved, even though we are trying to be minimalistic... but that's the theory :laughing: )

Ready for review again! Thanks! :wave:

afs:
Just checking here - subject to review, this is now ready to go into the codebase? If so - great - 4.6.0 where we come.

And now PR review ...

afs (approved):

kinow:
Yup, just pending a review with some testing.

If you find any other pending things here it shouldn't be hard to fix, and the best part is that this PR has added e2e tests.

So in case of issues/regressions, we can cover pretty much any issue with unit or e2e tests.

I have another branch with vue 3 on top of this PR. Just pending one small issue. So for next release I think we should be able to update all our dependencies to their latest too.

Thanks!

PR #1322, opened 2022-05-18
Add alternative java to more commands

seitenbau-govdata:
In #1185 the option to set the used java installation was added to most commands. For some commands this is still missing. Therefore we have added the missing ones.

### DIFF
--- a/apache-jena/bin/jena
+++ b/apache-jena/bin/jena
@@ -63,6 +63,25 @@ if [ -L "${JENA_HOME}" ]; then
   export JENA_HOME
 fi
 
+if [ -z "$JAVA" ]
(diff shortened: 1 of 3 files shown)


rvesse:
@afs Shouldn't these also have been generated from the templates in the first place?  Or was this an artefact of these commands coming in from a separate dev branch (particular for the `xload` stuff)

afs:
These commands are slightly different to the template.bin.

All the template ones vary only in the execute class name.

`jena` either prints details or shifts to find the command.
`xload-*` are the scripts for TDB1 xloader.

afs (approved):

PR #1323, opened 2022-05-18
fix a few logging typos

bvosburgh-tq:
When researching some problems, I came across a couple of bits of logging code that seemed to want to log a stack trace, but put the exception in the string format var-args and the exception was then ignored.


kinow (approved):
Thanks!!!

afs:
Thanks!

PR #1325, opened 2022-05-19
GH-1299: update GML XML namespace ont/gml => gml/3.2

### LINKED ISSUE #1299: Unable to parse actual GML as gmlLiteral due to xmlns
We are currently stumbling upon an issue with consuming GML exported from e.g. WFS

It cannot be used (directly) in Jena because the official [GML standard](https://portal.ogc.org/files/?artifact_id=20509) uses a different xmlns than the one Jena searches for, even though the local names of the elements are the same.

I've also opened https://github.com/opengeospatial/ogc-geosparql/issues/330 for clarification

SimonBin:
Fixes GH-1299 

A simple search&replace job

Based on opengeospatial/ogc-geosparql#334

### DIFF
--- a/jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/parsers/gml/GMLReader.java
+++ b/jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/parsers/gml/GMLReader.java
@@ -64,7 +64,7 @@ public class GMLReader implements ParserReader {
     private final CoordinateSequenceDimensions dims;
     private final DimensionInfo dimensionInfo;
 
-    private static final Namespace GML_NAMESPACE = Namespace.getNamespace("gml", "http://www.opengis.net/ont/gml");
+    private static final Namespace GML_NAMESPACE = Namespace.getNamespace("gml", "http://www.opengis.net/gml/3.2");
 
     /**
      * Aiming to achieve SF-0 of GML Simple Features Profile 2.0 [10-100R3].<br>
@@ -692,7 +692,7 @@ public class GMLReader implements ParserReader {
         return memberElements;
     }
 
-    private static final String EMPTY_GML_TEXT = "<gml:Point xmlns:gml='http://www.opengis.net/ont/gml' srsName=\"http://www.opengis.net/def/crs/OGC/1.3/CRS84\" />";
+    private static final String EMPTY_GML_TEXT = "<gml:Point xmlns:gml='http://www.opengis.net/gml/3.2' srsName=\"http://www.opengis.net/def/crs/OGC/1.3/CRS84\" />";
 
     public static GMLReader extract(String gmlText) throws JDOMException, IOException {
 
--- a/jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/parsers/gml/GMLWriter.java
+++ b/jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/parsers/gml/GMLWriter.java
@@ -32,7 +32,7 @@ import org.locationtech.jts.geom.*;
  */
 public class GMLWriter {
 
-    public static final Namespace GML_NAMESPACE = Namespace.getNamespace("gml", "http://www.opengis.net/ont/gml");
+    public static final Namespace GML_NAMESPACE = Namespace.getNamespace("gml", "http://www.opengis.net/gml/3.2");
 
     private static final XMLOutputter XML_OUTPUT = new XMLOutputter();
 
--- a/jena-geosparql/src/test/java/org/apache/jena/geosparql/configuration/GeoSPARQLOperationsTest.java
+++ b/jena-geosparql/src/test/java/org/apache/jena/geosparql/configuration/GeoSPARQLOperationsTest.java
@@ -77,7 +77,7 @@ public class GeoSPARQLOperationsTest {
         ORIGINAL_DATA.add(geometryA, Geo.AS_WKT_PROP, wktA);
         ORIGINAL_DATASET.setDefaultModel(ORIGINAL_DATA);
 
-        Literal gmlA = ResourceFactory.createTypedLiteral("<gml:Point xmlns:gml=\"http://www.opengis.net/ont/gml\" srsName=\"http://www.opengis.net/def/crs/EPSG/0/4326\"><gml:pos>1 0</gml:pos></gml:Point>", GMLDatatype.INSTANCE);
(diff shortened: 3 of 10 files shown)


galbiston:
I'm not familiar with all the details of GML standards but the GeoSPARQL standard `11-052r4` page 18 provides an example GML literal with namespace for GML 3.2 of `"http://www.opengis.net/ont/gml\"`. The current implementation of GML Simple Features Profile 2.0 `10-100r3` which is a profile of GML 3.2.1. 

@SimonBin could you expand upon why this needs to be changed? If it was changed would `http://www.opengis.net/gmlsf/2.0` not be more appropriate?

image

SimonBin:
hi @galbiston , it was a mistake in the standard; they confused the RDF prefix with the XML namespace

See the referenced issue https://github.com/opengeospatial/ogc-geosparql/pull/334

This should be changed because if we want to consume GML 3.2, lets say from a GeoServer or any other official tool, it will have the correct namespace (`http://www.opengis.net/gml/3.2`), and thus the current Jena+GeoSPARQL will be unable to process it.

XML namespaces are very much similar in nature to RDF Turtle/SPARQL prefix declarations; so if we have an XML document

```xml
<gml:Point 
        srsName="http://www.opengis.net/def/crs/OGC/1.3/CRS84"
        xmlns:gml="http://www.opengis.net/ont/gml">
    <gml:pos>-83.38 33.95</gml:pos>
</gml:Point>
```

its "fully expanded" structure could be imagined as (not valid XML, just to give an idea)

```
〈http://www.opengis.net/ont/gml#Point 
        srsName="http://www.opengis.net/def/crs/OGC/1.3/CRS84" 〉
    〈http://www.opengis.net/ont/gml#pos〉-83.38 33.95〈／http://www.opengis.net/ont/gml#pos〉
〈／http://www.opengis.net/ont/gml#Point〉
```

which is obviously completely different from

```
〈http://www.opengis.net/gml/3.2#Point 
        srsName="http://www.opengis.net/def/crs/OGC/1.3/CRS84" 〉
    〈http://www.opengis.net/gml/3.2#pos〉-83.38 33.95〈／http://www.opengis.net/gml/3.2#pos〉
〈／http://www.opengis.net/gml/3.2#Point〉
```

you can think of it like the two RDF Turtle documents

```turtle
@prefix gml: <http://www.opengis.net/gml/3.2#> .

[] a gml:Point ; gml:pos "-83.38 33.95" .
```

and

```turtle
@prefix gml: <http://www.opengis.net/ont/gml#> .

[] a gml:Point ; gml:pos "-83.38 33.95" .

```

They would result in these triples:

```
_:b0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.opengis.net/gml/3.2#Point> .
_:b0 <http://www.opengis.net/gml/3.2#pos> "-83.38 33.95" .
```

```
_:b0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.opengis.net/ont/gml#Point> .
_:b0 <http://www.opengis.net/ont/gml#pos> "-83.38 33.95" .
```

which are totally different properties and types

afs:
Every database using `gml:` is invalidated and needs reloading with corrected data?

That needs text for the release announcement.

galbiston:
Yes, as the namespace is encoded into the literal. Offering support through the parser would add complexity and result in a mixed datasets as new GML literals are created. However, `Dataset` conversion could be provided as a function in `configuration.GeoSPARQLOperations` as a one time operation to find all the incorrect namespace `gmlLiteral` and perform the string replacement. There are already similar functions to convert coordinate reference systems etc.

It does need to be made clear that the spatial relations are based on using geometries in the GML Simple Features Profile `10-100r3`, which is a subset of the geometries in GML 3.2. In the GML specification the compliance level is expressed in the document root but that doesn't exist in the GML literals as they are only fragments. So having the namespace as GML3.2 implies that any geometry can be used but many are out of scope of the GeoSPARQL standard.

The flexibility is being provided for full GML support but in practical terms only the GML Simple Features Profile seems to be relevant.

galbiston:
Any spatial indexes would also need to be deleted and rebuilt.

SimonBin:
unfortunately yes, but since it was previously impossible to consume GML created by other tools I hope that the real world impact should be small

afs:
@ Jena GeoSPARQL community: There needs to be text in the ANN for 4.6.0 that highlight this change.

Draft text:

GH-1299 - The GeoSPARQL has acknowledged a mistake in the example in their document
whereby the wrong namespace was being used for "gml:".

https://github.com/opengeospatial/ogc-geosparql/pull/334

Fixing this in jena-geosparql enables working with GML from other tools but unfortunately impacts impacts existing data. Any spatial indexes need to be deleted and rebuilt.

SimonBin:
> Fixing this in jena-geosparql enables working with GML from other tools but unfortunately impacts impacts existing data.

Existing gmlLiterals need to be migrated to the correct XML namespace

> Any spatial indexes need to be deleted and rebuilt.

... if you were previously making use of the gml in jena-geosparql


...

at least, as far as I can, tell this does not affect users that were only using wkt or the old wgs84

afs:
Draft text v2:
----
https://github.com/apache/jena/issues/1299 - The GeoSPARQL has acknowledged a mistake in the example in their document
whereby the wrong namespace was being used for "gml:".

https://github.com/opengeospatial/ogc-geosparql/pull/334

Fixing this in jena-geosparql enables working with GML from other tools but unfortunately impacts impacts existing data. 

If you were previously making use of the gml in jena-geosparql, any use of URI prefix `http://www.opengis.net/ont/gml` needs to change to `http://www.opengis.net/gml/3.2` and any spatial indexes need to be deleted and rebuilt
----
or should we just say "delete and rebuilt" in case the user operating the system, isn't the user providing the data?

afs:
I tried this PR on current main and got a test failure:
```
[INFO] Running org.apache.jena.geosparql.geof.nontopological.filter_functions.AsGeoJSONFFTest
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 2.658 s <<< FAILURE! - in org.apache.jena.geosparql.geof.nontopological.filter_functions.AsGeoJSONFFTest
[ERROR] testExec(org.apache.jena.geosparql.geof.nontopological.filter_functions.AsGeoJSONFFTest)  Time elapsed: 2.658 s  <<< FAILURE!
java.lang.AssertionError: 

Expected :LINESTRING (11.12011 46.06974, 11.12009 46.06982)
Actual   :LINESTRING EMPTY
Tolerance:1.0E-4
	at org.apache.jena.geosparql.geof.nontopological.filter_functions.AsGeoJSONFFTest.testExec(AsGeoJSONFFTest.java:54)
```

Could the PR be updated please?
Are there any other changes to jena-geosparql that have happened since the PR was created that might be problems?

SimonBin:
I updated the PR to current main, I hope it works now

afs:
Yes, it passes the build tests run now.

Thank you!

PR #1327, opened 2022-05-19
allow geo index search for literals

SimonBin:
I would like to open this for discussion

a relatively small change which is **not** part of the standard

it provides some nice real-world use: you can pass a Geometry literal in a SPARQL query and use the Geo-index to locate some Geometry/Feature, something which is not possible (afaik) to do with a filter(geof:*)


afs:
GeoSPARQL community - opinions?

Looking at the diff, it seems alright from my perspective but I'm not deeply into jena-geosparql.

It does needs tests and a documentation update.

SimonBin:
Hi @asf 

I added a test case

docs are on the jena-site only? I will prepare a PR

afs:
`s/@asf/@afs/` :smile:

> docs are on the jena-site 

Yes.

afs:
I'm getting a test failure at `SpatialObjectGeometryLiteral.testRetrieve`.

```
java.lang.NullPointerException
	at org.apache.jena.geosparql.geo.topological.SpatialObjectGeometryLiteral.retrieve(SpatialObjectGeometryLiteral.java:107)
	at org.apache.jena.geosparql.geo.topological.SpatialObjectGeometryLiteralTest.testRetrieve(SpatialObjectGeometryLiteralTest.java:77)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
. . .
```
The `testRetrieve` argument is null.

Should it be at `SpatialObjectGeometryLiteral:107`?
```
    if (targetSpatialObject != null && 
```
or is there a sync issue because of all the other changes in jena-geosparql recently?

SimonBin:
yes you're right

afs:
Merged (I squashed the commits to one that records the change with a message for the long term)

PR #1330, opened 2022-05-20
add geof:asGeoJSON function

SimonBin:
suggestion to add a GeoJSON conversion function

this will add a dependency on jts-io-common

it is not support for reading GeoJSON nor full implementation of GeoJSON data type, but it can be useful in SPARQL for web applications that wish to consume GeoJSON

co-authored with @LorenzBuehmann 

https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html#_function_geofasgeojson


jena-geosparql/src/test/java/org/apache/jena/geosparql/geof/nontopological/filter_functions/AsGeoJSONFFTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+import static org.junit.Assert.assertTrue;
+
+public class AsGeoJSONFFTest {
+    private static final double TOLERANCE = 0.0001;
+
+    public AsGeoJSONFFTest() {
+    }
+
+    @BeforeClass
```
afs, line 20: These don't need to be heer if they are empty.

afs:
Looks OK except for some details.

The build does not run.

Please add the Apache header (the same as in other java files in Apache Jena) to:

`AsGeoJSONFFTest.java`
`AsGeoJSONFF.java`
`GeoJSONDatatype.java`

You can run the build with `mvn clean install` or `mvn clean install -Pdev` (although that does not include jena-fuseki-geosparql).

GeoJSON is mentioned at https://jena.apache.org/documentation/geosparql/, (source: https://github.com/apache/jena-site/blob/main/source/documentation/geosparql/__index.md) which needs updating.

afs (approved):

Issue #1292, opened 2022-05-03
Add support for authentication with a bearer token


Aklakan:
A quick request to make the following single method public so that I can inject a bearer token - if possible for 4.5.0-SNAPSHOT. On the main branch I already get 4.6.0-SNAPSHOT - so I am not sure if PRs for 4.5.0 are still possible.

Because of the lack of interceptors in java HttpClient (I am underwhelmed...) this seems to be the only reliable way to make bearer token auth working with jena. De-facto, this AuthEnv is a kind of interceptor framework after all...
Of course adding a bearer token helper method to AuthLib would also be nice, but from my side this can happen in 4.6.0 if I can register my own auth mod for now.

```
public class AuthEnv {
    void registerAuthModifier(String requestTarget, AuthRequestModifier authModifier) {
        // Without query string or fragment.
        String serviceEndpoint = HttpLib.endpoint(requestTarget);
        //AuthEnv.LOG.info("Setup authentication for "+serviceEndpoint);
        authModifiers.put(serviceEndpoint, authModifier);
    }
}
```

afs:
Does `HttpLib.modifyByService` work for you? It can add the header.
https://jena.apache.org/documentation/sparql-apis/#customization-of-http-requests


4.5.0 has been built and is currently awaiting the VOTE from the PMC.
The build uses the Apache release configuration of the maven release plugin. If the version moves on,

Aklakan:
Hi Andy, thanks alot - the snippet below works:
```java
   String bearerToken = "SECRET";
   HttpRequestModifier modifier = (params, headers) -> {
       headers.put("Authorization", "Bearer " + bearerToken);
   };
   RegistryRequestModifier.get().add(url, modifier);
```

However, there is a minor follow up issue:
If the bearer token times out, then first a message is logged such as:
```
Bearer error="invalid_token", error_description="An error occurred while attempting to decode the Jwt: Jwt expired at 2022-05-03T20:22:50Z", error_uri="https://tools.ietf.org/html/rfc6750#section-3.1"
```

but then it causes a non-ideal NPE in AuthLib.authenticateHeader:

```java
public class AuthLib {
    private static AuthChallenge authenticateHeader(HttpResponse<?> httpResponse) {
        List<String> headers = httpResponse.headers().allValues("WWW-Authenticate");
        if ( headers.size() == 0 )
            return null;
        // Choose first digest or first basic.
        AuthChallenge aHeader = null;
        String result = null;
[...]

PR #1335, opened 2022-05-24
GH-1292: Support for bearer auth

afs:
This resolves GH-1292.

The PR covers mainly client-side with a thin interface for responding to bearer auth 401 challenges. The application must a function to provide the token. There is little a standard library can do. The server side has some preparation work but again there has to be custom code somewhere to validate tokens in the deployment environment. This is probaly better done with a Fuseki module - cusomt code for Fuseki main as released.

### DIFF
--- a/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
+++ b/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
@@ -24,6 +24,7 @@ import static org.apache.jena.fuseki.Fuseki.serverLog;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.*;
+import java.util.function.Function;
 import java.util.function.Predicate;
 
 import javax.servlet.Filter;
@@ -413,7 +414,10 @@ public class FusekiServer {
         private String                   httpsKeystore          = null;
         private String                   httpsKeystorePasswd    = null;
 
-        // Other servlets to add. Te pathspec for servlets must be unique.
+        // Bearer authentication : verify and extract the user for a request.
+        private Function<String, String> bearerVerifiedUser = null;
+
+        // Other servlets to add. The pathspec for servlets must be unique.
         // Order does not matter, the rules of pathspec dispatch are "exact match"
         // before "prefix match".
         private Map<String, HttpServlet> servlets           = new HashMap<>();
@@ -426,6 +430,8 @@ public class FusekiServer {
         private SecurityHandler          securityHandler    = null;
         private Map<String, Object>      servletAttr        = new HashMap<>();
 
+        // Setting and access methods also not active.
+        //private Context                  context            = null;
         // The default CORS settings.
         private static final Map<String, String> corsInitParamsDft = new LinkedHashMap<>();
         static {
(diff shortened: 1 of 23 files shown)


jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
```diff
@@ -873,6 +875,16 @@ public Builder realm(String realm) {
             return this;
         }
 
+//        /**
```
rvesse, line 878: Couldn't this be uncommented?

Also type `Berarer` in the javadoc here

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/auth/AuthBearerFilter.java
```diff
@@ -0,0 +1,132 @@
...
+            switch(authParser.getAuthScheme()) {
+                case BEARER : {
+                    if ( bearerToken == null ) {
+                        log.warn("Not a legal bearer token: "+authParser.getAuthArgs());
+                        response.sendError(HttpSC.BAD_REQUEST_400);
+                        return;
+                    }
+                    if ( verifiedUser == null ) {
+                        // No function to verify the token and extract the user.
+                        response.sendError(HttpSC.BAD_REQUEST_400);
```
rvesse, line 86: It's questionable whether this case is 400 Bad Request

This could arguably be a 500 Internal Server Error because the server is misconfigured if this filter is in use and there's no verifier function present

rvesse:
This PR is definitely a good first step and should get merged ASAP

> The server side has some preparation work but again there has to be custom code somewhere to validate tokens in the deployment environment. This is probably better done with a Fuseki module - custom code for Fuseki main as released.

I would somewhat disagree with this.  Certainly in the general case we probably can't support verifying every possible JWT that users might want to use (algorithms, key sources etc).

But for most everyday cases I think we could support a default verifier based on the following parameters that would be compatible with most OAuth2/OpenID Connect authentication providers:

- A JWKS file/URL, where JWKS is a JSON Web Key Set per [RFC 7517 Section 5](https://datatracker.ietf.org/doc/html/rfc7517#section-5), from which the public keys for verifying a token can be obtained
- Optionally the claim to obtain the username from.  This would default to `sub` but from practical experience the value of the `sub` field is often an internal system identifier rather than an actual username which is carried in a different claim.

This can be provided in a separate PR (and one which I'm happy to author myself)

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
afs: Bear authentication isn't support in the server builder. The key issue is how custom code gets added to a server installation.  It is not clear yet whether there can be server builder support or whether a Fuseki module is better.

So not in the public API at this time.

In the PR, the server builder does not do anything with `bearerVerifiedUser`. It is paired with `auth(AuthScheme.BEARER)` causing an exception.

javadoc fixed.

afs:
Environments where we might provide a solution are the major cloud providers.

The code can be delivered via a module that configures a servlet filter.

PR #1337, opened 2022-05-24
Quoted LOGGING variable in fuseki-server script.

### LINKED ISSUE #1336: Unquoted LOGGING variable in fuseki-server can make server fail to start
In version 4.5.0 of Apache Jena Fuseki, line 117 is:

```shell
exec "$JAVA" $JVM_ARGS $LOGGING -cp "$CP" org.apache.jena.fuseki.cmd.FusekiCmd "$@"
```

`$LOGGING` is set using `$FUSEKI_HOME`. I define the `FUSEKI_HOME` variable in my environment. It has an embedded space. Consequently, the `fuseki-server` script expands the `LOGGING` variable into multiple arguments. Java doesn't parse the arguments as the script intends. Can/should `$LOGGING` be surrounded in double quotes?

swartik:
Closes #1336

Addresses the problem of LOGGING having a value with spaces.

### DIFF
--- a/jena-fuseki2/apache-jena-fuseki/fuseki-server
+++ b/jena-fuseki2/apache-jena-fuseki/fuseki-server
@@ -113,8 +113,12 @@ then
     # Otherwise, the server will default to a built-in.
     LOGGING="-Dlog4j.configurationFile=$DFT_LOG_CONF"
 fi
+if [ -n "$LOGGING" ]
+then
+    LOGGING=${LOGGING@Q}
+fi
 
-exec "$JAVA" $JVM_ARGS "$LOGGING" -cp "$CP" org.apache.jena.fuseki.cmd.FusekiCmd "$@"
+exec "$JAVA" $JVM_ARGS $LOGGING -cp "$CP" org.apache.jena.fuseki.cmd.FusekiCmd "$@"
 
 ## Adding custom code to the Fuseki server:
 ##
--- a/jena-fuseki2/apache-jena-fuseki/fuseki-server
+++ b/jena-fuseki2/apache-jena-fuseki/fuseki-server
@@ -114,7 +114,7 @@ then
     LOGGING="-Dlog4j.configurationFile=$DFT_LOG_CONF"
 fi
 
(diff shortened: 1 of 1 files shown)


kinow (approved):
Thanks @swartik !

afs:
Hi @swartik -- a couple of problems:

* It's a sh script, not a bash script so `@Q` is not available. (We can switch to bash.)
* if it is bash, the substitution does not seem to work (reading the man age, I think it should but it doesn't. Bash version 5.1.16).
 
How about this approach:

```
if [ -n "$LOGGING" ]
then
    exec "$JAVA" $JVM_ARGS "$LOGGING" -cp "$CP" org.apache.jena.fuseki.cmd.FusekiCmd "$@"
else
    exec "$JAVA" $JVM_ARGS -cp "$CP" org.apache.jena.fuseki.cmd.FusekiCmd "$@"
fi
```

I've tested it with FUSKEI_HOME set to a directory path with a space in it. Could you confirm please? What in the path does your setup have the space?

There is an unrelated problem with some cases of FUSEKI_BASE including a path with a space. This is a separate issue not related to the script.

afs:
PR #1366 contains the suggested `sh` compatible fix.  @swartik Let us know if that works to your enviornment.

afs:
Closed by PR #1366.

swartik:
@afs Sorry for the delayed response – I was on vacation last week. Your solution works in my environment. Thanks for stepping in.

As you had some problems I didn't seem to experience, here's my set-up:

- I'm using cygwin, updated as of this morning.
- The bash version is 4.4.12(3)-release.
- The value I use for FUSEKI_HOME is `C:/Users/swartik/Apache Software Foundation/apache-jena-fuseki-4.5.0`.

I am curious why your version of bash didn't interpret `@Q` the same way mine did, but it's not important to resolving this issue, so unless you have lots of spare time we can leave that for another day.

afs:
I'm curious about as well `@Q` - in trying things out it sort of worked, and sort of didn't. But if some platforms work and some don't, it's a problem for the project.

PR #1340, opened 2022-05-27
JENA-2331: change JenaXMLInput to use the JDK XMLInputFactory

bvosburgh-tq:


jena-core/src/main/java/org/apache/jena/util/JenaXMLInput.java
```diff
@@ -72,16 +72,24 @@ public static XMLReader createXMLReader() throws ParserConfigurationException, S
      * Initialize an XMLInputFactory to jena settings.
      */
     public static void initXMLInputFactory(XMLInputFactory xf) {
+        // This disables DTDs entirely for the factory.
+        // All DTDs are silently ignored; takes precedence over ACCESS_EXTERNAL_DTD
+    	setXMLInputFactoryProperty(xf, XMLInputFactory.SUPPORT_DTD, Boolean.FALSE);
+
+        // Disable external DTDs (files and HTTP) - errors unless SUPPORT_DTD is false.
+        setXMLInputFactoryProperty(xf, XMLConstants.ACCESS_EXTERNAL_DTD, "");
```
afs, line 80: This could be moved to the end. I don't think order matters. Assuming all XMLInputFactory are supported by Woodstox.

jena-core/src/main/java/org/apache/jena/util/JenaXMLInput.java
```diff
@@ -72,16 +72,24 @@ public static XMLReader createXMLReader() throws ParserConfigurationException, S
...
-            xf.setProperty(XMLInputFactory.SUPPORT_DTD, false);
-
-            // Disable external DTDs (files and HTTP) - errors unless SUPPORT_DTD is false.
-            xf.setProperty(XMLConstants.ACCESS_EXTERNAL_DTD, "");
-            // disable external entities (silently ignore)
-            xf.setProperty("javax.xml.stream.isSupportingExternalEntities", false);
-        } catch(IllegalArgumentException ex){
+            xf.setProperty(name, value);
+        } catch(IllegalArgumentException ex) {
             Log.error(JenaXMLInput.class, "Problem setting StAX property", ex);
```
afs, line 93: Assuming this is now to catch specific case, the stacktrace isn't necessary (reduces the log output), if we have the exception message and the specific property `name` included instead.

afs (changes requested):

bvosburgh-tq:
> Only "JENA-2331: isolate setting of XMLInputFactory properties" makes sense for the long term history in git. One isn't relevant. Could you squash this PR please?

Done.

Thanks, Andy!

PR #1341, opened 2022-05-27
Improved Bearer Auth verifier support (GH-1292)

rvesse:


jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/auth/AuthBearerFilter.java
```diff
@@ -34,36 +37,129 @@
...
-    private final Function<String, String> verifiedUser;
+    private Function<String, String> userVerifier;
+
+    public static final String USER_VERIFIER_FUNCTION_ATTRIBUTE =
+            AuthBearerFilter.class.getCanonicalName() + ".userVerifier";
 
     @Override
-    public void init(FilterConfig filterConfig) {}
+    public void init(FilterConfig filterConfig) {
+        // For web.xml usage allow initializing in the following ways:
```
afs, line 91: Fuseki main does not use web.xml. It does not run webapps. It is a pure servlet filter stack with the static assets servlet to serve content requests.

`AuthBearerFilter` is added as a Fuseki server is built.

The code injection is by Fuseki module. They avoid the need to package Fuseki as distributed just to custom code.

rvesse: Yes but other $dayjob stuff does, and I would prefer to have a Jena provided filter that I can reuse in non-Fuseki contexts rather than being forced to implement another filter

rvesse: And the intent is to have the actual concrete implementation of a verifier function injected via Fuseki module

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/server/FusekiVocab.java
```diff
@@ -43,6 +43,9 @@
     public static final Property pPasswordFile              = property("passwd");
     public static final Property pRealm                     = property("realm");
     public static final Property pAuth                      = property("auth");
+    public static final Property pJwks                      = property("jwks");
```
afs, line 46: `FusekiVocab` is for RDF vocabulary used in the configuration file.

afs:
Context: Rob and I work at the same company.

Some improvements to `AuthBearerFilter` have gone in #1345.

The code currently has Fuseki dependencies so the goal of sharing code for bearer token handling between `AuthBearerFilter`, which is in Fuseki main, not Fuskei webapp, means that the separation isn't there yet.

Ensuring the bearer token handling is able to be deployed at cloud-providers, enterprise on-prem, and different trusted platforms will generate further requirements on the flexibility for general purpose code.

We'll close this as #1345 has some improvements and be back later.

Issue #1272, opened 2022-04-25
The IRI function in UpdateRequests ignores BASE IRIs


Aklakan:
The following query works as expected:
```sparql
BASE <https://foo.bar/> SELECT ?x { BIND(IRI('baz') AS ?x) }
```

```
-------------------------
| x                     |
=========================
| <https://foo.bar/baz> |
-------------------------
```

However, using the same graph pattern in the following update request ...
```sparql
BASE <https://foo.bar/> INSERT { ?x ?x ?x } WHERE { BIND(IRI('baz') AS ?x) }
CONSTRUCT WHERE { ?s ?p ?o }
```

... yields a result that suggests that the the base IRI was ignored:
```turtle
<file:///tmp/baz>  <file:///tmp/baz>  <file:///tmp/baz> .
```

The [sparql update spec](https://www.w3.org/TR/sparql11-update/) does not say anything about the IRI function so my assumption is that the behavior should be consistent with that for querying. As a consequence, the issue is with the behavior of Jena.

The reason is that the `E_IRI` function is only implemented to extract the base IRI from a currently running query - it lacks the logic to work with update requests.

```java
class E_IRI {
    @Override
    public NodeValue eval(NodeValue v, FunctionEnv env)
    { 
        String baseIRI = null ;
        if ( env.getContext() != null )
        {
            Query query = (Query)env.getContext().get(ARQConstants.sysCurrentQuery) ;
            if ( query != null )
                baseIRI = query.getBaseURI() ;
        }
        return NodeFunctions.iri(v, baseIRI) ;
    }
}
```

One solution would be to add another `ARQConstants.sysCurrentUpdate(Request)` constant.

Alternatively, UpdateRequest and Query both implement Prologue - so instead of sysCurrentQuery another constant such as ARQConstants.sysCurrent**Prologue** could be used.

Either way, running an update request would then have to set the appropriate context attribute and the E_IRI function would have to consider it.

afs:
`ARQConstants.sysCurrentPrologue` would seem the better fix.

Noticed:
`QueryExecDataset` and `QueryExecDatasetBuilder` both set `ARQConstants.sysCurrentQuery` which isn't right.

Aklakan:
There is now a best-effort PR - however I am not sure about the SPARQL spec and whether this actually opens a pandora's box:

[...]

PR #1343, opened 2022-05-28
GH-1272: Use static in-scope base for IRI()

afs:
Static scoped base for IRI("<x>").


afs:
There is one wrinkle to sort out. SSE parsers in two steps: text to general expression and general expression to update.

There is compatibility code in the PR.

Ideally, `(iri ...)` would take a new, optional arg to be the base. That said, SSE+updates isn't very common and SSE is primarily for Jena-dev use, it's just grown / been spotted.

Aklakan:
Just a minor remark: E_URI should have a public getBase() method so it can be accessed during expression traversal.

afs:
> Just a minor remark: E_IRI should have a public getBase() method

yes, and more! The second commit add a 2-argument form `IRI(baseExpr, rel)`. For update, I think that this general form might be useful for translating data. Experimental.

That'll need accessors to the state of E_IRI that "ExprFunction1" does not expose for printing (to SSE and to SPARQL) - WIP.

Whether it is better to be an "ExprFunction2" or "ExprFunctionN" isn't clear. `IRI` is still primarily a one-argument function. because it retains how it was setup (implicit base or explicit base), it isn't a true "ExprFunction2" either.

(And some of this code structure is really old!)

afs:
There are now 2 IRI functions: the standard "E_IRI" takes on argument and records the pasrer base and "E_IRI2" which has two arguments - `(base, rel)` - and because `base` can be relative, it also records the parser base.

These become functions `(iri)` and `(iri2)`.

PR #1371, opened 2022-06-10
Enabled 'decompressConcatenated' for bzip2 to prevent partial reads of input

Aklakan:
We just noticed missing data after loading a bzip2 file and realized that the flag was not set which caused the file to only be read partially.

### DIFF
--- a/jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
+++ b/jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
@@ -90,7 +90,7 @@ public class IO
         switch ( ext ) {
             case "":        return in;
             case ext_gz:    return new GZIPInputStream(in);
-            case ext_bz2:   return new BZip2CompressorInputStream(in);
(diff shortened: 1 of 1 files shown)


Aklakan:
We didn't generate the file but I have come across such concatenated streams several times.
I suppose it's a pretty standard feature; one can just do

```
bzip2 -c file1 > output.bz2
bzip2 -c file2 >> output.bz2
```

Note sure if the parallel versions of bzip2 (pbzip2 and lbzip2) generate them.too.

SimonBin:
as far as I know @LorenzBuehmann correct me if wrong such files are natively produced by [osm2rdf](https://github.com/ad-freiburg/osm2rdf), which uses multi-threaded streaming & compression

afs (approved):
(Doesn't look like it is intended to be a multi-file packaging)

Aklakan:
Note: A concatenated bzip2 file is NOT an archive - its just a stream of data - in our case triples. This is in contrast to `tar` or `zip` which includes multiple files. In bzip2 there are no extra file names to deal with.

Issue #1374, opened 2022-06-12
a convenient way to copy arq-context including registries.


sszuev:
### Version

4.6.0

### Suggestion

If we want to execute a query in isolated context with unique set of independent functions and p-functions, 
we will encounter some difficulties:

- the method `org.apache.jena.sparql.util.Context#copy` performs only shallow copying, not deep
- there is no straightforward way to pass the context while constructing `org.apache.jena.query.QueryExecution`

So we have to spend some time to find the proper way.
I think this can be simplified by adding two new helper-methods:

- `Context#copyWithFunctionRegistries(Context)` - the method to create a copy of `Context` with copied function and p-function registries
- `QueryExecutionFactory#create(Query, Graph, Context)` - the method to create `QueryExceution` for a `Graph` with the given context

### Are you interested in contributing a solution yourself?

Yes

afs:
Hi @sszuev ,

A copy of the context will share the original registries - the way to modify that is to modify the copy of the context. I'd put that code in the PropertyFunctionRegistry, not "teach" context about a specific setting - moie discussion for the PR - the copy operations look useful.

The best control over a query setup is via the new (4.3.x) builders which allow a change of context: `QueryExec` for graph level, `QueryExecution.create` for model level (the associated builder also some legacy compatibility about setting timeouts).

```
        Graph graph = GraphFactory.createDefaultGraph();

        PropertyFunctionRegistry pfr = new PropertyFunctionRegistry();
        PropertyFunction pf = new version();
        pfr.put("urn:test:version", (x)->pf);

        Context cxt = ARQ.getContext().copy();
        PropertyFunctionRegistry.set(cxt, pfr);

        String qs = """
                SELECT * { ?x <urn:test:version> ?y }
                """;
        RowSet rowSet = QueryExec.graph(graph)
                .context(cxt)
                .query(qs)
                .select();
        RowSetOps.out(rowSet);
```
Unrelated: there is a problem with some system provided property functions. There is a hardwired translation in the code.
```
        Graph graph = GraphFactory.createDefaultGraph();
        Context cxt = ARQ.getContext().copy();    
        PropertyFunctionRegistry pfr = new PropertyFunctionRegistry();
        PropertyFunctionRegistry.set(cxt, pfr);
        String qs = """
                PREFIX afn:     <http://jena.apache.org/ARQ/function#>
                PREFIX apf:     <http://jena.apache.org/ARQ/property#>
                SELECT * { ?x apf:version ?y }
                """;
[...]

PR #1375, opened 2022-06-12
GH-1374: add copyWithRegisties Context helper method

sszuev:
Feature request: a convenient way to execute a query against graph in a dedicated context.

If we want to execute a query in isolated context with unique set of independent functions and p-functions, 
we will encounter some difficulties:

- the method `org.apache.jena.sparql.util.Context#copy` performs only shallow copying, not deep
- there is no straightforward way to pass the context while constructing `org.apache.jena.query.QueryExecution`

So we have to spend some time to find the proper way.
I think this can be simplified by adding two new helper-methods:

- `Context#copyWithFunctionRegistries(Context)` - the method to create a copy of `Context` with copied function and p-function registries
- `QueryExecutionFactory#create(Query, Graph, Context)` - the method to create `QueryExceution` for a `Graph` with the given context

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/pfunction/PropertyFunctionRegistry.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/pfunction/PropertyFunctionRegistry.java
@@ -74,6 +74,19 @@ public class PropertyFunctionRegistry
         return reg;
     }
 
+    /**
+     * Copies the origin registry into a new one, or makes a fresh instance if the specified registry is {@code null).
+     * @param from {@link PropertyFunctionRegistry} or {@code null}
+     * @return {@link PropertyFunctionRegistry} a new instance
+     */
+    public static PropertyFunctionRegistry createFrom(PropertyFunctionRegistry from) {
+        PropertyFunctionRegistry res = new PropertyFunctionRegistry();
+        if (from != null) {
+            res.registry.putAll(from.registry);
+        }
+        return res;
+    }
+
     /** Insert an PropertyFunction class.
      *  Re-inserting with the same URI overwrites the old entry.
      *  New instance created on retrieval (auto-factory)
--- a/jena-arq/src/main/java/org/apache/jena/sparql/function/FunctionRegistry.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/function/FunctionRegistry.java
@@ -70,6 +70,19 @@ public class FunctionRegistry
         context.set(ARQConstants.registryFunctions, reg);
     }
 
+    /**
+     * Copies the origin registry into a new one, or makes a fresh instance if the specified registry is {@code null).
+     * @param from {@link FunctionRegistry } or {@code null}
+     * @return {@link FunctionRegistry} a new instance
+     */
+    public static FunctionRegistry createFrom(FunctionRegistry from) {
+        FunctionRegistry res = new FunctionRegistry();
+        if (from != null) {
+            res.registry.putAll(from.registry);
+        }
+        return res;
+    }
+
     public FunctionRegistry()
     {}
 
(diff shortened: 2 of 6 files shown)


Aklakan:
I suppose the ARQ method to copy the function registries should eventually also copy the service executor registries (one is already [available](https://github.com/apache/jena-site/blob/main/source/documentation/query/custom_service_executors.md) and another for bulk requests is [in the works](https://github.com/apache/jena/issues/1314)) - they are typically very small and I don't think it makes sense to exclude those when one explicitly asks for a deep copy of the context in order to add isolated extensions.

Maybe instead of naming the method `copyWithFunctionRegistries` it should be
more generally `copyWithRegistries`? (Alternatively something like copyWith{User,Plugin,Extension,...}Registries).
So the naming should be clarified with service executor registry in mind.

For consistency, the `ServiceExecutorRegistry` registry should provide a `createFrom` method too but that I can add with a separate PR.

jena-arq/src/main/java/org/apache/jena/sparql/pfunction/PropertyFunctionRegistry.java
```diff
@@ -77,7 +78,20 @@ public synchronized static PropertyFunctionRegistry get()
...
+     * @param from {@link PropertyFunctionRegistry} or {@code null}
+     * @return {@code PropertyFunctionRegistry} a new instance
+     */
+    public static PropertyFunctionRegistry createFrom(PropertyFunctionRegistry from) {
+        PropertyFunctionRegistry res = new PropertyFunctionRegistry();
+        if (from == null) {
+            return res;
+        }
+        res.registry.putAll(from.registry);
+        return res;
```
Aklakan, line 93: Redundant return can be removed if written that way:
```
        if (from != null) {
            res.registry.putAll(from.registry);
        }
        return res;
```

sszuev:
> ServiceExecutorRegistry



> I suppose the ARQ method to copy the function registries should eventually also copy the service executor registries (one is already [available](https://github.com/apache/jena-site/blob/main/source/documentation/query/custom_service_executors.md) and another for bulk requests is [in the works](https://github.com/apache/jena/issues/1314)) - they are typically very small and I don't think it makes sense to exclude those when one explicitly asks for a deep copy of the context in order to add isolated extensions.
> 
> Maybe instead of naming the method `copyWithFunctionRegistries` it should be more generally `copyWithRegistries`? (Alternatively something like `copyWith{User,Plugin,Extension,...}Registries).` So the naming should be clarified with the service executor registry in mind. Are there any further plugin registries in the context by default that would benefit from deep copying too?
> 
> For consistency, the `ServiceExecutorRegistry` registry should provide a `createFrom` method too but that I can add with a separate PR.

I think this is a good idea.

afs:
> I suppose the ARQ method to copy the function registries should eventually also copy the service executor registries 

Discussion on #1374.

Context should be intent neutral and not know the value of any symbols.

sszuev:
> Context should be intent neutral and not know the value of any symbols.

What is appropriate place to such functionality ? 
`org.apache.jena.sparql.util.Contexts` ?     
`org.apache.jena.sparql.util;ContextUtils` ?

This is static helper, instance of `Context` itself does not refer to any particular symbol. 
There are already static helpers like `fromDataset(DatasetGraph)`. 
I don't quite understand why `import org.apache.jena.sparql.core.DatasetGraph` is ok, but `import org.apache.jena.sparql.function.FunctionRegistry` is not?  
If we move it to some other place it would be harder to find it.

afs:
`ContextUtils` makes sense even if in `sparql.util`

`Context` is also used in RIOT - it was originally query-specific and mostly internal use but having such a mechanism, it get reused and escaped.

sszuev:
> `ContextUtils` makes sense even if in `sparql.util`
> 
> `Context` is also used in RIOT - it was originally query-specific and mostly internal use but having such a mechanism, it get reused and escaped.

Is there something else I can do in additional in the bounds of this PR ? Maybe marking all other static-helpers of `Context` with replacing and moving?

afs:
>Is there something else I can do in additional in the bounds of this PR ?

Thanks for asking. 

My first step is to look at #1381 to understand that - and it does currently look like there is a bug managing contexts in QueryExec. A fix may well have knock on-effects.

At the moment, there are several active and time-consuming PRs in flight in unrelated areas.

afs (approved):
Only thing is `{@code null)` => `{@code null}` each time in javadoc but I'll fix that.

PR #1385, opened 2022-06-15
Add maven default goal to build

kinow:
Small `pom.xml` change to add some default goals to Maven. Every time I sync the code I run `mvn clean test install -Pdev`, but sometimes I mess up part of the command… so with this `mvn -Pdev` is enough. WDYT?

Thanks

----

 - [ ] Tests are included.

----

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -818,6 +818,7 @@
   </dependencyManagement>
 
   <build>
+    <defaultGoal>clean test install</defaultGoal>
     <!-- Run the enforcer plugin automatically at compile time -->
     <plugins>
       <plugin>


rvesse (approved):

afs (approved):

afs:
No harm but "clean install" runs "test" anyway (or at least in the default lifecycle it does).

https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html

kinow:
> No harm but "clean install" runs "test" anyway (or at least in the default lifecycle it does).
> 
> https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html

Yup. I just got used to detailing each task. I think in Commons some projects use `clean verify` instead, which shouldn't go as far as installing the archive in the local repo. Feel free to update those goals if you'd like.

We started using it in Commons to also standardize the build. This way in GH Actions/Travis/Jenkins/etc we can simply `mvn`, without worrying about which targets each project uses (some projects may require running specific tasks, like rat/jacoco/javadoc/etc separately).

PR #1388, opened 2022-06-15
GH-1387 Improved custom service executor extension system

### LINKED ISSUE #1387: Improved custom service executor extension system
### Version

4.6.0-SNAPSHOT

### Suggestion

This is the issue for factoring out the extension system itself from #1314 so it can be reviewed in isolation.

Link to PR: https://github.com/apache/jena/pull/1388 (see it for detailed comments)

Yes

### LINKED ISSUE #1399: SERVICE fails in empty context.
### Version

4.6.0-SNAPSHOT

### What happened?

Executing a query involving `SERVICE` with an empty context does not work because there is no default service execution.

@Aklakan - for your improvements in this area, could you please ensure that it executes when the context does not have a setting for `ARQConstants.registryServiceExecutors`.

```shell
public static void main(String...args) {
        FusekiLogging.setLogging();
        FusekiServer server = FusekiServer.create()
                .port(0)
                .add("/ds",  DatasetGraphFactory.empty())
                .start();
        String SERVICE = "http://localhost:"+server.getPort()+"/ds";
        Node expected = NodeFactory.createLiteral("28181", XSDDatatype.XSDinteger);
        String queryString = "SELECT * { SERVICE <"+SERVICE+"> { VALUES ?x { 28181 } } }";
        // Empty context
        Context cxt = new Context();
        //Context cxt = ARQ.getContext().copy();
        RowSet rs = QueryExec.dataset(DatasetGraphFactory.empty())
                .query(queryString)
                .context(cxt)
                .select()
                .materialize();
        assertTrue(rs.hasNext());
        Node n = rs.next().get("x");
        assertEquals(expected, n);
    }
```

_No response_

Aklakan:
GitHub issue resolved #1387.
This fixes #1399.

Pull request Description: This PR adds the following changes to the custom service executor plugin system:

Improvements:

* **chaining** There is now `ChainingServiceExecutor` which allows for transparently forwarding a request to the next `ChainingServiceExecutor` instance in the registry. This way a custom service executor can modify the request and have it processed by the remainder of the chain.
* **bulk processing** The plugin system introduced with jena 4.5.0 only supports lookups with individual bindings. This PR adds support for plugins that want to process bindings in bulk. The main difference in the APIs is whether the `createExecution` method accepts a single `Binding` or a `QueryIterator`. There is now a `ServiceExecutorRegistryBulk` which by default has an registration that delegates to the non-bulk one.

Breaking changes:

* `ServiceExecutorRegistry.getFactories()` now returns a `List<ChainingServiceExecutor>` because this is its internal storage; previously it was `List<ServiceExecutorFactory>`. I am afraid re-establishing compatibility in this regard would require a completely new registry with a new context attribute.

Deprecations:
* Deprecated `ServiceExecutorFactory` in favor of `ServiceExecutor` (ServiceExecutorFactory extends from the latter). Use of the old code delegates to the new one.
    * In the new code the method that creates a `ServiceExecution` is now called `createExecution` (rather than `createExecutor`)
* ~Deprecated `ServiceExecutorRegistry.add` in favor of `prepend`. Both methods add executors to the beginning of the registry's list so they are considered before the default behavior.~ QueryEngineFactory also uses `add` to prepend; so `add` is consistent with existing naming.
* Legacy add method for `ServiceExecutorFactory` exists and wraps it as a `ChainingServiceExecutor`.
* Legacy remove method for `ServiceExecutorFactory` finds previously wrapped elements..
* Deprecated `ServiceExecution` because it doesn't add anything over `QueryIterator`.

----

 - [x] Javadoc exists and is up-to-data
 - [x] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/QueryIterator.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/QueryIterator.java
@@ -18,15 +18,13 @@
 
 package org.apache.jena.sparql.engine;
 
-import java.util.Iterator ;
-
-import org.apache.jena.atlas.lib.Closeable ;
+import org.apache.jena.atlas.iterator.IteratorCloseable;
 import org.apache.jena.sparql.engine.binding.Binding ;
 import org.apache.jena.sparql.util.PrintSerializable ;
 
 /** Root of query iterators in ARQ. */
 
-public interface QueryIterator extends Closeable, Iterator<Binding>, PrintSerializable
+public interface QueryIterator extends IteratorCloseable<Binding>, PrintSerializable
 {
     /** Get next binding */
     public Binding nextBinding() ;
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/main/iterator/QueryIterService.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/main/iterator/QueryIterService.java
@@ -31,12 +31,15 @@ import org.apache.jena.sparql.engine.iterator.QueryIterRepeatApply;
 import org.apache.jena.sparql.engine.iterator.QueryIterSingleton;
 import org.apache.jena.sparql.engine.main.QC ;
 import org.apache.jena.sparql.exec.http.Service;
-import org.apache.jena.sparql.service.ServiceExecution;
-import org.apache.jena.sparql.service.ServiceExecutorFactory;
 import org.apache.jena.sparql.service.ServiceExecutorRegistry;
+import org.apache.jena.sparql.service.single.ChainingServiceExecutor;
 import org.apache.jena.sparql.util.Context;
 
-
+/**
+ * This class continues to exist for compatibility with legacy service extensions.
+ * New code should register extensions at a {@link ServiceExecutorRegistry}.
+ */
+@Deprecated(since = "4.6.0")
 public class QueryIterService extends QueryIterRepeatApply
 {
     protected OpService opService ;
@@ -59,20 +62,21 @@ public class QueryIterService extends QueryIterRepeatApply
         ExecutionContext execCxt = getExecContext();
         Context cxt = execCxt.getContext();
         ServiceExecutorRegistry registry = ServiceExecutorRegistry.get(cxt);
-        ServiceExecution svcExec = null;
+        QueryIterator svcExec = null;
         OpService substitutedOp = (OpService)QC.substitute(opService, outerBinding);
 
         try {
             // ---- Find handler
             if ( registry != null ) {
-                for ( ServiceExecutorFactory factory : registry.getFactories() ) {
+                // FIXME This needs to be updated for chainable executors
+                for ( ChainingServiceExecutor factory : registry.getSingleChain() ) {
                     // Internal consistency check
                     if ( factory == null ) {
                         Log.warn(this, "SERVICE <" + opService.getService().toString() + ">: Null item in custom ServiceExecutionRegistry");
                         continue;
                     }
 
-                    svcExec = factory.createExecutor(substitutedOp, opService, outerBinding, execCxt);
+                    svcExec = factory.createExecution(substitutedOp, opService, outerBinding, execCxt, null);
                     if ( svcExec != null )
                         break;
                 }
@@ -81,8 +85,7 @@ public class QueryIterService extends QueryIterRepeatApply
             // ---- Execute
             if ( svcExec == null )
                 throw new QueryExecException("No SERVICE handler");
-            QueryIterator qIter = svcExec.exec();
-            qIter = QueryIter.makeTracked(qIter, getExecContext());
+            QueryIterator qIter = QueryIter.makeTracked(svcExec, getExecContext());
             // Need to put the outerBinding as parent to every binding of the service call.
             // There should be no variables in common because of the OpSubstitute.substitute
             return new QueryIterCommonParent(qIter, outerBinding, getExecContext());
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/main/OpExecutor.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/main/OpExecutor.java
@@ -45,6 +45,7 @@ import org.apache.jena.sparql.expr.Expr ;
 import org.apache.jena.sparql.expr.ExprList ;
 import org.apache.jena.sparql.procedure.ProcEval ;
 import org.apache.jena.sparql.procedure.Procedure ;
+import org.apache.jena.sparql.service.ServiceExecutorRegistry;
 
 /**
  * Turn an Op expression into an execution of QueryIterators.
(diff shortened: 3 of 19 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterRepeatApply.java
```diff
@@ -18,99 +18,26 @@
...
-
-        if ( getInput() == null )
-            return null;
-
-        if ( !getInput().hasNext() ) {
-            getInput().close();
-            return null;
-        }
-
+    protected QueryIterator nextStage(QueryIterator input) {
```
afs, line 38: Please explain what is going on here. 

If a change is needed, then having a superclass is not necessary. Just two `nextStage` methods with different signatures.

jena-arq/src/main/java/org/apache/jena/sparql/ARQConstants.java
```diff
@@ -292,6 +292,10 @@
     public static final Symbol registryServiceExecutors =
         SystemARQ.allocSymbol("registryServiceExecutors") ;
 
+    /** The service executor library registry key */
```
afs, line 295: Are two different keys necessary?

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterRepeatApply.java
Aklakan: QueryIterApply**Bulk**'s `nextStage(QueryIterator)` method flat-maps the input `QueryIterator` to an output `QueryIterator` until the input iterator is consumed - i.e. implementations of nextStage can "per stage" consume an arbitrary amount > 0 from the input iterator.

QueryIterApply implements `QueryIterApplyBulk.nextStage(QueryIterator)` such that it delegates to its own usual `QueryIterApply.nextStage(Binding)` method.

Note that `QueryIterApplyBulk` is also the base class for `QueryIterServiceBulk`. The latter only operates on the `QueryIterator` and therefore does not need a `nextStage(Binding)` method. I find this separation cleaner than having both `nextStage` methods on the same class together with an override that causes the binding-level one to no longer be called.

The only drawback I see with this separation is that reflection code that relied on `QueryIterApply.class.getDeclaredMethods()` would break.

jena-arq/src/main/java/org/apache/jena/sparql/ARQConstants.java
Aklakan: Very good question. Right now the bulk and non-bulk registries are two separate classes. I need to think about whether they could be combined into a single class.

Aklakan:
Just to clarify, my next step on this PR is to update and write javadoc; especially for what some of the classes are for. This should make reviewing easier.

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterRepeatApply.java
```diff
@@ -18,99 +18,26 @@
...
-        }
-    }
-
-    protected QueryIterator getCurrentStage() {
-        return currentStage;
     }
 
     protected abstract QueryIterator nextStage(Binding binding);
 
+    /** The caller guarantees that input has at least one more item */
```
afs, line 36: That is not the contract of `hasNextBinding`.

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterRepeatApply.java
afs: > I find this separation cleaner

My concern is that this is making the bulk case "normal" when in fact it is special to your service work.

Aklakan:
May latest attempt at simplifying things:

* Reverted QueryIterRepeatApply (it used inside of ServiceExecutorBulkToSingle which acts as the bridge between bulk and non-bulk)
* Combined bulk and non-bulk registries
* Deprecated ServiceExecution in favor using QueryIterator directly
* OpExecutor now directly calls the plugin system
* This way QueryIterService is actually not used anymore - service execution via the registry eventually by default ends up with a QueryIterRepeatApply. The still existing QueryIterService only uses the non-bulk part of the registry.

jena-arq/src/main/java/org/apache/jena/sparql/engine/QueryIterator.java
```diff
@@ -18,15 +18,13 @@
...
-
-import org.apache.jena.atlas.lib.Closeable ;
 import org.apache.jena.sparql.engine.binding.Binding ;
 import org.apache.jena.sparql.util.PrintSerializable ;
+import org.apache.jena.util.iterator.ClosableIterator;
 
 /** Root of query iterators in ARQ. */
 
-public interface QueryIterator extends Closeable, Iterator<Binding>, PrintSerializable
+public interface QueryIterator extends ClosableIterator<Binding>, PrintSerializable
```
afs, line 27: This change isn't necessary but if it is made, please use `IteratorCloseable`  See the comments in ClosableIterator which refer to the Model API and `ExtendedIterator` - let's keep `ClosableIterator` for jena-core.

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/iterator/QueryIterService.java
```diff
@@ -32,8 +32,8 @@
 import org.apache.jena.sparql.engine.main.QC ;
 import org.apache.jena.sparql.exec.http.Service;
 import org.apache.jena.sparql.service.ServiceExecution;
```
afs, line 34: This is now an unused import and can be removed.

afs:
Looks good - 2 small comments.

Then it's the checklist items : javadoc item and how you want to commits to look when merged to the codebase. e.g any squashing and and any use of "GH-1387: " to pick out key commits.

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/iterator/QueryIterService.java
Aklakan: I removed the import and added a deprecation annotation because QueryIterService is now no longer used but existing code may have sub-classed from it.

Aklakan:
I noticed that the overloaded `add` methods in ServiceExecutorRegistry would result in ambiguity with lambdas - which broke with (my) code based on the 4.2.0 api.
Before the naming gets frozen I now renamed the methods to:
```
add(ServiceExecutor) // Compatible with 4.2.0
addSingleLink(ChainingServiceExecutor) // New method; adds a link to the chain
addBulkLink(ChainingServiceExecutorBulk) // Also new
```
I updated javadoc and added a couple of ```@deprecated(since = "4.6.0")``` annotations.

afs (approved):

afs:
Are there any documentation changes for `jena-site`?

PR #1390, opened 2022-06-19
GH-1435 Missing implementations for RDFLinkAdapter

### LINKED ISSUE #1435: Missing implementations for RDFLinkAdapter
### Version

4.6.0-SNAPSHOT

### Feature

This is the issue for #1390.
It's about adding additional adapters for presenting RDFConnection-related objects as RDFLink-related ones.

Yes

Aklakan:
GitHub issue resolved #1435 
Pull request Description:

This PR proposes implementations of RDFLinkAdapter newQuery / newUpdate methods.
Attempting to use this methods on an instance created with `RDFLinkAdapter.adapt()` using a RDFConnection instance that is not backed by an RDFLink would raise UnsupportedOperationException.
A test case is provided ("TestRDFConnectionRewrapping")

For this purpose, the classes UpdateExecBuilderAdapter and QueryExecBuilderApdater have been introduced.
'adapt()' methods have been added and wired up.

There are two or three methods in QueryExecBuilder that presently cannot be delegated:
- getContext()
- initialTimeout()
- overallTimeout()

 The options I see are:
- extend QueryExecutionBuilder such that delegation becomes possible
- gracefully ignore
- raise UOE (UnsupportedOperationException)

----

 - [x] Tests are included.
 - [ ] (n/a) Documentation change and updates are provided for the [Apache Jena website](https://github.com/apache/jena-site/)

----


afs:
> RDFConnection instance that is not backed by an RDFLink

I'd like to understand where this arises. Since version 4.3, `RDFConnection` is a presentation and the machinery is all in `RDFLink`.

Aklakan:
In my case, its a mix of both legacy RDFConnection implementations from when the link layer did not exist - as well as custom decorations. For example, I was experimenting with a [getMetadata()](https://docs.oracle.com/javase/7/docs/api/java/sql/Connection.html#getMetaData()) method (analoguous to JDBC's connection) in order to allow for obtaining information about what the connection is connected to.


On top of that, once I get an RDFConnection from a 'data source' I sometimes wrap the connection with query transformations (sometimes paired with result set post processing).
This way the data source, query transformations and application logic don't have to know each other - at the cost of resulting in a bit of ugly wrapping but that's typically not really an issue - unless when the wrappers don't delegate properly.

```java
    public static RDFConnection wrapWithLinkDecorator(RDFConnection conn, Function<? super RDFLink, ? extends RDFLink> linkDecorator) {
        RDFLink oldLink = RDFLinkAdapter.adapt(conn);
        RDFLink newLink = linkDecorator.apply(oldLink);
        RDFConnection result = RDFConnectionAdapter.adapt(newLink);
        return result;
    }
```

Yes, ideally RDFConnection would just be a presenter for an RDFLink but the link/conn machinery making that assumption and therefore only providing an incomplete wrapping was again an experience like catching a splinter - because there is not really a way to workaround besides adding those classes of the PR.

Aklakan:
I also streamlined null handling in the wrappers
Existing code would use both variants:
```java
// class RDFLinkAdapter
    @Override
    public Graph get() {
        Model m = conn.fetch();
        if ( m == null ) return null;
        return m.getGraph();
    }
```
as well as
```java
    @Override
    public DatasetGraph getDataset() {
        return conn.fetchDataset().asDatasetGraph(); // fails if fetchDataset returned null
    }
```

I now changed that to:
```java
    @Override
    public DatasetGraph getDataset() {
        Dataset ds = conn.fetchDataset();
        return ds == null ? null : ds.asDatasetGraph();
    }
```

To motivate this change: There is a bit of unclarity in getDataset()
- It can delegate to fetchDataset() which downloads the dataset
- It can return a dataset **view** over the remote endpoint (where the find method is implemented as sending SPARQL select queries)
- It can return a local (in-memory) dataset

So far I used getDataset() mainly to expose a local in memory datasets where polymorphic views via `Resource.as(class)` are most meaningful.
Conversely, such views are not really meningful over remote views (each method call results in a remote request) or if getDataset delegates to fetchDataset which then attempts to download all of DBpedia.

Depending on the use case it thus makes sense for getDataset to return null, raise a UOP, return the local copy or a view over remote requests. The contract does not state whether null or UOP should be used - RDFLinkAdapter.get() suggest null is valid!
In any case, I don't think that the adapters are the place to judge which is the 'correct' usage so that's why I added null delegation.

afs:
The "GH-xxxx:" should refer to the issue.

The "GH-yyyy:" in commits will be a link and should refer to the issue, or the PR is there isn't an issue.

jena-arq/src/main/java/org/apache/jena/sparql/exec/QueryExecBuilderAdapter.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Override
+    public QueryExecMod overallTimeout(long timeout, TimeUnit timeUnit) {
+        builder = builder.timeout(timeout, timeUnit);
+        return this;
+    }
+
+    @Override
+    public Context getContext() {
+        // throw new UnsupportedOperationException("QueryExecBuilderAdapter.getContext()");
```
afs, line 77: `UnsupportedOperationException` for now.

afs, line 77: `UnsupportedOperationException`. 

In general, I'm not sure the context handling of `ExecHTTPBuilder` is quite right with this alteration of the use pattern to be less about a single use-once builder object. Any sharing of mutable state is ... tricky.

jena-arq/src/main/java/org/apache/jena/sparql/exec/QueryExecBuilderAdapter.java
```diff
@@ -0,0 +1,141 @@
...
+        return result;
+    }
+
+    public QueryExecutionBuilder getExecBuilder() {
+        return builder;
+    }
+
+    @Override
+    public QueryExecMod initialTimeout(long timeout, TimeUnit timeUnit) {
+        // Gracefully ignore?
```
afs, line 64: Why not pass to the adapted object? It covers the local dataset case which may not be of interest to you but dif we are going to have this adapter, it should be as complete as possible.

afs (changes requested):
The "adapter" pattern works properly when adapting immutable objects. 

If mutable, then for example two "Execution" adapters over one "Exec" adapter interfere through shared state.

Looking at the code, I think the context handling isn't quite right because it starts by a copy of `ARQ.getContext()`, not allowing a completely separate setup. But something to look at later.

jena-arq/src/main/java/org/apache/jena/sparql/exec/QueryExecBuilderAdapter.java
Aklakan: What do you mean here concretely? Do you mean I should remove the unwrapping `((QueryExecutionBuilderAdapter)builder).getExecBuilder()` in `adapt()` or am I overlooking something else?

afs (approved):

PR #1392, opened 2022-06-19
GH-1391 Support for context-local QueryEngineRegistries

Aklakan:
Pull request Description:
While the function, property function and service executor registries can be set per context, this is not possible with QueryEngineRegistry. So far engines thus have to be registered globally which can be brittle when a specific engine is wanted to evaluate a query.

For example, the default query engine always calls DatasetGraph.listGraphNodes for a pattern `GRAPH ?g { ?s ?p ?o }` - even if ?s, ?p or ?o were concrete and an index could be used instead.

The main use case is thus to allow for specifically using the quad-based engine (QueryEngineQuad + OpExecutorQuad) so that evaluation can be handled in the DatasetGraphs.find() method.
With a single global registry it is hard to guarantee that the desired engine will be chosen in the presence of other query engine factory plugins.

This PR makes it possible to override the global QueryEngineRegistry with a context-local one.

A test case for this use case using Jena's own quad-based engine is provided.

----

 - [x] Tests are included.

----


Aklakan:
- I just realized the method `getOrDefault` should be renamed to `chooseRegistry` - the same concept already exists with e.g.
`public static PropertyFunctionRegistry chooseRegistry(Context context)` (will update later today)

afs:
GH-1391 isn't the issue for this PR nor is it this PR.

Could you create an issue (or press send on a draft!) and label the commit please?

All this makes looking back over git history in several months/years time easier.

afs:
(sorry - misclick)

afs:
Instead of using `chooseRegistry` in `QueryExecutionFactory`, `QueryExecDatasetBuilder` and `QueryEngineFactoryWrapper`, I think we can give a simpler view by replacing `QueryEngineRegistry.get().find` with `QueryEngineRegistry.findFactory` (currently, the definition of `findFactory`) and changing `findFactory` to be `chooseRegistry(context).find(...)`.

`QueryEngineFactoryWrapper` can be left alone.
`QueryExecutionFactory` and `QueryExecDatasetBuilder` needs teh change `get().find` to `findFactory`.

afs:
The existing way to have a different query engine is to add a dataset-graph wrapper, and add an factory to the lookup chain.
That is very old and not really that useful.

Probably, `DatasetGraph` should have an `exec(op, context)` but that's not a change to make lightly. So this PR is a useful addition and exploration of the ideal design.

Aklakan:
> Instead of using `chooseRegistry` in `QueryExecutionFactory`, `QueryExecDatasetBuilder` and `QueryEngineFactoryWrapper`, I think we can give a simpler view by replacing `QueryEngineRegistry.get().find` with `QueryEngineRegistry.findFactory` (currently, the definition of `findFactory`) and changing `findFactory` to be `chooseRegistry(context).find(...)`.

Good point, makes it simpler.
 
> `QueryEngineFactoryWrapper` can be left alone. `QueryExecutionFactory` and `QueryExecDatasetBuilder` needs teh change `get().find` to `findFactory`.

Done and squashed.

PR #1404, opened 2022-06-24
JENA-2333 fix: sort langString first by value then by lang tags

litvinovg:
Pull request Description:
Pull request to fix  JENA-2333 issue, fixes comparison order for node values with language tags.

----
 - [x] Tests are included.
 - [JIRA Issue](https://issues.apache.org/jira/browse/JENA-2333)

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/expr/NodeValue.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/expr/NodeValue.java
@@ -734,10 +734,8 @@ public abstract class NodeValue extends ExprNode
                 // Two literals, both with language tags.
                 Node node1 = nv1.asNode() ;
                 Node node2 = nv2.asNode() ;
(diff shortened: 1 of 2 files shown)


afs:
Thanks - that look right.

It may take a while to fully dig into this because `NodeUtils.compareRDFTerms` looks odd as well -- this would relate to plain string and lang string comparison and lead to a related but different failure of transitive compare relationships.

afs:
This would fix the presenting issue but it creates more change because the current (4.5.0) path of sorting are "lang before lexicial".

More general discussion on the JIRA ticket. [JENA-2333](https://issues.apache.org/jira/browse/JENA-2333).

There is a SPARQL 1.2 discussion that cover this area https://github.com/w3c/sparql-12/issues/88.

afs:
Thanks for the PR - it's prompted a deeper look into ordering items that are of different categories.

PR #1406, opened 2022-06-27
JENA-2333: Sorting simple strings and language string.

afs:
This address the report of JENA-2333

It changes the ordering on terms of differing categories (value spaces). e.g. string and integers so that the string (simple and language) are before other datatypes. Previously (up to Jena 4.5.0) the comparison of differing categories was by lexical form first.

But two languages, compared by NodeValue, the value-sensitive comparison, was by language tag. This leads to an unstable sort (it is not a total ordering because the condition `A<B` and `B<C` implies `A<C` is not always true. JENA-2333 respots a case where this unstable sort leads to ant exception from `Arrays.sort` from the JDK.

This PR is an alternative to #1404.

----

 - [x] Tests are included.


afs:
This is "draft" at the moment. Currently, some tests and scripted tests fail and unalike datatype comparison is by old-style lexical form - switching to datatype-before-lexical makes sense design-wise.

afs:
[Email to dev@ about this change](https://lists.apache.org/thread/c65sbt9fvc005538xjplp34nqdzgb18t).

afs:
The tests have now updated for new sorting order.

PR #1438, opened 2022-07-13
GH-1432: Fix for adding text data after TDB2 compaction

afs:
GitHub issue resolved #1432

Pull request Description:

jena-text integrates with TDB2 by adding a `TransactionalComponent` and becoming part of the TDB/DBOW transaction system.

The bug is that the added `TransactionalComponent` is in the storage database, not the switchable database. The storage database is rewritten and rebuilt by a compaction and the added component gets lost.

----

 - [x] Tests are included.
 - [N/A] Documentation change and updates are provided for the [Apache Jena website](https://github.com/apache/jena-site/)
 - [x] Commits have been squashed to remove intermediate development commit messages.
 - [x] Key commit messages start with the issue number (GH-xxxx or JENA-xxxx)

By submitting this pull request, I acknowledge that I am making a contribution to the Apache Software Foundation under the terms and conditions of the [Contributor's Agreement](https://www.apache.org/licenses/contributor-agreements.html).

----

See the [Apache Jena "Contributing" guide](https://github.com/apache/jena/blob/main/CONTRIBUTING.md).

PR #1444, opened 2022-07-15
Improve handling of external transactional components

afs:
This is work mentioned in https://github.com/apache/jena/pull/1438#issuecomment-1183087146

The DBOE transaction system allows for transactional components from other sub-systems. An example of this is jena-text, where text index updates are coordinated with the RDF database updates when the database is TDB2.

The previous approach was specific to jena-text. This improves on that by making it a proper feature.

----


rvesse (commented):
Massively behind on reading through Jena PRs but this looks like a really nice solution to the problem

PR #1456, opened 2022-07-24
GH-1252: Hold up all possible write transactions while compacting.

### LINKED ISSUE #1252: Fuseki compact through java code
I'm on version 4.3.2 through docker secoresearch/fuseki and have a dataset with name '/fedora'. The database is growing like crazy after deletes and posts of the same element.

So I have understood I need to proactively compact the database, like described here: https://jena.apache.org/documentation/tdb2/tdb2_admin.html#compaction

`DatabaseMgr.compact(dataSet.asDatasetGraph(), true);`

But how to I get to the first parameter, `dataSet`?
I have a RDFConnection object initialized with username/password.

### LINKED ISSUE #1457: TIM: promoting while another writer is attempting to start may deadlock.
### Version

4.5.0

### What happened?

Restoring tests `promote_snapshot_06` and `promote_readCommitted_06` showed a TIM deadlock in `TestDatasetGraphInMemoryPromote`. The deadlock is timing sensitive.

_No response_

Yes

afs:
GitHub issue resolved #1252.
This fixes #1457.

Pull request Description:

This is a fix for  #1252 by having writers wait on a lock in the switchable container. Previous, they waited in the transaction coordinator, which is per storage, and changes during compaction. The issue #1252 is a bug when a write starts while compaction is going on. It was blocked (held on a lock) but inside the transaction coordinator for the storage about to be compacted.

#1457 found when some tests were restored.

----

----

### DIFF
--- a/jena-db/jena-dboe-storage/src/main/java/org/apache/jena/dboe/storage/system/DatasetGraphTxnCtl.java
+++ b/jena-db/jena-dboe-storage/src/main/java/org/apache/jena/dboe/storage/system/DatasetGraphTxnCtl.java
@@ -1,0 +1,351 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.jena.dboe.storage.system;
+
+import java.util.Objects;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import org.apache.jena.dboe.transaction.txn.TransactionException;
+import org.apache.jena.dboe.transaction.txn.TransactionalSystemControl;
+import org.apache.jena.query.ReadWrite;
+import org.apache.jena.query.TxnType;
+import org.apache.jena.sparql.core.DatasetGraph;
+import org.apache.jena.sparql.core.DatasetGraphWrapper;
+import org.apache.jena.sparql.util.Context;
+
+/**
+ * DatasetGraph wrapper controls entry and exit of transactions.
+ * <ul>
+ * <li>Exclusive access (no transactions active)
+ * <li>Read only database - No possible writers (write and promotable transactions)
+ * </ul>
+ */
+
+public class DatasetGraphTxnCtl extends DatasetGraphWrapper implements TransactionalSystemControl {
+
+    // ---- Controls
+
+    // Multiple state X or single state Y.
+    // Do not confuse with read/write transactions. We need a "one exclusive, or many
+    // other" lock which happens to be called {@code ReadWriteLock}.
+
+    // All transactions need "read" state X through out their lifetime.
+    // The "write" state Y is used for exclusive mode.
+    private ReadWriteLock exclusivitylock = new ReentrantReadWriteLock();
+
+    // Lock to guarantee only readers are present.
+    // Writers and promote transaction need to take a lock on entry.
+    // This is not reentrant.
+    private ReadWriteLock writeableDatabase = new ReentrantReadWriteLock();
+
+    // Better Lock naming?
+    //  MultipleOrSingle (MOS) == MRSW
+    //  MultipleAndSingle (MAS) == MR+SW
+
+    public DatasetGraphTxnCtl(DatasetGraph dsg) {
+        super(dsg);
+    }
+
+    public DatasetGraphTxnCtl(DatasetGraph dsg, Context context) {
+        super(dsg, context);
+    }
+
+    @Override
+    public void begin(TxnType txnType) {
+        Objects.nonNull(txnType);
+        checkNotActive();
+        enterTransaction(txnType, true);
+        super.begin(txnType);
+    }
+
+    final
+    protected void checkNotActive() {
+        if ( isInTransaction() )
+            throw new TransactionException(label("Currently in an active transaction"));
+    }
+
+    private String label(String msg) {
+        // Optional labelling
+        return msg;
+    }
+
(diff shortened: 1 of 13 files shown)


afs:
Still to do: review the changes after reflecting on the design.

jena-db/jena-dboe-storage/src/main/java/org/apache/jena/dboe/storage/system/DatasetGraphTxnCtl.java
```diff
@@ -0,0 +1,353 @@
...
+        endMultiMode(exclusivitylock);
+    }
+
+    /**
+     * Enter exclusive mode; block if necessary. There are no active transactions on
+     * return; new transactions will be held up in 'begin'. Return to normal (release
+     * waiting transactions, allow new transactions) with
+     * {@link #finishExclusiveMode}.
+     * <p>
+     * Do not call inside an existing transaction.
```
kinow, line 267: Isn't there a way to the code to tell whether it's running inside a transaction at this point? I think, if I recall correctly, I've seen some code doing that in.. JPA? Some Spring transaction code? But I am not sure how deep this code is in the execution context, nor if it'd be possible to expose the transaction at this point. Just curious :grimacing:

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/sys/DatabaseOps.java
```diff
@@ -258,46 +252,74 @@ private static void compact(DatasetGraphSwitchable container, Location loc1, Loc
...
+            // Old readers continue on db1.
+
         });
 
-        // New database running.
 
+        // This switches off the source database.
+        // It waits until all transactions (readers) have finished.
+        // This call is not undone.
+        // Database1 is no longer in use.
```
kinow, line 315: I think at this point we have two config files, and two databases. Are these left as-is, or is there a process to clean them up later?

kinow (commented):
Found a couple typos, also left some comments, but no blockers. More just I trying to learn a little more about this part of the code :grimacing: Thanks!

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/sys/DatabaseOps.java
afs: There's an option to delete now in the outer `compact(DatasetGraphSwitchable container, boolean shouldDeleteOld)` and Fuseki exposes this with `?deleteOld=true`. In Fuseki compaction is async tot he request - it's a background task, because it may take a while is held up by waiting for the database to be writer-free.

Another choice is to archive the old database and use as a snapshot/backup. It's fully functional and could be restored, or run elsewhere.

jena-db/jena-dboe-storage/src/main/java/org/apache/jena/dboe/storage/system/DatasetGraphTxnCtl.java
afs: In doing this bug fix, I scaled back from doing a large rework of transaction control. Not a good time just before a release!

There is `isInTransaction()` but at this point the real transaction which is on the storage database isn't started. I'm not sure the sequencing works. `DatasetGraphCxnCtl` is only controlling entry and exit to the switchable. The real transaction is in the storage `DatasetGraphTDB`. I'd like to split `TransactionCcordinator` into parts, the controls and the transaction mechanism itself.  I copied a more general `DatasetGraphCxnCtl` to a scratch area for later; it was getting too big a change just for the bug.

This is very low level messing around with the internals of the transaction system. An exception isn't great as it will naturally propagate out without other clearup. Maybe the javadoc should say "Don't call this operation" :smile:.

`isInTransaction` uses a ThreadLocal. ThreadLocals are easy to get wrong and slowly leak memory in a long running process (you have to explicitly remove them if threads are being created and which finish).

afs:
> More just I trying to learn a little more about this part of the code

Brave! In terms of the jumping in at the deep end, this also has whirlpools and sharks. The thread coordination is complicated. I have to set aside at least half-days to do much work on this area to get into the mindset to make changes.

jena-db/jena-dboe-storage/src/main/java/org/apache/jena/dboe/storage/system/DatasetGraphTxnCtl.java
kinow: >In doing this bug fix, I scaled back from doing a large rework of transaction control. Not a good time just before a release!
> (…) it was getting too big a change just for the bug.

:+1: sounds reasonable!

>This is very low level messing around with the internals of the transaction system. An exception isn't great as it will naturally propagate out without other clearup. Maybe the javadoc should say "Don't call this operation" smile.

I don't believe we will get many users navigating this deep into the code. We can leave the comment as-is now. Looks like you have a plan for improvements that may alter parts of this code too—and will probably change the comments anyway.

>isInTransaction uses a ThreadLocal. ThreadLocals are easy to get wrong and slowly leak memory in a long running process (you have to explicitly remove them if threads are being created and which finish).

:+1: good to be reminded of that! Every time I switch from JS or Python to Java I find that I forget things about parts that Java do not have in common with these languages (like threads, jee, memory management, etc) :bow:

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/sys/DatabaseOps.java
kinow: Thanks!

>it's a background task, because it may take a while is held up by waiting for the database to be writer-free.

Ah, I think I saw some comments in the mailing list about compaction running in the background. Do you know if the backup appears in the Fuseki UI, under Tasks?

kinow:
Yeah, I realized I would need a lot longer to really understand the code. But it still doesn't look too bad to understand what's going on locally (I think the hard bit is telling how these parts are used elsewhere :sweat_smile: ).

Thanks Andy!

jena-db/jena-dboe-storage/src/main/java/org/apache/jena/dboe/storage/system/DatasetGraphTxnCtl.java
afs: It's a nuisance these operations have to be public (short of JPMS modules). Really, it would be better with "friends of" declarations to span boundaries without being completely public. But the transaction system is not specific to TDB2 and so it's in a separate package and maven module.

jena-db/jena-tdb2/src/main/java/org/apache/jena/tdb2/sys/DatabaseOps.java
kinow: >Do you know if the backup appears in the Fuseki UI, under Tasks?

Ah, it does! Great!

image

kinow (approved):

PR #1458, opened 2022-07-27
Accumulated small changes

afs:
Comment fixes, supply "toString" for TDB1 abd TDB2, one protect against bad argument.

Preparing for release!

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/web/HttpNames.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/web/HttpNames.java
@@ -20,8 +20,9 @@ package org.apache.jena.riot.web;
 
 public class HttpNames
 {
-    // Switch to org.apache.http.HttpHeaders
-    // or org.apache.jena.ext.com.google.common.net
+    // See also Guava com.google.common.net.HttpHeaders
+    // (org.apache.jena.ext.com.google.common.net.HttpHeaders)
+
     public static final String hAccept              = "Accept" ;
     public static final String hAcceptCharset       = "Accept-Charset" ;
     public static final String hAcceptRanges        = "Accept-Ranges" ;
--- a/jena-base/src/main/java/org/apache/jena/atlas/lib/FileOps.java
+++ b/jena-base/src/main/java/org/apache/jena/atlas/lib/FileOps.java
@@ -35,7 +35,7 @@ public class FileOps {
 
     /**
      * Delete a file
-     * 
+     *
      * @param filename
      */
     public static void delete(String filename) {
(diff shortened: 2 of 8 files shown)


jena-arq/src/main/java/org/apache/jena/riot/web/HttpNames.java
```diff
@@ -20,8 +20,8 @@
 
 public class HttpNames
 {
-    // Switch to org.apache.http.HttpHeaders
-    // or org.apache.jena.ext.com.google.common.net
+    // See also Guava com.google.common.net.HttpHeaders
```
ajs6f, line 23: Maybe a little weird to put a Javadoc on a field, but maybe:
```
/**
* @see org.apache.jena.ext.com.google.common.net.HttpHeaders
```
?

jena-base/src/main/java/org/apache/jena/atlas/lib/FileOps.java
```diff
@@ -66,9 +66,11 @@ public static void delete(File f, boolean reportExistsAfter) {
     /** Delete all files in a directory */
     public static void clearDirectory(String dir) {
         File d = new File(dir) ;
-        for ( File f : d.listFiles() ) {
-            if ( f.isFile() )
-                delete(f, false) ;
+        if ( d.exists() && d.isDirectory() ) {
```
ajs6f, line 69: [isDirectory()](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#isDirectory--) actually seems to test for existence by itself, if I read "Return: true if and only if the file denoted by this abstract pathname exists and is a directory; false otherwise" correctly.

ajs6f (approved):
+1, just two super-minor thoughts. I'm trying to get back into the habit of reading and reviewing, really I am!

jena-arq/src/main/java/org/apache/jena/riot/web/HttpNames.java
afs: It's really an implementation note for the class. New line spacer added.

jena-base/src/main/java/org/apache/jena/atlas/lib/FileOps.java
afs: Thanks! Changed.

jena-base/src/main/java/org/apache/jena/atlas/lib/FileOps.java
```diff
@@ -63,12 +63,19 @@ public static void delete(File f, boolean reportExistsAfter) {
...
+     * @param dir
+     */
     public static void clearDirectory(String dir) {
         File d = new File(dir) ;
-        for ( File f : d.listFiles() ) {
-            if ( f.isFile() )
-                delete(f, false) ;
+        if ( d.isDirectory() ) {
+            for ( File f : d.listFiles() ) {
+                if ( f.isFile() )
```
rvesse, line 76: I take it this never needs to recurse into further levels of directory?

afs: This one only deletes files. There's a separate `clearAll` for recursive deletion.

It gets used mostly in testing (so, an expected controlled environment) and limits the damage in case things aren't as expected.

PR #1475, opened 2022-08-09
Improved implementation of PrefixMapStd

### LINKED ISSUE #1474: PrefixMapStd is very slow for lookups that 'miss'
### Version

4.6.0-SNAPSHOT

### What happened?

Another follow up on https://issues.apache.org/jira/browse/JENA-2309
I investigated the issue of slow PrefixMapStd further and it turns out that when there are 'hits' using it's fast track approach its very fast - however on 'misses' it gets extremely slow because then it falls back to iteration of all prefixes in the map.

Using a small benchmark runner (part of the PR) shows the the PrefixMap can be easily brought to require dozens of seconds for lookups that can be accomplished in a split second.
For that I created a new PrefixMap intended as a replacement that combines the best of both worlds: The fast track of PrefixMapStd and the trie backing (together with a guava cache) of the removed FastAbbreviatingPrefixMap.

Approach 0 = Jena's original PrefixMapStd:
```
Run 1 with base <http://example.org/> and separator / using approach 0: Static IRI lookups took 0.082 seconds
Run 1 with base <http://example.org/> and separator / using approach 0: Dynamic IRI lookups took 33.443 seconds
Run 1 with base <urn:foo:bar:> and separator : using approach 0: Static IRI lookups took 25.482 seconds
Run 1 with base <urn:foo:bar:> and separator : using approach 0: Dynamic IRI lookups took 31.088 seconds
```

Approach 1 = my optimized PrefixMap:
```
Run 1 with base <http://example.org/> and separator / using approach 1: Static IRI lookups took 0.070 seconds
Run 1 with base <http://example.org/> and separator / using approach 1: Dynamic IRI lookups took 0.401 seconds
Run 1 with base <urn:foo:bar:> and separator : using approach 1: Static IRI lookups took 0.099 seconds
Run 1 with base <urn:foo:bar:> and separator : using approach 1: Dynamic IRI lookups took 0.433 seconds
```

_No response_

Yes

Aklakan:
GitHub issue resolved #1474 

Reimplementation of PrefixMapStd to combine "fast-track" with trie backing.

----
 - [x] Tests are included. (All existing tests apply - no new ones created)

----


Aklakan:
This PR is ready for review.
Longest-prefix-lookup-cache invalidation now only happens if there is an actual change; adding duplicates or removing non-existing entries do not trigger invalidation.
Also, I wasn't sure about the thread-safety of PrefixMapStd - probably its better to have it so I added RWL locking.

Is there a place where the utility methods `calcWithLock(Lock, Supplier)` and `runWithLock(Lock, Runnable)` could be publicly added - or maybe they already exist in one of the dependencies?

Aklakan:
I converted back to draft because I have some pending updates which I need to benchmark against the well-working parts (i.e. IRIs with / or #) of the original PrefixMapStd first in order to determine whether performance-wise it would make sense to include them.

The general idea is to have PrefixMap implementation that auto-adapts to a given workload - such as parsing lots of queries without the overhead of updating reverse-lookup structures as well as updating them on-demand upon writing out RDF.

In essence the direction I am investigating is about building/updating the reverse-lookup (iri-to-prefix) structures lazily (upon abbreviating). This would buffer prefix-iri inserts/deletions similar to your BufferingPrefixMap; upon abbreviate only the delta is materialized into the reverse-lookup structures.

afs:
Another choice is to restrict to the basic case of prefix at the final "/", "#" and ":" (for URNs).  Only have the "fast path" abbreviate.

Do you have cases where abbreviation is not one of these?

If you are going for the complicated version,maybe the best way is to have a new PrefixMapCaching and leave PrefixMapStd.

Aklakan:
> Another choice is to restrict to the basic case of prefix at the final "/", "#" and ":" (for URNs). Only have the "fast path" abbreviate.

I think including ':' in the list and only using fast path (without resorting to scanning) would work efficiently and be sufficient for the vast majority of use cases. Without scanning, even relative IRIs (that do not contain any of the fast track chars) wouldn't cause problems.

> Do you have cases where abbreviation is not one of these?
Right now I only have some initial experiments where I abuse the prefix map as a poor-mans dictionary encoding in order to reduce the amount of bytes that need to be parsed in order to produce triples/quads. For this I am using trie-based lookups to encode the data (so IRIs can be split anywhere), but I have yet to evaluate whether this actually gives a noticeable performance boost.

afs:
@Aklakan -- is this PR still relevant?

PR #1478, opened 2022-08-12
GH-1477: Improved AsyncParser

### LINKED ISSUE #1477: Improved AsyncParser API
### Version

4.6.0-SNAPSHOT

### Issues
- Closing the iterator's returned by AsyncParser does not abort the parsing process. In fact, repeatedly abandoning iterators will cause parsing threads to silently pile up.
- AsyncParser's default chunk size of 100K tuples introduces a long delay unsuitable for content probing. Though the 100k chunk size gives me 5-10% better throughput than significantly lower values. 
- The EltStreamRDF is private. As mentioned in [JENA-2309](https://issues.apache.org/jira/browse/JENA-2309) those events would be useful in an hadoop/spark setting to scan for prefixes, thereby stopping the parser once only data is seen anymore.

### Improvements

PR https://github.com/apache/jena/pull/1478 adds the following improvements:
- Changed AsyncParser API to return IteratorCloseables whose close() method actually cancels parsing.
- Added a public E**v**tStreamRDF interface for the parsing events. The existing private E**l**tStreamRDF class remains as the internal data object. The naming is up for discussion :)
- Added a Builder that gives control over chunk and queue sizes. The builder can create "low-level" IteratorCloseables as well as java Streams. The latter allows for convenient use with  try-with-resources. The following snippet is from a new test case:
```java
try (Stream<Triple> s = AsyncParser.of(Channels.newInputStream(channel), Lang.TURTLE, null)
        .setChunkSize(100).streamTriples().limit(expectedLimit)) {
    // ...
}
```
- If a parser fails then all remaining parsers are still started with a destination in 'aborted state' in order for them to close their resources.

Yes

Aklakan:
GitHub issue resolved #1477

The API changes to AsyncParser are intended to be completely backwards-compatible.
I added a couple of additional tests that check the count of JVM threads and consumed bytes.

Overall the PR is ready for feedback such as whether the proposed changes are acceptable in the first place.

I also tested the integrating of this PR into our Spark setup to do the final real-world testing and and all is green.
I'll do a bit more experiments because it's just too easy to overlook some corner cases when fiddling with async + cancellation, but overall it looks good to me.

An example for a non-trivial setup now looks like:
```java
public class MyHadoopRecordReader {
     //  (Various logic)

    public Stream<Triple> parse(InputStream in) {
        return AsyncParser.of(in, lang, baseIri)
                .mutateSources(parser -> parser
                        .labelToNode(createLabelToNodeAsGivenOrRandom())
                        .resolver(newIRIxResolverAsGiven()))
                .setChunkSize(100)
                .streamTriples();
    }
}
```

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/system/AsyncParser.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/system/AsyncParser.java
@@ -20,25 +20,30 @@ package org.apache.jena.riot.system;
 
 import java.io.InputStream;
 import java.util.ArrayList;
-import java.util.Iterator;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
-import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
-import org.apache.jena.atlas.iterator.Iter;
+import org.apache.jena.atlas.iterator.IteratorCloseable;
 import org.apache.jena.atlas.iterator.IteratorSlotted;
 import org.apache.jena.atlas.lib.InternalErrorException;
 import org.apache.jena.atlas.logging.FmtLog;
 import org.apache.jena.graph.Node;
 import org.apache.jena.graph.Triple;
-import org.apache.jena.riot.*;
+import org.apache.jena.riot.Lang;
+import org.apache.jena.riot.RDFParser;
+import org.apache.jena.riot.RDFParserBuilder;
+import org.apache.jena.riot.RiotException;
+import org.apache.jena.riot.SysRIOT;
 import org.apache.jena.riot.lang.RiotParsers;
 import org.apache.jena.sparql.core.Quad;
+import org.apache.jena.util.iterator.ClosableIterator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -51,14 +56,31 @@ import org.slf4j.LoggerFactory;
  * <p>
  * There are overheads, so this is only beneficial in some situations. Delivery to
  * the StreamRDF has an initial latency while the first batch of work is accumulated.
+ * Using the {@link AsyncParserBuilder} gives control over the chunk size such that initial latency
+ * can be reduced at the cost of possibly decreasing the overall throughput.
+ * <p>
+ * Closing the returned {@link ClosableIterator}s terminates the parsing
+ * process and closes the involved resources.
  */
 public class AsyncParser {
 
-    private static Logger LOG = LoggerFactory.getLogger(AsyncParser.class);
-    private static int chunkSize = 100_000;
+    static final Logger LOG = LoggerFactory.getLogger(AsyncParser.class);
+    static final ErrorHandler dftErrorHandler = createDefaultErrorhandler(LOG);
+
+    static final int dftChunkSize = 100_000;
     // There is no point letting the parser get a long way ahead.
-    private static int queueSize = 10;
-    private static List<EltStreamRDF> END = List.of();
+    static final int dftQueueSize = 10;
+    static final List<EltStreamRDF> END = List.of();
+
+    private static final StreamRDF alwaysFailingStreamRdf = new StreamToElements(elt -> { throw new RuntimeException(new InterruptedException()); });
+
+    /* Destination states for handling concurrent abort */
+    /** Normal operation: Accept incoming data */
+    private static final int RUNNING  = 0;
+    /** Abort requested: The next chunk will trigger placing the poison on the queue and transition to STOPPED state */
+    private static final int ABORTING = 1;
+    /** Stopped (regardless whether due to normal or exceptional cause): Any further incoming data raises an exception */
+    private static final int STOPPED  = 2;
 
     private AsyncParser() {}
 
@@ -66,8 +88,8 @@ public class AsyncParser {
      * Function that reads a file or GETs a URL, parses the content on a separate
      * thread and sends the output to a StreamRDF on the callers thread.
      */
-    public static void asyncParse(String filesOrURL, StreamRDF output) {
-        asyncParse(List.of(filesOrURL), output);
+    public static void asyncParse(String fileOrURL, StreamRDF output) {
+        AsyncParser.of(fileOrURL).asyncParseSources(output);
     }
 
     /**
@@ -75,10 +97,7 @@ public class AsyncParser {
      * and sends the output to a StreamRDF on the callers thread.
      */
     public static void asyncParse(List<String> filesOrURLs, StreamRDF output) {
-        Objects.requireNonNull(filesOrURLs);
-        Objects.requireNonNull(output);
-        LOG.debug("Parse: "+filesOrURLs);
-        asyncParseSources(urlsToSource(filesOrURLs), output);
+        AsyncParser.ofLocations(filesOrURLs).asyncParseSources(output);
     }
 
     /**
@@ -86,10 +105,7 @@ public class AsyncParser {
      * and sends the output to a StreamRDF on the callers thread.
      */
     public static void asyncParse(InputStream input, Lang lang, String baseURI, StreamRDF output) {
-        Objects.requireNonNull(input);
-        Objects.requireNonNull(lang);
-        Objects.requireNonNull(output);
-        asyncParseSources(inputStreamToSource(input, lang, baseURI), output);
+        AsyncParser.of(input, lang, baseURI).asyncParseSources(output);
     }
 
     /**
@@ -97,52 +113,17 @@ public class AsyncParser {
      * send the output to a StreamRDF on the callers thread.
      */
     public static void asyncParseSources(List<RDFParserBuilder> sources, StreamRDF output) {
-        // Async thread
-        Logger LOG1 = LOG;
-        // Receiver
-        Logger LOG2 = LOG;
-        BlockingQueue<List<EltStreamRDF>> queue = new ArrayBlockingQueue<>(queueSize);
-        startParserThread(LOG1, sources, queue);
-        receiver(LOG2, queue, output);
-    }
-
-    private static List<RDFParserBuilder> urlsToSource(List<String> filesOrURLs) {
-        return filesOrURLs.stream().map(uriOrFile-> RDFParser.source(uriOrFile)).collect(Collectors.toList());
-    }
-
-    private static List<RDFParserBuilder> inputStreamToSource(InputStream input, Lang lang, String baseURI) {
-        return List.of(RDFParser.source(input).lang(lang));
-    }
-
-    private static Function<EltStreamRDF, Triple> elt2Triple = x-> {
-        if ( x.exception != null )
-            throw x.exception;
-        if ( x.quad != null ) {
-            Node g = x.quad.getGraph();
-            if ( g == Quad.tripleInQuad || Quad.isDefaultGraph(g) )
-                return x.quad.asTriple();
-            return null;
-        }
-        return x.triple;
-    };
-
-    private static Function<EltStreamRDF, Quad> elt2Quad = x-> {
-        if ( x.exception != null )
-            throw x.exception;
-        if ( x.triple != null )
-            return Quad.create(Quad.defaultGraphIRI, x.triple);
-        return x.quad;
-    };
-
-    /** Pull parser - triples */
-    public static Iterator<Triple> asyncParseTriples(String fileOrURL) {
-        return asyncParseTriples(List.of(fileOrURL));
+        AsyncParser.ofSources(sources).asyncParseSources(output);
     }
 
     /** Pull parser - triples */
-    public static Iterator<Triple> asyncParseTriples(List<String> filesOrURLs) {
-        Iterator<EltStreamRDF> source = asyncParseIterator(urlsToSource(filesOrURLs));
-        return Iter.iter(source).map(elt2Triple).removeNulls();
+    public static IteratorCloseable<Triple> asyncParseTriples(String fileOrURL) {
+        return AsyncParser.of(fileOrURL).asyncParseTriples();
+    }
+
+    /** Pull parser - triples */
+    public static IteratorCloseable<Triple> asyncParseTriples(List<String> filesOrURLs) {
+        return AsyncParser.ofLocations(filesOrURLs).asyncParseTriples();
     }
 
     /**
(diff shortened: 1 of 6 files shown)


Aklakan:
I just added a fix where the last batch was no longer dispatched if a parse error was encountered.

I missed that error before because it did not occur when I was using small chunk sizes.
I updated our hadoop/spark code to use small chunk sizes when probing for record offsets, and large ones once when parsing complete fragments and the test suite is now green.

I can add another test case for the described situation, but functionality-wise I'd say this PR is complete.

afs:
Please add the test.
The async parser is critical to TDB2 loading.
I'm finding it difficult to know what things are ready or not.

It would help me when trying to jump between several different PRs and issues (include RDF Delta):

1. Use "GH-" (capitals) in commits - having a common style makes grepping the commits logs easier.
2. Tidy the commits to what is important when looking back at this

You may think these are trivial - but for me, juggling multiple PRs and issues, common forms help.

Aklakan:
I searched jena-site for AsyncParser and extended the section about it: https://github.com/apache/jena-site/pull/115

Today I tested the `tdb2.tdbloader` tool built with this PR on a 10mio triples dataset and it was functional.
Tomorrow I will do a final performance comparison with the 4.5.0 loader to make sure I didn't accidentally mess something up.
This should then conclude the PR.

(Also the commit message now make use of use capital `GH-` prefix and are squashed).

Aklakan:
The recent values I get by comparing the phased loading strategy between 4.6.0 and 4.7.0-SNAPSHOT with this PR I get:
I incorrectly mentioned 4.5.0 in my last post; 4.6.0 is the most recent version without this PR.

Iteration means loading a dataset once.
I verified that the right number of triples were loaded into the store.
```
# 20 iterations with 13 quads in 2 graphs (a new graph every 10 quads)
The intention is check that small data that doesn't fill queue works.
jena 4.6.0:           39,355s
jena 4.7.0-SNAPSHOT:  40,553s
(with a single quad it's essentially also 40s which is mostly JVM/Jena startup time rather than actual insertions)

# 20 iterations with 9.876.543 quads in 99 graphs (a new graph every 100K quads)
jena: 4.6.0:         4094s
jena 4.7.0-SNAPSHOT: 3990s

# 10 iterations with 9.876.543 triples
jena 4.6.0           1946s
jena 4.7.0-SNAPSHOT: 1901s
```
On another day I had reversed figures such that 4.6.0 was slightly faster; but now that I repeated the runs it seems that there isn't a significant difference.

So to me it looks all good.

afs:
There are too many factors (e.g. JIT effects) to get really stable performance figures without very long warm-up and timing runs. Even when the rest of the machine is "quiet", there are things going on outside the JVM.

I don't think anything changed between 4.5.0 and 4.6.1 that directly impacts this area. But seemingly unrelated changes elsewhere can affect when the JIT decides to kick-in and can make for visible differences.

Aklakan:
> There are too many factors [...]

Yes, the only change I made to the processing code is the addition of an AtomicBoolean that is checked only after every chunk for whether to terminate the parser - so there is an extra check now but the overhead should be neglectable (considering the amount of work it takes to turn bytes into elements).

A current limitation is, that with large chunk sizes only taking the first few elements still has to wait for the first chunk. Though now one can at least control the chunk size.

In principle the builder API could in a future PR be extended to improve this, such as by adding support for a custom Predicate<EltStreamRDF> that allows for stopping the parser and dispatching the current chunk immediately.

afs:
Not a worry about the cost for the benefit here because it is per chunk but ... `AtomicBoolean` is not so negligible. It's a non-local memory load and they take forever.

rvesse:
If it's just used as a flag to check whether to continue or not could probably use a `private volatile boolean` instead?  Still has to go to main memory but should have marginally less overhead than an `AtomicBoolean`

jena-arq/src/main/java/org/apache/jena/riot/system/AsyncParser.java
```diff
@@ -199,7 +244,7 @@ protected X moveToNext() {
                         return null;
                     }
                     return x;
-                } catch (InterruptedException e) {
+                } catch (@SuppressWarnings("unused") InterruptedException e) {
```
afs, line 247: This `SuppressWarnings` causes a warning (an "info").

"At least one of the problems in category 'unused' is not analysed due to a compiler option being ignored"

The rest of the code does not use "unused".

Please remove.

jena-arq/src/main/java/org/apache/jena/riot/system/AsyncParser.java
```diff
@@ -239,40 +278,143 @@ public void fatal(String message, long line, long col) {
...
+            }
+        };
+        EltStreamBatcher batcher = new EltStreamBatcher(destination, chunkSize);
+        StreamRDF generatorStream = new StreamToElements(batcher);
 
         // Parser thread
-        Runnable task = ()->{
-            batcher.startBatching();
+        Runnable task = () -> {
+            batcher.startBatching(); // No try/catch block currently needed because this is a no-op
```
afs, line 312: Personally, this seems odd. It is a no-op currently but implementation may change. What is the comment recording?

afs, line 312: startBatching and finishBatching no longer line up.
finishBatching is conditional. Can it ever fail to be called? Comments needed.

jena-arq/src/main/java/org/apache/jena/riot/system/AsyncParser.java
```diff
@@ -239,40 +278,143 @@ public void fatal(String message, long line, long col) {
...
+                                // Parse error.
+                                EltStreamRDFImpl elt = new EltStreamRDFImpl();
+                                elt.exception = ex;
+                                batcher.accept(elt);
+                            }
+                            errorEncountered = true;
+                        }
+                    } catch (Throwable cause) {
+                        if (errorEncountered) {
+                            LOG1.debug("Suppressed exception", cause);
```
afs, line 349: Style point: other debug calls have  `if ( LOG1.isDebugEnabled() )`.

Even if unnecessary, a common style helps.

afs (changes requested):

jena-arq/src/main/java/org/apache/jena/riot/system/EltStreamRDF.java
```diff
@@ -0,0 +1,54 @@
...
+
+package org.apache.jena.riot.system;
+
+import org.apache.jena.graph.Triple;
+import org.apache.jena.sparql.core.Quad;
+
+/**
+ * Interface for streamRDF parsing elements.
+ */
+public interface EltStreamRDF {
```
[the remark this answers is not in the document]
Aklakan: Back in [JENA-2309](https://issues.apache.org/jira/browse/JENA-2309) I mentioned the use case for getting individual elements from the parser in order to scan the head of a file for prefixes, thereby stopping once only a configurable amount of triples/quads is seen.

The intent of the interface is to provide a public API to access elements, whereas EltStreamRDFImpl is a rather AsyncParser specific implementation - for example, the parser directly sets the (package visible) attributes without going through any getters/setters or ctors. Not sure if this is actually faster but this is one way it makes sense to implement. The public API however shouldn't be tied to that.
Also, the interface abstracts whether the element type is conditionally determined based on the non-null attributes (as it is done now) or whether it is stored in a dedicated attribute.

jena-arq/src/main/java/org/apache/jena/riot/system/AsyncParser.java
Aklakan: If startBatching failed that it would have to be handled in a try/catch block. If it failed in the code as it is now then the resources wouldn't be freed; even the poison (end marker) would never be placed on the queue resulting in infinite wait.

Aklakan: If close() is invoked and isAborted is false then isAborted is atomically set to true. Furthermore, the close()-logic proceeds to clear the queue and place the poison on it.
I essence, if isAborted is true, then it means that the poison is (about to be) placed on the queue - and therefore finishBatching is no longer needed.

In that case finishBatching is suppressed because it would redundantly try to place another poison on the queue.

afs: But does it do any harm?

Thinking long term - clear code such as "start-finish" pairing, or "start-finish|abort" or "start-shutdown(current "finish")-finish" is useful to have for the next person (who may be yourself in a few PRs time!)

jena-arq/src/main/java/org/apache/jena/riot/system/EltStreamRDFImpl.java
```diff
@@ -0,0 +1,105 @@
...
+    @Override public Quad      getQuad()      { return quad; }
+    @Override public boolean   isPrefix()     { return prefix != null; }
+    @Override public String    getPrefix()    { return prefix; }
+    @Override public boolean   isBase()       { return prefix == null && iri != null; }
+    @Override public String    getIri()       { return iri; }
+    @Override public boolean   isException()  { return exception != null; }
+    @Override public Throwable getException() { return exception; }
+
+    @Override
+    public EltStreamRDFType getType() {
```
afs, line 62: This class is really a tagged union value (immutable).

Suggestions
* have a slot for the type and set the type when constructed. 
* make all fields final.
* use `Objects.requireNonNull` in constructors to catch problems early. Then there is a guarantee one and only one slot is non-null.

Its tedious writing out the duplication but it is not visible.

And maybe it will be a Java `record` at Java17 (which also means it can't be an interface - see the implementation-of-one comment).

jena-arq/src/main/java/org/apache/jena/riot/system/AsyncParserBuilder.java
```diff
@@ -0,0 +1,162 @@
...
+    public AsyncParserBuilder() {
+        this.chunkSize = dftChunkSize;
+        this.queueSize = dftQueueSize;
+        this.daemonMode = true;
+        this.sources = Collections.emptyList();
+    }
+
+    public AsyncParserBuilder(List<RDFParserBuilder> sources) {
+        this();
+        this.sources = sources;
```
afs, line 55: This can be isolated with `this.sources = List.copyOf(sources);` then changes to the argument have no effect or concurrency issue. WDYT?

jena-arq/src/main/java/org/apache/jena/riot/system/EltStreamRDFImpl.java
Aklakan: > which also means it can't extends an interface

I guess I have to try that out myself - so far I haven't. From the snippets I find on the web I would have thought something like this might work:
```
interface Foo { String name(); }
record Bar (String name) implements Foo;
```
Its not clear to me whether bean properties (methods with set/get/is prefixes) would still auto-map to the record fields.

In any case, I can add your suggestions which would make EltStreamRDFImpl "clean" to use publicly.
Though there is then still the question whether the ctors should be public or whether construction should only be possible with the named static methods such as `EltStreamRDF.triple(t)` - rather than `new EltStreamRDF(t)`.

The intent of the interface was to decouple client code from these implementation considerations so that the implementation can evolve independently - but I am fine with removing it if it turns out that one implementation can be effectively authorative.

jena-arq/src/main/java/org/apache/jena/riot/system/AsyncParserBuilder.java
Aklakan: Fine with me.

jena-arq/src/main/java/org/apache/jena/riot/system/AsyncParser.java
Aklakan: > But does it do any harm?

Right now the chunk-destination raises an exception if it is in abort state intended to forcefully kill the parser thread once it sends out another chunk.
Calling finishBatching with isAborted=true would attempt to dispatch the remaining batch and poison to the destination which is guaranteed to raise an exception.
Maybe in order to improve the structure, the `Runnable task = ()` would have to be factored out into its own class (that implements Runnable), so that parts of the process can be put into their own methods.

Aklakan:
I just realized that the atomic boolean is not sufficient to synchronize between
```
Iterator.close() { if (!isAborted.get()) { isAborted.set(true); queue.clear()} }
Destination.receive(chunk) { if(!isAborted.get() { queue.add(chunk); }}
```

When calling close(), the control flow in the destination may be after abort check but before the queue.add.
So it can result in the queue being cleared by close, then one more chunk is added, and only then the poison is added.
This can effectively lead to garbage data becoming returned. This needs a synchronized block.

jena-arq/src/main/java/org/apache/jena/riot/system/EltStreamRDFImpl.java
afs: The getters for records are `field()` style, no `get`. There are no setters because records are value-based/immutable.

I was wrong - records can implement interface- they can't extend anything.

Yes - this does seem to be authoritive as it is the output of the AsyncParser.

afs:
Not sure I fully understand that. Could you give a concrete timeline diagram? 

Is the problem having two sides calling "add" (both consumer and producer)? We already have is the AtomicBoolean for a back channel. A bit of extra work in what should be an error situation so as not to impact the normal processing is a reasonable tradeoff.

jena-arq/src/main/java/org/apache/jena/riot/system/EltStreamRDF.java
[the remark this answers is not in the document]
afs: That is an argument for a public top-level thing, not an interface.

`EltStreamRDF` is effectively specific to AsyncParser.

java generics don't help here - the interaction of generics and subclassing/impl interfaces doesn't make things easy (using casts is IMO not nice!)

jena-arq/src/main/java/org/apache/jena/riot/system/AsyncParserBuilder.java
Aklakan: done

jena-arq/src/main/java/org/apache/jena/riot/system/EltStreamRDFImpl.java
Aklakan: > There are no setters because records are value-based/immutable.

Whoops, I mentioned setters just out of habit.

I replaced the EltStreamRDF interface with the implementation and attempted to make it 'record-like' for future-proofness, i.e. field() rather than getField(). I implemented all your suggestions.

jena-arq/src/main/java/org/apache/jena/riot/system/AsyncParser.java
Aklakan: done

jena-arq/src/main/java/org/apache/jena/riot/system/AsyncParser.java
Aklakan: I factored out `Runnable task = () -> {...};` into its own `Task` class. There the run method has a clearer structure making use of private helper methods.

Aklakan: Revised in the factored out `Task` class.

jena-arq/src/main/java/org/apache/jena/riot/system/AsyncParser.java
Aklakan: done

jena-arq/src/main/java/org/apache/jena/riot/system/EltStreamRDF.java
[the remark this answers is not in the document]
Aklakan: EltStreamRDF Interface replaced with the implementation

jena-arq/src/main/java/org/apache/jena/riot/system/AsyncParser.java
```diff
@@ -51,143 +56,183 @@
  * <p>
  * There are overheads, so this is only beneficial in some situations. Delivery to
  * the StreamRDF has an initial latency while the first batch of work is accumulated.
+ * Using the {@link Builder} gives control over the chunk size such that initial latency
```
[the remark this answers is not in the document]
Aklakan: fixed

Aklakan:
I revised the previous comment with line numbers and a better description of the issue.
The most recent commits have the AtomicBoolean replaced with an AtomicInteger representing 3 states: RUNNING, ABORTING (= abort requested) and ABORTED.

These states affect what happens when the destination receives the next chunk:
- When in ABORTING state, then there is an atomic transition into ABORTED state. Subsequently, the chunk is ignored and instead the queue is cleared and the poison is placed on it.
- In ABORTED state and further chunk raises an exception.
- In RUNNING state the chunk is just normally added to the queue

A concurrent abort() request transitions from RUNNING into ABORTING state and interrupts the parser thread.
- If the parser was waiting at a full queue then it should be interrupted and do the abort action (clear queue + place poison)
- Otherwise, at latest when the next chunk arrives then the abort action will trigger

I hope I didn't overlook anything so that this approach is now sound. The "art" here is essentially getting away with only a single atomic value that is checked per chunk.

afs:
There are a few changes going it. What's the status?

Aklakan:
```
10 iterations of inserting 5,000,000 quads
4.6.0: 24m32,261s
4.7.0: 24m56,504s

20 iterations of inserting 5,432,109 quads
4.6.0: 53m10,034s
4.7.0: 51m39,403s
```

So no significant difference, also I found no issues in our hadoop setup.
It should be done.

afs (approved):
I will clean up the unused imports (TestAsyncParser).

afs:
@Aklakan 
FYI: After this is merged, the code reorg #1539 PR will go in which is a small bump to forked repos.

PR #1483, opened 2022-08-15
Improved http content type issue exception message

### LINKED ISSUE #1482: Improved content type issue exception message
PR: #1483 

### Version

4.6.0-SNAPSHOT

### Feature

With queries against DBpedia (http://dbpedia.org/sparql) I occasionally get 
`Endpoint returned Content-Type: text/html which is not recognized for SELECT queries.` which does not show what the server is actually returning.

I extended the message to include status code, status message and an excerpt of the body in order to ease investigation of the issue such that the message now looks like:
```
Endpoint returned Content-Type: text/html which is not recognized for SELECT queries. Status Code 200 OK, Body: <!DOCTYPE html>
<html lang="en">
<head>...
```
So in this specific case it seems it's really DBpedia returning the web page as a response to a SPARQL request.

Yes

Aklakan:
GitHub issue resolved #1482 

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/exec/http/QueryExecHTTP.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/exec/http/QueryExecHTTP.java
@@ -26,6 +26,7 @@ import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpRequest.BodyPublishers;
 import java.net.http.HttpResponse;
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.concurrent.TimeUnit;
@@ -40,6 +41,7 @@ import org.apache.jena.atlas.lib.InternalErrorException;
 import org.apache.jena.atlas.lib.Pair;
 import org.apache.jena.atlas.logging.Log;
 import org.apache.jena.atlas.web.HttpException;
+import org.apache.jena.atlas.web.MediaType;
 import org.apache.jena.graph.Graph;
 import org.apache.jena.graph.Triple;
 import org.apache.jena.http.HttpEnv;
@@ -58,6 +60,7 @@ import org.apache.jena.sparql.engine.http.QueryExceptionHTTP;
 import org.apache.jena.sparql.exec.QueryExec;
 import org.apache.jena.sparql.exec.RowSet;
 import org.apache.jena.sparql.util.Context;
+import org.apache.jena.web.HttpSC;
 
 /**
  * A {@link QueryExec} implementation where queries are executed against a remote
@@ -165,7 +168,8 @@ public class QueryExecHTTP implements QueryExec {
         // Use the explicitly given header or the default selectAcceptheader
         String thisAcceptHeader = dft(appProvidedAcceptHeader, selectAcceptheader);
 
-        HttpResponse<InputStream> response = performQuery(thisAcceptHeader);
+        HttpRequest request = effectiveHttpRequest(thisAcceptHeader);
+        HttpResponse<InputStream> response = executeQuery(request);
         InputStream in = HttpLib.getInputStream(response);
         // Don't assume the endpoint actually gives back the content type we asked for
         String actualContentType = responseHeader(response, HttpNames.hContentType);
@@ -190,10 +194,15 @@ public class QueryExecHTTP implements QueryExec {
 
         // Map to lang, with pragmatic alternatives.
         Lang lang = WebContent.contentTypeToLangResultSet(actualContentType);
-        if ( lang == null )
-            throw new QueryException("Endpoint returned Content-Type: " + actualContentType + " which is not recognized for SELECT queries");
-        if ( !ResultSetReaderRegistry.isRegistered(lang) )
-            throw new QueryException("Endpoint returned Content-Type: " + actualContentType + " which is not supported for SELECT queries");
+        boolean unknownLang = lang == null;
+        boolean unsupportedFormat = !unknownLang && !ResultSetReaderRegistry.isRegistered(lang);
+        if ( unknownLang || unsupportedFormat ) {
+            String errorTerm = unknownLang ? "recognized" : "supported";
+            String errorMsg = String.format("Endpoint returned Content-Type: %s which is not %s for SELECT queries",
+                    actualContentType, errorTerm);
+            raiseException(errorMsg, request, response, in);
+        }
+
         // This returns a streaming result set for some formats.
         // Do not close the InputStream at this point.
         ResultSet result = ResultSetMgr.read(in, lang);
@@ -205,7 +214,8 @@ public class QueryExecHTTP implements QueryExec {
         checkNotClosed();
         check(QueryType.ASK);
         String thisAcceptHeader = dft(appProvidedAcceptHeader, askAcceptHeader);
-        HttpResponse<InputStream> response = performQuery(thisAcceptHeader);
+        HttpRequest request = effectiveHttpRequest(thisAcceptHeader);
+        HttpResponse<InputStream> response = executeQuery(request);
         InputStream in = HttpLib.getInputStream(response);
 
         String actualContentType = responseHeader(response, HttpNames.hContentType);
@@ -227,8 +237,9 @@ public class QueryExecHTTP implements QueryExec {
             else if ( actualContentType.equals(WebContent.contentTypeJSON))
                 lang = ResultSetLang.RS_JSON;
         }
-        if ( lang == null )
-            throw new QueryException("Endpoint returned Content-Type: " + actualContentType + " which is not supported for ASK queries");
+        if (lang == null) {
+            raiseException("Endpoint returned Content-Type: " + actualContentType + " which is not supported for ASK queries", request, response, in);
+        }
         boolean result = ResultSetMgr.readBoolean(in, lang);
         finish(in);
         return result;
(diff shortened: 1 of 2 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/exec/http/QueryExecHTTP.java
```diff
@@ -190,10 +191,20 @@ private RowSet execRowSet() {
...
+
+        if ( unknownLang || unsupportedFormat ) {
+            int statusCode = response.statusCode();
+            String statusCodeMsg = HttpSC.getMessage(statusCode);
+            String bodyStr = IO.abbreviateAsUTF8(in, 1024, "...");
+            String errorTerm = unknownLang ? "recognized" : "supported";
+
+            throw new QueryException(String.format(
+                    "Endpoint returned Content-Type: %s which is not %s for SELECT queries. Status code %d %s, body: %s",
+                    actualContentType, errorTerm, statusCode, statusCodeMsg, bodyStr));
```
rvesse, line 205: Would it be worth dumping the whole body into a string and putting that in an inner exception, or creating a new derived subclass of `QueryException`?

In your example the abbreviated content isn't particularly enlightening so having the full response available somewhere for advanced users to further inspect might be useful

Aklakan: In the code the limit is hard coded to 1024 characters; may example only showed an excerpt of the excerpt. Also, the threshold could be made configurable via the context.
What I want to avoid is unconditionally loading everything into a string because for large incorrect responses this could again make investigation difficult.

I was also thinking that QueryExceptionHTTP would be more appropriate - but the original code used QueryExeception; so I thought maybe @afs had something in mind when he did not use the former (such as that a non-handleable response content type is not really an HTTP issue).

Furthermore, there seems to be more strange things going on with Jena over Virtuoso. In my specific case the problem is that Virtuoso sometimes returns plain text/html for queries - rather than some virtuoso error. But with this PR at least this becomes visible. So it's seems to be related to that Virtuoso instance (DBpedia) rather than Jena.

I also intend to extend this PR with logging the HTTP request headers on error during this week - because some of my queries sometimes hang - but without knowing the exact headers I can't reproduce it - if I paste the query into the Web form it works.

afs: In practice, I think developers will often check the endpoint with tools such as curl and wget to get more information. The whole thing might be very large for a log item.

(and, yes, Virtuoso does seem to return text/html - it seems to snoop the request to determine whether it is a programmatic query or now, and also users don't always find its endpoint but the SPARQL web UI.)

> QueryExectipionHTTP

I _think_ it is because it is not strictly an HTTP-level error. The HTTP mechanisms are fine. It's the application usage that is mismatched.

afs (changes requested):

jena-arq/src/main/java/org/apache/jena/sparql/exec/http/QueryExecHTTP.java
Aklakan: > I think developers will often check the endpoint with tools such as curl and wget to get more information. 
That's a chicken & egg problem:  Sometimes queries fail for no apparent reason - and because Jena so far does not log details about failed request it's nearly impossible to reproduce the exact requests with curl and wget for further investigation.

> I think it is because it is not strictly an HTTP-level error.

Yes, that's what I thought was the reason - the question is whether reuse of QueryExceptionHTTP would still be justified because its still a QueryException *related* to an HTTP request. Also, that exception has a response field that (IIRC) does not show on the stack-trace but can be printed out on demand.

afs: The response code is 200 to get to this point in the code. 3xx, 4xx and 5xx are already handled elsewhere.

Aklakan:
I just squashed the commits and added the HTTP headers to the message.
This is an example where a forced an unsupported content type error. It still looks ugly, so wrapping the details up in a subclass of QueryException would still be good - but at least there is now a lot of useful information included:
I guess the query string is not strictly needed, because this is the one thing the application knows.

```
Exception in thread "main" org.apache.jena.query.QueryException: Endpoint returned Content-Type: application/sparql-results+json which is not recognized for SELECT queries. Status code 200 OK, Method GET, Headers: {Accept=[application/sparql-results+json, application/sparql-results+xml;q=0.9, text/tab-separated-values;q=0.7, text/csv;q=0.5,application/json;q=0.2,application/xml;q=0.2,*/*;q=0.1]}, Remaining body: 
{ "head": { "link": [], "vars": ["s", "p", "o"] },
  "results": { "distinct": false, "ordered": true, "bindings": [
    { "s": { "type": "uri", "value": "http://www.openlinksw.com/virtrdf-data-formats#default-iid" }	, "p": { "type": "uri", "value": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type" }	, "o": { "type": "uri", "value": "http://www.openlinksw.com/schemas/virtrdf#QuadMapFormat" }},
    { "s": { "type": "uri", "value": "http://www.openlinksw.com/virtrdf-data-formats#default-iid-nullable" }	, "p": { "type": "uri", "value": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type" }	, "o": { "type": "uri", "value": "http://www.openlinksw.com/schemas/virtrdf#QuadMapFormat" }},
    { "s": { "type": "uri", "value": "http://www.openlinksw.com/virtrdf-data-formats#default-iid-blank" }	, "p": { "type": "uri", "value": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type" }	, "o": { "type": "uri", "value": "http://www.openlinksw.com/schemas/virtrdf#QuadMapFormat" }},
    { "s": { "type": "uri", "value": "http://www.ope...
	at org.apache.jena.sparql.exec.http.QueryExecHTTP.raiseException(QueryExecHTTP.java:472)
	at org.apache.jena.sparql.exec.http.QueryExecHTTP.execRowSet(QueryExecHTTP.java:204)
	at org.apache.jena.sparql.exec.http.QueryExecHTTP.select(QueryExecHTTP.java:161)
	at org.apache.jena.sparql.exec.QueryExecutionAdapter.execSelect(QueryExecutionAdapter.java:117)
	at org.apache.jena.sparql.exec.http.QueryExecHTTP.main(QueryExecHTTP.java:647)

```

jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
```diff
@@ -441,13 +441,50 @@ public static String readWholeFileAsUTF8(InputStream in) {
...
+     * Append a given marker such as "..." if more data remains in the stream.
+     * Closing the stream is the caller's responsibility.
+     *
+     * @param in           InputStream to be read
+     * @param maxWidth     Maximum number of bytes (rather than characters) to read from the input stream
+     * @param abbrevMarker A string to append to the result if the input stream was not fully consumed
+     * @return      String
+     */
+    public static String abbreviateAsUTF8(InputStream in, int maxWidth, String abbrevMarker) throws IOException {
+        return abbreviate(in, StandardCharsets.UTF_8, maxWidth, abbrevMarker);
```
afs, line 454: Use `IO.asUT8` here.

jena-arq/src/main/java/org/apache/jena/sparql/exec/http/QueryExecHTTP.java
```diff
@@ -441,13 +455,28 @@ private static long asMillis(long duration, TimeUnit timeUnit) {
         return (duration < 0) ? duration : timeUnit.toMillis(duration);
     }
 
+    private void raiseException(String errorMsg,
+            HttpRequest request, HttpResponse<InputStream> response, InputStream in) {
```
afs, line 459: Format glitch.

jena-arq/src/main/java/org/apache/jena/sparql/exec/http/QueryExecHTTP.java
```diff
@@ -441,13 +455,28 @@ private static long asMillis(long duration, TimeUnit timeUnit) {
...
+        int bodySummaryLength = 1024;
+        int statusCode = response.statusCode();
+        String statusCodeMsg = HttpSC.getMessage(statusCode);
+        String bodyStr;
+        try {
+            bodyStr = in == null ? "(no data supplied)" : IO.abbreviateAsUTF8(in, bodySummaryLength, "...");
```
afs, line 465: Should use the `Content-Type` of the response for the character set and default to UTF-8. 

Especially as this might be natural language text, JSON or HTML here.

jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
```diff
@@ -441,13 +441,50 @@ public static String readWholeFileAsUTF8(InputStream in) {
...
+
+    /** Read a given number of characters from a stream as UTF-8.
+     * Append a given marker such as "..." if more data remains in the stream.
+     * Closing the stream is the caller's responsibility.
+     */
+    public static String abbreviate(Reader reader, int maxWidth, String abbrevMarker) throws IOException {
+        char[] buffer = new char[maxWidth + 1];
+        int n = IOUtils.read(reader, buffer);
+        StringBuilder sb = new StringBuilder();
+        if (n > 0) {
```
afs, line 473: Minor: The "if" test isn't needed.

If there is a test, it could be:
```
    int n = IOUtils.read(reader, buffer);
    if ( n == 0 )
       return "";
```

afs (changes requested):

jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
Aklakan: I removed the UTF8 specific methods because the charset is now always passed as an argument anyway.

jena-arq/src/main/java/org/apache/jena/sparql/exec/http/QueryExecHTTP.java
Aklakan: Hm, the squashing breaks the code references... Anyway, the raiseException method now attempts to use the response charset and falls back to UTF8.

jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
Aklakan: Removed the if statement (I was incorrectly assuming it could return -1 like the read methods on InputStream)

afs (approved):

PR #1489, opened 2022-08-24
Clean javadoc for doclint "syntax,html"

afs:
Pull request Description:

This cleans the code so that `<doclint>syntax,html` is clean except: jena-iri and jena-core which ave some large HTML tables which doclint does not like but seem valid, albeit old HTML. 

But.

The top level is set to `<doclint>none` because while Java17 passes, Java11 checks the "missing" category (despite not being asked to) and also rejects `<br/>`.

The Jena release is currently using Java17 compiling for Java11 because the javadoc search fails.

But we ought to ensure that java11 does build Jena as we support Java11.

----

### DIFF
--- a/jena-core/pom.xml
+++ b/jena-core/pom.xml
@@ -183,6 +183,7 @@
           <!-- OK for Java17 (compiling to Java11); not OK for Java11
                <doclint>syntax</doclint>
           -->
+          <doclint>none</doclint>
           <author>true</author>
           <version>true</version>
           <show>public</show>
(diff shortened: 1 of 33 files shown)


kinow (approved):
Minor typo in comment. Otherwise looks good!

afs:
Thanks.

Looking at the issue tracker for Java, some of the javadoc issues claim to be fixed in the version of java11 on my machine `openjdk version "11.0.16" 2022-07-19` but they are still present running goal javadoc:javadoc.

Issue #1501, opened 2022-08-31
Buffer the input to bz2 decompression.


afs:
### Version

4.6.0

### Feature

https://lists.apache.org/thread/z8cl2kwsdmg47l5o05xsyvsnc3s3ozzp

> We checked code and the Apache Commons Compress docs, a colleague spotted the hint at https://commons.apache.org/proper/commons-compress/examples.html#Buffering :
> 
> > The stream classes all wrap around streams provided by the calling code and they work on them directly without any additional buffering. On the other hand most of them will benefit from buffering so it is highly recommended that users wrap their stream in Buffered(In|Out)putStreams before using the Commons Compress API.
> we were curious about this statement, checked org.apache.jena.atlas.io.IO class and added one line in openFileEx
> 
> in = new BufferedInputStream(in);
> 
> which wraps the file stream before its passed to the decompressor streams 

### Are you interested in contributing a solution yourself?

Yes

PR #1503, opened 2022-08-31
GH-1501: Buffer bz2 decompression

afs:
GitHub issue resolved: #1501

Pull request Description: #1501

----

### DIFF
--- a/jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
+++ b/jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
@@ -88,11 +88,31 @@ public class IO
         }
         InputStream in = new FileInputStream(filename);
         String ext = getExtension(filename);
+
+        // Input is a file stream.
+        // https://commons.apache.org/proper/commons-compress/examples.html#Buffering :
+        // """
+        // The stream classes all wrap around streams provided by the calling
+        // code and they work on them directly without any additional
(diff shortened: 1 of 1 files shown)


jena-base/src/main/java/org/apache/jena/atlas/io/IO.java
```diff
@@ -86,13 +86,34 @@ static public InputStream openFileEx(String filename) throws IOException, FileNo
...
-            case ext_bz2:   return new BZip2CompressorInputStream(in, true);
-            case ext_sz:    return new SnappyCompressorInputStream(in);
+            case "":
+                return in;
+            case ext_gz:
+                // Makes a small improvement (<5%) to use 8K.
+                return new GZIPInputStream(in, 8*1024);
+            case ext_bz2:
+                // Make a huge improvement. x10 faster.
+                in = IO.ensureBuffered(in);
```
kinow, line 113: @afs this part is looking weird.

We have earlier called `InputStream in = IO.ensureBuffered(in0);`. Then if bz2, we call `in = IO.ensureBuffered(in);` again? Like twice calling `IO.ensureBuffered`?

rvesse: `ensureBuffered()` checks whether the passed input stream is already buffered and if so is a no-op so don't think it matters if we call it twice

Issue #1614, opened 2022-11-10
Property path handling in query optimizer and timeout handler


LorenzBuehmann:
### Version

4.7.0-SNAPSHOT

### Question

Hi,

- Jena 4.7.0-SNAPSHOT
- TDB2
- Fuseki

### Part 1

We deployed Wikidata truthy dump and got a query which runs "forever" (probably):

``` sparql
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
PREFIX wikibase: <http://wikiba.se/ontology#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>

SELECT ?item ?itemLabel WHERE {
  ?item (rdfs:label|skos:altLabel) ?itemLabel;
    rdf:type wikibase:Property.
  FILTER(CONTAINS(LCASE(?itemLabel), "border"@en))
}
LIMIT 10
```
There are in total 7 results (when the reordered query terminates).

You can see the usage of a property path in the first triple pattern. It looks like the BGP is never reordered here? We also computed the TDB stats:

|         Pattern                     |         Count      |
|------------------------------|---------------:|
| `rdfs:label`                 | 510 188 728 |
| `skos:altLabel`              | 102 554 842 |
| `rdf:type wikibase:Property` | 9003        |

 the size of the second triple pattern would be `9000`, so fairly small.

Reordering the triple pattern indeed helps, but the user would have to know this. I can imagine property path query result estimation is difficult ...

### Part 2
With the same query we recognized that the query execution timeout is never considered with that query, so it looks like it's in a branch where query termination isn't checked for maybe when calling `GraphUtils.allNodes(graph)` in `PathLib` L245 ??

relevant part of JStack dump:

```
"qtp123458189-28" #28 prio=5 os_prio=0 cpu=45409.83ms elapsed=198.44s tid=0x00007f3831342000 nid=0x3b runnable  [0x00007f1eacefa000]
   java.lang.Thread.State: RUNNABLE
	at java.io.RandomAccessFile.seek0(java.base@11.0.16.1/Native Method)
	at java.io.RandomAccessFile.seek(java.base@11.0.16.1/RandomAccessFile.java:591)
	at org.apache.jena.dboe.base.file.BinaryDataFileRandomAccess.seek(BinaryDataFileRandomAccess.java:95)
[...]

Issue #1604, opened 2022-11-05
Do not apply algebra BGP recording by default.


afs:
### Change

[JENA-2317](https://issues.apache.org/jira/browse/JENA-2317) - BGP reordering in the algrebra stage of optimization to help filter placement - is causing some problems.

Until these are resolved, it is better to default behaviour to not running this step. Optimization of BGPs still occurs at the execution stage (TDB1, TDB2, TIM and all general datasets).

Turn this optimization step on if running the BSBM benchmark for query 5 with 
```ARQ.getContext().set(ARQ.optReorderBGP, true);```
or
```--set arq:optReorderBGP=true```

### Are you interested in contributing a pull request for this task?

Yes

PR #1616, opened 2022-11-14
calculate ungrounded path starting set lazily

SimonBin:
Issue improved: GH-1614

- the GraphUtils.allNodes used in the worst case determineUngroundedStartingSet was previously computing the contents of the whole data set. this has two drawbacks, (1) not lazy (2) not easy to abort, i.e queryTimeout was not respected

  changed it to Iter.distinct (unfortunately that means in worst case it will still require to load everything into memory)

- the `PathLib.execUngroundedPath` and `execUngroundedPathSameVar` were both consuming the iterator fully while creating new QueryIterConcats. Drawbacks are the same as above, additionally LIMIT could not be pushed down. Changed it to be lazy instead

co-authored with @LorenzBuehmann

----

 - [ ] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIter.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIter.java
@@ -20,11 +20,13 @@ package org.apache.jena.sparql.engine.iterator;
 
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Function;
 
 import org.apache.jena.atlas.io.IndentedWriter ;
 import org.apache.jena.sparql.core.Var ;
 import org.apache.jena.sparql.engine.ExecutionContext ;
 import org.apache.jena.sparql.engine.QueryIterator ;
+import org.apache.jena.sparql.engine.binding.Binding;
 import org.apache.jena.sparql.serializer.SerializationContext ;
 
 /**
@@ -66,6 +68,15 @@ public abstract class QueryIter extends QueryIteratorBase
 
     public static QueryIterator map(QueryIterator qIter, Map<Var, Var> varMapping) {
         return new QueryIteratorMapped(qIter, varMapping);
+    }
+
+    /** flatmap.
+     * The mapper may return null to signal "no iterator for this binding"
+     */
+    public static QueryIter flatMap(QueryIterator input, Function<Binding, QueryIterator> mapper, ExecutionContext execCxt) {
+        return new QueryIterRepeatApply(input, execCxt) {
+            @Override protected QueryIterator nextStage(Binding binding) { return mapper.apply(binding); }
+        };
     }
 
     @Override
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterExtendByVar.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterExtendByVar.java
@@ -20,6 +20,7 @@ package org.apache.jena.sparql.engine.iterator;
 
 import java.util.Iterator ;
 
+import org.apache.jena.atlas.iterator.Iter;
 import org.apache.jena.graph.Node ;
 import org.apache.jena.sparql.ARQInternalErrorException ;
 import org.apache.jena.sparql.core.Var ;
@@ -28,7 +29,7 @@ import org.apache.jena.sparql.engine.binding.Binding ;
 import org.apache.jena.sparql.engine.binding.BindingFactory ;
 
 /**
- * Yield new bindings, with a fixed parent, with values from an iterator. 
+ * Yield new bindings, with a fixed parent, with values from an iterator.
  */
 public class QueryIterExtendByVar extends QueryIter
 {
@@ -36,10 +37,10 @@ public class QueryIterExtendByVar extends QueryIter
     private Binding binding ;
     private Var var ;
     private Iterator<Node> members ;
-    
+
     public QueryIterExtendByVar(Binding binding, Var var, Iterator<Node> members, ExecutionContext execCxt) {
         super(execCxt);
-        if ( true ) { // Assume not too costly.
+        if ( true ) {
             if ( binding.contains(var) )
                 throw new ARQInternalErrorException("Var " + var + " already set in " + binding);
         }
@@ -61,7 +62,9 @@ public class QueryIterExtendByVar extends QueryIter
     }
 
     @Override
-    protected void closeIterator() {}
+    protected void closeIterator() {
+        Iter.close(members);
+    }
 
     @Override
     protected void requestCancel() {}
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterRepeatApply.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterRepeatApply.java
@@ -34,8 +34,8 @@ public abstract class QueryIterRepeatApply extends QueryIter1 {
     private QueryIterator currentStage;
     private volatile boolean cancelRequested = false;
 
-    public QueryIterRepeatApply(QueryIterator input, ExecutionContext context) {
-        super(input, context);
+    public QueryIterRepeatApply(QueryIterator input, ExecutionContext execCxt) {
+        super(input, execCxt);
         this.currentStage = null;
 
         if ( input == null ) {
--- a/jena-arq/src/main/java/org/apache/jena/sparql/path/PathLib.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/path/PathLib.java
@@ -22,6 +22,7 @@ import java.util.ArrayList ;
 import java.util.Iterator ;
 import java.util.List ;
 import java.util.Objects;
+import java.util.function.Function;
 import java.util.function.Predicate;
 
 import org.apache.jena.atlas.iterator.Iter ;
@@ -41,9 +42,7 @@ import org.apache.jena.sparql.engine.ExecutionContext ;
 import org.apache.jena.sparql.engine.QueryIterator ;
 import org.apache.jena.sparql.engine.binding.Binding ;
 import org.apache.jena.sparql.engine.binding.BindingFactory ;
-import org.apache.jena.sparql.engine.iterator.QueryIterConcat ;
-import org.apache.jena.sparql.engine.iterator.QueryIterPlainWrapper ;
-import org.apache.jena.sparql.engine.iterator.QueryIterYieldN ;
+import org.apache.jena.sparql.engine.iterator.*;
 import org.apache.jena.sparql.mgt.Explain ;
 import org.apache.jena.sparql.path.eval.PathEval ;
 import org.apache.jena.sparql.pfunction.PropertyFunctionFactory ;
@@ -78,7 +77,7 @@ public class PathLib
         op = flush(bp, op);
         return op;
     }
-    
+
     static private Op flush(BasicPattern bp, Op op) {
         if ( bp == null || bp.isEmpty() )
             return op;
@@ -104,14 +103,14 @@ public class PathLib
             Path path = new P_Link(triplePath.getPredicate());
             triplePath = new TriplePath(triplePath.getSubject(), path, triplePath.getObject());
         }
-        
-        return execTriplePath(binding, 
+
+        return execTriplePath(binding,
                               triplePath.getSubject(),
                               triplePath.getPath(),
                               triplePath.getObject(),
                               execCxt) ;
     }
-    
+
     public static QueryIterator execTriplePath(Binding binding, Node s, Path path, Node o, ExecutionContext execCxt) {
         Explain.explain(s, path, o, execCxt.getContext()) ;
         s = Var.lookup(binding, s) ;
@@ -143,13 +142,13 @@ public class PathLib
         }
         return evalGroundedOneEnd(binding, iter, endNode, execCxt);
     }
-    
+
     private static QueryIterator evalGroundedOneEnd(Binding binding, Iterator<Node> iter, Node endNode, ExecutionContext execCxt) {
         List<Binding> results = new ArrayList<>() ;
-        
+
         if (! Var.isVar(endNode))
-            throw new ARQInternalErrorException("Non-variable endnode in _execTriplePath") ;
-        
+            throw new ARQInternalErrorException("Non-variable endNode in evalGroundedOneEnd") ;
+
         Var var = Var.alloc(endNode) ;
         // Assign.
         for (; iter.hasNext();) {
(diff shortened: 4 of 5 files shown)


namedgraph:
Was `TransformPathFlattern` meant to be called `TransformPathFlattener`?

jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformPathFlattern.java
```diff
@@ -53,9 +58,60 @@ public TransformPathFlattern(PathCompiler pathCompiler)
...
+        public AltTransformer(PathCompiler pathCompiler) {
+            this.pathCompiler = pathCompiler;
+        }
+
+        @Override
+        public Op transform(OpPath opPath) {
+            TriplePath triplePath = opPath.getTriplePath();
+            Path path = triplePath.getPath();
+            if (path instanceof P_Alt) {
+                Node s = triplePath.getSubject();
```
rvesse, line 96: So I don't believe this transformation is semantically safe in all cases.

At a minimum it needs to consider whether either end of the path is a blank node variable as pushing those into a union will make them unique variables relative to the outside of the union and change the semantics of the query.

i.e. a query like `SELECT * WHERE { ?s ex:foo / (ex:a | ex:b) ?o }` can have the first part of the path rewritten to yield `SELECT * WHERE { ?s ex:foo _:bnode . _:bnode (ex:a | ex:b) ?o }`, if you then rewrite the 2nd part into a union you change the semantics -> `SELECT * WHERE { ?s ex:foo _:bnode . { _:bnode ex:a ?o . } UNION { _:bnode ex:b ?o } }`

From the [SPARQL 1.1 specification](https://www.w3.org/TR/sparql11-query/#QSynBlankNodes):

> Blank nodes are indicated by either the label form, such as "_:abc", or the abbreviated form "[]". A blank node that is used in only one place in the query syntax can be indicated with []. A unique blank node will be used to form the triple pattern. Blank node labels are written as "_:abc" for a blank node with label "abc". The same blank node label cannot be used in two different basic graph patterns in the same query.

So that's technically an illegal query you've generated. Evaluation wise those two `_:bnode` variables are now in different scopes and so rather than joining the results of the BGP and the Union you'll be generating a cross-product

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterConcatIterator.java
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.jena.sparql.engine.binding.Binding;
+import org.apache.jena.sparql.serializer.SerializationContext;
+
+import java.util.*;
+
+
+/**
+ * A query iterator that joins two or more iterators into a single iterator. */ 
+
+public class QueryIterConcatIterator extends QueryIter
```
rvesse, line 34: @afs Is there not already a query iterator that supported this pattern?

I see a `QueryIterConcat` referenced in the code deletions so why was that not suitable?  Could that have been modified to solve the issue rather than introducing a new class

jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformPathFlattern.java
```diff
@@ -32,7 +37,7 @@
  * For example,
  * <ul>
  * <li>Path seq {@literal ->} BGPs or a (sequence)
- * <li>"|" is not expanded into a union.
+ * <li>"|" is expanded into a union.
```
rvesse, line 40: Need to add new optimiser test cases that verify the modified transform is having the desired effects, and to verify it doesn't make semantically unsafe transforms as noted in other comments

rvesse (changes requested):
Thanks for the PR, I'm concerned about the semantic safety of one aspect which is detailed in my review comments

jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformPathFlattern.java
Aklakan: Hm, The transform happens on the algebra level so all bnodes should have been renamed to variables with appropriate scoping - right?

rvesse: I was using the SPARQL syntax to illustrate the issue but this applies at the algebra nonetheless.

This transform is happening after any scope renaming has happened.  If rewriting an alt to a `union` operator that's also introducing a `join` into the algebra which if there's a blank node variable involved breaks the semantics AFAICT

SimonBin: Thank you for the thorough analysis and concern. In my testing, I got the expected results when querying e.g. ` { ?s ex:foo _:bnode . _:bnode (ex:a | ex:b) ?o }` as you suggested. Do you have any additional pointers how to verify the safety of the transform?

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterConcatIterator.java
SimonBin: the QueryIterConcat can be used to make a materialised list of many QueryIterators. In this case, the (nested) QueryIterators themselves should be constructed lazily, that's why I didn't see a good way to make use of the existing class

afs: Isn't that then a flatmap? 
`QueryIterRepeatApply` provides flatmap for `QueryIterator`.

jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformPathFlattern.java
afs: > In my testing

Which tests are you referring to? There don't seem to be any in the PR. 
`TestTransformPathFlatten`

SimonBin: manual testing based on the input provided by @rvesse , I will add some tests to TestTransformPathFlatten

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterConcatIterator.java
SimonBin: should I delete this class and replace it with QueryIterPlainWrapper.create(Iter.flatMap(...))?

afs:
`TransformPathFlatternAlt` performs `P_Alt` transformation. 

IIRC it interacts with the BGP extension mechanism when considering beyond simple entailment. We;ll need to investigate this.

Things to consider:

* It might be better to restrict optimization to the case of the specific and important case of `(property | property)`, not general path. This is mentioned on [JENA-2325](https://issues.apache.org/jira/browse/JENA-2325).
* It might be better to do performance improvements in the evaluation stage to preserve semantics.

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterConcatIterator.java
afs: `QueryIterRepeatApply` if you want cancellation at that granularity.

jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformPathFlattern.java
```diff
@@ -53,9 +58,60 @@ public TransformPathFlattern(PathCompiler pathCompiler)
...
-        PathBlock pattern = pathCompiler.reduce(opPath.getTriplePath()) ;
-        // Any generated paths of exactly one to triple; convert to Op.
-        return PathLib.pathToTriples(pattern) ;
+        Op op = transformReduce(opPath, pathCompiler);
+        return op;
     }
-}
+
+    static Op transformReduce(Op op, PathCompiler pathCompiler) {
+        return Transformer.transform(
```
afs, line 66: I'm confused - this is only called with an `OpPath` as the first argument.

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterConcatIterator.java
SimonBin: sorry, I don't understand. RepeatApply is an abstract class that takes a QueryIterator and repeatedly calls nextStage, how will that help here?

Aklakan: You can rearrange the code in `PathLib` along the lines of first turning the Nodes to Bindings, and then wrap that as a QueryIterator which serves as input to QueryIterRepeatApply:

```java
Iterator<Node> itNode = determineUngroundedStartingSet(graph, path, execCxt) ;
Iterator<Binding> itBinding = Iter.iter(itNode).map(n -> BindingFactory.binding(binding, sVar, n));
QueryIterator lhs = QueryIterPlainWrapper.wrap(itBinding);
QueryIterator combined = new QueryIterRepeatApply(lhs) {
  QueryIterator nextStage() {
                Iterator<Node> pathIter = PathEval.eval(graph, b2.get(sVar), path, execCxt.getContext());
            return evalGroundedOneEnd(b2, pathIter, oVar, execCxt);
  }
};
```

SimonBin: would that have any advantage over or be significantly differnt from QueryIterPlainWrapper(Iter.flatMap)?

jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformPathFlattern.java
SimonBin: so am I! fixed

LorenzBuehmann:
> * It might be better to restrict optimization to the case of the specific and important case of `(property | property)`, not general path. This is mentioned on [JENA-2325](https://issues.apache.org/jira/browse/JENA-2325).

I would be fine with this - currently only `p+` and `p*` are handled smarter, but very common usages of `p|q` or even `p|^p` are not covered at all, leading to worst case scenario with an unbound triple pattern by gathering as seed nodes all subjects and objects in the graph and collecting those in a set. Note, we're more and more working on larger dataset like Wikidata, thus, it's definitely something that won't work at all in that case.
We also tried with at least handling basic cases of p_alt only contain p_oneormore, p_link or inverted p_links, this would cover probably most real world cases. It looks a bit clumsy still and maybe bringing the p_alt to some kind of normal form first would make it easier. @Aklakan also suggested to build an NFA, but I think we should keep it as simply as possible, we only need the first hop (or last to get objects) in the automaton to have seeds nodes. The rest of the evaluation I'd keep as is.  

Also as a comment, the PR addressed multiple things afaik:

- consider `LIMIT X` due to fully working on iterators, thus, being as lazy as possible
- usage of QueryIterators such that the query execution context is used to be aware of query termination when a timeout happened
- handling of p_alt

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterConcatIterator.java
afs: > You can rearrange the code in `PathLib`

There is also `QueryIterExtendByVar`.

afs:
> an unbound triple pattern by gathering as seed nodes all subjects and objects in the graph

Part 1 of #1614 is about a lack of reordering across paths and isn't in this PR.

We need to check we aren't heading off in the wrong direction with changes in path eval when there is a bigger picture.

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterConcatIterator.java
SimonBin: hummmm... I can see how the execUngroundedPath could be rewritten with ExtendByVar + implementing RepeatApply but for the  execUngroundedPathSameVar the code is more cumbersome as we want to go from Node -> (x, node) -> (x, binding) -> QueryIterator but the RepeatApply only takes a binding

LorenzBuehmann:
> Part 1 of #1614 is also about a lack of reordering across paths and that isn't in this PR.

True, but a query with only a single triple pattern having an alt_p would still benefit from smarter handling of basic property path cases to gather seed nodes. that was my point, we could easily add handle those without changing any other code but add more lines to method `PathLib::determineUngroundedStartingSet`

It would be a huge performance gain already:
Wikidata
```sparql
SELECT * {
?s wd:P10564|wd:P4496 ?industry_code
}
```
where both properties are used just in 1746 and 997 triples compared to millions of subjects and objects being gathered and then evaluated otherwise

rvesse:
Per @afs earlier comment there is actually an alternative path transform already in the code base that already covered more complex path flattening such as alternatives.

I've opened PR #1620 to make it possible to toggle that transform on with the existing optimiser via an ARQ symbol leaving the existing simple transform unchanged.  That way users can opt into the more extensive path flattening if they wish but the existing default behaviour remains.

SimonBin:
my concern would be that this completely different path flatterning in #1620  has never been put to wide testing and per the comment in the source file, `It does not produce very nice execution structures so ARQ uses
 a functional equivalent, but different, transformation.`. One obvious instance from your tests, `:p1{2}` now becomes a join instead of a simple bgp (although I guess the other optimisers will turn that back into a sequence so maybe it doesn't matter)

jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformPathFlattern.java
```diff
@@ -53,9 +58,53 @@ public TransformPathFlattern(PathCompiler pathCompiler)
...
+    }
+
+    static Op transformReduce(OpPath opPath, PathCompiler pathCompiler) {
         // Flatten down to triples where possible.
-        PathBlock pattern = pathCompiler.reduce(opPath.getTriplePath()) ;
+        PathBlock pattern = pathCompiler.reduce(opPath.getTriplePath());
         // Any generated paths of exactly one to triple; convert to Op.
-        return PathLib.pathToTriples(pattern) ;
+        Op op = PathLib.pathToTriples(pattern);
+        return transformAlt(op, pathCompiler);
```
afs, line 70: A request for explanation: 

We seem to have two mechanism for the same thing here - PathCompiler and AltTransformer+transformReduce.

What advantages does this bring?

SimonBin: the suggestion of this change is as follows: we leave the original pathCompiler unchanged (`transformReduce`/pathCompiler.reduce) -- it can only flatten a path to PathBlocks and does not handle P_Alt. Then, we add another Transformer (transformAlt/AltTransformer) which only transforms any P_Alt to OpUnion

afs:
Please fix commit 900dfc7676ba2d1e8a18d63c3c7d5eb5dc2a2dac - it has the whole original issue description (before it was fixed it up) and has people's names and ids in it.

Put "GH-1616: " at the start of the commit message. It will then refer to the PR.

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterConcatIterator.java
afs: > the difference between cancel() and close() 

`cancel` is async e.g. query timeout. `close` is called from creating thread a a controlled point.

afs: https://gist.github.com/afs/cb8e4d806d4efd9239682ec9220f936f

rvesse:
> my concern would be that this completely different path flatterning in #1620 has never been put to wide testing and per the comment in the source file,

And my point is that neither is yours.  You're changing the default behaviour for all existing users based on testing on your own datasets.

As the recent flurry of issues around another new transform that was introduced has shown optimisations that benefit one dataset don't necessarily benefit all, and ultimately led to it being disabled by default (#1604).  Hence the suggestion of enabling an opt-in to the alternative transform rather than changing the existing default.

> `It does not produce very nice execution structures so ARQ uses a functional equivalent, but different, transformation.`. One obvious instance from your tests, `:p1{2}` now becomes a join instead of a simple bgp (although I guess the other optimisers will turn that back into a sequence which maybe isn't quite as bad?)

Yes the full optimiser converts joins of adjacent BGPs or triple operators back into a single BGP, expanded my test cases to explicitly cover this now.

> 
> Another thought about Std should be given w.r.t inverse paths like `^:p1+`, they are reversed by the current optimiser but remain in reverse in the Std implementation. This might then cause worse behaviour in the determineUngroundedStartingSet which does not handle them at the moment afaik (fall-back to allNodes)

Well if both ends of the path are ungrounded whether we reverse it or not is irrelevant, and regardless we're conflating the issue of deficiencies in the path evaluator with algebra transformation.

afs:
@SimonBin - do you have time to return this PR into step 1 of #1629? (that is, remove the `TransformPathFlattern`, squash to one commit). Then step 2 can be based on code after the step 1 change.

If you don't have time, say so and I'll put in a new PR starting from the diff of this one.

SimonBin:
I removed the Transformer. Note, step 1  of  #1629 would mean to change PathEval to iters (see my comment there, starting point would be PathEval.eval$), that is not implemented in this PR

afs:
Hi - could you squash the "AFS" commit into the  main one. I don't want my name in the commit log (the commit was originally an example!).

Re: PathEval 
Unlikely for the 4.7.0. Fortunately, that's a separable piece of work.

I have some code for step 2 covering better starting sets of all the path operators I can see how to do.
I believe optimizing these is more important.

It will however require testing on larger data. I hope you group can do that for the cases you(plural) have raised.

The objective is "better" not "perfect".

afs:
@simonbin - Thanks!

I'll do a PR for better ungrounded starting based on this PR.

afs (approved):

PR #1631, opened 2022-11-20
GH-1615: LATERAL join

### LINKED ISSUE #1615: LATERAL join
# LATERAL join

_Proposed experimental feature._

_Work-in-progress: the issue description is being edited in-place._

A `LATERAL` join is like a foreach loop, looping on the results from the left-hand side (LHS), the pattern before the `LATERAL` keyword, and executing the right-hand side (RHS) query pattern once for each row, with the variables from the RHS in-scope during each RHS evaluation.

A regular join only executes the RHS once, and the variables from the LHS are only used for the join condition after evaluation of the left and right sub-patterns.

Another way to think of a lateral join is as a `flatmap`.

Examples:
```
## Get exactly one label for each subject in a row.
SELECT * {
   ?s ?p ?o
   LATERAL {
     SELECT * { ?s rdfs:label ?label } LIMIT 1
   }
}
```

```
## Get zero or one labels for each subject.
SELECT * {
   ?s ?p ?o
   LATERAL { OPTIONAL { SELECT * ?s rdfs:label ?label } LIMIT 1}
   }
}
```
`{ OPTIONAL ...` is the same as writing `{ {} OPTIONAL ...`.
`{ }` evaluates to the join identity, a table of one row of zero columns.

### Syntax

The `LATERAL` keyword which has the graph pattern so far (from the `{` starting the current block) and a `{ }` block afterwards. 

Possible addition: `LATERAL ( ?var1 ?var2 ...)` to specify certain variables to expose to the RHS. Other variables would be (inner)joined as usual. This may be an unnecessary feature.

### Scope

A sub-select may have variables of the same name that are not lateral-joined to a variable of the same name from the LHS.

```
SELECT * {
   ?s ?p ?o
   LATERAL {
     SELECT ?label { ?s rdfs:label ?label } LIMIT 1
   }
}
```
The inner `?s` in the `SELECT ?label` is not the outer `?s` because the `SELECT ?label` does not pass out `?s`. As a sub-query the `?s` could be any name except `?label` for the same results.

This is the same situation as a sub-query in other situations.

There needs to be a new syntax restriction: there can no variable introduced by `AS` (`BIND`, or sub-query) or `VALUES` in-scope at the top level of the `LATERAL` RHS, that is the same name as  any [in-scope](https://www.w3.org/TR/sparql11-query/#variableScope) variable from the LHS.

```
## ** Illegal **
SELECT * {
   ?s ?p ?o
   LATERAL { BIND( 123 AS ?o) }
}
```

See [SPARQL Grammar note 12](https://www.w3.org/TR/sparql11-query/#sparqlGrammar).

In ARQ, [LET](https://www.w3.org/TR/sparql11-query/#variableScope) would work.
`LET` for a variable that is bound acts like a filter.

### Evaluation

[Substituting variables](https://afs.github.io/substitute.html) from the LHS into the RHS (with the same restrictions), then executing the pattern, gives the evaluation of `LATERAL`

### Notes

There is a similarity to filter `NOT EXISTS`/`EXISTS` expressed as the not-legal `FILTER ( ASK { pattern } )` where the variables of the row being filtered are available to "pattern". This is similar to ab SQL [correlated subquery](https://en.wikipedia.org/wiki/Correlated_subquery).

### Elsewhere

* https://github.com/w3c/sparql-12/issues/100
* Jena's [SERVICE <loop:>](https://jena.apache.org/documentation/query/service_enhancer.html)

* Oxigraph: [oxigraph/issues/267](https://github.com/oxigraph/oxigraph/issues/267), [oxigraph/pull/274](https://github.com/oxigraph/oxigraph/pull/274)
* https://docs.stardog.com/query-stardog/stored-query-service#correlated-subqueries

* https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-LATERAL
* https://dev.mysql.com/doc/refman/8.0/en/lateral-derived-tables.html
* https://en.wikipedia.org/wiki/Correlated_subquery

### Syntax

`LATERAL` is added to the SPARQL grammar at rule `[[56] GraphPatternNotTriples](https://www.w3.org/TR/sparql11-query/#rGraphPatternNotTriples)`. As a syntax form, it is similar to `OPTIONAL`.

```
[56]  	GraphPatternNotTriples	  ::=  	GroupOrUnionGraphPattern | OptionalGraphPattern | LateralGraphPattern | ...
[57]  	OptionalGraphPattern	  ::=  	'OPTIONAL' GroupGraphPattern
[  ]  	LateralGraphPattern	  ::=  	'LATERAL' GroupGraphPattern
```

### Algebra

The new algebra is operator is `lateral` which takes two expressions

```
  SELECT * {
    ?s  ?p  ?o
    LATERAL
      { ?a  ?b  ?c }
}
```
is translated to:
```
  (lateral
    (bgp (triple ?s ?p ?o))
    (bgp (triple ?a ?b ?c)))
```

### Evaluation

To evaluate `lateral`:

* Evaluate the first argument (left-hand side from syntax) to get a multiset of solution mappings.
* For each solution mapping ("row"),
    inject variable bindings into the second argument
    Evaluate this pattern
    Add to results

Outline:
```
Definition: Lateral

Let Ω be a multiset of solution mappings. We define:

Lateral(Ω, P) = { μ | union of Ω1 where 
           foreach μ1 in Ω:
               pattern2 = inject(pattern, μ1)
               Ω1 = eval(D(G), pattern2)
	       result Ω1
	   }
```
where `inject` is the [corrected `substitute`](https://afs.github.io/substitute.html) operation.

An alternative style is to define Lateral more like "evaluate P such that μ is in-scope" in some
way, rather than rely on `inject` which is a mechanism. 

```
Definition: Evaluation of Lateral

eval(D(G), Lateral(P1, P2) = Lateral(eval(D(G), P1), P2)
```

afs:
GitHub issue resolved #1615

Pull request Description:
Implementation of LATERAL join.

Currently, there are different commits for the different implementation steps but each step does not necessarily work on its own.  That helps inspect the PR. 

The PR will be cleaned up before merge.

----

 - [x] Tests are included.


afs:
Rebased to current main branch.

jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformFilterImplicitJoin.java
```diff
@@ -120,10 +120,14 @@ private static Op apply(ExprList exprs, Op subOp) {
...
-        // hence eliminate all rows. Return the empty table.
+        // LATERAL : This is not longer true.
+//        // Special case : deduce that the filter will always "eval unbound"
+//        // hence eliminate all rows. Return the empty table.
+//        if (testSpecialCaseUnused(subOp, joins, remaining))
+//            return OpTable.empty();
+        // But simply skipping this causes (filter) to become (assign) which fails as (assign) does not handle errors.
         if (testSpecialCaseUnused(subOp, joins, remaining))
-            return OpTable.empty();
+            return null;
```
rvesse, line 130: Would an alternative fix be overriding the transform of `OpLateral` in this class to return a direct copy of the original `opLateral` ignoring any transforms this might have generated?

afs: That's a possibility - I think it seems to designing for handling a bad query case (if not LATERAL), rather than optimization in LATERAL.

It does seem to be a risk more widely, now and in future changes and not just in LATERAL - e.g. optimizing then substituting (e.g. parameterized queries). (NOT)EXISTS as well which is related to substitution. The fix for JENA-500 is a specific fix for the whole-query case.

The transform walk is bottom up so it will run over the code - the walker needs to be changed to avoid certain ops.

rvesse: Yes the bottom up property means the optimisation would be attempted but then you could discard it when you reached the `OpLateral` level but that does mean unnecessary work is done.

Having the walker be able to decide what operations to descend through could be a useful mechanism in general for limiting the scope of transformations

afs: Changing the walker is big change - I've tried it before. I'd like to get this PR in without resorting to that.

I believe the special case is also a potential impact in other places. We haven't seen it reported because e.g. EXISTS {} is often simple.

Is there any pointers as to when it has been beneficial?

I tried skipping it rather than `return null` but `TestSemanticEquivalence.implicitJoinEvaluation3` and on a SPARQL WG test fail.

In `implicitJoinEvaluation3`, it only fails on the query string part - the next two algebra level tests work. This doesn't make sense to me -- the algebra steps are supposed to cover the possibilities of the SPARQL query.

afs: Query timeouts will catch that (we hope).

EXISTS uses the same injection of bindings as JENA-500 except it does it dynamically at runtime (like LATERAL - EXIST is the ASK version of LATERAL). Whether that's the current iffy form or the better one written up.

It's is less likely to occur in EXISTS due to usage but it is possible. And like LATERAL, complicated because it may be bound sometimes and not others.

Can we merge as-is which is functionally safe?

---

I'm still mystified by `TestSemanticEquivalence.implicitJoinEvaluation3`.

Maybe the test is wrong and comparing to zero happens for the wrong reasons but I can't see a difference to the algebra sub-tests.

PR #1641, opened 2022-11-30
Use distinctAdjacent for TDB 2 graph iteration (GH-1639)

rvesse:
When the primary index is G??? based use distinctAdjacent() rather than distinct() when listing graph nodes for TDB 2.  This provides some speedup and a constant memory footprint for this iteration as we only have to remember the previous graph name not all the graph names.

This partially addresses #1639, in my testing it yields about a 20% speed up on some synthetic test datasets.

----


rvesse:
Yes I see your PR #1642 covers the other possible code path in a similar fashion

afs (approved):

PR #1645, opened 2022-11-30
GH-1643: Replace commons-fileload

### LINKED ISSUE #1643: Replace Apache Commons FileUpload with Servlet MultipartConfig
### Change

The functionality of the Apache Commons FileUploader can nowadays be fulfilled using features of the Servlet API.

This then unblocks upgrading from Jetty10 (javax packages) to Jetty11 (jakarta packages).

Yes

afs:
GitHub issue resolved #1643

Pull request Description:
See #1643

Tested for Jetty 10.

It uses Servlet API 3.0 which Tomcat has supported since version 7.

This PR removes the `commons-fileupload` dependency which is no longer used. 

https://issues.apache.org/jira/browse/FILEUPLOAD-309

----

 - [x] Tests are included. Existing tests cover usage.

### DIFF
--- a/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/system/DataUploader.java
+++ b/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/system/DataUploader.java
@@ -102,15 +102,33 @@ public class DataUploader {
         }
     }
 
+    // @MultipartConfig but for ServletFilters
+    // * @MultipartConfig only wokrs on servlets.
+    // * Jetty: attribute setting.
+    // * Tomcat: META-INF/context.xml setting
+
     // Jetty requires a setting of this annotation object as a request attribute.
-     private static MultipartConfigElement multipartConfigElement = new MultipartConfigElement("");
-     private static String multipartAttributeName = org.eclipse.jetty.server.Request.__MULTIPART_CONFIG_ELEMENT;
+    private static final MultipartConfigElement multipartConfigElement = new MultipartConfigElement("");;
+    private static final String jettyMultipartAttributeName;
+
+    static {
+        String x;
+        // May not be on the classpath
+        try {
+            Class.forName("org.eclipse.jetty.server.Request");
+            x = org.eclipse.jetty.server.Request.__MULTIPART_CONFIG_ELEMENT; }
+        catch (ClassNotFoundException th) { x = null; }
+        jettyMultipartAttributeName = x;
+    }
+
+
+    // Tomcat: In the WAR file: META-INF/context.xml
+    // <Context allowCasualMultipartParsing="true">
+    // </Context>
 
     /**
      * Process an HTTP upload of RDF files (triples or quads) with content type
      * "multipart/form-data" or "multipart/mixed".
-     * <p>
-     * Form data (content-disposition: form-data; name="...") is rejected.
      * <p>
      * Data is streamed straight into the destination graph or dataset.
      * <p>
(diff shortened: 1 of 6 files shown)


jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/system/DataUploader.java
```diff
@@ -113,29 +117,16 @@ public static UploadDetails incomingData(HttpAction action, StreamRDF dest) {
...
-                    // This code is currently used to put multiple files into a single destination.
-                    // Additional field/values do not make sense.
-                    ServletOps.errorBadRequest(format("Only files accepted in multipart file upload (got %s=%s)", fieldName, value));
-                    // errorBadRequest does not return.
-                    return null;
-                }
-
-                InputStream input = fileStream.openStream();
+            for ( Part part : request.getParts() ) {
+                InputStream input = part.getInputStream();
```
kinow, line 127: :ok_man: that's some nice simplification!

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/system/DataUploader.java
```diff
@@ -102,6 +102,10 @@ public static UploadDetails incomingData(HttpAction action, StreamRDF dest) {
         }
     }
 
+    // Jetty requires a setting of this annotation object as a request attribute.
+     private static MultipartConfigElement multipartConfigElement = new MultipartConfigElement("");
+     private static String multipartAttributeName = org.eclipse.jetty.server.Request.__MULTIPART_CONFIG_ELEMENT;
```
kinow, line 107: Do you know if this `__MULTIPART_CONFIG_ELEMENT` attribute was created this way to discourage external use?

kinow: Hm, can't find any note saying that, so probably just a different way to name constants I guess. https://github.com/eclipse/jetty.project/blob/410e9db837c964bfcb27ff01693eb77a45dcd60f/jetty-server/src/main/java/org/eclipse/jetty/server/Request.java#L144

kinow (approved):

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/system/DataUploader.java
afs: > Do you know if this __MULTIPART_CONFIG_ELEMENT attribute was created this way to discourage external use?

I'll check some more. Using the annotation `@MultipartConfig` on FusekiFilter does not work (it's not a servlet).

The name is Jetty specific which is why it is protected by the `getAttribute`.  Jetty source code likes leading `_`'s. See `org.eclipse.jetty.server.Request`.

I'm unsure if Tomcat needs anything special. I haven't managed to find a mention one way or the other.

PS: https://stackoverflow.com/questions/8047173/how-to-use-httpservletrequestgetparts-in-a-servlet-filter-running-on-tomcat

got me to:

https://tomcat.apache.org/tomcat-7.0-doc/config/context.html#Common_Attributes

Set `allowCasualMultipartParsing`.

afs: There is no rush to complete this.

@kinow - thanks for catching the issues - I hadn't noticed that the constant was Jetty-specific.

With various other things doing on, clarifying the situation for Tomcat may not get done for 4.7.0.

There may be a way to automatically programtically  set `allowCasualMultipartParsing` in Tomcat, otherwise the user needs to set it in their `context.xml`.

Marked as "draft" to avoid accidentally merging!

afs:
Not for 4.7.0. 

Until the Tomcat situation is cleared up, this can wait.
It would be nice not to require uses to modify the tomcat deployment configuration.

(On the complete migration to "jakarta", the war file wil only run in Tomcat10, whereas no it only runs in Tomcat <=9 - ther is a converter tool.)

afs:
Now supports Tomcat 8 and 9 as well as Jetty 10.

Tomcat 10 uses packages from `jakarta.*` so is incompatible with Fuseki at the moment.

kinow (approved):

PR #1655, opened 2022-12-01
Use a skip scan based iterator for listing graph names in TDB2 (GH-1639)

rvesse:
Adds a new `BPTreeDistinctKeyPrefixIterator` that allows iterating only records which are considered distinct based on a portion of their key.  It is effectively a skip scan based iterator that can avoid reading portions of the B+Tree where all records share the same key prefix.  This is used to improve performance of `DatasetGraphTDB.listGraphNodes()` and `SolverLibTDB.graphNames()`, in some cases dramatically so.

Used a couple of different test scenarios with calling `listGraphNodes()`:

- 100k named graphs comprising a total of 125 million quads
   - On 4.6.1 this takes an average of 21,480 milliseconds
   - With this patch this is reduced to 18,235 milliseconds
- 10 named graphs comprising a total of 177 million quads
   - On 4.6.1 this takes an average of 27,856 milliseconds 
   - With the previous patches (#1641 and #1642) this was reduced to an average of 25,185 milliseconds
   - With this new patch this is reduced to around 5 milliseconds 😄 

This resolves #1639

### DIFF
--- a/jena-db/jena-dboe-trans-data/src/main/java/org/apache/jena/dboe/trans/bplustree/BPTreeDistinctKeyPrefixIterator.java
+++ b/jena-db/jena-dboe-trans-data/src/main/java/org/apache/jena/dboe/trans/bplustree/BPTreeDistinctKeyPrefixIterator.java
@@ -49,10 +49,15 @@ class BPTreeDistinctKeyPrefixIterator implements Iterator<Record> {
     BPTreeDistinctKeyPrefixIterator(BPTreeNode node, int keyPrefixLength) {
         if (keyPrefixLength < 1) {
             throw new IllegalArgumentException("keyPrefixLength must be >= 1");
+        } else if (keyPrefixLength > node.params.recordFactory.keyLength()) {
+            throw new IllegalArgumentException(
+                    "Maximum keyPrefixLength for this B+Tree is " + node.params.recordFactory.keyLength());
         }
+
         this.keyPrefixLength = keyPrefixLength;
         this.minRecord = node.minRecord();
         this.maxRecord = node.maxRecord();
+
         BPTreeRecords r = loadStack(node);
         current = getRecordsIterator(r);
     }
@@ -116,8 +121,7 @@ class BPTreeDistinctKeyPrefixIterator implements Iterator<Record> {
             // iterator for this entire subtree and skip further recursion
             Record subtreeMin = n.minRecord();
             Record subtreeMax = n.maxRecord();
-            if (Arrays.compare(subtreeMin.getKey(), 0, this.keyPrefixLength, subtreeMax.getKey(), 0,
-                               this.keyPrefixLength) == 0) {
+            if (haveSamePrefix(subtreeMin, subtreeMax)) {
                 return Iter.singleton(subtreeMin);
             }
             r = loadStack(n);
@@ -127,25 +131,29 @@ class BPTreeDistinctKeyPrefixIterator implements Iterator<Record> {
         return getRecordsIterator(r);
     }
 
+    protected final boolean haveSamePrefix(Record a, Record b) {
+        return Arrays.compare(a.getKey(), 0, this.keyPrefixLength, b.getKey(), 0, this.keyPrefixLength) == 0;
+    }
+
     // ---- Places we touch blocks.
 
     protected Iterator<Record> getRecordsIterator(BPTreeRecords records) {
         records.bpTree.startReadBlkMgr();
         Iterator<Record> iter;
         if (!records.hasAnyKeys()) {
-            return Iter.nullIterator();
-        }
-        // Check whether we need to scan the whole page or can process it by skipping or singleton yield
-        Record lowRecord = records.getLowRecord();
-        if (Arrays.compare(lowRecord.getKey(), 0, this.keyPrefixLength, records.getHighRecord().getKey(),
-                           0, this.keyPrefixLength) == 0) {
-            // If the low and high keys for this page have the same prefix then just return a singleton iterator
-            iter = Iter.singleton(lowRecord);
+            iter = Iter.nullIterator();
         } else {
-            // Otherwise need to scan the whole page
-            iter = records.getRecordBuffer().iterator();
-            records.bpTree.finishReadBlkMgr();
+            // Check whether we need to scan the whole page or can process it by skipping or singleton yield
+            Record lowRecord = records.getLowRecord();
+            if (haveSamePrefix(lowRecord, records.getHighRecord())) {
+                // If the low and high keys for this page have the same prefix then just return a singleton iterator
+                iter = Iter.singleton(lowRecord);
+            } else {
+                // Otherwise need to scan the whole page
+                iter = records.getRecordBuffer().iterator();
+            }
         }
+        records.bpTree.finishReadBlkMgr();
         return iter;
     }
 
@@ -161,7 +169,7 @@ class BPTreeDistinctKeyPrefixIterator implements Iterator<Record> {
             if (it == null || !it.hasNext()) {
                 continue;
             }
-            BPTreePage p = it.next();
+            it.next();
             stack.push(it);
         }
         BPTreePage p = steps.get(steps.size() - 1).page;
--- a/jena-db/jena-dboe-trans-data/src/main/java/org/apache/jena/dboe/trans/bplustree/BPTreeDistinctKeyPrefixIterator.java
+++ b/jena-db/jena-dboe-trans-data/src/main/java/org/apache/jena/dboe/trans/bplustree/BPTreeDistinctKeyPrefixIterator.java
@@ -19,6 +19,7 @@
 package org.apache.jena.dboe.trans.bplustree;
 
 import org.apache.jena.atlas.iterator.Iter;
+import org.apache.jena.atlas.lib.Bytes;
 import org.apache.jena.atlas.lib.InternalErrorException;
 import org.apache.jena.dboe.base.record.Record;
 import org.apache.jena.dboe.trans.bplustree.AccessPath.AccessStep;
@@ -34,7 +35,24 @@ class BPTreeDistinctKeyPrefixIterator implements Iterator<Record> {
     static Logger log = LoggerFactory.getLogger(BPTreeDistinctKeyPrefixIterator.class);
 
     public static Iterator<Record> create(BPTreeNode node, int keyPrefixLength) {
-        return new BPTreeDistinctKeyPrefixIterator(node, keyPrefixLength);
+        if (keyPrefixLength < 1) {
+            throw new IllegalArgumentException("keyPrefixLength must be >= 1");
+        } else if (keyPrefixLength > node.params.recordFactory.keyLength()) {
+            throw new IllegalArgumentException(
+                    "Maximum keyPrefixLength for this B+Tree is " + node.params.recordFactory.keyLength());
+        }
+        Record minRecord = node.minRecord();
+        Record maxRecord = node.maxRecord();
+        // If there's no records just return a null iterator
+        if (minRecord == null) {
+            return Iter.nullIter();
+        }
+        // If there's only a single distinct prefix then just return a singleton
+        if (Arrays.compare(minRecord.getKey(), 0, keyPrefixLength, maxRecord.getKey(), 0, keyPrefixLength) == 0) {
+            return Iter.singleton(minRecord);
+        }
+        // Otherwise need the full iterator
+        return new BPTreeDistinctKeyPrefixIterator(node, keyPrefixLength, minRecord, maxRecord);
     }
 
     // Convert path to a stack of iterators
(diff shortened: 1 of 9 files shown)


afs:
It is great to see this but it is now very close to 4.7.0.

Aklakan:
Thanks for this work! I recently had to fight with postgresql because native skip scans are a missing issue there. The performance improvement (down to a few milliseconds) is what I experienced there. Also, e.g. one of the postgres extensions greatly advertised their skip scan implementation in [this blog post](https://www.timescale.com/blog/how-we-made-distinct-queries-up-to-8000x-faster-on-postgresql/).

rvesse:
> It is great to see this but it is now very close to 4.7.0.

Yeah I think this one will have to push out and be given more time to be refined and settled down.  The change breaks a few tests so there's clearly some corner cases I am not catching correctly yet!

SimonBin:
very nice work!!

```
SELECT ?e {
  GRAPH ?g {
    ?e spatial:withinBoxGeom ("POLYGON((19.49 50.62,26.87 50.626,26.87 46.43,19.49 46.43,19.49 50.62))"^^geo:wktLiteral 10)
  }
}
```

0.2 seconds

```
SELECT (count(distinct ?thing) as ?count)
WHERE {
    graph ?g {
        ?thing a/rdfs:subClassOf* coy:Powerplant .
    }
}
```

0.1 seconds


`select distinct ?g { graph ?g {} }` unchanged

jena-tdb2/src/main/java/org/apache/jena/tdb2/store/DatasetGraphTDB.java
```diff
@@ -157,4 +178,61 @@ public NodeTupleTable chooseNodeTupleTable(Node graphNode) {
             // Includes Node.ANY and union graph
             return getQuadTable().getNodeTupleTable();
     }
+
+    private static class DistinctKeyPrefixIterator implements Iterator<Node> {
```
SimonBin, line 182: leftover?

SimonBin:
>  The change breaks a few tests
didnt see any?

rvesse:
> > The change breaks a few tests
> 
> didn't see any?

Already fixed them with my force push

jena-tdb2/src/main/java/org/apache/jena/tdb2/store/DatasetGraphTDB.java
rvesse: Yes this was a first pass implementation before I went with the low level iterator

- [ ] Remove discarded implementation prototype

SimonBin:
the quick + dirty way to make select ?g { graph ?g {} } fast:

https://github.com/AKSW/jena/commit/2b064dfbd4e32dbdd6c49f8a5031c3bd839c86e3?w=1#diff-e82707e0fbbe7ca027b63ecbac2d24d5607432e492061ff8d1cc3f5c6318e2feR162-R167

not sure what to do about the filter though.......

Aklakan:
Imho there should also be some optimization for mapping the common query DISTINCT ?g { GRAPH ?g { ?s ?p ?o } } to effectively OpDatasetNames so that the skip scan can be leveraged further.

* Where would be the best place for that? Maybe a new optimizer under `org.apache.jena.sparql.algebra.optimize`?
* Is there a mechanism to ask a DatasetGraph's metadata for whether empty graphs are allowed (maybe a context attribute?) - such that in this case the optimizer would refuse the transform?

afs:
> the common query `DISTINCT ?g { GRAPH ?g { ?s ?p ?o } }`

Is it? Or is it a mistake for `GRAPH ?g {}`?

> Would it make sense to handle it as a query rewrite

Adding too many "maybe" optimizations slows down fast, small queries. (We know this from BSBM.)

`GRAPH ?g {}` is distinct graph names.

The implementation exception might be `DatasetGraphMapLink` and `DynamicDatasetGraph`.

In `DatasetGraphMap.listGraphNodes`:
```
    // Hide empty graphs.
```
so it looks like it is a matter of copying that to `DatasetGraphMapLink` (caveat inference graphs).

`DynamicDatasetGraph` might be able to do better:: do `listGraphNames` on the wrapped dataset and filter. But this might be worse. Numbers matter.

The right thing to do is address in the implementations in a separate PR, starting with some test cases.

Aklakan:
> Is it? Or is it a mistake for `GRAPH ?g {}`?

I really meant `GRAPH ?g { ?s ?p ?o} `. On \*cough\* DBpedia:
*  [`GRAPH ?g { }`](https://dbpedia.org/sparql?default-graph-uri=&qtxt=SELECT+%3Fg+{+GRAPH+%3Fg+{+}+}&format=text%2Fhtml&timeout=10000&signal_unconnected=on) -> no results
* [`GRAPH ?g { ?s ?p ?o }`](https://dbpedia.org/sparql?default-graph-uri=&qtxt=SELECT+DISTINCT+%3Fg+{+GRAPH+%3Fg+{+%3Fs+%3Fp+%3Fo+}+}+LIMIT+5&format=text%2Fhtml&timeout=10000) -> expected results

afs:
> On *cough* DBpedia:

The SPARQL endpoint or you local load into TDB2 with this PR applied?

Aklakan:
> The SPARQL endpoint or you local load into TDB2 with this PR applied?

On the SPARQL endpoint - the graph patterns in my post are links to DBpedia.

afs:
> So the "portable" query is the spo variant.

It's not portable - it's a workaround. It might be a poor choice on another store.

There's an issues list for Virtuoso and the user list for Virtuoso is on SourceForge - has it been reported?
The users list is where they ask for bug reports judging by the SO responses of "use the emailing list".

We're not here to fix DBpedia. It has several deviations from the specification.

Jena is open and you can submit reports - that can be overused. Email would be better.
Time spent looking at Jena code was wasted. Sigh. That's another step to 4.7.0 delayed.

jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/SolverLibTDB.java
```diff
@@ -145,24 +147,49 @@ static QueryIterator testForGraphName(DatasetGraphTDB ds, Node graphNode, QueryI
...
+            // Have a generic Graph based index, can at least optimise slightly by only needing distinct adjacent as we
+            // know the same Graph Names will be stored in a sequence within the index. This also keeps memory footprint
+            // for the distinct constant to a single Graph Name
             iter1 = idx.find(TupleANY);
-            needDistinct = false;
+            distinctMode = GraphNamesDistinctMode.ADJACENT;
+        } else {
+            // Fall back to a find all which will use the most appropriate index available
+            // Need full distinct calculation as no guarantee that the index is ordered with respect to the graph name
+            // Memory usage will scale with number of distinct graph names though this is a
```
rvesse, line 190: - [ ] Fix incomplete comment

jena-db/jena-dboe-trans-data/src/test/java/org/apache/jena/dboe/trans/bplustree/TestBPTreeDistinctKeys.java
```diff
@@ -0,0 +1,337 @@
...
+
+    @Test
+    public void bptree_distinct_by_key_11() {
+        BPlusTree bpt = makeRangeIndex(this.treeOrder, 0);
+        RecordFactory factory = bpt.getRecordFactory();
+        for (int key : randomData) {
+            bpt.insert(factory.create(Bytes.packInt(key)));
+        }
+
+        // Find only the keys with a unique 3 byte prefix
```
rvesse, line 284:
```suggestion
        // Find only the first keys with a unique 3 byte prefix
```

jena-db/jena-dboe-trans-data/src/test/java/org/apache/jena/dboe/trans/bplustree/TestBPTreeDistinctKeys.java
```diff
@@ -0,0 +1,337 @@
...
+
+    @Test
+    public void bptree_distinct_by_key_10() {
+        BPlusTree bpt = makeRangeIndex(this.treeOrder, 0);
+        RecordFactory factory = bpt.getRecordFactory();
+        for (int key : randomData) {
+            bpt.insert(factory.create(Bytes.packInt(key)));
+        }
+
+        // Find only the keys with a unique 2 byte prefix, since we already ensured with our data generation that the
```
rvesse, line 261:
```suggestion
        // Find only the first keys with a unique 2 byte prefix, since we already ensured with our data generation that the
```

jena-db/jena-dboe-trans-data/src/test/java/org/apache/jena/dboe/trans/bplustree/TestBPTreeDistinctKeys.java
```diff
@@ -0,0 +1,337 @@
...
+        this.randomSize = randomSize;
+        this.treeOrder = treeOrder;
+
+        // Generate random data
+        Random random = new Random();
+        randomData = new ArrayList<>();
+        int maxKey = (int) Math.pow(2, 24);
+        while (randomData.size() < this.randomSize) {
+            int x = random.nextInt(maxKey);
+            if (!randomData.contains(x)) {
```
rvesse, line 88: - [ ] Add explanatory comments about the random data generation

rvesse:
Now 4.7.0 is out we should be able to get this reviewed and merged so that users have time to start testing the updated SNAPSHOTs with this improvement

Aklakan:
We have the skip scan in use in a graph with around 1 billion triples and graph listings are super fast :+1: 

What would be eventually needed is also make this feature publicly accessible in the various  Tuple/DatasetGraph interfaces.
My proposal looks like this:

```java
// D = domain tuple type (e.g. Quad or Tuple<NodeId), C = component type (e.g. Node or NodeId)
interface QuadStore<D, C> {
  TupleStreamer<D, C> find(C g, C s, C p, C o, boolean distinct, int ... projectedColumns);
}

interface TupleStreamer<D, C> {
  Iterator<C> asComponents(); // e.g. Node or NodeIds
  Iterator<D> asDomainTuples(); e.g. Quad
  Iterator<Tuple<C>> asGenericTuples();
}
```

```sparql
SELECT DISTINCT ?p { GRAPH ?g { ?s ?p ?o } }
```

could then map to a `find()` call such as
```
Iterator<Node> distinctPredicates = datasetGraph.find(ANY, ANY, ANY, ANY, true, 2).asComponents();
```

Furthermore, I wonder if the way TDB indexes data would be suitable for a skip scan for the case to retrieve a resource's distinct predicates:
```sparql
SELECT DISTINCT ?p { GRAPH ?g { <concreteS> ?p ?o } }
```
The background is, that we have resources with 4mio+ statements (yeah not that usual) where a scan for distinct predicates takes seconds - maybe with the skip scan it would also be possible to speed this case up?

rvesse:
> We have the skip scan in use in a dataset with around 1 billion triples and graph listings are super fast 👍
> 
> What would be eventually needed is also make this feature publicly accessible in the various Tuple/DatasetGraph interfaces. My proposal looks like this but maybe someone has already better ideas:
> 
> ```java
> // D = domain tuple type (e.g. Quad or Tuple<NodeId), C = component type (e.g. Node or NodeId)
> interface TupleMatcher4<D, C> {
>   TupleStreamer<D, C> find(C g, C s, C p, C o, boolean distinct, int ... projectedColumns);
> }
> 
> interface TupleStreamer<D, C> {
>   Iterator<C> asComponents(); // e.g. Node or NodeIds
>   Iterator<D> asDomainTuples(); // e.g. Quad
>   Iterator<Tuple<C>> asGenericTuples();
> }
> ```

I don't suspect that this sort of low level execution optimisation is ever going to bubble up into the high level end-user APIs like `DatasetGraph`

I don't disagree that this iterator could be used to optimise execution of other query patterns **but** the goal here is to start small and incrementally improve.

> 
> This way a request for e.g. distinct predicates:
> 
> ```sparql
> SELECT DISTINCT ?p { GRAPH ?g { ?s ?p ?o } }
> ```
> 
> could then map to a `find()` call such as
> 
> ```
> Iterator<Node> distinctPredicates = datasetGraph.find(ANY, ANY, ANY, ANY, true, 2).asComponents();
> ```
> 
> Furthermore, I wonder if the way TDB indexes data would be suitable for a skip scan for the case to retrieve a resource's distinct predicates:
> 
> ```sparql
> SELECT DISTINCT ?p { GRAPH ?g { <concreteS> ?p ?o } }
> ```
> 
> The background is, that we have resources with 4mio+ statements (yeah not that usual) where a scan for distinct predicates takes seconds - maybe with the skip scan it would also be possible to speed this case up?

So as I used to tell a solution architect I worked closely with in a past $dayjob that optimisation is fundamentally a trade off. The goal of a general purpose optimiser is to apply optimisations that are generally useful to most users and most data yielding performance improvements for the general case.

Detecting whether an optimisation is applicable or not has a non-zero cost to it and for some query/dataset patterns that are unusual, e.g. a subject with 4 million statements, a general purpose optimiser shouldn't try to optimise for that because it's so outside it's normal expectations.

I would suggest for this kind of optimisation, where you have a specific query pattern that runs poorly on your dataset(s), you consider creating your own custom optimiser (based off ARQ's default one) adding in optimisations for specific query patterns you need specialised optimisation for.  You can transform those into custom `Op` instances (derived from ARQs `OpExt` extension point) and provide suitable `eval()` implementations that can call into the relevant low level APIs as appropriate.

That gives you a way to experiment with some of these things outside of the Jena codebase and then potentially contribute them back later if they prove to be of more general value.  But right now it seems like there's a lot of stuff that's very specific to your use cases that may not be generally applicable.

afs (approved):
One unused import.

jena-db/jena-dboe-trans-data/src/main/java/org/apache/jena/dboe/trans/bplustree/BPTreeDistinctKeyPrefixIterator.java
```diff
@@ -0,0 +1,221 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.jena.dboe.trans.bplustree;
+
+import org.apache.jena.atlas.iterator.Iter;
+import org.apache.jena.atlas.lib.Bytes;
```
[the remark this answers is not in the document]
rvesse: Fixed

PR #1683, opened 2022-12-16
configurable yasgui prefix and query

SimonBin:
GitHub issue resolved #1680

Pull request Description:

fusekiUI will look for a dataset named yasgui-config to load Example Queries and Prefixes based on the current dataset name

----

 - [ ] Tests are included.

----

### DIFF
--- a/jena-fuseki2/jena-fuseki-ui/src/services/fuseki.service.js
+++ b/jena-fuseki2/jena-fuseki-ui/src/services/fuseki.service.js
@@ -22,6 +22,39 @@ import { BUS } from '@/events'
 
 const DATASET_SIZE_QUERY_1 = 'select (count(*) as ?count) {?s ?p ?o}'
 const DATASET_SIZE_QUERY_2 = 'select ?g (count(*) as ?count) {graph ?g {?s ?p ?o}} group by ?g'
+const CONF_DATASET_NAME = 'yasgui-config'
+const CONF_EXAMPLE_QUERIES_QUERY = `PREFIX conf: <https://yasgui.triply.cc/config#>
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#first>
+PREFIX sd: <http://www.w3.org/ns/sparql-service-description#>
+PREFIX list: <http://jena.apache.org/ARQ/list#>
+JSON {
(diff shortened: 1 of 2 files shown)


SimonBin:
if it's decided that we need to avoid using JSON then I can work on manual unwrapping json-resultset, but I wanted to open the PR first

afs:
Just to be clear - this is too late for 4.7.0.

SimonBin:
no worries, there's always a next version

kinow:
Not necessarily what we are doing here, but I thought interesting that [prez](https://github.com/RDFLib/prez#data) uses a `triplestore` dataset that documents the existing datasets in the triple store. Then, through existing vocabularies (dcat/prof/skos/etc) it tries to automatically know what users can do with the dataset, and it renders templates for the user (they are working on a python + vue web app refactoring).

Maybe the idea of a global "triplestore" dataset of sorts, storing some data common to all vocabularies like example queries, and maybe settings, could be helpful for this issue too.

afs:
Fuseki admin has a system database. Its not used much (flexible data representation not required! and on disk file are easier or user to work with). That could be repurposed.

kinow:
> Fuseki admin has a system database. Its not used much (flexible data representation not required! and on disk file are easier or user to work with). That could be repurposed.

Didn't know that. +1 !

SimonBin:
I didn't see the system database in the dataset list so it might be harder to edit the config from the UI(?)

afs:
For managing fixed structures, an RDF database is probably over-engineering.

A basic HTTP PUT/GET to a "prefix store" (persisted by a text file - i.e. human editable) and a URL scheme may be better (simpler, easier to maintain, does not require RDF handling in the UI code (potentially yet more dependencies)).

But please ensure the parts of the UI that provide the "workbench" also work when just a SPARQL query endpoint is the target - i.e. it works for any triplestore and that includes a Fuseki server where the UI user is not in charge of. Specific features may not be available i a graceful way.

Yasgui has some configuration features - presumably it uses browser-local storage.

kinow:
I think this one requires further discussion and thinking on how to implement it. Also some work on YASGUI integration with Vue.js.

SimonBin:
sure, that's why I opened the discussion :-)

for the example in this pull request, it provides the possibility to change the prefix and sample query list per dataset in the Fuseki UI, and you can edit it using the Fuseki UI built-in graph editor, so that was a plus for us

PR #1711, opened 2023-01-07
GH-1710 Add note about isomorphism support in RDF-star

### LINKED ISSUE #1710: Graph.isIsomorphicWith doesn't work with blank nodes in quoted triples
### Version

4.7.0

### What happened?

When checking for isomorphism between two `Graph`s, Jena is supposed not to check for exact equality between blank nodes, but rather use a fancier isomorphism algorithm. This, however, seems not to be not true for quoted triples.

File `data/source.nt`:
```
_:b0 <http://example/p> <http://example/o> .
<< _:b0 <http://example/p> <http://example/o> >> <http://example/q> "ABC" .
```

Code:
```java
package org.example;

import org.apache.jena.riot.Lang;
import org.apache.jena.riot.RDFDataMgr;

public class Main {
        var g1 = RDFDataMgr.loadGraph("data/source.nt", Lang.NT);
        var g2 = RDFDataMgr.loadGraph("data/source.nt", Lang.NT);
        System.out.println(g1.isIsomorphicWith(g2));
    }
}
```

This prints "false" instead of "true".

I encountered this problem when writing a test suite where I wanted to ensure that the input and output graphs are isomorphic. Quoted blank nodes seem to be the only case where this fails (all other RDF-star works fine).

I also tried to circumvent this by first converting both graphs into their reified forms (with `org.apache.jena.system.RDFStar`), but this failed too.

I have no idea how to even attempt fixing this, so I'll probably be of not much more help. But at least I can say that the issue is not critical to me :)

_No response_

I don't understand the relevant internals, no.

Ostrzyciel:
GitHub issue resolved #1710

Pull request Description:
Updated the javadoc of the `Graph` class to make it clear that the isomorphism algorithm implemented there does not fully support RDF-star. The reader is pointed to another, slower implementation that correctly implements this.

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/graph/Graph.java
+++ b/jena-core/src/main/java/org/apache/jena/graph/Graph.java
@@ -154,6 +154,12 @@ public interface Graph
 	 * <a href="http://www.w3.org/TR/rdf-concepts#section-Graph-syntax">
      * http://www.w3.org/TR/rdf-concepts#section-Graph-syntax
      * </a>
+     *
+     * Note: this implementation does not handle correctly blank nodes in
+     * quoted triples (RDF-star). If you need to work with RDF-star,
+     * use the slower implementation in
+     * {@link org.apache.jena.sparql.util.IsoMatcher}.
+     *
 	 * @param g Compare against this.
 	 * @return boolean True if the two graphs are isomorphic.
 	 */


kinow (approved):

afs:
Thanks!

PR #1728, opened 2023-01-24
fix(sec): upgrade org.apache.shiro:shiro-web to 1.11.0

wjl110:
### What happened？
There are 1 security vulnerabilities found in org.apache.shiro:shiro-web 1.10.1
- [CVE-2023-22602](https://www.oscs1024.com/hd/CVE-2023-22602)

### What did I do？
Upgrade org.apache.shiro:shiro-web from 1.10.1 to 1.11.0 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
     <ver.lucene>9.4.2</ver.lucene>
 
     <ver.jetty>10.0.13</ver.jetty>
-    <ver.shiro>1.10.1</ver.shiro>
+    <ver.shiro>1.11.0</ver.shiro>
 
     <ver.protobuf>3.21.12</ver.protobuf>
     <ver.libthrift>0.17.0</ver.libthrift>


kinow (approved):
Other changes https://shiro.apache.org/blog/2023/01/13/apache-shiro-1110-released.html

You must be logged in to view the Shiro changes in that link, here's a screenshot for others without a login:

image

rvesse:
Worth noting that [CVE-2023-22602](https://www.cve.org/CVERecord?id=CVE-2023-22602) applies specifically to the combination of Apache Shiro and Spring Boot which aren't used in Jena anywhere

Regardless we should still take the upgrade

afs:
I'm surprises dependabot hasn't generated a change. It sent one another project but failed to notice when it was done and the PR didn't close.

As @rvesse says, it does not affect Jena directly but it is good to update.

> Ideally, no insecure libs should be used.

That is impossible for javascript!
Dependencies don't update to pick up their own dependencies for security patches.

kinow:
>That is impossible for javascript!
>Dependencies don't update to pick up their own dependencies for security patches.

:fire: sad but true

PR #1752, opened 2023-02-11
GH-1697: Buffering dataset, buffering graph

### LINKED ISSUE #1697: Buffering Datasets
### Version

4.7.0

### Feature

A buffering dataset wraps another dataset. It collects changes (add and delete for triples, quads and prefixes), presents a dataset API that is the base dataset as if the chnages had been applied. It has a flush operation that executes the changes on the base dataset (normally, as a single transaction).

Use case: _adding true transactional abort to any dataset_

General purpose datasets can not provide MR+SW transaction and can not provide abort. (see [`DatasetGraph.supportsTransactionAbort`](https://jena.apache.org/documentation/javadoc/arq/org.apache.jena.arq/org/apache/jena/sparql/core/DatasetGraph.html#supportsTransactionAbort()). By delaying changes to the underlying dataset until a commit is performed, abort can be provided by using a buffered dataset that stores the changes until the commit; an abort is simply forget the buffered changes.

Use case: _batching many small transactions into a single transaction on the base storage_

Many small updates (add/delete one or two triples) can result in more work being in the transaction mechanism than in the updates themselves. A buffered dataset (with an associated journal to make each small transaction safe - see "RDF Patch".) 

Use case: _Rule-derived (inference) overlay_

The buffering layer can keep inferred triples separately from the base data.

_Notes_

* For the use cases above, the buffering dataset uses JVM heap to store the ongoing differences. If larger than that, it is probably better to use a proper persistent dataset directly.

* With a switchable buffering dataset, the update of the underlying dataset may be done asynchronously.

Interfaces: `DatasetGraphBuffering`, `BufferingCtl`.

Classes: `BufferingDatasetGraph`, `BufferingGraph`, `BufferingPrefixMap`, `BufferingPrefixMapping`

Once the machinery design is done, providing a standalone graph case is simple. The general design is a pattern where there is an in-memory `DatasetGraph` that records additions, and a set of items deleted, that is used to filter the `find()` iterator.

Yes

afs:
GitHub issue resolved #1697

Pull request Description:

Buffering DatasetGraph (integrated with the dataset transaction lifecycle).
Buffering Graph (not integrated with the graph transaction lifecycle).

----

 - [x] Tests are included.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/other/G.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/other/G.java
@@ -18,18 +18,23 @@
 
 package org.apache.jena.riot.other;
 
+import static org.apache.jena.graph.Node.ANY;
+
 import java.util.*;
 import java.util.function.Consumer;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 
 import org.apache.jena.atlas.iterator.Iter;
 import org.apache.jena.graph.*;
+import org.apache.jena.rdf.model.impl.Util;
 import org.apache.jena.riot.out.NodeFmtLib;
 import org.apache.jena.sparql.core.DatasetGraph;
 import org.apache.jena.sparql.core.Quad;
 import org.apache.jena.sparql.graph.NodeConst;
 import org.apache.jena.sparql.util.graph.GNode;
 import org.apache.jena.sparql.util.graph.GraphList;
+import org.apache.jena.util.IteratorCollection;
 import org.apache.jena.util.iterator.ExtendedIterator;
 
 /**
@@ -65,6 +70,7 @@ public class G {
     public static boolean isResource(Node n)    { return n != null && (n.isURI()||n.isBlank()); }
     public static boolean isNodeTriple(Node n)  { return n != null && n.isNodeTriple(); }
     public static boolean isNodeGraph(Node n)   { return n != null && n.isNodeGraph(); }
+    public static boolean isNullOrAny(Node n)   { return n == null || Node.ANY.equals(n); }
 
     /** Convert null to Node.ANY */
     public static Node nullAsAny(Node x) { return nullAsDft(x, Node.ANY); }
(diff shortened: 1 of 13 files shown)


jena-arq/src/main/java/org/apache/jena/riot/other/G.java
```diff
@@ -845,4 +850,110 @@ public static <X> X calcTxn(Graph graph, Supplier<X> action) {
...
+        // Two literals.
+        if ( ! Util.isLangString(data) || ! Util.isLangString(match) )
+            // Not both strings with lang tags.
+            return (match==null) || (match == Node.ANY) || match.equals(data) ;
+
+        // Language tags compare case insensitively.
+        String lex1 = data.getLiteralLexicalForm();
+        String lex2 = data.getLiteralLexicalForm();
+        String lang1 = data.getLiteralLanguage();
+        String lang2 = data.getLiteralLanguage();
```
kinow, line 956: Doing just a quick review as I don't know much about this part of the code (and it's late) so this comment could be something silly, :bow: 

@afs, won't `lex1` and `lex2` be identical? I haven't checked the code to see if `getLiteralLexicalForm` will return a new object, but maybe one was supposed to use `match` instead? Ditto for the `lang` vars?

jena-arq/src/main/java/org/apache/jena/system/buffering/BufferingPrefixMapping.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Override
+    protected void remove(String prefix) {
+        deleted.add(prefix);
+        added.removeNsPrefix(prefix);
+    }
+
+    @Override
+    public int numPrefixes() {
+        return added.numPrefixes() - deleted.size() + other.numPrefixes();
```
kinow, line 87: Was curious if `numPrefixes` could return a negative value but I think that's not possible (or at least I couldn't come up with a scenario for that).

What about concurrency? Is there any risk of `numPrefixes` accessing `deleted.size()` while it's updated concurrently, and does it matter?

kinow (commented):
Don't know much about this part of the code, so just read docs and asked a couple things (that could be silly, feel free to ignore it :+1: )

jena-arq/src/main/java/org/apache/jena/riot/other/G.java
afs: Good catch - 1 should be "match".

jena-arq/src/main/java/org/apache/jena/system/buffering/BufferingPrefixMapping.java
afs: All classes `Buffering*` are not thread safe. The idea is a thread wraps a graph/dataset to store up changes. graph support isn't transaction aware - dataset support is.

If anything more, then multithreaded apps need to use the base dataset properly!

kinow: Thanks Andy!

kinow (approved):

PR #1802, opened 2023-03-17
GH-1801: Prevents needless materialization of lexical forms

### LINKED ISSUE #1801: NodeValue needlessly materializes lexical forms
### Change

NodeValue's `_setByValue` method only handles xsd datatypes however it eagely materializes the lexical form even of non-xsd namespace'd datatypes. This introduces a noticeable performance overhead (with my workload of many small json objects around 5-10%) when dealing with datatype extensions such as large geometries or json objects which are only used as intermediary values.

NodeValue itself bears the following comment

> 2. Conversely, delaying turning a value into a graph node is
> valuable because intermediates, like the result of 2+3, will not
> be needed as nodes unless assignment (and there is no assignment
> in SPARQL even if there is for ARQ).
> Node level operations like str() don't need a full node.

The simple solution is to defer materialization of the lexical form after having ensured the given Node has a datatype in the xsd namespace.

As a question, I wonder if it is really necessary for `_setByValue` to always go via the lexical form for all XSD types, or whether as a future improvement it would be possible to reuse the LiteralLabel's Java object.

Profile without enhancement:
image

Profile with enhancement:
image

Yes

Aklakan:
GitHub issue resolved #1801 

Defer materialization of lexical form after it is ensured that the node's datatype is in the XSD namespace.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/expr/NodeValue.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/expr/NodeValue.java
@@ -671,7 +671,6 @@ public abstract class NodeValue extends ExprNode
         if ( NodeUtils.hasLang(node) )
             return new NodeValueLang(node) ;
         LiteralLabel lit = node.getLiteral() ;
-        String lex = lit.getLexicalForm() ;
         RDFDatatype datatype = lit.getDatatype() ;
 
         // Quick check.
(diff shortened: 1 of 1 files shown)


rvesse (approved):
LGTM, seems like a simple but useful improvement

Probably want @afs to look to at this to make sure there's not any unforeseen consequences of this

kinow (approved):
Looks good to me! Thanks!

jena-arq/src/main/java/org/apache/jena/sparql/expr/NodeValue.java
afs, line 1: While you are at it, please also change line new-690 s/lit.getLexicalForm()/lex/.

PR #1805, opened 2023-03-18
GH-1804: Configurable fuseki modules

### LINKED ISSUE #1804: Fuseki modules: allow a server to be built using a given choice of modules.
Currently, Jena 4.7.0, [Fuseki modules](https://jena.apache.org/documentation/fuseki2/fuseki-modules) are loaded during Jena start-up. The module itself is a single object with methods executed during the build step for Fuseki Main.

There is a system-wide set of modules with code loaded using the Java ServiceLoader. Modules can be added without needing any Java code by placing jars on the classpath.

A module is can be a shaded jar with necessary dependencies but it must not include the server Fuseki code. It is quite fragile to mange dependency exclusion. It is version sensitive. 

It is an extra step where errors can arise such as failure to supply the dependencies or get out of step with versions.

This issue is for an additional way to use Fuseki modules; it does require Java code.

By adding the ability to specific a list of modules to the build step, servers built from Java code can have modules which are not handled by ServiceLoader but are regular object instances. No special step to build a module jar; dependency management is all part of the development (maven/gradle).

The default list would be the system-wide loaded list of modules  - in other words, the default behaviour is as before.

### Are you interested in contributing a solution?

Yes

afs:
GitHub issue resolved #1804 

Pull request Description:
Provide programmatic control of modules when building a server.

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
+++ b/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
@@ -49,9 +49,7 @@ import org.apache.jena.fuseki.auth.AuthPolicy;
 import org.apache.jena.fuseki.build.FusekiConfig;
 import org.apache.jena.fuseki.ctl.*;
 import org.apache.jena.fuseki.main.cmds.FusekiMain;
-import org.apache.jena.fuseki.main.sys.FusekiErrorHandler;
-import org.apache.jena.fuseki.main.sys.FusekiModuleStep;
-import org.apache.jena.fuseki.main.sys.JettyLib;
+import org.apache.jena.fuseki.main.sys.*;
 import org.apache.jena.fuseki.metrics.MetricsProviderRegistry;
 import org.apache.jena.fuseki.server.*;
 import org.apache.jena.fuseki.servlets.*;
@@ -151,15 +149,18 @@ public class FusekiServer {
     private int httpsPort;
     private final String staticContentDir;
     private final ServletContext servletContext;
+    private final FusekiModules modules;
 
     private FusekiServer(int httpPort, int httpsPort, Server server,
                          String staticContentDir,
+                         FusekiModules modules,
                          ServletContext fusekiServletContext) {
-        this.server = server;
+        this.server = Objects.requireNonNull(server);
         this.httpPort = httpPort;
         this.httpsPort = httpsPort;
         this.staticContentDir = staticContentDir;
-        this.servletContext = fusekiServletContext;
+        this.servletContext = Objects.requireNonNull(fusekiServletContext);
+        this.modules = Objects.requireNonNull(modules);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/FusekiServer.java
```diff
@@ -1037,6 +1051,26 @@ public Builder addFilter(String pathSpec, Filter filter) {
             return this;
         }
 
+        /**
+         * Set the {@link FusekiModule Fuseki Module} for a server.
+         * If no modules are added to a builder, then the system-wide default set (found by loading FusekiModule
+         * via Java's {@link ServiceLoader} mechanism, is used.
```
kinow, line 1057: missing closing `)`?

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/sys/FusekiModules.java
```diff
@@ -18,67 +18,56 @@
...
-import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Consumer;
 
-import org.apache.jena.base.module.Subsystem;
-
-/** Registry of modules */
+/**
+ * List of {@linkplain FusekiModule Fuseki modules}.
+ * This is the immutable unit of a modules for a server.
```
kinow, line 26: s/of a modules/of modules?

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/sys/FusekiModules.java
```diff
@@ -18,67 +18,56 @@
...
+ * <p>
+ * @see FusekiModulesLoaded
+ */
 public class FusekiModules {
 
-    private static Object lock = new Object();
+    /** A Fuseki module with no members. */
+    public static FusekiModules empty = FusekiModules.create();
```
kinow, line 33: Make it `final` maybe? Doesn't seem to be changed in the class methods.

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/sys/FusekiModules.java
```diff
@@ -18,67 +18,56 @@
...
-        registry = new ArrayList<>();
-        subsystem = new Subsystem<FusekiModule>(FusekiModule.class);
-        subsystem.initialize();
-        synchronized(lock) {
-            subsystem.forEach(registry::add);
-        }
+    private final List<FusekiModule> modules;
+
+    private FusekiModules(FusekiModule ... modules) {
+        this.modules = List.of(modules);
```
kinow, line 48: NPE here if `modules` is `null` (coming from the public `#create` method).

kinow (approved):
Couple comments, but no blockers :+1:

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/sys/FusekiModules.java
afs: Done. Always good to check early - this is unlikely to be performance critical code.

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/sys/FusekiModules.java
afs: Yes!

PR #1806, opened 2023-03-19
Misc

afs:
* github for Windows with Java17 :: avoid an issue with javadoc generation when running maven with Java11.
* Improvements for running the original ARP when using RIOT.

----

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -30,4 +30,4 @@ jobs:
         distribution: 'temurin'
         java-version: '11'
     - name: Build with Maven
-      run: mvn -B verify -Pdev --file pom.xml
+      run: mvn -B -o --file pom.xml -Dmaven.javadoc.skip=true verify
--- a/jena-arq/src/main/java/org/apache/jena/sparql/expr/E_IRI.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/expr/E_IRI.java
@@ -52,11 +52,11 @@ public class E_IRI extends ExprFunction1 {
     }
 
     public E_IRI(String parserBaseURI, Expr relExpr) {
+        // The base is the base of the parser at the time this E_IRI was created.
         this(parserBaseURI, relExpr, sparqlPrintName, sseFunctionName);
     }
 
     protected E_IRI(String baseStr, Expr relExpr, String sparqlName, String sseName) {
-        //super(relExpr, sparqlName, sseName);
         super(relExpr, sseName);
         this.parserBase = baseStr;
         this.relExpr = relExpr;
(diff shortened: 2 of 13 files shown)


.github/workflows/maven.yml
```diff
@@ -30,4 +30,4 @@ jobs:
         distribution: 'temurin'
         java-version: '11'
     - name: Build with Maven
-      run: mvn -B verify -Pdev --file pom.xml
+      run: mvn -B -o --file pom.xml -Dmaven.javadoc.skip=true verify
```
kinow, line 33: Shouldn't we leave one platform with javadoc at least? Or is that being built on Jenkins?

jena-arq/src/main/java/org/apache/jena/sparql/expr/E_IRI.java
```diff
@@ -74,14 +74,15 @@ protected NodeValue evalSpecial(Binding binding, FunctionEnv env) {
...
     /*package*/ static NodeValue resolve(NodeValue relative, String baseIRI, FunctionEnv env) {
         if ( baseIRI == null ) {
-            // Legacy
             if ( env.getContext() != null ) {
                 Query query = (Query)env.getContext().get(ARQConstants.sysCurrentQuery);
                 if ( query != null )
                     baseIRI = query.getBaseURI();
+                // If still null, NodeFunctions.iri will use the system base.
+//                if ( baseIRI == null )
+//                    baseIRI = IRIs.getBaseStr();
```
kinow, line 83: I think these two last lines can be removed? Looks like not needed anymore?

kinow (approved):

.github/workflows/maven.yml
afs: The release has to use Java17 outputting Java11 byte code because of Javadoc probems. Fixes have not been fully backported to Java11.

That means one check is needed - does the code compile on Java11 - not the javadoc.

It' probably less of an issue these days but it used to be possible to have JDK library calls at the level of the compiler, not the target, and then they weren't available at runtime.

This job checks that Java11 can compile the code.

Jenkins does make Javadoc, as do the other github actions (Linux, MacOS). It's this one case (Windows+Java11) that seems more unstable.

jena-arq/src/main/java/org/apache/jena/sparql/expr/E_IRI.java
afs: It's not necessary, it does document an assumption about the behaviour of NodeFunctions.iri.

.github/workflows/maven.yml
kinow: >Jenkins does make Javadoc, as do the other github actions (Linux, MacOS). It's this one case (Windows+Java11) that seems more unstable.

:+1: thanks!

PR #1818, opened 2023-03-29
Misc

afs:
Pull request Description:
Miscellaneous small things accumulated while 

[Count triples in a quads stream](https://github.com/apache/jena/commit/9864de292da5b7808f9609fd545021ba6ebcb853) which is a bug fix for reporting upload stats in Fuseki.

[Fix generating file name when running on MS Windows](https://github.com/apache/jena/commit/e3c02a6612f0daad703176e9b3574e5773481c17) is test related and should make the GH/Windows action work.

[Document and test bnode-iri round trip](https://github.com/apache/jena/commit/a1cf0685c3bd8610dad76a24d9cd29cbc49fb718) adds library functions that are missing.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/riot/out/NodeFmtLib.java
+++ b/jena-arq/src/main/java/org/apache/jena/riot/out/NodeFmtLib.java
@@ -35,6 +35,7 @@ import org.apache.jena.riot.system.RiotChars;
 import org.apache.jena.shared.PrefixMapping;
 import org.apache.jena.sparql.ARQConstants;
 import org.apache.jena.sparql.core.Quad;
+import org.apache.jena.sys.JenaSystem;
 
 /** Presentation utilities for Nodes, Triples, Quads and more.
  * <p>
@@ -45,6 +46,8 @@ import org.apache.jena.sparql.core.Quad;
  */
 public class NodeFmtLib
 {
+    static { JenaSystem.init(); }
+
     // Replaces FmtUtils
     // See and use EscapeStr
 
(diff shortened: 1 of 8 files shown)


jena-arq/src/main/java/org/apache/jena/riot/out/NodeFmtLib.java
```diff
@@ -195,19 +198,24 @@ private static String displayStrNodes(Node...nodes) {
...
+     */
 
-    private static char LabelLeadingLetter = 'B';
+    private static final char LabelLeadingLetter = 'B';
 
     public static String encodeBNodeLabel(String label) {
-        StringBuilder buff = new StringBuilder();
+        // A UUID string is 36 characters including 4 dashes.
+        // Together with the leading 'B', a total of 45 characters + any additional encoding needed.
+        StringBuilder buff = new StringBuilder(48);
```
kinow, line 218: >A UUID string is 36 characters 

`36`

>including 4 dashes.

`36 + 4 = 40`

>Together with the leading 'B'

`40 + 1 = 41 ?`

>a total of 45 characters + any additional encoding needed.

`45`?

>new StringBuilder(48);

`48`? :sweat_smile: 

Is the `48` just to have some buffer?

jena-arq/src/main/java/org/apache/jena/riot/system/RiotLib.java
```diff
@@ -85,6 +85,16 @@ public static String blankNodeToIriString(Node node) {
         throw new RiotException("Not a blank node or URI");
     }
 
+    /**
+     * Convert a ARQ-encoded blank node URI nod to a blank node, otehrwise return the argument node unchanged.
```
kinow, line 89: s/a ARQ/an ARQ if we say it like ark/arc? I actually say it that way but no idea what's the correct way.

And I think the `nod` can be deleted?

And s/otehrwise/otherwise

jena-arq/src/main/java/org/apache/jena/riot/system/RiotLib.java
```diff
@@ -107,6 +117,15 @@ public static boolean isBNodeIRI(String iri) {
         return iri.startsWith(bNodeLabelStart);
     }
 
+    /**
+     * Test whether a node is an ARQ-encoded blank node IRI.
+     */
+    public static boolean isBNodeIRI(Node node) {
+        if ( ! node.isURI() )
+            return false;
+        return isBNodeIRI(node.getURI());
```
kinow, line 126: Or `return node.isURI() && isBNodeIRI(node.getURI())`

jena-arq/src/main/java/org/apache/jena/riot/out/NodeFmtLib.java
afs: The import thing is to avoid reallocation in the StringBuilder in common usage (UUID strings).

Dashes get X-encoded so the overall is 45 -- a dash is "X2D".

`878b5787-2c84-4624-a237-ad6fc35ec260`
becomes 
`B878b5787X2D2c84X2D4624X2Da237X2Dad6fc35ec260` which is 45.

So the buffer must be >=45 for UUIDs.

48 is a multiple of 8 (possible cache effect) and more than 45.

The label isn't required to be a UUID.

😄

jena-arq/src/main/java/org/apache/jena/riot/system/RiotLib.java
afs: Style choice. The `if` is a guard on the argument, then a test on the `getURI`. Think of it as lisp `(cond)` inspired‼️

I'm not a fan of (IMO) unnecessary brevity in compound expressions because multiple lines given multiple debug points.

jena-arq/src/main/java/org/apache/jena/riot/system/RiotLib.java
afs: Done

kinow (approved):

jena-arq/src/main/java/org/apache/jena/riot/out/NodeFmtLib.java
kinow: s/gerater/greater

PR #1831, opened 2023-04-01
GH-1830: Fuseki admin, create dataset - write instantiated template.

### LINKED ISSUE #1830: Fuseki/webapp writes the configuration file with unnecessary additional triples
### Version

4.7.0

### What happened?

When a new database is created in the UI, the configuration file is written to `run/configuration/<nbame>.ttl`. This contains the instantiated template but also additional triples (system definitions, some RDF statements). The latter are unnecessary.

_No response_

Yes

afs:
GitHub issue resolved #1830

Pull request Description:
Write the instantiated template as the configuration file.

----

### DIFF
--- a/jena-fuseki2/jena-fuseki-webapp/src/main/java/org/apache/jena/fuseki/mgt/ActionDatasets.java
+++ b/jena-fuseki2/jena-fuseki-webapp/src/main/java/org/apache/jena/fuseki/mgt/ActionDatasets.java
@@ -141,8 +141,8 @@ public class ActionDatasets extends ActionContainerItem {
 
         try {
             // Where to build the templated service/database.
-            Model model = ModelFactory.createDefaultModel();
-            StreamRDF dest = StreamRDFLib.graph(model.getGraph());
+            Model modelData = ModelFactory.createDefaultModel();
+            StreamRDF dest = StreamRDFLib.graph(modelData.getGraph());
 
             if ( hasParams || WebContent.isHtmlForm(ct) )
                 assemblerFromForm(action, dest);
@@ -151,6 +151,8 @@ public class ActionDatasets extends ActionContainerItem {
             else
                 assemblerFromBody(action, dest);
 
+            Model model = ModelFactory.createDefaultModel();
+            model.add(modelData);
             AssemblerUtils.addRegistered(model);
 
             // ----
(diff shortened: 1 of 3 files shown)


jena-fuseki2/jena-fuseki-webapp/src/main/java/org/apache/jena/fuseki/mgt/ActionDatasets.java
```diff
@@ -151,14 +151,16 @@ else if ( WebContent.isMultiPartForm(ct) )
             else
                 assemblerFromBody(action, dest);
 
+            Model model = ModelFactory.createDefaultModel();
+            model.add(modelData);
```
kinow, line 155: Looks like there will be two `Model` instances here, but `model` is never used, except for creating a copy of `modelData` to be used in `AssemblerUtils.addRegistered`. The rest of the code uses `modelData`.

If that's correct, maybe that could be `AssemblerUtils.addRegistered(ModelFactory.createDefaultModel().add(modelData));` to avoid risking the `model` instance being used accidentally. Or maybe `Model` could have a `.copy()` method?

kinow (approved):
One question, but no blocker. Looks good to me!

jena-fuseki2/jena-fuseki-webapp/src/main/java/org/apache/jena/fuseki/mgt/ActionDatasets.java
afs: It does get used but its via the actions of the Model API and `FusekiConfig.buildDataAccessPoint`.

`model` is used first by `findService` and then `FusekiConfig.buildDataAccessPoint(subject, registry)` which walks the model. A `Resource` knows which model it came from. (e.g. `subject.listProperties` needs the model of the subject).

`AssemblerUtils.addRegistered(model)` modifies the model argument. We want to write `modelData` but after the assembling is done and so is known to be successful (there is also a model write at L162, `systemFileCopy` which is really logging.) Being after, we need the unmodified and modified versions around.

`model.add(anOtherModel)` is not that expensive - it shares strings used in URIs, strings in literals and blank nodes.
(Adding one graph into another is cheaper.)

PR #1851, opened 2023-04-20
incorrect integer cast in scripting.NV

SimonBin:
GitHub issue resolved #

Pull request Description: Script Function would crash with number exception when trying to return a big integer from javascript (like 100000000000000000000000000)

as far as I can tell it was also trying to use the wrong variable (x instead of d)

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-arq/src/test/java/org/apache/jena/sparql/function/scripting/TestNV.java
+++ b/jena-arq/src/test/java/org/apache/jena/sparql/function/scripting/TestNV.java
@@ -66,7 +66,7 @@ public class TestNV {
         String largeInteger = "1" + "0".repeat(22);
         NodeValue nodeValue = NV.toNodeValue(new BigInteger(largeInteger).doubleValue());
         NV nv = new NV(nodeValue);
-        assertEquals(nv.getValue(), largeInteger);
+        assertEquals(largeInteger, nv.getValue());
         assertTrue("is a number", nv.isNumber());
     }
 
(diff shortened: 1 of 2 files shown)


kinow (commented):
Looks sensible to me. Maybe we could add a unit test for this too? :+1: Thanks!

afs:
That does not seem to be the code path for `100000000000000000000000000` (larger than a long).

`NV.fromNodeValue` seems to need something as well.

jena-arq/src/test/java/org/apache/jena/sparql/function/scripting/TestNV.java
```diff
@@ -58,7 +61,15 @@ public class TestNV {
         assertEquals(nv.getLabel(), nv.getValue());
         assertEquals("BlankNode", nv.getTermType());
     }
-    
+
+    @Test public void nv_14() {
+        String largeInteger = "1" + "0".repeat(22);
+        NodeValue nodeValue = NV.toNodeValue(new BigInteger(largeInteger).doubleValue());
+        NV nv = new NV(nodeValue);
+        assertEquals(nv.getValue(), largeInteger);
```
afs, line 69: Arguments the wrong way round? It's "expected" , "actual".

(This test does not work for your original example but that's OK because very large numbers will always be a problem between double and exact representation.)

afs, line 69:
```suggestion
        assertEquals(largeInteger, nv.getValue());
```

afs (approved):

PR #1865, opened 2023-05-11
Jena issue 1279

### LINKED ISSUE #1279: Improve the implementation of in-memory, general purpose, non-transactional graphs.
See PR #1273.

arne-bdt:
GitHub issue resolved #1279: Improve the implementation of in-memory, general-purpose, non-transactional graphs.

Pull Request Description:
- Improved performance of GraphMem for: Graph#find, Graph#stream, Graph#contains
- The performance improvement partly stems from implementing Iterator#forEachRemaining (new since Java 8) for many iterators and helpers
- Introduced a new module for benchmarks, especially JMH benchmarks, and prepared for regression tests against previous releases. This allows the impact of the changes to be evaluated.

Motivation:
My work on an alternative to GraphMem is progressing slowly. Some of my performance improvements, when compared to GraphMem, stem from Iterator and Spliterator implementations. Since I mainly work with sparse arrays, I was able to apply some of my work directly to GraphMem and its helper classes. 
To avoid any negative impact from my work, it was necessary to compare the performance of the tuned GraphMem with the previous release. So, I restructured my JMH benchmarks and added them to the project. Now, anyone can easily evaluate the impact on their specific graph. The new modules also enable other developers to easily add their JMH tests or regression tests to Jena.

----

Summary of first commit:
- Improved performance of GraphMem for:
  - Graph#find
    - Slightly by tuning the filter predicates
    - Significantly when results are processed via Iterator#forEachRemaining
  - Graph#stream
    - Slightly by tuning the filter predicates
    - Significantly for most stream operations
    - java.util.stream.BaseStream#parallel is now fully supported.
      (Before these changes, parallel execution was even slower than single-threaded execution)
  - Graph#contains
    - Only for non-concrete (fluent) patterns
- GraphMem used a variety of different Iterator implementations, helper classes, and wrappers:
  -> None of them supported #forEachRemaining
  -> It seemed appropriate to almost universally implement #forEachRemaining to avoid a wrapper or helper
     from breaking the newly gained performance advantages of #forEachRemaining

Summary of second commit:
- new module: jena-benchmarks, with two sub-modules:
  - jena-benchmarks-shadedJena480
    - This module wraps some dependencies from the Apache Jena Release 4.8.0
    - The wrapped module is used to implement regression tests against older releases
  - jena-benchmarks-jmh
    - contains JMH benchmarks for different Graph operations
      - Allows for the comparison of the performance of different Graph implementations and different releases of the same implementation
    - contains JMH benchmarks for SparseArraySpliterator and SparseArraySubSpliterator 

----

Details of first commit:

Implemented Iterator#forEachRemaining in an optimized way for many iterators throughout the Jena project:
- Replaced hasNext();next(); calls by #forEachRemaining in some promising places (not all)

Optimized NiceIterator:
- NiceIterator#hasNext now avoids redundant calls to current.hasNext()
- NiceIterator#andThen has optimized code to handle expensive hasNext calls of wrapped iterators.

Tuned GraphMem:
- Removed unused classes in the 'mem' namespace: 
  - GraphMemTripleStore, NodeToTripleMap, SetBunch and WrappedHashMap
- Iterators:
  - Tuned Iterator implementations, mainly by adding forEachRemaining implementations
  - Optimized code in BasicKeyIterator. The new iterator works in reverse order, so I had to adapt HashCommon#removeFrom
  - Optimized code in HashedBunchMap#iterator
  - TrackingTripleIterator#forEachRemaining now simply calls super.forEachRemaining() and sets current to null
    -> This had a significant impact on performance
- Spliterators:
  - Created specialized spliterators SparseArraySpliterator and SparseArraySubSpliterator (+unit tests)
  - Replaced Spliterator implementation within HashCommon and HashedBunchMap with SparseArraySpliterator
  - Implemented Spliterators to support fast stream operations, where the SparseSpliterator*
    implementations were not suitable
- Replaced usage of org.apache.jena.graph.impl.GraphBase#containsByFind with optimized implementations:
  - Introduced org.apache.jena.graph.impl.TripleStore#containsMatch
  - Implemented org.apache.jena.mem.NodeToTriplesMapBase#containsMatch using the spliterator

- Filter operations:
  - Introduced org.apache.jena.graph.Triple.Field#filterOnConcrete to avoid double-checking of Node#isConcrete (+unit tests)
  - Created org.apache.jena.mem.FieldFilter to efficiently build filter predicates only when needed, and only with
    the required conditions (+unit tests)
  - Used org.apache.jena.mem.FieldFilter#filterOn in NodeToTriplesMap#iterator, NodeToTriplesMapBase#stream,
    and NodeToTriplesMapBase#containsMatch to only filter when a filter is needed.
    For example: For find(sub, ANY, ANY), there is no need for a filter in the underlying TripleBunch

----

Latests Bechmarks:
JDK 17.0.7, OpenJDK 64-Bit Server VM, 17.0.7+7 -Xmx12G
Notice: 
- The Legend unfortunately hides the corresponding release of GraphMem. It is always this PR (Rows 1, 3, 5) vs. Jena 4.8.0 (Rows 2, 4, 6).
- The scale is logarithmic since the first two sample graphs (cheeses-0.1.ttl and pizza.owl.rdf) are very small compared to the third graph (bsbm-1m.nt.gz)

Graph#contains with fluent (non-concrete) triples:
image

Graph#find all triples with Iterator#forEachRemaining:
image

Graph#find all triples with Iterator#hasNext + Iterator#Next:
image

Graph#find by pattern with single evaluation of Iterator#hasNext:
image

Graph#stream all triples - single threaded and parallel
image

Graph#stream by pattern with Stream#findAny:
image

----

 - [X] Tests are included.
 - [nothing to do?] Documentation change and updates are provided for the [Apache Jena website](https://github.com/apache/jena-site/)
 - [X] Commits have been squashed to remove intermediate development commit messages.
 - [X] Key commit messages start with the issue number (GH-xxxx or JENA-xxxx)

----


afs:
Good to see this!

arne-bdt:
Found minor bugs in test-suite. (not in produtive code)
I will fix them today or tomorrow and squash the commits again.

jena-arq/src/main/java/org/apache/jena/riot/protobuf/ProtobufRDF.java
afs, line 1: Need to remove the now unused import of org.apache.jena.sparql.engine.binding.Binding;

jena-base/src/main/java/org/apache/jena/atlas/iterator/Iter.java
```diff
@@ -460,6 +475,14 @@ public T next() {
             return t;
         }
 
+        @Override
+        public void forEachRemaining(Consumer<? super T> action) {
+            stream.forEachRemaining(item->{
+                this.action.accept(item);
```
afs, line 481: Two calls to action?

jena-benchmarks/jena-benchmarks-jmh/src/test/java/org/apache/jena/JenaVersion.java
afs, line 1: Not sure why this is showing as being derived from TestWrappedSetTripleBunch.

Would `JenaVersion` be better in jena-core/main?

jena-tdb1/src/main/java/org/apache/jena/tdb/sys/DatasetControlMRSW.java
```diff
@@ -126,6 +127,13 @@ public T next() {
             }
         }
 
+        @Override
+        public void forEachRemaining(Consumer<? super T> action) {
+            iter.forEachRemaining(action);
+            close();
+            checkIterConcurrentModification();
```
afs, line 134: Should this be at the start of the method?

jena-benchmarks/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+
+  <licenses>
+    <license>
+      <name>Apache 2.0 License</name>
+      <url>https://www.apache.org/licenses/LICENSE-2.0</url>
+    </license>
+  </licenses>
+
+  <modules>
+    <module>jena-benchmarks-jmh</module>
```
afs, line 51: I'm having an IDE build problem.

Code in `TestGraph...` classes depends on a package rooted at `shadedJena480`

e.g.
```
private List<org.apache.shadedJena480.graph.Triple> triples480;
```
which Eclipse isn't finding.

jena-core/src/main/java/org/apache/jena/mem/NodeToTriplesMapBase.java
```diff
@@ -19,6 +19,9 @@
 package org.apache.jena.mem;
 
 import java.util.*;
+import java.util.function.Consumer;
```
afs, line 22: Unused import.

jena-base/src/main/java/org/apache/jena/atlas/iterator/Iter.java
arne-bdt: Indeed. org.apache.jena.atlas.iterator.Iter.IterOperate is initialised with a `Consumer<T> action`. 
When consuming via org.apache.jena.atlas.iterator.Iter.IterOperate#hasNext, the consumer is also called, so when consuming via `org.apache.jena.atlas.iterator.Iter.IterOperate#forEachRemaining` there are two consumers, the one the class was initialised with and the one that is a parameter of the method.

jena-benchmarks/jena-benchmarks-jmh/src/test/java/org/apache/jena/JenaVersion.java
arne-bdt: That's weird. Recently my IDE came up with strange ideas about how new files were variants of existing files in git.
But `JenaVersion` was added in a commit and I can't see a link to 'TestWrappedSetTripleBunch' here.
I must confess that I am not too familiar with the git cli and mostly rely on my IDE. 

The `JenaVersion` enum is only used in jena-benchmarks-jmh to distinguish the different versions of jena for regression testing. So I see no reason to move it to jena-core/main.

jena-benchmarks/pom.xml
arne-bdt: I have the same problem with InteliJ IDEA. 
I'm not sure, but it seems to be a general problem with shaded libraries, as I had the same problem with jena-shaded-guava. 
My solution was to manually add "jena\jena-benchmarks\jena-benchmarks-shadedJena480\target\jena-benchmarks-shadedJena480-4.9.0-SNAPSHOT.jar" as a library to jena-bechmarks-jmh.
I had to do the same before with jena-shaded-guave.
--> However, I would be grateful for a real solution to this problem.

jena-core/src/main/java/org/apache/jena/mem/NodeToTriplesMapBase.java
arne-bdt: As far as I can see, there is no unused import here.
The `Consumer `ist used in `org.apache.jena.util.iterator.NiceIterator#forEachRemaining`.

jena-tdb1/src/main/java/org/apache/jena/tdb/sys/DatasetControlMRSW.java
arne-bdt: Sorry, I intruduced  a bug here, because `close `"disables" `checkIterConcurrentModification` in this class. So `close` have to be done after `checkIterConcurrentModification`.

I took a look into the Java `ArrayList` and `HashMap` doc and implementatons for my decision to put the check at the end.
The doc header of `HashMap` says;

> Note that the fail-fast behavior of an iterator cannot be guaranteed
>  as it is, generally speaking, impossible to make any hard guarantees in the
>  presence of unsynchronized concurrent modification.  Fail-fast iterators
>  throw `ConcurrentModificationException` on a best-effort basis.
>  Therefore, it would be wrong to write a program that depended on this
>  exception for its correctness: _the fail-fast behavior of iterators_
>  _should be used only to detect bugs_.

My conlusions:
`Iterator#hasNext` --> should not check for concurrent modifications
'Iterator#next' --> should check for concurrent modifications
'Iterator#forEachRemaining' -->  should usualy be checked at the end because this is the point where it is most useful to detect errors. Checking only at the beginning is almost useless. Checking at each iteration is expensive, but can be useful if it avoids index errors at the same time.

Here is one example for checking at the end: `java.util.HashMap.KeySet#forEach`.

jena-arq/src/main/java/org/apache/jena/riot/protobuf/ProtobufRDF.java
arne-bdt: Import removed.

jena-core/src/test/java/org/apache/jena/graph/test/TestNodeToTriplesMapMem.java
```diff
@@ -23,26 +23,27 @@
 import junit.framework.TestSuite;
 import org.apache.jena.graph.* ;
 import org.apache.jena.graph.Triple.* ;
-import org.apache.jena.mem.NodeToTriplesMap ;
+import org.apache.jena.mem.NodeToTriplesMapMem ;
+import org.apache.jena.util.iterator.ExtendedIterator;
```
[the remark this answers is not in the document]
arne-bdt: Import removed.

jena-core/src/test/java/org/apache/jena/mem/SparseArraySpliteratorTest.java
```diff
@@ -0,0 +1,528 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.jena.mem;
+
+import org.apache.jena.atlas.iterator.ActionCount;
```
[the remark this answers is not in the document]
arne-bdt: Import removed.

jena-core/src/main/java/org/apache/jena/mem/SparseArraySubSpliterator.java
```diff
@@ -0,0 +1,220 @@
...
+        final int entriesCount = pos - fromIndex;
+        if (entriesCount < 2) {
+            return null;
+        }
+        if (this.estimateSize() < 2L) {
+            return null;
+        }
+        final int toIndexOfSubIterator = this.pos;
+        this.pos = fromIndex + (entriesCount >>> 1);
+        return new SparseArraySubSpliterator(entries, this.pos, toIndexOfSubIterator, fillRatio, checkForConcurrentModification);
```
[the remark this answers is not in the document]
arne-bdt: I fixed it similar to the solution above.

afs (approved):

jena-core/src/main/java/org/apache/jena/mem/SparseArraySpliterator.java
```diff
@@ -0,0 +1,203 @@
...
+    public Spliterator<E> trySplit() {
+        if (pos < 2) {
+            return null;
+        }
+        if (this.estimateSize() < 2L) {
+            return null;
+        }
+        final int toIndexOfSubIterator = this.pos;
+        this.pos = pos >>> 1;
+        return new SparseArraySubSpliterator(entries, this.pos, toIndexOfSubIterator, fillRatio, checkForConcurrentModification);
```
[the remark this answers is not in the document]
afs: Without `<>` or `<E>` I was getting an Eclipse warning. Either should work.

jena-benchmarks/pom.xml
afs: I find that simply not including the projects with shading works. We can see how this goes for everyone when merged.

jena-benchmarks/jena-benchmarks-jmh/pom.xml
```diff
@@ -0,0 +1,161 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.jena</groupId>
+            <artifactId>jena-core</artifactId>
```
afs, line 117: Duplicate of one above.

arne-bdt: duplication removed

jena-tdb1/src/main/java/org/apache/jena/tdb/sys/DatasetControlMRSW.java
afs: Thank you for the analysis.

A CCME is only "advice" - and then more for the same thread/nested changes case than multithread concurrent access.

arne-bdt: Thanks. 
If it is more for the same thread/nested changes, should I put it at the beginning of the implementation of "forEachRemaining" in "DatasetControlMRSW"?

afs: In the `forEachRemaining`, when not doing it fine grained in `next`, either start or finish is fine. It is not going to trap the moment the nested change happens, nor if the `forEachRemaining` is beign nested but that's OK.

Ready to merge?

arne-bdt: Yes, I would be happy if you would merge this PR.

afs:
Sorry - just noticed one small thing when about to merge.
This is my mistake - I should have seen this earlier.

The commit messages use "JENA-1279". JENA-? refers to JIRA (in some places, this gets a link in the UI; I think this is some kind of Apache local addition).

https://issues.apache.org/jira/browse/JENA-1279 - which is slightly different!

GH-1279 links to the github issue, using github's normal mechanisms.

Please could you change the messages on the commits?

I've put in [a change for the PR template](https://github.com/apache/jena/pull/1899). Our JIRA days are long gone, and we have it as record and as a legacy channel.

arne-bdt:
> Sorry - just noticed one small thing when about to merge. This is my mistake - I should have seen this earlier.
> 
> The commit messages use "JENA-1279". JENA-? refers to JIRA (in some places, this gets a link in the UI; I think this is some kind of Apache local addition).
> 
> https://issues.apache.org/jira/browse/JENA-1279 - which is slightly different!
> 
> [GH-1279](https://github.com/apache/jena/issues/1279) links to the github issue, using github's normal mechanisms.
> 
> Please could you change the messages on the commits?
> 
> I've put in [a change for the PR template](https://github.com/apache/jena/pull/1899). Our JIRA days are long gone, and we have it as record and as a legacy channel.

No problem, I'm just learning how it all works here.
I hope the commit comments are ok now?

afs:
Yes - all good now.

Thank you very much for this significant improvement to Jena.

PR #1898, opened 2023-06-01
GH-1897: More control for Fuseki modules setup

### LINKED ISSUE #1897: Improve configuration of Fuseki Modules.
### Version

4.8.0

### Feature

The current mechanism for Fuseki Modules is by using the java ServiceLoader.

Current state:
* Fuseki Modules are instantiated at init time.
* It is undefined whether there is one module object per JVM or one per server build.
* It is not possible, in a practical way, to make a selection of modules to build a server that includes modules defined in java but without service loader configuration.
* modulepath+ServiceLoader works slightly differently to classpath+ServiceLoader

Proposal:
* Discover, but do not instantiate modules at init time with the service loader.
* Introduce a unit - a collection of module objects - used to build a server.
* One way to build a collection of module objects is to instantiate the discovered set
* The FusekiServer.Builder allows setting the modules - it defaults to instantiating the discovered set (that is, drop-ins work as before)
* It is also possible to define the collection of modules for detailed selection and control in Java code, including controlling sharing, or not.

Yes

afs:
GitHub issue resolved #1897

Pull request Description: see #1897

This PR is based on using Fuseki as a configurable triple store, and adding functionality.
The Fuseki module system doesn't allow, in a practical way, enough control. The ServiceLoader
mechanisms is convenient for some use cases, but fixed and inflexible.

----

 - [x] Tests are included.

----


afs:
This should give both drop-in jars or programmatic control to set up the Fuseki modules to assigned when building a server.

kinow:
Any tips on how to review it? The code looks OK, but would it be worth building this branch and taking the new Fuseki for a spin? Maybe using different configurations?

afs:
You can pull the PR and build locally.

```
ID=1898
NEWBRANCH=fuseki-modules

git fetch "$REPO" pull/$ID/head:$NEWBRANCH
echo "Checkout branch: $NEWBRANCH"
git switch "$NEWBRANCH"
```

As things stand at the moment, we're not well setup for branch builds that deliver artifacts to download because they will go to the same SNAPSHOT maven repo as the daily build of main SNAPSHOT. People would pick up either one or the other - what every ran last! - unless they know which dated snapshot to pick up.

As it has been quite rare for use to have this situation, just using the development has been simpler. This feature, if not used, shouldn't interfere in any way. No one has reported using modules - which is not surprising as they are "experimental". I hope to be able to be confident to remove that label . There is one last API affecting change to investigate.

I have been tested out these changes on the £dayjob codebase. Telicent use Fuseki with ABAC security provided by a module (open source) as well as other modules for OpenTelemetry, AWS bearer authentication and [jena-fuseki-kafka](https://github.com/Telicent-io/jena-fuseki-kafka). The designs are becoming stable enough to publicise more.

I also have tried out with the the admin+UI as a automatically loaded module.

afs:
There are some examples that might help. One is 

https://github.com/apache/jena/blob/main/jena-fuseki2/jena-fuseki-main/src/test/java/org/apache/jena/fuseki/main/examples/ExFusekiMain_3_FusekiModule.java

afs:
This last set of changes makes more significant changes to API. If there are going to be changes at 4.9.0 while it's "experimental", then getting the changes done seems worthwhile.

- Fresh instances of modules of automatically loaded modules are created for each build.
- Differentiate between automatically loaded modules and modules provided directly as code.

It does mean more substantial changes including `FusekiModule` as it currently is, to be `FusekiAutoModule`, which means any ServiceLoader files change.

In `META-INF/services/`, 
`org.apache.jena.fuseki.main.sys.FusekiModule` becomes `org.apache.jena.fuseki.main.sys.FusekiAutoModule`

This is unfortunate but sub/super classes does not apply to service loader file name:

Not renaming the service file, and the interface it is associated with, can be made to work without this change (commit ad2487b2193) but the long term structure and naming is wrong. Not changing it now makes that compromise permanent. So some pain now while the "experimental" label applies.

I hope the label can be removed now at 4.9.0 or 5.0.0.

----

As a final and code-isolated trail run, this has now been tested downstream using this PR.

kinow:
>I hope the label can be removed now at 4.9.0 or 5.0.0.

Reading your comment I started wondering if we should release it with 5.x instead of 4.x, glad you considered it.

I **think** this should probably be considered, as i) I think it's hard to give a nice & user-friendly error message when the ServiceLoader fails to load due to a key name change, ii) this could impact users deploying multiple services (e.g. fuseki-main's or containers), and iii) some users may complain about having to deal with that in 4.x (I think there was a recent-ish discussion choosing the JVM of Jena; I guess the same could happen about choosing when to add breaking changes in the ServiceLoaders?

afs:
Now looks for and logs a warning if old style Fuseki modules are found.

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/sys/FusekiAutoModules.java
```diff
@@ -0,0 +1,230 @@
...
+            }
+            return newServiceLoader;
+        }
+
+        private void discoveryWarnLegacy() {
+            Class<FusekiModule> moduleClass = FusekiModule.class;
+            try {
+                ServiceLoader<FusekiModule> newServiceLoader = ServiceLoader.load(moduleClass, this.getClass().getClassLoader());
+                newServiceLoader.stream().forEach(provider->{
+                    FmtLog.warn(FusekiAutoModules.class, "Legacy: \"%s\" : Interface FusekiModule changed to FusekiAutoModule", provider.type().getSimpleName());
```
rvesse, line 176: Might want to be explicit in this warning that the legacy module is ignored as a result

afs:
Fixed conflicts with `main`.

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/sys/FusekiAutoModules.java
afs: Done

PR #1967, opened 2023-07-21
Capabilities and locks

### LINKED ISSUE #1965: Reorganise Capabilities so the default is "equality by term".
### Change

Currently the default is "equality by value" but this is supported in only one place, the original GraphMem.

With an eye to removing "handleLiteralDatatypes", reorganise so the default is "equality by term", and have GraphMem override the default.

Ensure reasoners are sensitive to the choice of "same term"/"same value" in their data graphs.

Yes

### LINKED ISSUE #1966: Deprecate Model.getLock()
### Version

4.9.0

### Feature

Deprecate `Model.getLock()`. (It defaults to a MRSW lock).

Instead, encourage use of transactions or use of `Model.enterCriticalSection`.

Yes

afs:
GitHub issue resolved #1965.
GitHub issue resolved #1966.

----


jena-core/src/main/java/org/apache/jena/rdf/model/impl/ModelCom.java
```diff
@@ -1048,32 +1009,26 @@ public int numPrefixes()
...
-
-        If the prefix maps to multiple URIs, then we discard it completely.
-
-        @param m Model who's namespace is to be updated
-        @param ns the namespace map to add to the Model
+     * Service method to update the namespaces of a Model given the mappings from
+     * prefix names to sets of URIs. If the prefix maps to multiple URIs, then we
+     * discard it completely.
+     *
+     * @param m Model who's namespace is to be updated
```
kinow, line 1016: s/who's/whose (I think)

jena-core/src/main/java/org/apache/jena/rdf/model/impl/ModelCom.java
```diff
@@ -1604,9 +1542,8 @@ public GraphListener adapt( final ModelChangedListener L )
...
       @Override
-      public Model notifyEvent( Object e )
-      {
-          getGraph().getEventManager().notifyEvent( getGraph(), e );
+      public Model notifyEvent(Object e) {
+          getGraph().getEventManager().notifyEvent(getGraph(), e);
           return this;
       }
-}
+  }
```
kinow, line 1549: GH UI shows this as the last line of the file, I think. If it really is, then maybe this space prefix is not needed?

kinow (approved):

jena-core/src/main/java/org/apache/jena/rdf/model/impl/ModelCom.java
afs: Done!

jena-core/src/main/java/org/apache/jena/rdf/model/impl/ModelCom.java
afs: Good ctach. There's a general misalignment in the latter part of the file. I'll fix that and use the opportunity to reformat he the code.

afs:
I was thinking that an MRSW transaction would work as a migration path but it isn't as straight forward as that.

The problem is that graphs have their own API to transactions which does have a distinction of "read" and "write" that can be mapper to "read lock" and "write lock".  Lock promotion isn't supported by Jena's `ReentrantReadWriteLock`.

`SimpleTransactionHandler` does not support transactions. 

For now, it is better to leave `Model.getLock` alone, just revise its javadoc.

I've just come across [StampedLock](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/StampedLock.html) introduced in Java8 which might help.

afs:
Deprecating Model.getLock isn't so simple - see the note on the issue #1966.

So I'll remove that part of this PR.

PR #1989, opened 2023-08-15
rdfdiff fixes

william-vw:
GitHub issue resolved #

Pull request Description:

What I believe are some fixes for rdfdiff so it works for some of my test cases (unsure whether it will work for all cases, though).

----

 - [X] Tests are included.

----

### DIFF
--- a/jena-cmds/src/main/java/arq/rdfdiff.java
+++ b/jena-cmds/src/main/java/arq/rdfdiff.java
@@ -18,8 +18,7 @@
 
 package arq;
 
-import static org.apache.jena.atlas.logging.LogCtl.setLogging;
-
+import java.io.FileInputStream;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Comparator;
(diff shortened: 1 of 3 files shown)


rvesse:
Could you please describe the bugs this PR is fixing?

I can't see what those are from reading the PR diff.  There's a lot of whitespace change in this PR as well which makes understanding what's changed very difficult even with Hide Whitespace mode enabled so describing what's actually changed would be useful.

From what I can follow you are changing the output format which I don't think we'd want to do because that potentially breaks all existing users of this tool.  The current output format is intentionally following classic `diff` style.  Perhaps consider putting the alternative output format behind a new flag/option so we retain backwards compatibility while allowing users like yourself to opt into a newer output format?

afs:
Ping @william-vw

afs:
Ping @william-vw

william-vw:
@afs apologies, I'm lacking the cycles these days to properly format this PR so I will close it for now.

afs:
I completely understand!

It can be reopened.

Issue #1992, opened 2023-08-17
[question] unexpected ConcurrentModificationException while working with GraphMem


sszuev:
### Version

4.9.0

### Question

`GraphMem` does not guarantee that the iterator will not be broken if it is interrupted by a modification operation.
So this is just a question.

But in some scenarios when we do so we may expect that it will not be broken if search pattern and concrete triple do not match.

In any case, I think that additional documentation would not hurt.

See:
```java
import org.apache.jena.graph.Graph;
import org.apache.jena.graph.GraphMemFactory;
import org.apache.jena.graph.Node;
import org.apache.jena.graph.NodeFactory;
import org.apache.jena.graph.Triple;
import org.apache.jena.vocabulary.OWL;
import org.apache.jena.vocabulary.RDF;

import java.util.Iterator;

public class TmpGraphMem {

    public static void main(String... args) {
        Graph graph = GraphMemFactory.createGraphMem2();
        for (int i = 0; i < 42; i++) {
            graph.add(Triple.create(NodeFactory.createURI("a" + i), RDF.type.asNode(), OWL.Class.asNode()));
        }

        Triple pattern = Triple.create(Node.ANY, RDF.type.asNode(), OWL.Class.asNode());
        Triple triple1 = Triple.create(NodeFactory.createURI("x"), NodeFactory.createURI("p"), NodeFactory.createURI("W"));
        Triple triple2 = Triple.create(NodeFactory.createURI("x"), NodeFactory.createURI("p"), OWL.Class.asNode());
        System.out.println(pattern.matches(triple1)); // false
        System.out.println(pattern.matches(triple2)); // false

        Iterator<Triple> it = graph.find(pattern);
        it.next();
        it.next();

        graph.add(triple1);
        it.next(); // no ConcurrentModificationException
        it.next(); // no ConcurrentModificationException

        graph.add(triple2);

        it.next(); // ConcurrentModificationException
        it.next(); // ConcurrentModificationException
    }
}
```

afs:
This is standard Java iterator behaviour.

Any modification to an underlying datastructure may invalidate an iterator.

Filtering is not a factor - the filter is navigation of the underlying datastructure.

The same is true for JDK lists and maps.

For example, in a HashMap, the whole hash vector maybe resized, moving everything around.

If you want isolation, consume the iterator to a list and iterate on the list, or keep a record of the changes to make and make them afterwards.

sszuev:
[...]

Issue #1961, opened 2023-07-18
[question] concurrent `org.apache.jena.graph.Graph` implementations


sszuev:
### Version

4.x.x

### Question

In [ONT-API](https://github.com/owlcs/ont-api) we need thread-safe graph.
For this purpose, I created a separate simple library: https://github.com/sszuev/concurrent-rdf-graph
It contains `SynchronizedGraph` & `ReadWriteLockingGraph`.
It would be convenient to have only Jena in dependencies (instead of Jena + this library).
So, the question to Apache Jena community: do we need such kinds of Graphs in Jena?
Even we don't need, I think it is good to have this issue for a record.

sszuev:
Well, the code is open for both [ONT-API](https://github.com/owlcs/ont-api) and [concurrent-rdf-graph](https://github.com/sszuev/concurrent-rdf-graph).
ONT-API is an implementation of [OWL-API](https://github.com/owlcs/owlapi) api, we can't use transactions, there are RW-lock mechanism. We have to follow OWL-API rules. It supports concurrent access to OWL via `java.util.concurrent.locks.ReadWriteLock`.

Also, I don't quite understand how transactions could help if we have Streams and don't want to put everything in memory.
Usually transactions are used in the top-level, maybe with help of AOP. 
In ONTAPI we have neither top-level nor AOP. [ReadWriteLockingGraph](https://github.com/sszuev/concurrent-rdf-graph/blob/main/src/main/kotlin/ReadWriteLockingGraph.kt) offers such way. If we have a lot of read-operations, and few write ones, no so much memory is required.
In concurrent-rdf-graph there are tests.

afs:
Thanks for raising this issue.

Jena provides: `GraphTxn` from `GraphFactory.createGraphTxn` (new name `Factory` ->  `GraphFactory`).

`GraphTxn` is thread-safe on each single method call even without calling transaction begin-commit/abort from the application.

> I don't quite understand how transactions could help if we have Streams and don't want to put everything in memory.

With transactions, iterators from this graph are consistent - they iterator over the data at the time `find()` was called regardless of changes by other threads. It is not a copy. It is MR+SW - readers and writers can run at the same time. TDB2 and TIM (the dataset for transactional use in-memory behind `GarphTxn`) do not throw away the state of the database until all usage of that epoch of the database have finished with it.

> ONT-API is an implementation of [OWL-API](https://github.com/owlcs/owlapi) api, we can't use transactions, there are RW-lock mechanism.

[...]

PR #1994, opened 2023-08-17
jena-core: add more javadocs about Graph-mem thread-safety and ConcurrentModificationException

sszuev:
…rentModificationException (see #1992, #1961)

GitHub issue resolved #

Pull request Description:

----

 - [ ] Tests are included.

----

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/graph/GraphMemFactory.java
+++ b/jena-core/src/main/java/org/apache/jena/graph/GraphMemFactory.java
@@ -30,8 +30,20 @@ import org.apache.jena.util.iterator.NullIterator ;
 /**
  * A factory class for creating memory Graphs.
  * <p>
- * Apache Jena is migrating to term semantics graph for consistency across all in-memory and persistent storage graphs
- *
+ * Apache Jena is migrating to term semantics graph for consistency across all in-memory and persistent storage graphs.
+ * <p>
+ * All the graphs that this factory creates are <strong>not thread-safe</strong>.
+ * Note that if the memory Graph is structurally modified at any time after
+ * the iterator has been created by any of the {@code find*} or {@code stream*} methods, the iterator may throw
+ * a {@link java.util.ConcurrentModificationException ConcurrentModificationException}
+ * if continued with it after this modification.
+ * This may happen even if the queried data does not relate directly to the modified data
+ * (i.e. when triple search pattern does not match added or deleted triple).
+ * <p>
+ * The good practice is to explicitly close any {@link ExtendedIterator} immediately after a read operation.
+ * For GraphMem implementations {@code ExtendedIterator}'s materializing methods (such as {@link ExtendedIterator#toList()})
+ * could be used safely without explicit close. The same is true for {@link java.util.stream.Stream Java Stream}'s
+ * terminal operations.
  */
 public class GraphMemFactory
 {
--- a/jena-arq/src/main/java/org/apache/jena/sparql/graph/GraphFactory.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/graph/GraphFactory.java
@@ -33,7 +33,9 @@ public class GraphFactory {
     }
 
     /**
-     * Create a graph that is a Jena memory graph
+     * Create a graph that is a Jena memory graph.
+     * The created graph is <strong>not thread safe</strong>.
+     * Inappropriate use of graph iterators and streams may cause {@code ConcurrentModificationException}.
      *
      * @see #createDefaultGraph
      */
@@ -42,7 +44,7 @@ public class GraphFactory {
     }
 
     /**
-     * Create an in-memory, transactional graph.
+     * Create an in-memory, thread-safe, transactional graph.
      * <p>
      * This fully supports transactions, including abort to roll-back changes. It
      * provides "autocommit" if operations are performed outside a transaction. The
--- a/jena-arq/src/main/java/org/apache/jena/sparql/graph/GraphTxn.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/graph/GraphTxn.java
@@ -33,7 +33,7 @@ import org.apache.jena.util.iterator.ExtendedIterator;
 import org.apache.jena.util.iterator.WrappedIterator;
 
 /**
- * In-memory, transactional graph.
+ * In-memory, thread-safe, transactional graph.
  *
  * @implNote
  * The implementation uses the default graph of {@link DatasetGraphInMemory}.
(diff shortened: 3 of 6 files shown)


kinow (approved):

jena-core/src/main/java/org/apache/jena/graph/GraphMemFactory.java
```diff
@@ -30,8 +30,17 @@
...
+ * <p>
+ * All the graphs that this factory creates are <strong>not thread-safe</strong>.
+ * Note that if the memory Graph is structurally modified at any time after
+ * the iterator has been created by any of the {@code find*} or {@code stream*} methods, the iterator may throw
+ * a {@link java.util.ConcurrentModificationException ConcurrentModificationException}
+ * if continued with it after this modification.
+ * This may happen even if the queried data does not relate directly to modified data
+ * (i.e. when triple search pattern does not match added or deleted triple).
+ * The good practice is to explicitly close iterators immediately after a read operation or
+ * use terminate operations such as {@link ExtendedIterator#toList()}.
```
rvesse, line 43:
```suggestion
 * use materialising operations such as {@link ExtendedIterator#toList()}.
```

sszuev: In java-stream API such operations are called `terminal`, so it's typo. 

maybe `materializing (or terminal, in terms of Java Stream API)` ?

afs: "terminal" is slightly different - it does not necessary read the whole stream (e.g. `findFirst`) but there is an internal "steam close" call .  

`ExtendedIterator` does not this hidden close for partially consumed iterators. It's run-to-end or call `ExtendedIterator.close`. `materialising` is running to the end.

sszuev: When we call `Stream.findFirst`, we can't use this stream anymore, so for Graph-mem it seems safe even without explicit close. Although this is a good practice.

As far as I can see, there is still no `org.apache.java.util.iterator.ExtendedIterator.findFirst` method and other terminal methods. Such operations are available only in the form of utilities (`org.apache.java.util.iterator.Iter.findFirst`) and they do not call close. But, by the way, it makes them not so useful.

The existing "ExtendedIterator" materializing methods are also terminal. So I don't see any logical error here.

In addition, `Graph` now supports Java Streams, so we also have to say a few words about Streams.

I think some word could be added directly to `Graph` interface, and additional explanations could be added to `GraphMemFactory` as well, to make it less ambiguous.

By the way, why we use `org.apache.jena.atlas.lib.Closeable` instead of `java.lang.AutoCloseable`?
I think we can do something like this:
```java
interface org.apache.jena.atlas.lib.Closeable extends java.lang.AutoCloseable {

    @Override
    void close(); // no exception
}
```
\+ In ONTAPI we also have utils (com.github.sszuev.jena.ontapi.utils.Iterators)[https://github.com/sszuev/jena-owl2/blob/main/src/main/java/com/github/sszuev/jena/ontapi/utils/Iterators.java].
What do you think about adding these functionality to Jena's `Iter`?

sszuev:
@afs 

Would it be a good idea to add a mention to the factory-method and directly to the particular implementations documentation that the `ConcurrentModificationException` may not occur in some cases? A client application (such as `ReadWriteLockingGrapg`) may use this circumstance for its own purposes. For example, for optimization purposes (see issues https://github.com/apache/jena/issues/1992 and https://github.com/sszuev/concurrent-rdf-graph/issues/1)

For factory-method docs could look like this:
```html
<strong>Note</strong>: Interrupting Graph Iterators with modification operations 
may result in a `ConcurrentModificationException` due to structural changes.
It is <strong>strictly</strong> not recommended in the client code violate this rule. 
However, in some cases, interruption is possible in some cases, but it depends on the implementation. 
See the documentation for the specific GraphMem implementation. 
This behavior can be changed at any time, 
so please read the documentation of specific implementations 
if your code still requires interrupting iterations contrary to the recommendations.
```

jena-core/src/main/java/org/apache/jena/graph/GraphMemFactory.java
afs: `AutoCloseable.close` is required. Jena allows exhausting the iterator.

afs:
IMO the javadoc should focus on the API contract. If it hints at more, a few application writers will chance it!

For iterators, testing or seeing modifications work isn't enough to guarantee the code. Different amounts of data change whether a modification inside the processing loop succeeds or crashes. Blank nodes mean two runs on the same data files may behave differently.

Implementations can change - but code writing to the contract will work.

When the material gets into discussion, for example when it says "may", that feels more like advice which is more suited for a [How To](https://jena.apache.org/documentation/notes/index.html) on the subject. The javadoc can link to the material.

The contract  is:

* run to exhaustion or close the iterator
* use `finally` to be sure
* No modification inside the processing loop of the iterator
* `ExtendedIterator.remove` is not guaranteed and for general code, should not be relied upon.

jena-core/src/main/java/org/apache/jena/graph/GraphMemFactory.java
sszuev: It looks like I still don't quite understand the functional difference between `ExtendedIterator` and `Stream`. 

Both closeable, both have terminal and intrermediate operations (e.g. intermediate: `ExtendedIterator.filterKeep`, `Stream.filter`), both can be exhausted or not), both lazy. 
What am I missing out on?

`AutoCloseable` is not required, it is also "good practice", but in Java we can use `try-with-resources` to make our code safer. 
Also, switching to `AutoCloseable` will not break user code.

sszuev:
> IMO the javadoc should focus on the API contract. If it hints at more, a few application writers will chance it!
> 
> For iterators, testing or seeing modifications work isn't enough to guarantee the code. Different amounts of data change whether a modification inside the processing loop succeeds or crashes. Blank nodes mean two runs on the same data files may behave differently.
> 
> Implementations can change - but code writing to the contract will work.
> 
> When the material gets into discussion, for example when it says "may", that feels more like advice which is more suited for a [How To](https://jena.apache.org/documentation/notes/index.html) on the subject. The javadoc can link to the material.
> 
> The contract is:
> 
>     * run to exhaustion or close the iterator
> 
>     * use `finally` to be sure
> 
>     * No modification inside the processing loop of the iterator
> 
>     * `ExtendedIterator.remove` is not guaranteed and for general code, should not be relied upon.

Agree.

afs (approved):

PR #1996, opened 2023-08-25
JENA-2354: Set Caffeine initial cache size to 25%

afs:
Issue resolved : [JENA-2354](https://issues.apache.org/jira/browse/JENA-2354)

Pull request Description:
Set the initial size of the Caffeine cache to 25%, not 50%.

----


kinow (approved):
Small change, looks good. I guess you will wait for dobrist's reply in Jira (or merge noe and tweak again later if needed).

afs:
The OPs best course of action is to control the cache size with StoreParams.  Setting the initial cache size isn't safe for databases that are in-use as the cache can grow later on. The initial sizing only delays the pain.

It is possible the OP 's setup never filled half the cache in which case the initial setting would over allocate. The size was chosen without a detailed analysis - it seemed like a good idea at the time! There isn't a contract.

PR #2059, opened 2023-10-22
Fix IRIs.asBase

afs:
Better fix for 41acb38a34

There are two problems to be fixed, not one!

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/irix/IRIs.java
+++ b/jena-core/src/main/java/org/apache/jena/irix/IRIs.java
@@ -76,24 +76,32 @@ public class IRIs {
     /**
      * Given a candidate baseURI string, which may be a filename,
      * turn it into a IRI suitable as a base IRI.
+     * This includes encoding characters in a filename (e.g. spaces).
      */
-    public static String toBase(String baseURI) {
-        if ( baseURI == null )
+    public static String toBase(String uriForBase) {
+        if ( uriForBase == null )
             return getBaseStr();
-        String scheme = scheme(baseURI);
+        String scheme = scheme(uriForBase);
         if ( Sys.isWindows ) {
             // Assume a scheme of one letter is a Windows drive letter.
             if ( scheme != null && scheme.length() == 1 )
                 scheme = "file";
         }
-        // File scheme if the URI is "file:" or there is no scheme
-        // and the system base is a "file:" URI
-        boolean isFile = ( scheme != null )
-                ? scheme.equals("file")
-                : IRIs.getSystemBase().hasScheme("file");
-        if ( isFile )
-            baseURI = IRILib.encodeFileURL(baseURI);
-        return IRIs.getSystemBase().resolve(baseURI).toString();
+        if ( scheme == null  ) {
+            // Relative name: it the base is a file: URI, encode the relative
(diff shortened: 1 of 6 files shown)


kinow (commented):
I can check out with the IDE later, but the change appears to be only the variable name, from looking at the GitHub UI. And also a comment.

afs:
Good catch.

The Jenkins jobs are showing some double encoding going on which I haven't been able to produce locally.

There are two problems - but seems I've only found one of them :-)

afs:
This makes the test suite run when the current directory, or a potential base URI,  has a space in it.

There is an unrelated issue with jena-fuseki-geosparql; resource URLs return a string space in them that do not the original resource with a space in its name.

jena-core/src/main/java/org/apache/jena/irix/IRIs.java
```diff
@@ -76,24 +76,32 @@ public static String getBaseStr() {
...
-        boolean isFile = ( scheme != null )
-                ? scheme.equals("file")
-                : IRIs.getSystemBase().hasScheme("file");
-        if ( isFile )
-            baseURI = IRILib.encodeFileURL(baseURI);
-        return IRIs.getSystemBase().resolve(baseURI).toString();
+        if ( scheme == null  ) {
+            // Relative name: it the base is a file: URI, encode the relative
+            // name if it does not look like it is already encoded.
+            boolean isFileBase = ( scheme == null ) && IRIs.getSystemBase().hasScheme("file");
```
kinow, line 93: This `( scheme == null )`  will always be `true`, because of the `if ( scheme == null )` above.

jena-fuseki2/jena-fuseki-geosparql/src/main/java/org/apache/jena/ext/io/github/galbiston/rdf_tables/file/FileConverter.java
```diff
@@ -101,6 +101,7 @@ public static void writeToModel(File inputFile, Model model, char delimiter, Boo
         } catch (IOException | RuntimeException | CsvValidationException ex) {
 
             LOGGER.error("FileConverter: Line - {}, File - {}, Exception - {}", lineNumber, inputFile.getAbsolutePath(), ex.getMessage());
+            ex.printStackTrace();
```
kinow, line 104: Currently debugging it? Or if intended to leave in the code maybe better to be wrapped in another exception thrown?

kinow (approved):
Quickly reading the code it looks good. Left a couple of comments, nothing major :+1:

jena-core/src/main/java/org/apache/jena/irix/IRIs.java
afs: Fixed!

jena-fuseki2/jena-fuseki-geosparql/src/main/java/org/apache/jena/ext/io/github/galbiston/rdf_tables/file/FileConverter.java
afs: Good point. Moved the exception into the `AssertionError`. Previously, it was hiding the information.

PR #2055, opened 2023-10-21
GH-2054: Upgrade UI dependencies to latest available

kinow:
GitHub issue resolved #2054

Pull request Description:

Used `ncu` to upgrade deps. First just minor & bug fixes.

```bash
kinow@ranma:~/Development/java/jena/jena/jena-fuseki2/jena-fuseki-ui$ npx ncu -u
Using yarn
Upgrading /home/kinow/Development/java/jena/jena/jena-fuseki2/jena-fuseki-ui/package.json
[====================] 43/43 100%

 @cypress/code-coverage             ^3.10.0  →  ^3.12.4     
 @cypress/vue                        ^4.2.0  →   ^6.0.0     
 @vitejs/plugin-vue                  ^4.0.0  →   ^4.4.0     
 @vitest/coverage-c8                ^0.27.1  →  ^0.33.0     
 @vue/compiler-sfc                  ^3.2.39  →   ^3.3.6     
 @vue/test-utils                     ^2.0.2  →   ^2.4.1     
 concurrently                        ^7.6.0  →   ^8.2.2     
 cypress                            ^12.1.0  →  ^13.3.2     
 cypress-vite                        ^1.3.0  →   ^1.4.2     
 eslint                             ^8.23.1  →  ^8.52.0     
 eslint-plugin-cypress              ^2.12.1  →  ^2.15.1     
 eslint-plugin-import               ^2.26.0  →  ^2.28.1     
 eslint-plugin-promise               ^6.0.1  →   ^6.1.1     
 eslint-plugin-vue                   ^9.5.1  →  ^9.17.0     
 express                            ^4.18.1  →  ^4.18.2     
 jsdom                              ^21.0.0  →  ^22.1.0     
 json-server                        ^0.17.0  →  ^0.17.4     
 nodemon                            ^2.0.20  →   ^3.0.1     
 sass                               ^1.54.9  →  ^1.69.4     
 sinon                              ^14.0.0  →  ^17.0.0     
 vite                                ^4.0.5  →   ^4.5.0     
 vite-plugin-istanbul                ^4.0.0  →   ^5.0.0     
 vitest                             ^0.27.1  →  ^0.34.6     
 @fortawesome/fontawesome-svg-core   ^6.2.0  →   ^6.4.2     
 @fortawesome/free-solid-svg-icons   ^6.2.0  →   ^6.4.2     
 @fortawesome/vue-fontawesome        ^3.0.1  →   ^3.0.3     
[...]

Issue #2054, opened 2023-10-21
[jena5] upgrade UI deps


kinow:
### Change

Placeholder issue for upcoming PR

### Are you interested in contributing a pull request for this task?

Yes

afs:
The build failure seems to be: 
```
[INFO] --- frontend:1.14.2:yarn (yarn install) @ jena-fuseki-ui ---
[INFO] Running 'yarn install --frozen-lockfile' in /home/jenkins/workspace/Jena/Jena_Development_Test/jena-fuseki2/jena-fuseki-ui
[INFO] node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by node)
```
so not a Jena build issue.

Looks like:
https://stackoverflow.com/questions/72921215/getting-glibc-2-28-not-found

afs:
Te build situation has already been reported: https://issues.apache.org/jira/browse/INFRA-23954

The issue is `nanoid` 5.* 
```
[INFO] error nanoid@5.0.2: The engine "node" is incompatible with this module. Expected version "^18 || >=20". Got "16.20.2"
```
It is a nanoid v4->v5 change.

Maybe we can use nanoid v4.0.2.

afs:
We have some overrides:
```
  "overrides": {
    "terser": "^5.14.2",
    "loader-utils": "^1.4.2",
    "cookiejar": "^2.1.4",
    "json5": "^1.0.2"
  }
```
Are these still necessary?

kinow:
>It is a nanoid v4->v5 change.
>
>Maybe we can use nanoid v4.0.2.

Not sure, haven't seen that error before -- I always use GH Actions to test Node apps, where I just say

```yaml
      - name: Setup Node
        uses: actions/setup-node@v4
        with:
          node-version: '18'
          cache: yarn
```

>We have some overrides:
>Are these still necessary?

Hard to say. The overrides are handy as they solve dependency issues like the ones we had with security, but at the cost that later you need to inspect the dependency tree with the resolution, check NPM for releases, modify your `package.json`, check the dependency tree again, and confirm that the override can be removed.

Created https://github.com/apache/jena/issues/2082 to track that :+1: 

Thanks!

PR #2084, opened 2023-11-11
Downgrade dependencies to work with node 16

afs:
This PR chooses versions to work with node 16, and hence earlier versions of libc (e.g. as found by default on Ubuntu 18.04).

While the build controls the version of node, it does not have control over the version of libc provided by the base OS.

See #2054 and #2055.

----

### DIFF
--- a/jena-fuseki2/jena-fuseki-ui/package.json
+++ b/jena-fuseki2/jena-fuseki-ui/package.json
@@ -29,7 +29,6 @@
     "bootstrap": "^5.3.2",
     "follow-redirects": "^1.15.3",
     "mitt": "^3.0.1",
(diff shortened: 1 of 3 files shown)


kinow (approved):

afs:
Hopefully, this will unblock the ASF build.

Ubuntu 18.04 EOL was May 2023 but is still around.  Maybe ASF Jenkins will upgrade sometime ([INFRA-23954](https://issues.apache.org/jira/browse/INFRA-23954)) and we can reverse this PR.

kinow:
All good. It doesn't affect the final JS created for Jena, only the development part and we can workaround it until infra sorts out the versions in Jenkins :shrug: Thanks for the PR!

PR #2096, opened 2023-11-16
ParameterizedSparqlString: Fixed flaky tests due to HashMap

wang3820:
Pull request Description:
### Issue

Tests like `org.apache.jena.query.TestParameterizedSparqlString#test_param_string_injection_10` makes more than one injections. `ParameterizedSparqlString.getVars()` is used to retrieve the order of injections, `getVars()` calls `this.params.keySet().iterator()` on the `ParameterizedSparqlString` object and the implementation of `params` of `ParameterizedSparqlString` is a `HashMap` which does not guarantee the order. So what is returned is not necessarily by the injection order.

https://github.com/apache/jena/blob/fc01d33e6c11eed68784f820bf2dd161868d1d37/jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java#L1138

https://github.com/apache/jena/blob/fc01d33e6c11eed68784f820bf2dd161868d1d37/jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java#L141

### Fix

The implementation of `params` of `ParameterizedSparqlString` is change to `LinkedHashMap` which guarantees the order returned is order of injection

----

 - [ ] Tests are included.

----

### DIFF
--- a/jena-arq/src/test/java/org/apache/jena/query/TestParameterizedSparqlString.java
+++ b/jena-arq/src/test/java/org/apache/jena/query/TestParameterizedSparqlString.java
@@ -1512,6 +1512,16 @@ public class TestParameterizedSparqlString {
         pss.setLiteral(second, " . } ; DROP ALL ; INSERT DATA { <s> <p> ");
 
         pss.asUpdate();
+        // Due to the unpredictability of the order of parameters in this.params,
+        // which is a HashMap, here the reverse order is checked to make sure
+        // an ARQException is thrown
+        pss.setLiteral("var", "a");
+        pss.setLiteral("var2", "b");
+
+        pss.setLiteral(second, "?" + first);
+        pss.setLiteral(first, " . } ; DROP ALL ; INSERT DATA { <s> <p> ");
+
+        pss.asUpdate();
         Assert.fail("Attempt to do SPARQL injection should result in an exception");
     }
 
@@ -1583,6 +1593,16 @@ public class TestParameterizedSparqlString {
 
         pss.setLiteral(first, " ?" + second + " ");
         pss.setLiteral(second, " . } ; DROP ALL ; INSERT DATA { <s> <p> ");
+
+        pss.asUpdate();
+        // Due to the unpredictability of the order of parameters in this.params,
+        // which is a HashMap, here the reverse order is checked to make sure
+        // an ARQException is thrown
+        pss.setLiteral("var", "a");
+        pss.setLiteral("var2", "b");
+
+        pss.setLiteral(second, " ?" + first + " ");
+        pss.setLiteral(first, " . } ; DROP ALL ; INSERT DATA { <s> <p> ");
 
         pss.asUpdate();
         Assert.fail("Attempt to do SPARQL injection should result in an exception");


rvesse:
Can you describe why this test is flaky, or show an environment where it fails?

The test case you mention is already designed to be resistant to variable ordering of the returned injections so not sure what this is gaining us here

wang3820:
> Can you describe why this test is flaky, or show an environment where it fails?
> 
> The test case you mention is already designed to be resistant to variable ordering of the returned injections so not sure what this is gaining us here

**Nondex** was used to check and locate the flakiness in the test. The test can be reproduced using the following command:
```shell
mvn install -pl jena-arq -am -DskipTests
mvn -pl jena-arq test -Dtest=org.apache.jena.query.TestParameterizedSparqlString#test_param_string_injection_10
mvn -pl jena-arq edu.illinois:nondex-maven-plugin:2.1.7-SNAPSHOT:nondex -Dtest=org.apache.jena.query.TestParameterizedSparqlString#test_param_string_injection_10
``` 

How nondex works is that it shuffles collection objects that does not guarantee order, like `HashMap`, while running the test and see it fails because of some assumptions were made based on the order is maintained. From my understanding, since the variable ordering is retrieved from the `HashMap`, the returned injection order could be wrong.

afs:
What's the repo URL to get to `edu.illinois:nondex-maven-plugin`?

wang3820:
https://github.com/TestingResearchIllinois/NonDex; however, for JAVA 17 projects, you might need to use 
```shell
git clone https://github.com/kaiyaok2/NonDex
cd NonDex
git checkout support_java_17
mvn clean install -DskipTests
```
to install version 2.1.7. 

Thanks for your patience.

wang3820:
> > How nondex works is that it shuffles collection objects that does not guarantee order, like `HashMap`, while running the test and see it fails because of some assumptions were made based on the order is maintained. From my understanding, since the variable ordering is retrieved from the `HashMap`, the returned injection order could be wrong.
> 
> So its actively shuffling the collection while the JVM is running, something that would **never** happen under normal circumstances?

Yes, that is correct; however, while active shuffling of collections would never happen under normal circumstances, reordering of elements in a `HashMap` could happen. As it states in the [documentation](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html): "This class makes no guarantees as to the order of the map; in particular, it does not guarantee that the order will remain constant over time." However, as I understand, the probability of that happening is not high. The nondex tool simply significantly increase the chance. Since the insertion order really matters here, `LinkedHashMap` may be a better choice.

afs:
If a HashMap resizes through updates by "put", the iteration order can change.
If a HashMap is not changing, it will not reorder.

It is the test that is making the ordering assumption not the code under test.

The defaults for a HashMap are size 16 default load factor 0.75. No reorder for 2-4 items. With nondex it is as if it is.

Would it be possible to try both orders?

rvesse:
> It is the test that is making the ordering assumption not the code under test.

I think this is the key point in this discussion.  This test is **deliberately** detecting the current iteration order in order to make a **deliberate** attempt to circumvent the SPARQL injection protections that the `ParameterisedSparqlString` provides.

The underlying code does not need a consistent order to function, in the test scenario in question doing things in the opposite order actually means that the attempted attack vector has no chance of working.  In a sense having an unpredictable iteration order is actually of benefit to this particular class

afs:
We're agreed that there is no issue with main-ParameterizedSparqlString.

The test determines an order ATM.

Would it be better to try both orders and expect at least (or exactly) one failure?

wang3820:
Apologies for the late reply, it was the Thanksgiving holidays in the US. Yes, I agree that there is no issue with main-ParameterizedSparqlString. I think it'd be better to try both orders.

wang3820:
I'm not entirely sure why the other commits shows up here... Still rather new to git. So sorry about this. How should I fix this?

jena-arq/src/test/java/org/apache/jena/query/TestParameterizedSparqlString.java
```diff
@@ -1585,6 +1595,15 @@ public void test_param_string_injection_15() {
...
         pss.asUpdate();
+        // Due to the unpredictability of the order of parameters in this.params,
+        // which is a HashMap, here the reverse order is checked to make sure
+        // an ARQException is thrown
+        pss.setLiteral("var", "a");
+        pss.setLiteral("var2", "b");
+
+        pss.setLiteral(second, " ?" + first + " ");
+        pss.setLiteral(first, " . } ; DROP ALL ; INSERT DATA { <s> <p> ");
+
```
rvesse, line 1606: Missing a `pss.asUpdate()` call here as the injection doesn't get attempted until the user tries to convert the `ParameterizedSparqlString` into an update

rvesse (changes requested):

jena-arq/src/test/java/org/apache/jena/query/TestParameterizedSparqlString.java
wang3820: FIxed! Thanks

PR #2097, opened 2023-11-16
Fix flaky prefixmap

wang3820:
**Nondex** was used to check and locate the flakiness in the test. The test can be reproduced using the following command:
```shell
mvn install -pl jena-arq -am -DskipTests
mvn -pl jena-arq test -Dtest=org.apache.jena.riot.system.TestPrefixMapOverPrefixMapping#prefixMap_abbrev_10
mvn -pl jena-arq edu.illinois:nondex-maven-plugin:2.1.7-SNAPSHOT:nondex -Dtest=org.apache.jena.riot.system.TestPrefixMapOverPrefixMapping#prefixMap_abbrev_10
``` 
----

 - [ ] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
+++ b/jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
@@ -138,7 +138,7 @@ public class ParameterizedSparqlString implements PrefixMapping {
 
     private StringBuilder cmd = new StringBuilder();
     private String baseUri;
-    private Map<String, Node> params = new LinkedHashMap<>();
+    private Map<String, Node> params = new HashMap<>();
     private Map<Integer, Node> positionalParams = new HashMap<>();
     private PrefixMapping prefixes;
     private Map<String, ValueReplacement> valuesReplacements = new HashMap<>();
--- a/jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
+++ b/jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
@@ -138,7 +138,7 @@ public class ParameterizedSparqlString implements PrefixMapping {
 
     private StringBuilder cmd = new StringBuilder();
     private String baseUri;
-    private Map<String, Node> params = new HashMap<>();
+    private Map<String, Node> params = new LinkedHashMap<>();
     private Map<Integer, Node> positionalParams = new HashMap<>();
     private PrefixMapping prefixes;
     private Map<String, ValueReplacement> valuesReplacements = new HashMap<>();
--- a/jena-core/src/main/java/org/apache/jena/shared/impl/PrefixMappingImpl.java
+++ b/jena-core/src/main/java/org/apache/jena/shared/impl/PrefixMappingImpl.java
@@ -19,6 +19,7 @@
 package org.apache.jena.shared.impl;
 
 import java.util.ArrayList ;
+import java.util.LinkedHashMap;
 import java.util.List ;
 import java.util.Map ;
 import java.util.Map.Entry;
@@ -50,7 +51,7 @@ public class PrefixMappingImpl implements PrefixMapping
         // ConcurrentHashMaps protects against breaking each datastructure
         // but does not protect against inconsistency.
         prefixToURI = new ConcurrentHashMap<>();
-        URItoPrefix = new ConcurrentHashMap<>(); 
+        URItoPrefix = new ConcurrentHashMap<>();
         }
     
     protected void set(String prefix, String uri) {
@@ -187,7 +188,7 @@ public class PrefixMappingImpl implements PrefixMapping
         
     @Override
     public Map<String, String> getNsPrefixMap()
-        { return CollectionFactory.createHashedMap( prefixToURI ); }
+        { return new LinkedHashMap<>( prefixToURI ); }
         
     @Override
     public String getNsURIPrefix( String uri )


jena-core/src/main/java/org/apache/jena/shared/impl/PrefixMappingImpl.java
```diff
@@ -49,8 +50,8 @@ public PrefixMappingImpl()
         { 
         // ConcurrentHashMaps protects against breaking each datastructure
         // but does not protect against inconsistency.
-        prefixToURI = new ConcurrentHashMap<>();
-        URItoPrefix = new ConcurrentHashMap<>(); 
+        prefixToURI = new ConcurrentSkipListMap<>();
+        URItoPrefix = new ConcurrentSkipListMap<>();
         }
```
afs, line 55: The order of entries in a PrefixMapping is not guaranteed. This part of the contract of for `PrefixMapping`.

There are other implementations of `PrefixMapping`.

Where is this a problem?

jena-arq/src/main/java/org/apache/jena/query/ParameterizedSparqlString.java
```diff
@@ -138,7 +138,7 @@ public class ParameterizedSparqlString implements PrefixMapping {
 
     private StringBuilder cmd = new StringBuilder();
     private String baseUri;
-    private Map<String, Node> params = new HashMap<>();
+    private Map<String, Node> params = new LinkedHashMap<>();
     private Map<Integer, Node> positionalParams = new HashMap<>();
```
afs, line 142: This is #2096, not `PrefixMapping`.

wang3820: Changed it back. Sorry about this.

jena-core/src/main/java/org/apache/jena/shared/impl/PrefixMappingImpl.java
wang3820: I understand. Thanks for pointing it out.

jena-core/src/main/java/org/apache/jena/shared/impl/PrefixMappingImpl.java
```diff
@@ -187,7 +189,7 @@ public String getNsPrefixURI( String prefix )
         
     @Override
     public Map<String, String> getNsPrefixMap()
-        { return CollectionFactory.createHashedMap( prefixToURI ); }
+        { return new LinkedHashMap<>( prefixToURI ); }
```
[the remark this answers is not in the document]
wang3820: I think the problems is in `pmTestAbbrev`. For test `org.apache.jena.riot.system.TestPrefixMapOverPrefixMapping#prefixMap_abbrev_10 `, when this line is called, 
https://github.com/apache/jena/blob/fc01d33e6c11eed68784f820bf2dd161868d1d37/jena-arq/src/test/java/org/apache/jena/riot/system/AbstractTestPrefixMap.java#L313
either `(p0,b)` or `(p1,)` could be returned, as order in `prefixToURI` is not predictable. But the implemetation of this function
https://github.com/apache/jena/blob/fc01d33e6c11eed68784f820bf2dd161868d1d37/jena-arq/src/test/java/org/apache/jena/riot/system/AbstractTestPrefixMap.java#L326
would only return `"p0:b"`. So if line 313 returns (p1,), the assertion here would fail.
https://github.com/apache/jena/blob/fc01d33e6c11eed68784f820bf2dd161868d1d37/jena-arq/src/test/java/org/apache/jena/riot/system/AbstractTestPrefixMap.java#L328


However, I do agree that putting the `LinkedHashMap` there is not a good fix. All that does is to make nondex, the tool I was using, not to shuffle that map so that the test does not fail. 

I hope this offers some clarification.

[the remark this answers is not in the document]
afs: Thank you for the clarification. 

The test is wrong.
It even says so!

https://github.com/apache/jena/blob/fc01d33e6c11eed68784f820bf2dd161868d1d37/jena-arq/src/test/java/org/apache/jena/riot/system/AbstractTestPrefixMap.java#L325

BTW The test is about `PrefixMap` not `PrefixMapping`. They are different.

`PrefixMapping` is the original for a model API.
It is not suitable for large scale storage even with the `PrefixMappingImpl` simplified system interface. from the way it manages the reverse mapping and because, by contract, obeys XML qname rules for abbreviation.

`PrefixMap` is the result of that experience. Simpler, lower-level. 
`PrefixMapping` is stil used in the Model API for compatibility.

afs:
#2098 fixes the test.

wang3820:
Thanks for letting me know. Should I close this draft now?

afs:
If `AbstractTestPrefixMap` is the only place it occurs for PrefixMap/PrefixMapping (I'm guessing the test failure was in subclass `TestPrefixMapOverPrefixMapping`), then yes, I think we're done here.

And you've altered us to be watchful in the future because the situation may yet also occur elsewhere.

Tests are mostly small scale and reordering due to rehashing is unlikely.

wang3820:
Great thanks.

afs:
Thanks!

wang3820:
> If `AbstractTestPrefixMap` is the only place it occurs for PrefixMap/PrefixMapping (I'm guessing the test failure was in subclass `TestPrefixMapOverPrefixMapping`), then yes, I think we're done here.
> 
> And you've altered us to be watchful in the future because the situation may also occur elsewhere.
> 
> Tests are mostly small scale and reordering due to rehashing is unlikely.

Actually this problem also occurs in subclass`TestPrefixMapTDB1`, but #2098 fixes this too.

PR #2152, opened 2024-01-04
GH-2150 Fix for bad query AST builder generated

rvesse:
GitHub issue resolved #2150

In some cases using the query builder API to create a query/update could result in an apparently valid query (when printing as a string) but the actual AST was bad.  When attempting to actually execute the query/update locally ARQ would fail to compile it as a result.

This commit fixes BuildElementVisitor to avoid an oversimplification of some AST elements, specifically an ElementGroup containing a single ElementFilter that was leading to the bad AST.

----

 - [ ] Tests are included.

----

### DIFF
--- a/jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/rewriters/BuildElementVisitor.java
+++ b/jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/rewriters/BuildElementVisitor.java
@@ -125,8 +125,16 @@ public class BuildElementVisitor implements ElementVisitor {
             // noting to do
             result = el;
         } else if (lst.size() == 1) {
-            lst.get(0).visit(this);
-            // result is now set properly
+            // GH-2150 If the single element is an ElementFilter need to leave it as-is otherwise some built queries
+            // and updates will fail when we attempt to execute them as the Element AST will be considered invalid by
+            // AlgebraGenerator
+            Element singleton = lst.get(0);
+            if (singleton instanceof ElementFilter) {
+                result = el;
+            } else {
+                lst.get(0).visit(this);
+                // result is now set properly
+            }
         } else {
             updateList(lst);
             result = el;
--- a/jena-extras/jena-querybuilder/src/test/java/org/apache/jena/arq/querybuilder/UpdateBuilderTest.java
+++ b/jena-extras/jena-querybuilder/src/test/java/org/apache/jena/arq/querybuilder/UpdateBuilderTest.java
@@ -29,12 +29,15 @@ import org.apache.jena.graph.Node;
 import org.apache.jena.graph.NodeFactory;
 import org.apache.jena.graph.Triple;
 import org.apache.jena.graph.impl.CollectionGraph;
+import org.apache.jena.query.DatasetFactory;
+import org.apache.jena.query.ReadWrite;
 import org.apache.jena.rdf.model.Literal;
 import org.apache.jena.rdf.model.Model;
 import org.apache.jena.rdf.model.ModelFactory;
 import org.apache.jena.rdf.model.Property;
 import org.apache.jena.rdf.model.Resource;
 import org.apache.jena.rdf.model.ResourceFactory;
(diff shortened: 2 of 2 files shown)


jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/rewriters/BuildElementVisitor.java
```diff
@@ -125,8 +125,16 @@ public void visit(ElementGroup el) {
             // noting to do
             result = el;
         } else if (lst.size() == 1) {
```
afs, line 127: @Claudenw What's the background to make the size=1 case special? (noting 2017 was a some time ago!)
It was introduced around Jena 3.7.0 which aligns with #2150.


I don't think the query builder should be making lst of 1 a special case.

There is no harm leaving groups of one. The parser generates them.

The algebra converter, query pretty printing (and OpAsQuery) all have to understand the group-of-one case. The algebra optimizer deals with them. The problem is that it does not deal with non-parser generated structures so well.

`ElementFilter` isn't the only example though the algebra converter copes with bare occurrences of other elements properly.

It is more valuable if the querybuilder generates ASTs like the parser. `Query.equals` then applies for the case `querybuilder -> string -> parse-> query` and `querybuilder.build -> query`.

If that then triggers querybuilder test failures, there may be a wider issue lurking.
(PS it does `WhereHandlerTest.testAddGraph` but `WhereValidator` seems to have its own idea of equality.

afs (changes requested):

jena-extras/jena-querybuilder/src/test/java/org/apache/jena/arq/querybuilder/UpdateBuilderTest.java
```diff
@@ -491,6 +495,38 @@ public void testInsertAndDeleteWithVariableNode() {
...
+
+        conn.begin(ReadWrite.WRITE);
+        try {
+            conn.update(updateRequest);
+            conn.commit();
+        } finally {
+            conn.end();
+        }
+    }
+
```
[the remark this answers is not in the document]
afs: Updated to an shorter example - `SELECT * WHERE { FILTER(true) }`.

jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/rewriters/BuildElementVisitor.java
rvesse: It looks like a lot of the test suite assumes this simplification, removing it generates 67 failures of the module so possibly your approach from #2155 of allowing these forms to compile might be more broadly useful especially since users could be building query ASTs in other ways

@afs Just discovered while working on $dayjob code that's using query builder this morning that similar cases exists for other elements.  E.g. create a query with just a `BIND` also exhibits the same problem

afs: Then better to solve in the `AlgebraGenerator`.
I'll add `ElementBind` to #2155 and look for a general solution.

A separate matter which would need test case changes  is whether querybuilder should create ASTs the parser might generate and have the equivalance `querybuilder -> string -> parse-> query` and `querybuilder.build -> query`.

When used over the wire, queries from the querybuilder are turned into strings.

Issue #2156 for whether the querybuilder ought to try for parser-compatible syntax structures.

afs: I don't see a a general solution because syntax alone does not have a hierarchy that identifies elements into general categories (that's in the algebra).

rvesse:
Closing in favour of PR #2155 which improves the `AlgebraGenerator` to handle these non-parser ASTs that the Query Builder (or other programmatic query construction) may create

afs:
@rvesse -- Should the branch in the Jena repo be deleted?

PR #2164, opened 2024-01-10
[Minor] Code clean-up

TelicentPaul:
Removing unused imports, fixes to some incorrect string formatting, correcting a case-sensitive switch option & an infinite recursion.

----

----


jena-db/jena-dboe-trans-data/src/main/java/org/apache/jena/dboe/trans/data/TransBinaryDataFile.java
```diff
@@ -166,7 +166,7 @@ protected void _abort(TxnId txnId, TxnBinFile txnResetState) {
             // Internal consistency check.
             // (Abort after commit would trigger the warning.)
             if ( txnResetState.length != x )
-                Log.warn(this, format("Mismatch: state.length = %d,  committedLength = %d", txnResetState.length != x));
+                Log.warn(this, format("Mismatch: state.length = %d,  committedLength = %d", txnResetState.length, x));
```
kinow, line 169: Looks like a copy-pasta bug! Good catch!

jena-tdb2/src/main/java/org/apache/jena/tdb2/store/NodeIdFactory.java
```diff
@@ -102,7 +102,7 @@ private static NodeId create(int v1, long v2) {
         int t = v1 >> 24;
         NodeIdType type = NodeIdType.intToEnum(t);
         if ( type == NodeIdType.SPECIAL )
-            throw new TDBException(String.format("Attempt to create a special from a long: 0x%016", v2));
+            throw new TDBException(String.format("Attempt to create a special from a long: 0x%016X", v2));
```
kinow, line 105: These two changes were the only ones I was not sure. Can't recall ever using this syntax. The rest of the changes look good!

kinow (approved):
Looks OK to me. Not sure about one change in a file (syntax), but +1 assuming CI tests pass. Thanks @TelicentPaul !

afs:
The files  ARQParser.java and ARQParserTokenManager.java are automatically generated - same for all the pairs like that.

```
/* Generated By:JavaCC: Do not edit this line. ARQParser.java */
```
JavaCC isn't run in the build. In each directory there a shell script that does several things - clear up old stuff, generate the parser, generate the txt file that can be used to produce complete HTML (jjdoc does not), then some patching up such as adding _SuppressWarnings ("unused")_.

1. People can build Jena without needing to have javacc
1. The source is available when debugging
1. The files do not change every often.

jena-tdb2/src/main/java/org/apache/jena/tdb2/store/NodeIdFactory.java
TelicentPaul: It's admittedly an odd one.  Without the X the format method doesn't know what it's doing and will throw an UnknownFormatConversionException. With the X it knows that it's hexadecimal, and the preceding numbers tell it how many characters to . 

I had to plug it into a wee online [java compiler](https://www.programiz.com/java-programming/online-compiler/) just to be sure.

`
        System.out.println(String.format("Will print a 16 character hex: 0x%016X", 13L));
        System.out.println(String.format("Will throw an exception: 0x%016", 13L));
    }
`

TelicentPaul:
I'm happy to rollback the changes in the auto-generated files since (a) it's only imports (b) it'll presumably be overridden next time the process runs. Just let me know.

afs:
We might as well leave the PR as is. Yes, they get overwritten. Not a lot we can do about that short of an amount more of hackery. "Mostly harmless". 🤷🏻 It is as the comment in the source files says!

Looks like the class level _SuppressWarnings ("unused")_ is added manually at least sometimes. IIRC it is too fragile to script exact fixes like unused imports always - javacc has changed its output across versions in the past or if the javacc setup changes (which is very rare).

afs:
What is odd is that this PR fixes a couple of things that Eclipse does not pick up and there doesn't seem to be an preference option for.

- some unused static imports
- unnecessary import of static inner class

Bad news is I tries some other warning settings and found a real issue. Another PR ...

afs (approved):

TelicentPaul:
Can I ask that as part of your PR you fix the "configutation" typo in FusekiMain.java?
It caught my eye but doesn't merit it's own PR.

afs:
> Can I ask that as part of your PR you fix the "configutation" typo in FusekiMain.java?
> It caught my eye but doesn't merit it's own PR.

#2166 is just changes in jena-core.

I've changed my local working copy on an on-going branch (about TDB commands) that is close-ish to jena-fuseki-main. OK?

PR #2234, opened 2024-02-01
fix(sec): upgrade org.apache.jena:jena-arq to 4.9.0

suguds:
### What happened？
There are 1 security vulnerabilities found in org.apache.jena:jena-arq 4.8.0
- [CVE-2023-32200](https://www.oscs1024.com/hd/CVE-2023-32200)

### What did I do？
Upgrade org.apache.jena:jena-arq from 4.8.0 to 4.9.0 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### How can we automate the detection of these types of issues?
By using the [GitHub Actions](https://github.com/murphysecurity/actions) configurations provided by murphysec, we can conduct automatic code security checks in our CI pipeline.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/jena-benchmarks/jena-benchmarks-shadedJena480/pom.xml
+++ b/jena-benchmarks/jena-benchmarks-shadedJena480/pom.xml
@@ -70,7 +70,7 @@
         <dependency>
             <groupId>org.apache.jena</groupId>
             <artifactId>jena-arq</artifactId>
-            <version>4.8.0</version>
+            <version>4.9.0</version>
         </dependency>
     </dependencies>
 


rvesse:
Thanks for the PR but in this case we cannot accept it.

The module containing the `4.8.0` dependency is purely a benchmarking module used to compare performance numbers between the old version of one of our core API implementations with a newer version that was introduced from `4.9.0` onwards.  So the usage of `4.8.0` is a) intentional and b) only for performance benchmarking purposes to ensure no substantative performance regressions.  The usage of `4.8.0` within the benchmarking is tightly scoped to the API under test and does not use any of the portions of the API affected by CVE-2023-32200

As such there is no security risk involved here and I will close this issue

Please note that for future reference any security issues with any Apache project should be reported using the [Apache Security Process](https://www.apache.org/security/) and not via public PRs/issues.

PR #2253, opened 2024-02-06
GH-2233: Improved handling of specifically set HTTP headers in RDFLinkHTTP

### LINKED ISSUE #2233: Explicit Accept headers on RDFConnectionRemote not always forwarded
### Version

5.0.0-SNAPSHOT

### What happened?

Setting headers explicitly on `RDFConnectionRemoteBuilder` such as using `.acceptHeaderSelectQuery(WebContent.contentTypeResultsXML)` does not work consistently across the APIs.

Only `RDFConnection.query(queryObject)` works as expected.

* For query *strings*, internally the query type is always considered QueryType.UNKNOWN and the custom header is ignored - even if query parsing is enabled (it is by default):
  https://github.com/apache/jena/blob/0fd96b64e1f404477eb8f95cefa49c776d907c27/jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java#L256
* For query *objects*, using `RDFConnection.newQuery()` does not forward the header:
  https://github.com/apache/jena/blob/0fd96b64e1f404477eb8f95cefa49c776d907c27/jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java#L268

The code below includes the respective logged headers in the comments.
```java
class Test {
    // Force log output of jena's init before our actual experiments.
    JenaSystem.init();

    try (RDFConnection conn = RDFConnectionRemote.newBuilder()
      .queryEndpoint("https://dbpedia.org/sparql")
      .acceptHeaderSelectQuery(WebContent.contentTypeResultsXML)
      .build()) {

      String queryStr = "SELECT * { ?s <urn:bar> <urn:baz> }";

      System.out.println("String - Direct");
      // DEBUG org.apache.jena.http.HTTP: > GET https://dbpedia.org/sparql?query=SELECT%20%2A%20%7B%20?s%20%3Curn:bar%3E%20%3Curn:baz%3E%20%7D
      // DEBUG org.apache.jena.http.HTTP:   Accept          application/sparql-results+json, application/sparql-results+xml;q=0.9, text/tab-separated-values;q=0.7, text/csv;q=0.5,application/json;q=0.2,application/xml;q=0.2,*/*;q=0.1
      try (QueryExecution qe = conn.query(queryStr)) {
        ResultSet rs = qe.execSelect().materialise();
        System.out.println(ResultSetFormatter.asText(rs));
      }

      System.out.println("String - Builder");
      // DEBUG org.apache.jena.http.HTTP: > GET https://dbpedia.org/sparql?query=SELECT%20%2A%20%7B%20?s%20%3Curn:bar%3E%20%3Curn:baz%3E%20%7D
      // DEBUG org.apache.jena.http.HTTP:   Accept          application/sparql-results+json, application/sparql-results+xml;q=0.9, text/tab-separated-values;q=0.7, text/csv;q=0.5,application/json;q=0.2,application/xml;q=0.2,*/*;q=0.1
      try (QueryExecution qe = conn.newQuery().query(queryStr).build()) {
        ResultSet rs = qe.execSelect().materialise();
        System.out.println(ResultSetFormatter.asText(rs));
      }


      Query query = QueryFactory.create(queryStr);

      // THIS CASE WORKS AS EXPECTED
      System.out.println("Query - Direct");
      // DEBUG org.apache.jena.http.HTTP: > GET https://dbpedia.org/sparql?query=SELECT%20%20%2A%0AWHERE%0A%20%20%7B%20?s%20%20%3Curn:bar%3E%20%20%3Curn:baz%3E%20%7D%0A
      // DEBUG org.apache.jena.http.HTTP:   Accept          application/sparql-results+xml
      try (QueryExecution qe = conn.query(query)) {
        ResultSet rs = qe.execSelect().materialise();
        System.out.println(ResultSetFormatter.asText(rs));
      }

      System.out.println("Query - Builder");
      // DEBUG org.apache.jena.http.HTTP: > GET https://dbpedia.org/sparql?query=SELECT%20%20%2A%0AWHERE%0A%20%20%7B%20?s%20%20%3Curn:bar%3E%20%20%3Curn:baz%3E%20%7D%0A
      // DEBUG org.apache.jena.http.HTTP:   Accept          application/sparql-results+json, application/sparql-results+xml;q=0.9, text/tab-separated-values;q=0.7, text/csv;q=0.5,application/json;q=0.2,application/xml;q=0.2,*/*;q=0.1
      try (QueryExecution qe = conn.newQuery().query(query).build()) {
        ResultSet rs = qe.execSelect().materialise();
        System.out.println(ResultSetFormatter.asText(rs));
      }
    }
  }
}
```

_No response_

Maybe

### LINKED ISSUE #2255: Support for preconfiguring the query/update send modes on RDFLinkHTTP
### Version

5.0.0-SNAPSHOT

### Feature

Some triple stores, such as older virtuoso open source versions, do not support `QuerySendMode.asGetWithLimitBody`.
While it is possible to create a link and connection to them using the code below, actually executing a query on the link will fail for such stores.

```java
RDFLink link = RDFLinkHTTP.newBuilder()
    .queryEndpoint("https://example.org/sparql")
    .build();
```

A concrete instance of a public triple store where the default send mode fails is mentioned in https://github.com/apache/jena/issues/2233#issuecomment-1930965743

The proposal is to extend the builder with methods to preconfigure the default query and update send modes for the sake of abstraction. When an RDFLink is passed to the application logic, then the latter should be able to execute queries successfully - without having to worry about how to actually configure the query executions created from the link.

```java
RDFLink link = RDFLinkHTTP.newBuilder()
    .queryEndpoint("https://example.org/sparql")
    .querySendMode(QuerySendMode.asGetWithLimitForm)
    .updateSendMode(UpdateSendMode.asPostForm)
    .build();
```

Yes

Aklakan:
GitHub issue resolved #2233 
GitHub issue resolved #2255 

Pull request Description: This PR proposes 
* opening up  ARQ's QueryExecHTTPBuilder for subclassing, such that RDFLinkHTTP can override the buildX() method to set the effective header according to configuration.
* making the send modes for queries and updates configurable on `RDFLinkHTTPBuilder`.

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
+++ b/jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
@@ -39,7 +39,9 @@ import org.apache.jena.sparql.exec.http.DSP;
 import org.apache.jena.sparql.exec.http.GSP;
 import org.apache.jena.sparql.exec.http.QueryExecHTTP;
 import org.apache.jena.sparql.exec.http.QueryExecHTTPBuilder;
+import org.apache.jena.sparql.exec.http.QuerySendMode;
 import org.apache.jena.sparql.exec.http.UpdateExecHTTPBuilder;
+import org.apache.jena.sparql.exec.http.UpdateSendMode;
 import org.apache.jena.sparql.util.Context;
 import org.apache.jena.system.Txn;
 import org.apache.jena.update.UpdateFactory;
@@ -79,6 +81,9 @@ public class RDFLinkHTTP implements RDFLink {
     // Whether to check SPARQL updates given as strings by parsing them.
     protected final boolean parseCheckUpdates;
 
+    protected final QuerySendMode querySendMode;
+    protected final UpdateSendMode updateSendMode;
+
     /** Create a {@link RDFLinkHTTPBuilder}. */
     public static RDFLinkHTTPBuilder newBuilder() {
         return new RDFLinkHTTPBuilder();
@@ -103,7 +108,8 @@ public class RDFLinkHTTP implements RDFLink {
                           String acceptDataset, String acceptGraph,
                           String acceptSparqlResults,
                           String acceptSelectResult, String acceptAskResult,
-                          boolean parseCheckQueries, boolean parseCheckUpdates) {
+                          boolean parseCheckQueries, boolean parseCheckUpdates,
+                          QuerySendMode querySendMode, UpdateSendMode updateSendMode) {
         // Any defaults.
         HttpClient hc =  httpClient!=null ? httpClient : HttpEnv.getDftHttpClient();
         if ( txnLifecycle == null )
@@ -124,6 +130,8 @@ public class RDFLinkHTTP implements RDFLink {
         this.acceptAskResult = acceptAskResult;
         this.parseCheckQueries = parseCheckQueries;
         this.parseCheckUpdates = parseCheckUpdates;
+        this.querySendMode = querySendMode;
+        this.updateSendMode = updateSendMode;
     }
 
     @Override
@@ -365,7 +373,7 @@ public class RDFLinkHTTP implements RDFLink {
     private QueryExecHTTPBuilderOverRDFLinkHTTP createQExecBuilder() {
         checkQuery();
         QueryExecHTTPBuilderOverRDFLinkHTTP builder = new QueryExecHTTPBuilderOverRDFLinkHTTP();
-        builder.endpoint(svcQuery).httpClient(httpClient);
+        builder.endpoint(svcQuery).httpClient(httpClient).sendMode(querySendMode);
         return builder;
     }
 
@@ -376,7 +384,6 @@ public class RDFLinkHTTP implements RDFLink {
             sBuff.append(", ");
         sBuff.append(acceptString);
     }
-
 
     /**
      * Return a {@link UpdateExecBuilder} that is initially configured for this link
@@ -392,7 +399,8 @@ public class RDFLinkHTTP implements RDFLink {
 
     /** Create a builder, configured with the link setup. */
     private UpdateExecHTTPBuilder createUExecBuilder() {
-        return UpdateExecHTTPBuilder.create().endpoint(svcUpdate).httpClient(httpClient);
+        return UpdateExecHTTPBuilder.create().endpoint(svcUpdate).httpClient(httpClient)
+                .sendMode(updateSendMode);
     }
 
     @Override
--- a/jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
+++ b/jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
@@ -37,8 +37,10 @@ import org.apache.jena.sparql.engine.binding.Binding;
 import org.apache.jena.sparql.exec.*;
 import org.apache.jena.sparql.exec.http.DSP;
 import org.apache.jena.sparql.exec.http.GSP;
+import org.apache.jena.sparql.exec.http.QueryExecHTTP;
 import org.apache.jena.sparql.exec.http.QueryExecHTTPBuilder;
 import org.apache.jena.sparql.exec.http.UpdateExecHTTPBuilder;
+import org.apache.jena.sparql.util.Context;
 import org.apache.jena.system.Txn;
 import org.apache.jena.update.UpdateFactory;
 import org.apache.jena.update.UpdateRequest;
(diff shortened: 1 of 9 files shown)


jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
```diff
@@ -241,77 +250,122 @@ public QueryExec query(Query query) {
...
         }
 
-        // Make sure it was set somehow.
-        if ( requestAcceptHeader == null )
-            throw new JenaConnectionException("No Accept header");
-        if ( requestAcceptHeader != null )
-            builder.acceptHeader(requestAcceptHeader);
-        // Delayed creation so QueryExecution.setTimeout works.
+        @Override
+        protected QueryExecHTTP buildX(HttpClient hClient, Query queryActual, String queryStringActual, Context cxt) {
```
Aklakan, line 289: I see that some of the attributes I changed from final to protected are passed to this method. The implementation here uses the attributes rather than the parameters - perhaps that should be changed.

Aklakan: Code changed to use the parameters.

rvesse (approved):
LGTM

Think @afs has modified this area of the code most recently so probably worth waiting for his review before proceeding

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
```diff
@@ -241,77 +258,125 @@ public QueryExec query(Query query) {
...
+                    // However, use it to derive the queryType
+                    parsedQuery = QueryFactory.create(queryStringActual);
+                    if (QueryType.UNKNOWN.equals(finalQueryType)) {
+                        finalQueryType = parsedQuery.queryType();
+                    }
+                }
+            }
+
+            // Use the query string as provided if possible, otherwise serialize the query.
+            String queryStringToSend = ( queryStringActual != null ) ? queryStringActual : queryActual.toString();
```
Aklakan, line 320: Redundant, handled by the base class.

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
```diff
@@ -241,77 +258,125 @@ public QueryExec query(Query query) {
...
+                        break;
+                }
+                // Make sure it was set somehow.
+                if ( requestAcceptHeader == null )
+                    throw new JenaConnectionException("No Accept header");
+
+                this.acceptHeader(requestAcceptHeader);
+            }
+
+            // Delayed creation so QueryExecution.setTimeout works.
```
Aklakan, line 364: Redundant, handled by the base class.

Aklakan:
I made one more round of cleanup for the `buildX` method in the inner class `QueryExecHTTPBuilderOverRDFLinkHTTP` within `RDFLinkHTTP` and squashed the commits.

Aklakan:
Just clarifying that this PR is ready for (re-)review - no hurry though :)

afs:
Noted :-) Slight overload what with Jena5 and RDF 1.2.

The PR does look clearer. There may also be another approach which I'll code up. 

Fundamentally, RDFLink and QueryExecHTTP have slight different approaches: RDFLink has four accept choices and chooses late when the query is known; QueryExecHTTP has one and can be set late.

I'm investigating setting up the queryExec builder with the combined four accept choices - they happen to be disjoint. But maybe there is really a fundamental mismatch that might lead to a deep change for the better.

Aklakan:
Note, that the `QueryExecHTTPBuilderOverRDFLinkHTTP` is a non-static(!) subclass of QueryExecHTTPBuilder which effectively introduces the four additional header setting fields - by implicitly inheriting those from RDFLinkHTTP. 

The whole logic could - as you seem to suggest - also go to QueryExecHTTPBuilder directly, removing the need for QueryExecHTTPBuilderOverRDFLinkHTTP.

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
```diff
@@ -241,77 +243,111 @@ public QueryExec query(Query query) {
...
-        if ( requestAcceptHeader != null )
-            builder.acceptHeader(requestAcceptHeader);
-        // Delayed creation so QueryExecution.setTimeout works.
+        @Override
+        protected QueryExecHTTP buildX(HttpClient hClient, Query queryActual, String queryStringActual, Context cxt) {
+            Query parsedQuery = queryActual;
+            if ( parsedQuery == null ) {
+                if ( RDFLinkHTTP.this.parseCheckQueries ) {
+                    // Don't retain the query
+                    // However, use it to derive the queryType
```
afs, line 291: The comment needs clarifying. It is being retained!

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
```diff
@@ -241,77 +243,111 @@ public QueryExec query(Query query) {
...
     // Create the QExec
 
     private QueryExec queryExec(Query query, String queryString, QueryType queryType) {
-        checkQuery();
-        if ( query == null && queryString == null )
-            throw new InternalErrorException("Both query and query string are null");
-        if ( query == null ) {
-            if ( parseCheckQueries )
-                // Don't retain the query.
-                QueryFactory.create(queryString);
```
afs, line 256: Something has changed because this:
```
    public static void main(String... args) {
        RDFLink link = RDFLinkHTTP.service("http://localhost:3030/ds").parseCheckSPARQL(true).build();
        try ( link; ) {
            link.query("Custom");
        }
        System.out.println("Done");
    }
```
throws an exception in Jena5/main but not in this PR. The parse error has moved to the point of execution.

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
```diff
@@ -241,77 +243,111 @@ public QueryExec query(Query query) {
...
+     * An internal subclass of {@link QueryExecHTTPBuilder} that overrides
+     * {@link #buildX(HttpClient, Query, String, Context)}
+     * in order to have the accept header field derived based on this {@link RDFLinkHTTP} configuration.
+     */
+    /* package */ class QueryExecHTTPBuilderOverRDFLinkHTTP
+        extends QueryExecHTTPBuilder
+    {
+        protected QueryType presetQueryType;
+
+        public QueryExecHTTPBuilderOverRDFLinkHTTP(QueryType presetQueryType) {
```
afs, line 280: `public` -> `protected`

afs (changes requested):

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
Aklakan: Done.

Aklakan:
* This is now an internal method `QueryExecHTTPBuilderOverRDFLinkHTTP.queryStringValidated` which sets a query string and validates it if if `parseCheckQueries` is enabled.
* I changed `QueryExecHTTPBuilder.setQuery` from private to protected in order to be able to update the state of the builder consistently when a query object or string is set.

A couple of notes for future reference:

1. It turns out that so far there is no public API on `QueryExecHTTPBuilder` to set a query string and have at processed according to the RDFLinkHTTP config:

* The `query(String)` method are all expected to parse the query string so that e.g. substitution can be performed
* The `queryString(String)` method on `ExecHTTPBuilder` explicitly states that `The queryString is not interpreted and it may contain SPARQL syntax`.

2. `QueryType` does not discriminate between construct triples and construct quads. If it did, then it would no longer be needed to pass the `parsedQuery` around. The constant `CONSTRUCT_QUADS` is commented out because apparently for other use cases it was not needed.

afs (approved):

afs:
Thanks!

PR #2275, opened 2024-02-17
GH-2267: Display all the pagination options

### LINKED ISSUE #2267: Fuseki UI broken for "edit" tab when number of graphs is large
### Version

4.10.0

### What happened?

Reported on mailing list: https://lists.apache.org/thread/zdtjtl3qlypl7d713z7bd1117kfjq149

The Fuseki UI is broken when we work on a large number of graphs and open the "edit" tab, see
image

You can reproduce it with

``` bash
for i in {1..100..1}; do echo "<http://example.org/s> <http://example.org/p> <http://example.org/o> <http://example.org/g$i> ." >> /tmp/data.nq; done
$JENA_HOME/bin/tdb2.tdbloader --loc /tmp/tdb2/debug /tmp/data.nq
$FUSEKI_HOME/fuseki-server --loc /tmp/tdb2/debug --port 3333 /ds
```

_No response_

No

kinow:
GitHub issue resolved #2267

Pull request Description:

@LorenzBuehmann first tentative to fix the issue. I was going to change the `Pagination` Vue component, so that it had an option like `maxItems` (or max displayed, etc). And set it to some value like `5` or `6` as that would fix in the screen, and show the first `2` or `3`, followed by `[…]` , and then the last `2` or `3`.

But I know sometimes users may want to know the page they are in, and also be able to jump from page 23 to page 4, which wouldn't be possible doing that.

Not the most beautiful, but maybe this simpler approach to just display everything visibly works OK for this page? Or should we keep it one line only, abbreviating the list? What do you think?

image

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-fuseki2/jena-fuseki-ui/package.json
+++ b/jena-fuseki2/jena-fuseki-ui/package.json
@@ -61,7 +61,7 @@
     "sinon": "^17.0.0",
     "vite": "^5.0.12",
     "vite-plugin-istanbul": "^6.0.0",
-    "vitest": "^0.34.6",
+    "vitest": "^1.3.1",
     "wait-on": "^7.0.1"
   },
   "overrides": {
--- a/jena-fuseki2/jena-fuseki-ui/src/components/dataset/Pagination.vue
+++ b/jena-fuseki2/jena-fuseki-ui/src/components/dataset/Pagination.vue
@@ -16,110 +16,124 @@
 -->
 
 <template>
-  <div class="row g-0">
-    <div class="col-12">
-      <ul
-        role="menubar"
-        aria-disabled="false"
-        aria-label="Pagination"
-        class="pagination mb-2 mx-0 justify-content-center"
+  <ul
+    role="menubar"
+    aria-disabled="false"
+    aria-label="Pagination"
+    class="pagination flex-wrap mb-2 mx-0 p-0 justify-content-center"
+  >
+    <!-- pages backward controls -->
+    <li
+      :aria-hidden="isBackLinkAriaDisabled"
+      :class="getBackLinkClass('page-item')"
+      role="presentation"
+    >
+      <button
+        :aria-disabled="isBackLinkAriaDisabled"
+        :class="getBackLinkClass('page-link')"
+        @click="goToPage(1)"
+        type="button"
+        role="menuitem"
+        tabindex="-1"
+        aria-label="Go to first page"
       >
-        <!-- pages backward controls -->
-        <li
-          :aria-hidden="getBackLinkAriaDisabled()"
-          :class="getBackLinkClass('page-item')"
-          role="presentation"
-        >
-          <button
-            :aria-disabled="getBackLinkAriaDisabled()"
-            :class="getBackLinkClass('page-link')"
-            @click="goToPage(1)"
-            type="button"
-            role="menuitem"
-            tabindex="-1"
-            aria-label="Go to first page"
-          >
-            «
-          </button>
-        </li>
-        <li
-          :aria-hidden="getBackLinkAriaDisabled()"
-          :class="getBackLinkClass('page-item')"
-          role="presentation"
-        >
-          <button
-            :aria-disabled="getBackLinkAriaDisabled()"
-            :class="getBackLinkClass('page-link')"
-            @click="goToPage(currentPage - 1)"
-            type="button"
-            aria-label="Go to previous page"
-            role="menuitem"
-          >
-            ‹
-          </button>
-        </li>
-        <!-- pages -->
-        <li
-          v-for="page in numberOfPages"
-          :key="page"
-          role="presentation"
-          class="page-item"
-        >
-          <span
-            :aria-label="`Go to page ${ page }`"
-            :class="getPageLinkClass(page)"
-            :aria-checked="page === currentPage"
-            @click="goToPage(page)"
-            role="menuitemradio"
-            type="button"
-            aria-posinset="1"
-            aria-setsize="2"
-            tabindex="0"
-          >{{ page }}</span>
-        </li>
-        <!-- pages forward controls -->
-        <li
-          :aria-hidden="getNextLinkAriaDisabled()"
-          :class="getNextLinkClass('page-item')"
-          role="presentation"
-        >
-          <button
-            :aria-disabled="getNextLinkAriaDisabled()"
-            :class="getNextLinkClass('page-link')"
-            @click="goToPage(currentPage + 1)"
-            type="button"
-            role="menuitem"
-            aria-label="Go to last page"
-          >
-            ›
-          </button>
-        </li>
-        <li
-          :aria-hidden="getNextLinkAriaDisabled()"
-          :class="getNextLinkClass('page-item')"
-          role="presentation"
-        >
-          <button
-            :aria-disabled="getNextLinkAriaDisabled()"
-            :class="getNextLinkClass('page-link')"
-            @click="goToPage(numberOfPages)"
-            type="button"
-            role="menuitem"
-            aria-label="Go to next page"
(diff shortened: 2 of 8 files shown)


LorenzBuehmann:
isn't the common approach to to limit the number of pages and simply show a "Next" button or something? And optionaly the last page?
I have no experience with Vue, but there is something built-in: https://bootstrap-vue.org/docs/components/pagination

Like this maybe?
image

kinow:
Yup, that's another approach, common too. I was just wondering whether we should allow users to select any page number, but I think limiting it to a only a few options is the way to go. Let me change it, show a preview here, and if you think that's looking OK I'll write a test or two and mark as ready for review, @LorenzBuehmann 

> I have no experience with Vue, but there is something built-in: https://bootstrap-vue.org/docs/components/pagination

Yup, and Bootstrap Vue provides other components that we could use. But I chose to use the plain `bootstrap` dependency instead, which brings just the CSS/JS from the original Bootstrap code to [our list of JS dependencies](https://github.com/apache/jena/blob/63dc60bd9f931d0ea09bcdc52b263a5afa3a4369/jena-fuseki2/jena-fuseki-ui/package.json#L21-L34).

The main reason for that is that in JS once you add a dependency, it normally brings `N` transitive dependencies (and `N` may be in the tens, hundreds, thousands :exploding_head: ). Fuseki UI is rather small, so crafting these small components like tables, pagination, forms, is not too hard.

The only complex part is the query editor, but we use an external dependency for that one. Less dependencies, less overhead to maintain security & upgrades.

Thanks!

jena-fuseki2/jena-fuseki-ui/package.json
```diff
@@ -61,7 +63,7 @@
     "sinon": "^17.0.0",
     "vite": "^5.0.12",
     "vite-plugin-istanbul": "^5.0.0",
-    "vitest": "^0.34.6",
+    "vitest": "^1.3.1",
```
kinow, line 66: I wonder why dependabot didn't pick this up. Maybe because it was a major update.

kinow: I hope to be able to replace istabul by vitest as vite - written by same creator of Vue - is becoming more widespread, even among React and Angular projects, and more and more tools too.

kinow:
The current coverage report works well, but the one provided by Vitest UI is like Surefire + Jacoco + a graph with the code, and only using Vitest (i.e. easy to configure). Had to add more dev dependencies, but I hope that's OK.

Current coverage doesn't look bad.

image

But the Pagination and Table are the only ones lacking now.

image

image

Going to add a few ones to make sure we don't have any regressions here, even though the code is easy to debug and fix if anything goes wrong.

kinow:
And of course writing tests I already found some bugs :bug: :sweat_smile: Going to fix and fix up the commits.

jena-fuseki2/jena-fuseki-ui/src/components/dataset/Pagination.vue
```diff
@@ -16,110 +16,124 @@
...
+        »
+      </button>
+    </li>
+  </ul>
 </template>
 
 <script>
+function range (size, startAt) {
+  return [...Array(size).keys()].map(i => i + startAt)
+}
```
kinow, line 131: image

Similar to Python's `range` function. If other components need it, this can be moved to a separate module and shared by them. For now keeping here in the `Pagination` component to simplify it.

jena-fuseki2/jena-fuseki-ui/src/components/dataset/Pagination.vue
```diff
@@ -16,110 +16,124 @@
...
+
 export default {
   name: 'Pagination',
 
-  prop: {
+  props: {
```
kinow, line 136: Ops :bug:

jena-fuseki2/jena-fuseki-ui/src/components/dataset/TableListing.vue
```diff
@@ -70,8 +70,8 @@
         </jena-table>
       </div>
     </div>
-    <div class="row g-0">
-      <div class="col-12">
+    <div class="row">
+      <div class="col-12 p-0">
```
kinow, line 74: It had some padding, causing the pagination component to be slightly smaller than the table. Now they can both grow/expand to the maximum available width, reaching the same width (or the pagination may wrap to the next line if the viewport is too small, which I believe is OK)

jena-fuseki2/jena-fuseki-ui/tests/unit/components/dataset/pagination.vue.spec.js
```diff
@@ -0,0 +1,125 @@
...
+
+    // Wait until $emits have been handled
+    await component.vm.$nextTick()
+
+    const inputEventsEmitted = component.emitted().input[0]
+    const singleEventEmitted = inputEventsEmitted[0]
+
+    expect(singleEventEmitted).toBe(theAnswer)
+  })
+})
```
kinow, line 125: The tests in Vitest UI (see `Pagination.vue`):

image

afs:
Looks great. Tried "vitest --coverage" as well. 😀 

Would it be better to have the graph names sorted so they appear in an predicable, albeit arbitrary,  order?
It can make it possible to guess how far through a long list to start looking for a particular graph and it'll be roughly the same as graphs are added and deleted, otherwise its "hashset" order.

That can be done in the server - not part of this PR.

afs (approved):

kinow:
>Would it be better to have the graph names sorted so they appear in an predicable, albeit arbitrary, order?
It can make it possible to guess how far through a long list to start looking for a particular graph and it'll be roughly the same as graphs are added and deleted, otherwise its "hashset" order.

Oh, snap. I looked at that so many times and never thought of that :rofl: big plus 1 for that.

>That can be done in the server - not part of this PR.

:+1: 

Another future improvement, but now for the UI, would be to keep the page number. I'd have to add the page number to the URL (e.g. `/current-url#page=2` or `/current-url?page=2`), read that param, validate the page exists, and update the component state.

Not too hard, and not a major issue. But nice to have for the future too.

Thanks @afs !

@LorenzBuehmann want to take a final look before it's merged and considered fixed?

afs:
One of the recent dependency updates seems to caused git conflict on package.json and yarn.lock.
Looking at the conflicts, I'm not sure the SHA512 has is right.

kinow:
> One of the recent dependency updates seems to caused git conflict on package.json and yarn.lock. I _think_ the second choice is correct except that `"vitest": "^1.3.1"` in this PR.

Hadn't noticed the conflict, thanks. I rebased, accepted `main`'s versions, then re-added `vitest` during rebase interactive, and `yarn install`. I think it's all done again.

kinow:
Spoke too soon, accidentally downgraded `vite-plugin-istanbul`, so had to re-add the latest version from `main`, and re install, edit commit :face_exhaling: Now it's done! :grimacing:

afs:
Thank you!

afs (approved):

kinow:
Going to merge this one, but @LorenzBuehmann feel free to re-open the issue, or create a new one or comment here or on the closed issue if you have any feedback :+1: Cheers

afs:
While everything seems to work, I get a non-fatal-to-the-build test failure on a timeout #2331. 

It reminds me of what we've had in the past.

It seems to be in a different area.

datasets -- after creating new datasets -- Visualizes the dataset information (Info View, tab) (failed)

kinow:
I wonder if that's the viewport. If you want to test you can try.... hmmm, to increease the browser window during the text.  If that doesn't solve it then maybe it's failing because I changed the table component (so just need to update the test :crossed_fingers: ). Thanks Andy!

PR #2314, opened 2024-03-07
GH-2311: Fix for union syntax

### LINKED ISSUE #2311: Typo in org.apache.jena.sparql.serializer.FormatterElement
### Version

4.10.0

### What happened?

I think the comparison should be `el.getElements().size() == 1` in the code snippet below.

https://github.com/apache/jena/blob/e54eb5e45fbb546342681046049509a76df58cc9/jena-arq/src/main/java/org/apache/jena/sparql/serializer/FormatterElement.java#L251-L255

Considering that `ElementUnion` always has at least two elements, this issue may not affect the correctness of the program.

_No response_

None

afs:
This closes #2311

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/serializer/FormatterElement.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/serializer/FormatterElement.java
@@ -248,7 +248,7 @@ public class FormatterElement extends FormatterBase implements ElementVisitor {
             return ;
         }
 
-        if ( el.getElements().size() == 0 ) {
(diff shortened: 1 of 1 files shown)


kinow (approved):
p.s. does it need a test? Or not easy/worth?

afs:
> does it need a test?

Tricky because it would work anyway (the loop will do the right thing for a union-of-one). Since the code already called out (wrongly - it can't be excuted) the union-of-one case, leaving the intended path seems clearer. YMMV.

PR #2329, opened 2024-03-11
GH-2328: Fuseki tests on MS Windows

### LINKED ISSUE #2328: Unstable build on MS Windows in some environments.
### Change

Running the build and tests on MS Windows with github actions or on Apache Jenkins sometimes fails in jena-fuseki-main, jena-fuseki-webapp, jena-integration tests.

What seems to be in common is test setup uses a single sever for the whole suite. This used to be necessary due to OS resources - it probably isn't anymore.

Yes

afs:
GitHub issue resolved #2328

Pull request Description:
Rewrite jena-fuseki-main tests to create a per-test Fuseki server.

----


afs:
Finger crossed. The test suite is less vulnerable to heavily loaded build servers causing test interference.  🤞🏻

jena-geosparql/src/test/java/org/apache/jena/ext/io/github/galbiston/expiring_map/ExpiringMapTest.java
```diff
@@ -441,6 +437,7 @@ public void testExpiry_change_expiry_interval_reduced() throws InterruptedExcept
      *
      * @throws java.lang.InterruptedException
      */
+    @Ignore // This test is not stable on loaded CI servers.
```
kinow, line 440: Is it for all CI servers or just Windows?

Maybe we could replace this by `@EnabledIfEnvironmentVariable` instead? We can have a bunch of tests like this (or fragile/unstable) and enable with `SOME_VAR=1 mvn test blabla`? https://junit.org/junit5/docs/5.2.0/api/org/junit/jupiter/api/condition/EnabledIfEnvironmentVariable.html & https://www.baeldung.com/junit-5-conditional-test-execution

Easier than using test suites or Maven profiles, I think.

kinow (approved):
Looks grand. One comment but not important. Thanks

jena-geosparql/src/test/java/org/apache/jena/ext/io/github/galbiston/expiring_map/ExpiringMapTest.java
afs: This test has been unstable on all OSes. It's weird - it goes through phases of failing over a period of a few days/weeks for a particular environment, then nothing, then back again. All I can think of is that it is the particular server it gets allocated and that's sticky.

This test is time related - it has built-in fixed time intervals and expected timeouts so long stalls in the test suite trigger the unexpected failure. I don't think it is that likely to affect a downstream build but it could.

My experience is that the effect of a heavily loaded server is different between Linux and Windows - there seems to be tendency to longer pauses on Windows - no idea why.

kinow (approved):

PR #2359, opened 2024-03-18
Gh 2358 from() and fromNamed() take object

### LINKED ISSUE #2358: from() and fromNamed() should take an Object rather than a String
### Version

5.x

### Feature

Currently the QueryBuilder framework has both from() and fromNamed() methods that take a String argument.   There are cases where another object (a Node object for instance) contains the name.  To make the methods symmetric with the other methods that convert things to Nodes these methods should take Objects and attempt to convert them to Strings.

Yes

Claudenw:
GitHub issue resolved #2358

Pull request Description:
Changes QueryBuilder `from()` and `fromNamed()` methods to accept Objects.

No documentation change required.
----

 - [x] Tests are included.

----

### DIFF
--- a/jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/UpdateBuilder.java
+++ b/jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/UpdateBuilder.java
@@ -123,8 +123,8 @@ public class UpdateBuilder {
     }
 
     /**
-     * Checks that deletes or inserts have been added to the builder.
-     * @return true if there are any delete or insert statements.
+     * Checks that no deletes or inserts have been added to the builder.
+     * @return true if there are no delete or insert statements.
      */
     public boolean isEmpty() {
         return deletes.isEmpty() && inserts.isEmpty();
--- a/jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/UpdateBuilder.java
+++ b/jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/UpdateBuilder.java
@@ -123,6 +123,14 @@ public class UpdateBuilder {
     }
 
     /**
+     * Checks that deletes or inserts have been added to the builder.
+     * @return true if there are any delete or insert statements.
+     */
+    public boolean isEmpty() {
+        return deletes.isEmpty() && inserts.isEmpty();
+    }
+
+    /**
      * Build the update.
      *
      * <b>Note: the update does not include the prefix statements</b> use
@@ -132,7 +140,7 @@ public class UpdateBuilder {
      */
     public Update build() {
 
-        if (deletes.isEmpty() && inserts.isEmpty()) {
+        if (isEmpty()) {
             throw new IllegalStateException("At least one delete or insert must be specified");
         }
 
--- a/jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/handlers/DatasetHandler.java
+++ b/jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/handlers/DatasetHandler.java
@@ -21,7 +21,10 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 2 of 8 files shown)


jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/UpdateBuilder.java
```diff
@@ -122,6 +122,14 @@ private ExtendedIterator<Quad> getQuads(Collection<QuadHolder> holders) {
         return result;
     }
 
+    /**
+     * Checks that deletes or inserts have been added to the builder.
+     * @return true if there are any delete or insert statements.
```
kinow, line 127: Returns true if there are no deletes and no inserts? Because the return is checking `isEmpty` for both....

jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/handlers/DatasetHandler.java
```diff
@@ -54,24 +57,42 @@ public void addAll(DatasetHandler datasetHandler) {
...
      *
-     * @param graphNames The from names to add.
+     * @param graphName The graph name to add.
      */
-    public void fromNamed(Collection<String> graphNames) {
-        for (String uri : graphNames) {
-            query.addNamedGraphURI(uri);
+    public void fromNamed(Object graphName) {
+        if (graphName instanceof Collection) {
+            ((Collection<?>)graphName).forEach(this::fromNamed);
```
kinow, line 89: Couldn't this be used to create one of those request bombs that keep iterating until it causes a StackOverflowError? That's probably fine since external users won't have access to this methods, I think, only application developers... but looks a bit dangerous (maybe worth adding a note in the docs that it's a recursive method... or replace the recursion by a stack+while, etc...). Same the `from` method.

kinow (commented):
A couple of comments, but otherwise looks good to me.

rvesse (approved):

jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/handlers/DatasetHandler.java
```diff
@@ -54,24 +57,42 @@ public void addAll(DatasetHandler datasetHandler) {
...
+     * Add a graph name to the from named list.
      *
-     * @param graphNames The from names to add.
+     * @param graphName The graph name to add.
      */
-    public void fromNamed(Collection<String> graphNames) {
-        for (String uri : graphNames) {
-            query.addNamedGraphURI(uri);
+    public void fromNamed(Object graphName) {
+        if (graphName instanceof Collection) {
```
afs, line 88: Coudl be a java17 

```
if (graphName instanceof Collection collection) {
   collection.forEach(this:fromNamed);
}
```

afs (approved):
One observation

jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/handlers/DatasetHandler.java
Claudenw: I had problems converting the javadoc to version 17 , cause a problem that I will have to work out.

jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/UpdateBuilder.java
Claudenw: good catch.  fixed

kinow (approved):

jena-extras/jena-querybuilder/src/main/java/org/apache/jena/arq/querybuilder/handlers/DatasetHandler.java
afs: Javadoc - yes :-( It can be a bit bumpy.

The build uses Java17 so language features are available if you want. Doesn't requite it all converted. Use, or not, as desired.

The javadoc seems to be built properly.

PR #2395, opened 2024-04-07
Fix for query exec when abort is called before exec

### LINKED ISSUE #2394: QueryExec: abort before exec is ignored.
### Version

5.1.0-SNAPSHOT

### What happened?

The following test does not raise the expected QueryCancelledException, in fact, `abort()` is ignored.

```java
    @Test(expected = QueryCancelledException.class)
    public void test_Cancel_API_5() {
        model.add(RDF.type, RDF.type, RDF.type);
        try (QueryExecution qe = QueryExecutionFactory.create("SELECT * { ?s ?p ?o }", model)) {
            qe.abort();
            ResultSetFormatter.consume(qe.execSelect());
        }
    }
```

Note, that because `abort` may be called asynchronously, abort signals must be handled even if the e.g. the thread that called `execSelect()` has not finished the internal QueryIterator setup yet.

_No response_

Yes

Aklakan:
GitHub issue resolved #2394

Pull request Description: Cancel signal is moved to class level and is set to true when abort is called.

----

 - [x] Tests are included.
 - [x] Commits have been squashed to remove intermediate development commit messages. (test case and fix are separate commits)

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/exec/QueryExecDataset.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/exec/QueryExecDataset.java
@@ -89,6 +89,7 @@ public class QueryExecDataset implements QueryExec
     private long                     timeout2         = TIMEOUT_UNSET;
     private final AlarmClock         alarmClock       = AlarmClock.get();
     private long                     queryStartTime   = -1; // Unset
+    private AtomicBoolean            cancelSignal     = new AtomicBoolean(false);
 
     protected QueryExecDataset(Query query, String queryString, DatasetGraph datasetGraph, Context cxt,
                                QueryEngineFactory qeFactory,
@@ -142,6 +143,7 @@ public class QueryExecDataset implements QueryExec
 
     @Override
     public void abort() {
+        cancelSignal.set(true);
         synchronized (lockTimeout) {
             // This is called asynchronously to the execution.
             // synchronized is for coordination with other calls of
@@ -166,11 +168,9 @@ public class QueryExecDataset implements QueryExec
     }
 
     private RowSet execute() {
-        execInit();
         startQueryIterator();
-        Iterator<Binding> iter = queryIterator;
         List<Var> vars = query.getResultVars().stream().map(Var::alloc).collect(Collectors.toList());
-        return RowSetStream.create(vars, iter);
+        return RowSetStream.create(vars, queryIterator);
     }
 
     // -- Construct
@@ -357,12 +357,6 @@ public class QueryExecDataset implements QueryExec
     }
 
     class TimeoutCallback implements Runnable {
-        private final AtomicBoolean cancelSignal;
-
-        public TimeoutCallback(AtomicBoolean cancelSignal) {
-            this.cancelSignal = cancelSignal;
-        }
-
         @Override
         public void run() {
             synchronized (lockTimeout) {
@@ -379,11 +373,9 @@ public class QueryExecDataset implements QueryExec
     }
 
     private class QueryIteratorTimer2 extends QueryIteratorWrapper {
-        private final AtomicBoolean cancelSignal;
 
-        public QueryIteratorTimer2(QueryIterator qIter, AtomicBoolean cancelSignal) {
+        public QueryIteratorTimer2(QueryIterator qIter) {
             super(qIter);
-            this.cancelSignal = cancelSignal;
         }
 
         long yieldCount = 0;
@@ -400,7 +392,7 @@ public class QueryExecDataset implements QueryExec
                 // So nearly not needed.
                 synchronized(lockTimeout)
                 {
-                    TimeoutCallback callback = new TimeoutCallback(cancelSignal);
+                    TimeoutCallback callback = new TimeoutCallback();
                     expectedCallback.set(callback);
                     // Lock against calls of .abort() or of timeout1Callback.
 
@@ -437,10 +429,32 @@ public class QueryExecDataset implements QueryExec
             queryStartTime = System.currentTimeMillis();
     }
 
-    /** Start the query iterator, setting timeouts as needed. */
+    /** Wrapper for starting the query iterator but also dealing with cancellation */
     private void startQueryIterator() {
-        if ( queryIterator != null )
+        synchronized (lockTimeout) {
+            if (cancelSignal.get()) {
+                // Fail before starting the iterator if cancelled already
+                throw new QueryCancelledException();
+            }
+
+            startQueryIteratorActual();
+
+            if (cancelSignal.get()) {
+                queryIterator.cancel();
+
+                // Fail now if cancelled already
+                throw new QueryCancelledException();
+            }
+        }
+    }
+
+    /** Start the query iterator, setting timeouts as needed. */
+    private void startQueryIteratorActual() {
+        if ( queryIterator != null ) {
             Log.warn(this, "Query iterator has already been started");
+            return;
+        }
+
         execInit();
 
         /* Timeouts:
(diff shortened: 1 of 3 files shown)


namedgraph:
Reminds me of this issue I had but I never got around to isolating it...
https://www.mail-archive.com/users@jena.apache.org/msg20564.html

Aklakan:
I added another test case for concurrent abort which should (occasionally) hang prior to this patch - the reason it should hang is that sometimes abort happens to be called before the execution starts - this scenario is also covered by the non-concurrent test case.

if it doesn't hang, then maybe the test's task count per cpu needs to be increased - feedback is welcome.

afs:
Improving this is good to have.

To try my understanding out, I wrote some tests. Maybe I'm misunderstanding the intention.

Whole file: [TestCancelQueryAFS.java.txt](https://github.com/apache/jena/files/14980629/TestCancelQueryAFS.java.txt)

It is what happens for CONSTRUCT et al, and whether abort before exec does cancel when there is no iteration.

`test_cancel_select_1()` does not cancel. It does not touch the iterator.
`test_cancel_select_2()` does cancel.

In the tests file there are tests for CONSTRUCT and ASK. I haven't figured out why these don't cancel - I would have thought the iterator was run.

There is also a copy of test_Cancel_API_5 and variations to show I am runnign the PR code :smile:

```java
    @Test
    public void test_cancel_select_1() {
        cancellationTest("SELECT * {}", exec->exec.select());
    }

    @Test
    public void test_cancel_select_2() {
        cancellationTest("SELECT * {}", exec->exec.select().hasNext());
    }

. . .

    static void cancellationTest(String queryString, Consumer<QueryExec> execAction) {
        DatasetGraph dsg = DatasetGraphFactory.createTxnMem();
        dsg.add(SSE.parseQuad("(_ :s :p :o)"));

        try(QueryExec aExec = QueryExec.dataset(dsg).query(queryString).build()) {
            aExec.abort();
            assertThrows(QueryCancelledException.class, ()-> execAction.accept(aExec));
        }
    }
```

Aklakan:
test_cancel_select_1 does not raise the exception because the iterator is not accessed - the cancel flag is set though.
IMO QueryCancelledException should only be thrown on iterator access - if it were raised directly on `select()` then in general one would need two exception handlers - one on the iterator and one on the query exec level. The try-with-resources block closes resources of a query exec whose iterator has not been touched - so there is shouldn't be a resource leak.

I have so far only looked at the select case (not construct, describe, etc) - should I look into this and extend the PR or will you work on it?

Aklakan:
> So I'd say that the correct behavior

Hm, "correctness" is the wrong wording - its an API contract question. The query exec methods that do not return an iterator are expected to fail right away (because they'll internally touch the iterator). However, if an iterator is returned I am uncertain whether the creation of the iterator should already fail due to cancellation - or whether the place to check for cancellation of an iterator is solely on next() (IIRC hasNext returns true even if cancelled, but picking the item up with next() will raise the exception.).

jena-arq/src/main/java/org/apache/jena/sparql/exec/QueryExecDataset.java
```diff
@@ -89,6 +89,7 @@ public class QueryExecDataset implements QueryExec
     private long                     timeout2         = TIMEOUT_UNSET;
     private final AlarmClock         alarmClock       = AlarmClock.get();
     private long                     queryStartTime   = -1; // Unset
+    private AtomicBoolean            cancelSignal     = new AtomicBoolean(false);
```
afs, line 92: Because this has moved to the class, it does not need to be passed as an argument to the constructors of the inner classes.  

Because the classes have members of the same name, I'm getting a compiler warning.

afs:
Thanks for pointing out that failing at `select()` is a visible change.

I'm personally in favour of consistency (my perceived idea of consistency!) which is fail ASAP. The `QueryExec` has been created and the abort operation is on that.

```
    qexec.abort();
    qexec.select();
```

`select()` does already get questions asked so why is it quick and then the query is slow.

`construct()` isn't failing nor is `ask()`. `describe()` is.

`ask()` calls `.next()` once and the comment says something about timeouts.

`.hasNext()` is often the place where the real work has to happen because to know whether there is a "next" , the engine has to do the work!

Suggestion: merge this PR as a step forward (a couple of warnings have appeared - `cancelSignal` is now a member and being hidden by the inner class timeout) addressing #2394 and do more in another issue/PR.

Aklakan:
> I'm personally in favour of consistency (my perceived idea of consistency!) which is fail ASAP. 

Right, I agree that fail fast is typically easier to debug.

I noticed that `execute()` was only used for select and describe, whereas `startQueryIterator()` is the method actually common to all operations - so I moved the abort check there.

I also extended your iterator-based test cases to make two assertions whether abort happens before or after obtaining the iterator.

The test cases now work according to our understanding.

jena-arq/src/main/java/org/apache/jena/sparql/exec/QueryExecDataset.java
Aklakan: Removed the superfluous ctor argument from the inner classes.

jena-arq/src/test/java/org/apache/jena/sparql/api/TestCancelQueryAFS.java
```diff
@@ -0,0 +1,155 @@
+package org.apache.jena.sparql.api;
+
+/*
+ *  Copyright (c) Telicent Ltd.
```
Aklakan, line 4: License header likely needs adjustment (this is the only file with a Telicent header)

jena-arq/src/test/java/org/apache/jena/sparql/api/TestCancelQueryAFS.java
```diff
@@ -0,0 +1,155 @@
...
+import org.apache.jena.rdf.model.Model;
+import org.apache.jena.rdf.model.ModelFactory;
+import org.apache.jena.sparql.core.DatasetGraph;
+import org.apache.jena.sparql.core.DatasetGraphFactory;
+import org.apache.jena.sparql.exec.QueryExec;
+import org.apache.jena.sparql.sse.SSE;
+import org.apache.jena.vocabulary.RDF;
+import org.junit.Test;
+
+public class TestCancelQueryAFS {
```
Aklakan, line 38: Merge into TestQueryExecutionCancel?

afs:
Looks good.

I'll do some clearup - like not call a test class "... AFS". I've been looking at the Java memory model recently and there a volatile which could at least be an `AtomicReference` and maybe use fences, not full volatile semantics. It won't make an observable difference in this situation but it least the pattern will be recorded - not going as far as `VarHandles` which look quite complicated and fragile for general code.

jena-arq/src/test/java/org/apache/jena/sparql/api/TestCancelQueryAFS.java
afs: Yes. And drop the tests that are copies (`test_Cancel_API_5`).

afs (approved):

jena-arq/src/test/java/org/apache/jena/sparql/api/TestCancelQueryAFS.java
Aklakan: Methods are now merged into TestQueryExecutionCancel. Duplicated tests are removed.

Aklakan:
I removed the class TestQueryAFS and moved the methods to TestQueryExecutionCancel.

Are you going to adapt the code from volatile to AtomicReference?

afs:
Excellent.

Let's merge this and do any refinement as and when time permits.

PR #2409, opened 2024-04-10
Migrating Base 64 operations from Apache Commons Codec to Util package.

TelicentPaul:
GitHub issue resolved # N/A

Pull request Description:

Updating repo to use the Util's Base 64 class (introduced in JDK 8) which has greater performance than Apache Commons Codec.

----

 - [ ] Tests are included.

----

### DIFF
--- a/jena-integration-tests/src/test/java/org/apache/jena/http/AuthBearerTestLib.java
+++ b/jena-integration-tests/src/test/java/org/apache/jena/http/AuthBearerTestLib.java
@@ -80,7 +80,7 @@ public class AuthBearerTestLib {
     private static String enc64(String x) {
         byte[] bytes = x.getBytes(StandardCharsets.UTF_8);
         // URL encoding, no padding, no chunking line breaks.
-        String s = Base64.getUrlEncoder().encodeToString(bytes);
+        String s = Base64.getEncoder().encodeToString(bytes);
         return s;
     }
 
--- a/jena-integration-tests/src/test/java/org/apache/jena/http/AuthBearerTestLib.java
+++ b/jena-integration-tests/src/test/java/org/apache/jena/http/AuthBearerTestLib.java
@@ -19,9 +19,8 @@
 package org.apache.jena.http;
 
 import java.nio.charset.StandardCharsets;
+import java.util.Base64;
 import java.util.Objects;
-
-import org.apache.commons.codec.binary.Base64;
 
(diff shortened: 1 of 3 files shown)


kinow (approved):

afs:
See [RFC 4648](https://datatracker.ietf.org/doc/html/rfc4648) for the discussion and tables.

There is also [RFC 2045](https://datatracker.ietf.org/doc/html/rfc2045)

* [`xsd:base64Binary`](https://www.w3.org/TR/xmlschema-2/#base64Binary), which is defined to be RFC 2045 with the addition of XML white space making it RFC 4686 section 4.
* Bearer tokens `enc64` uses RFC4648 table2 (RFC 4686 section 5) -- this may be wrong.

The two areas of Base64 are different alphabets --  `XSDbase64Binary.java`, and `makeSkolem.java` is about `xsd:base64binary` which is RFC 2045 with white space and breaks at 76 characters.

`AuthBearerTestLib.java` `enc64` uses is based around `base64url` (`encodeBase64URLSafeString`).

OAuth and JWT seem to be `base64` (white space handling isn't clear).

This is hidden because Apache Commons Codec Base64 decoding is universal.

I am unclear what the right combination is.

Performance links:
* https://aws.amazon.com/blogs/developer/follow-up-on-base64-codec-performance
* https://dkomanov.medium.com/base64-encoding-performance-jdk-vs-apache-commons-3fb83323414b

jena-integration-tests/src/test/java/org/apache/jena/http/AuthBearerTestLib.java
```diff
@@ -81,7 +80,7 @@ public static String generateTestToken(String user) {
     private static String enc64(String x) {
         byte[] bytes = x.getBytes(StandardCharsets.UTF_8);
         // URL encoding, no padding, no chunking line breaks.
-        String s = org.apache.commons.codec.binary.Base64.encodeBase64URLSafeString(bytes);
+        String s = Base64.getUrlEncoder().encodeToString(bytes);
```
afs, line 83: OAuth at least is not `base64url`, but is `base64`.
This bug is concealed by using the general decoder. If changing the decoder to only `base64`, this test needs to be fixed.

```suggestion
        String s = Base64.getEncoder().encodeToString(bytes);
```

afs (changes requested):

PR #2413, opened 2024-04-11
GH-2412: Improved JoinClassifier for unbound values

### LINKED ISSUE #2412: Incorrect JoinClassifier results with unbound values.
### Version

5.1.0-SNAPSHOT

### What happened?

The following two variants incorrectly return 1 binding whereas 0 would be expected.
The reason is that JoinClassifier.isLinear returns true because VarFinder does not consider the unbound values.

A linear join substitutes the (in-scope) variables on the right-hand-side (rhs) with each binding from the left-hand-side (lhs). This causes the incorrect substitution of ?x on the rhs with `0` such that `bound(?x)` yields true.

```sparql
SELECT * {
  VALUES ?x {
    0
  }
  {
    VALUES ?x {
      UNDEF
    }
    FILTER(bound(?x))
  }
}
```

```sparql
SELECT * {
  BIND(0 AS ?x)
  {
    BIND(coalesce() AS ?x)
    FILTER(bound(?x))
  }
}
```

_No response_

Yes

Aklakan:
GitHub issue resolved #2412

Pull request Description: Additional heuristics and test cases in JoinClassifier and VarFinder for dealing with unbound values.

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/main/VarFinder.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/main/VarFinder.java
@@ -25,18 +25,26 @@ import static org.apache.jena.sparql.util.VarUtils.addVarsFromTriple ;
 import static org.apache.jena.sparql.util.VarUtils.addVarsFromTriplePath ;
 
 import java.io.PrintStream ;
+import java.util.ArrayList;
 import java.util.HashSet ;
+import java.util.Iterator;
 import java.util.List ;
 import java.util.Set ;
 
+import org.apache.jena.atlas.iterator.Iter;
 import org.apache.jena.atlas.lib.SetUtils ;
 import org.apache.jena.sparql.algebra.Op ;
 import org.apache.jena.sparql.algebra.OpVisitor ;
+import org.apache.jena.sparql.algebra.Table ;
 import org.apache.jena.sparql.algebra.op.* ;
 import org.apache.jena.sparql.core.BasicPattern ;
 import org.apache.jena.sparql.core.Var ;
 import org.apache.jena.sparql.core.VarExprList ;
+import org.apache.jena.sparql.engine.binding.Binding ;
+import org.apache.jena.sparql.expr.E_Bound;
+import org.apache.jena.sparql.expr.E_Coalesce;
 import org.apache.jena.sparql.expr.Expr ;
+import org.apache.jena.sparql.expr.ExprFunction;
 import org.apache.jena.sparql.expr.ExprList ;
 import org.apache.jena.sparql.expr.ExprVars;
 import org.apache.jena.sparql.util.VarUtils ;
@@ -329,10 +337,43 @@ public class VarFinder
 
         private void processAssignVarExprList(VarExprList varExprList) {
             varExprList.forEachVarExpr((v,e)-> {
-                defines.add(v) ; // Expression may eval to error -> unset?
-                if ( e != null )
+                if ( e != null ) {
+                    boolean isLikelyToBeDefined = isLikelyToBeDefined(e);
+                    if (isLikelyToBeDefined) {
+                        defines.add(v) ; // Expression may eval to error -> unset?
+                    } else {
+                        optDefines.add(v);
+                    }
                     ExprVars.nonOpVarsMentioned(assignMentions, e);
+                }
             }) ;
+        }
+
+        /**
+         * Heuristic to check whether the expression's evaluation result is likely
+         * to always be defined (i.e. never null / type error).
+         * Does not detect runtime type errors, such as when
+         * expr is (?x < 1) and ?x is assigned a string.
+         */
+        private boolean isLikelyToBeDefined(Expr expr) {
+            boolean result;
+            if (expr.isVariable()) {
+                result = defines.contains(expr.asVar());
+            } else if (expr.isConstant() || expr instanceof E_Bound) {
+                result = true;
+            } else if (expr.isFunction()) {
+                ExprFunction fn = expr.getFunction();
+                List<Expr> args = fn.getArgs();
+                if (fn instanceof E_Coalesce) {
+                    result = Iter.anyMatch(args.iterator(), this::isLikelyToBeDefined);
+                } else {
+                    result = Iter.allMatch(args.iterator(), this::isLikelyToBeDefined);
+                }
+            } else {
+                // Should never happen
+                result = false;
+            }
+            return result;
         }
 
         @Override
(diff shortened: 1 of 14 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/engine/main/VarFinder.java
```diff
@@ -351,9 +389,26 @@ public void visit(OpProject opProject) {
...
+                while (varIt.hasNext()) {
+                    Var var = varIt.next();
+                    if (!row.contains(var)) {
+                        varIt.remove();
+                        optVars.add(var);
+                    }
+                }
+            }
+            defines.addAll(definedVars);
+            optDefines.addAll(optVars);
```
Aklakan, line 411: Are the arguments against treating unbound table values as optDefines?

jena-arq/src/test/java/org/apache/jena/sparql/core/AbstractTestDynamicDataset.java
```diff
@@ -204,7 +204,17 @@ public void dynamic_dft_7() {
                   "{ GRAPH <urn:x-arq:DefaultGraph> { ?s <uri:p> ?o . FILTER ( ?o IN ( 1, 2) ) } }",
                   2, dataset);
     }
-    
+
+    @Test
```
Aklakan, line 208: This class is not an ideal fit for the join tests. Any suggestion?

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/VarFinder.java
```diff
@@ -329,12 +336,43 @@ public void visit(OpExtend opExtend) {
 
         private void processAssignVarExprList(VarExprList varExprList) {
             varExprList.forEachVarExpr((v,e)-> {
-                defines.add(v) ; // Expression may eval to error -> unset?
-                if ( e != null )
+                if ( e != null ) {
+                    boolean isLikelyToBeDefined = isLikelyToBeDefined(e);
+                    if (isLikelyToBeDefined) {
+                        defines.add(v) ; // Expression may eval to error -> unset?
```
Aklakan, line 342: If the variable may be unbound it probably still needs to be tracked - e.g. in optDefines?

Aklakan:
I just removed some left-over code that had slipped into this PR.
The proposed changes are ready for discussion.

jena-arq/src/test/java/org/apache/jena/sparql/core/AbstractTestDynamicDataset.java
rvesse: Can you not make the `testCount()` method `public static` and then create a new test class that simply calls it?

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/VarFinder.java
```diff
@@ -351,9 +389,26 @@ public void visit(OpProject opProject) {
...
+            Table table = opTable.getTable();
+            Set<Var> definedVars = new HashSet<>(table.getVars());
+            List<Var> optVars = new ArrayList<>();
+            Iterator<Binding> rowIt = table.rows();
+            while (rowIt.hasNext()) {
+                Binding row = rowIt.next();
+                Iterator<Var> varIt = definedVars.iterator();
+                while (varIt.hasNext()) {
+                    Var var = varIt.next();
+                    if (!row.contains(var)) {
```
rvesse, line 404: This seems overly aggressive at first glance.  If I am reading this correctly any row that does not contain a value triggers the join linearisation being blocked. What happens if you have multiple rows with some values and some `UNDEF`, is that going to create bad results if we allow linearisation?

Perhaps a different solution is needed around handling of linear joins for `OpTable` to just never substitute over an `UNDEF` though guess that might be significantly more involved?

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/VarFinder.java
rvesse: No, makes sense to me

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/VarFinder.java
Aklakan: > If I am reading this correctly any row that does not contain a value triggers the join linearisation being blocked.

Its per variable - so if there is an UNDEF value for a variable then linearization for that variable gets blocked.
Conversely, if the variable is defined in all rows its added to the defined set.
I was creating test cases for skewed multi-variable joins (i.e. joins on multiple variables where some of them are unbound) when I noticed that I get incorrect results.

jena-arq/src/test/java/org/apache/jena/sparql/core/AbstractTestDynamicDataset.java
Aklakan: With my question I was mainly referring to whether I have overlooked an already existing suitable place for these tests - if there is none so far I could of course create a new class and register it with the test suite.

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/VarFinder.java
afs: Yes - that seems likely. 
Not putting it in is quite safe.
Doing so and tests all pass.

Aklakan: Done

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/VarFinder.java
Aklakan: Added variable to optDefines if it is not considered defined.

jena-arq/src/test/java/org/apache/jena/sparql/core/AbstractTestDynamicDataset.java
Aklakan: There is now a AbstractTestQueryExec hierarchy analogous to AbstractTestDynamicDataset.

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/VarFinder.java
```diff
@@ -329,12 +336,43 @@ public void visit(OpExtend opExtend) {
...
+            boolean result;
+            if (expr.isVariable()) {
+                result = defines.contains(expr.asVar());
+            } else if (expr.isConstant() || expr instanceof E_Bound) {
+                result = true;
+            } else if (expr.isFunction()) {
+                ExprFunction fn = expr.getFunction();
+                List<Expr> args = fn.getArgs();
+                if (fn instanceof E_Coalesce) {
+                    result = args.stream().anyMatch(this::isLikelyToBeDefined);
```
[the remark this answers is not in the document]
Aklakan: Done

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/VarFinder.java
```diff
@@ -329,12 +336,43 @@ public void visit(OpExtend opExtend) {
...
+                } else {
+                    result = args.stream().allMatch(this::isLikelyToBeDefined);
```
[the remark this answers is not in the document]
Aklakan: Done

Aklakan:
I think I addressed all changed. You probably still want to check the AbstractTestQueryExec / TestQueryExec* classes that I added.

afs:
Looks good.

Attached are some more tests `TestVarFinder2` for `VarFinder` directly.
If you agree with my understanding, please add the file and add to `TS_Algebra`.
(5 tests fail in Jena without this PR).
[TestVarFinder2.java.txt](https://github.com/apache/jena/files/15049216/TestVarFinder2.java.txt)

One clearing thing:
`AbstractTestDynamicDataset` - the `join_*` test are still there.

Aklakan:
The test cases in TestVarFinder2 look good to me.
I added one more assertion to `varfind_extend_6`: `assertSetVars(vf.getOpt(), var_x);`

> AbstractTestDynamicDataset - the join_* test are still there.

Whoops, removed.

afs (approved):

afs:
Thanks!

PR #2405, opened 2024-04-08
GH-2404: Support for multi-variable join keys

Aklakan:
GitHub issue resolved #2404

Pull request Description: This PR proposes introduction of the class MultiHashProbeTable which creates HashProbleTable indexes on demand based on the lookup requests.

~I still need to test the performance with~ @LorenzBuehmann 's examples from the [mail thread](https://www.mail-archive.com/users@jena.apache.org/msg20755.html). The examples have been added to the `jena-benchmarks-jmh` module.

----

 - [x] Tests are included.
 - [x] Benchmarks are included.
 - [x] Commits have been squashed to remove intermediate development commit messages.
 - [x] Key commit messages start with the issue number (GH-xxxx)

By submitting this pull request, I acknowledge that I am making a contribution to the Apache Software Foundation under the terms and conditions of the [Contributor's Agreement](https://www.apache.org/licenses/contributor-agreements.html).

----

See the [Apache Jena "Contributing" guide](https://github.com/apache/jena/blob/main/CONTRIBUTING.md).

PR #2536, opened 2024-06-12
GH-2535: Always use hash joins when joining VALUES blocks

### LINKED ISSUE #2535: Always use hash joins when joining VALUES blocks
### Version

5.1.0-SNAPSHOT

### Feature

This is one more follow up towards support for multi variable join keys as triggered by the mail thread around https://www.mail-archive.com/users@jena.apache.org/msg20755.html

Related issues:  #2412 and #2404

The mail mentions example sparql queries, such as the one below, that use value blocks to exemplify the issue with multi-variable joins.

```sparql
# test.rq
select (count(*) as ?C) where {
  {
    select ?X ?Y (struuid() as ?UUID) where {
      values ?X_i { 0 1 2 3 4 5 6 7 8 9 }
      values ?X_j { 0 1 2 3 4 5 6 7 8 9 }
      bind ( ?X_i + 10 * ?X_j as ?X)
      values ?Y_i { 0 1 2 3 4 5 6 7 8 9 }
      values ?Y_j { 0 1 2 3 4 5 6 7 8 9 }
      bind ( ?Y_i + 10 * ?Y_j as ?Y) 
    } 
  } {
    select ?X ?Y where {
      {
        select ?X ?Y (rand() as ?RAND) where {
          values ?X_i { 0 1 2 3 4 5 6 7 8 9 }
          values ?X_j { 0 1 2 3 4 5 6 7 8 9 }
          bind ( ?X_i + 10 * ?X_j as ?X)
          values ?Y_i { 0 1 2 3 4 5 6 7 8 9 }
          values ?Y_j { 0 1 2 3 4 5 6 7 8 9 }
          bind ( ?Y_i + 10 * ?Y_j as ?Y) 
        } 
      } filter (?RAND < 0.95) 
    } 
  } 
}
```
Jena's JoinClassifier so far linearizes joins between values blocks, such as the given SPARQL example, which gets very slow for larger values blocks.
An extra flag is necessary to force hash joins:

```bash
arq --explain --time --set arq:optIndexJoinStrategy=false --query test.rq
```

The goal of this issue is to modify JoinClassifier such that joins between table-based operands are automatically processed as hash joins rather than linear joins.

Yes

Aklakan:
GitHub issue resolved #2535 

Pull request Description: Update to JoinClassifier such that joins between tables are not linearized.
The downside is, that this PR adds extra passes over the algebra tree to JoinClassifier in order to check for the operand types.

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/main/JoinClassifier.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/main/JoinClassifier.java
@@ -18,8 +18,10 @@
 
 package org.apache.jena.sparql.engine.main ;
 
+import java.util.Iterator;
 import java.util.Set ;
 
+import org.apache.jena.atlas.iterator.Iter;
 import org.apache.jena.atlas.lib.SetUtils ;
 import org.apache.jena.sparql.algebra.Op ;
 import org.apache.jena.sparql.algebra.OpVisitor ;
@@ -62,6 +64,12 @@ public class JoinClassifier
 
         // Lateral is different.
         if ( right instanceof OpLateral )   return false ;
+
+        // Do not linearize when effectively joining tables - i.e. force hash joins between tables.
+        Basis leftBasis = getBasis(left);
+        Basis rightBasis = getBasis(right);
+        if ( leftBasis == Basis.TABLE && rightBasis == Basis.TABLE )
+            return false ;
 
         // Assume something will not commute these later on.
         return check(left, right) ;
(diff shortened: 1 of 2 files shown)


arne-bdt:
This looks amazing, but it is also at the heart of the SPARQL engine. 
Could there be cases where joins are slower now than before? 
Do we need regression benchmarks for join performance? 
I simply don't feel confident enough to review this PR.

Aklakan:
> Do we need regression benchmarks for join performance?
My intention was to use this PR as a basis for regression testing of #2405, where the benchmark queries contain VALUE blocks that join on more than one variable.
However, without this PR, the linear joins are so slow that a workaround is needed to enforce hash joins.
For example, hash joins are enforced by wrapping the right-hand-side with a LIMIT:

```
// The following extra test cases for TestClassify.java are GREEN _without_ this PR:
    @Test public void testClassify_Join_values_plain() {
        TestClassify.classifyJ(
        """
        {
          VALUES ?z { 0 }
          VALUES ?z { 1 }
        }
        """,
        true); // true means linear join
    }

    @Test public void testClassify_Join_values_wrapped() {
        TestClassify.classifyJ(
        """
        {
          VALUES ?z { 0 }
          { SELECT * { VALUES ?z { 1 } } LIMIT 1000000000 }
        }
        """,
        false); // false means no linear join (i.e.executed as hash join)
    }
```
With this PR, the workaround is no longer needed.

> Could there be cases where joins are slower now than before?
That's the central question, and it would be good to get feedback on this :)

So far I could not think of any good examples. The PR introduces a slight overhead for checking the operand types but I'd say this is outweighed by the overall performance improvement for queries that inline data using VALUE blocks.

As a further improvement for this PR, instead of adding the logic to detect table-operands to JoinClassifier, it could also be moved to [TransformJoinStrategy](https://github.com/apache/jena/blob/f666547086d4a474ef62a07e8d1a4a2b336022fa/jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformJoinStrategy.java#L39) - In that case, only if the original JoinClassifier returned "linear join" then an extra check for table operands would be made that could change the verdict to "hash join".

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/JoinClassifier.java
```diff
@@ -235,4 +243,85 @@ private static boolean safeModifier(Op op) {
...
+     * If that side of the join is a property function then {@link Basis#PFUNCTION} is returned.
+     *
+     * <p>
+     * The special handling of property functions is due to that
+     * OpJoin(TABLE, PFUNCTION) needs to be linearized to OpSequence(TABLE, PFUNCTION).
+     *
+     * <p>
+     * This method resolves OpExt to its effective op.
+     */
+    private static Basis getBasis(Op op) {
```
afs, line 268: This will look better when switch/case/instanceof is available! (Java21, not Java17)

afs (approved):

afs:
Sorry this has taken so long.

Aklakan:
> Sorry this has taken so long.

No worries - thanks for your time for looking into this!

Issue #1793, opened 2023-03-09
Fuseki UI: custom SPARQL query prefixes or examples


nichtich:
### Version

4.7.0

### Feature

The default SPARQL query 

~~~sparql
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
SELECT * WHERE {
  ?sub ?pred ?obj .
} LIMIT 10
~~~

Should be extendable with custom prefixes used in a database or graph. An optional set of sample queries would also help. Open question is where to store these configuration. Fuseki and database configurations beyond list of services are not made available other than via /$/datasets. An option would be to introduce a JSON configuration file in the webapp root at http://localhost:3030/config.json. This would allow for additional run-time configuration of the user interface (no need to rebuild) such as custom CSS, title, logo etc.

### Are you interested in contributing a solution yourself?

Yes

nichtich:
Looks like this is covered by https://github.com/apache/jena/pull/1683

afs:
@nichtich If you think it is covered elsewhere, please close the issue.

PR #2538, opened 2024-06-14
GH-2543: Prefixes service

### LINKED ISSUE #2543: Prefixes Service
### Version

5.1.0-SNAPSHOT

### Feature

A service allowing for prefix storage and management supporting prefix and namespace lookup, update, and remove operations.
The requests to the API could have 3 parameters: prefix, uri, and removePrefix, the combination of which would define the operation performed. The read, HTTP GET, operations would be fetching the prefix by URI, fetching the URI by prefix, and fetching all. The read-write, invoked via HTTP POST, operations are then removing and updating prefix-URI pairs.

Yes

karolina-telicent:
This resolves #2543

Pull request Description:

This pull request contains the addition of the prefixes service to Jena. The prefixes service allows for prefix management on a given dataset provided in the config.ttl file. When making requests to the API the url can have 3 parameters: prefix, uri, and prefixToRemove, the combination of which defines the operation performed. The read operations are fetching the prefix by URI, fetching the URI by prefix, and fetching all. The read-write operations are removing and updating prefix-URI pairs.

----

 - [x] Tests are included.

----


jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ActionPrefixesR.java
```diff
@@ -0,0 +1,156 @@
...
+                            FmtLog.info(action.log, "[%d] - %s", action.id, new JsonObject(entry.getKey(), entry.getValue()));
+                        });
+                action.setResponseContentType(WebContent.contentTypeJSON);
+                action.getResponseOutputStream().print(String.valueOf(allJsonPairs));
+
+                ServletOps.success(action);
+                action.endRead();
+                return;
+            }
+            if (prefix != null && uri == null) {
```
rvesse, line 89: Stylistic (and entirely optional): Rather than these if checks which aren't particularly clear why not have a `chooseResponseType()` (or similarly named method) that does those in a short method returning an `enum` value indicating the response type selected and then using a `switch` statement here for clarity?

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ActionPrefixesRW.java
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    @Override
+    protected void doDelete(HttpAction action) {
+        validateDelete(action);
+        action.beginWrite();
+
+        try {
+            String prefixToRemove = action.getRequestParameter("removeprefix");
```
rvesse, line 55: Since this is a `DELETE` request would it make more sense to just use the same `prefix` parameter as the read endpoints?

The use of HTTP verb already differentiates this request from a read request so using a different parameter name seems unnecessary

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ActionPrefixesR.java
```diff
@@ -0,0 +1,156 @@
...
+    protected void doOptions(HttpAction action) {
+        ActionLib.setCommonHeadersForOptions(action);
+        action.setResponseHeader(HttpNames.hAllow, "GET,OPTIONS");
+        ServletOps.success(action);
+    }
+
+    public void validateGet(HttpAction action) {
+        validate(action);
+        // check if the combination of parameters is legal
+        String prefix = action.getRequestParameter("prefix");
```
rvesse, line 51: Can we have all the parameter and JSON field names defined as constants somewhere please?

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ActionPrefixesR.java
```diff
@@ -0,0 +1,156 @@
...
+                action.setResponseContentType(WebContent.contentTypeJSON);
+                action.getResponseOutputStream().print(String.valueOf(allJsonPairs));
+
+                ServletOps.success(action);
+                action.endRead();
+                return;
+            }
+            if (prefix != null && uri == null) {
+
+                if (prefix.isEmpty()) {
```
rvesse, line 91: @afs Does it make sense to move some of these checks into `validate()` method or does it not make much difference from a Fuseki request lifecycle perspective?

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ActionPrefixesRW.java
```diff
@@ -0,0 +1,141 @@
...
+            action.end();
+        }
+    }
+
+
+    public void validatePost(HttpAction action) {
+        validate(action);
+        String prefix = action.getRequestParameter("prefix");
+        String uri = action.getRequestParameter("uri");
+        String prefixToRemove = action.getRequestParameter("removeprefix");
```
rvesse, line 84: So I guess this is why `removeprefix` is a separate parameter because we have `POST` allowing for both update and delete, might be cleaner to just limit `POST` to updating a prefix and keep `DELETE` for deleting a prefix

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/PrefixesService.java
```diff
@@ -0,0 +1,57 @@
...
+
+public class PrefixesService {
+    public static final Operation operationPrefixesRW;
+    public static final Operation operationPrefixesR;
+
+    private static final ActionService procPrefixR;
+    private static final ActionService procPrefixRW;
+
+    static {
+        operationPrefixesR = Operation.alloc("http://telicent.io/fuseki/operation/prefixes-r",
```
rvesse, line 42: Needs to change to be a Jena controlled namespace

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/prefixes/ActionProcPrefixes.java
```diff
@@ -0,0 +1,228 @@
...
+import org.apache.jena.sparql.core.Transactional;
+
+import java.io.IOException;
+import java.util.*;
+
+
+//validate and unpack the HTTP Get request,
+//call the storage's fetch URI by the prefix function from the HTTP request
+
+public class ActionProcPrefixes extends BaseActionREST {
```
rvesse, line 35: There seems to be two different implementations of the same thing - this and the `ActionPrefixesR`/`ActionPrefixedRW` combination - can some/all of the logic be shared between these to reduce the duplication?

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/prefixes/ActionProcPrefixes.java
```diff
@@ -0,0 +1,228 @@
...
+        validatePost(action);
+        Transactional transactional = storage.transactional();
+
+        transactional.begin(TxnType.WRITE);
+        try {
+            String prefix = action.getRequestParameter("prefix");
+            String uri = action.getRequestParameter("uri");
+
+            if(prefix.isEmpty()) {
+                String prefixToRemove = action.getRequestParameter("removeprefix");
```
rvesse, line 203: Again same comment about overloading `POST` to be both update and delete while also having a separate `DELETE` operation, would be cleaner to keep `POST` as update only

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/prefixes/PrefixesAccess.java
```diff
@@ -0,0 +1,42 @@
...
+    Optional<String> fetchURI(String prefix);
+
+    Transactional transactional();
+
+    void updatePrefix(String prefix, String uri);
+    void removePrefix(String prefixToRemove);
+
+    Map<String, String> getAll();
+
+    /** Fetches the prefixes assigned o the provided URI. There can be multiple in the JsonArray.**/
```
rvesse, line 40:
```suggestion
    /** Fetches the prefixes assigned to the provided URI. There can be multiple in the list**/
```

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/prefixes/PrefixesRDF.java
```diff
@@ -0,0 +1,220 @@
...
+ * ]
+ */
+
+public class PrefixesRDF implements PrefixesAccess {
+
+    DatasetGraph dataset = DatabaseMgr.createDatasetGraph();
+    public void setDataset() {
+
+        dataset.begin(ReadWrite.WRITE) ;
+        try {
```
rvesse, line 58: Is this method leftover debugging code?

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/prefixes/PrefixesRDF.java
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+    }
+
+    public Transactional transactional() { return dataset; }
+
+    @Override
+    public Optional<String> fetchURI(String prefix) {
+
+        String query = "PREFIX prefixes: <http://jena.apache.org/prefixes#> " +
```
rvesse, line 99: Could be better to use a parameterised string here i.e. https://jena.apache.org/documentation/query/parameterized-sparql-strings.html

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/prefixes/PrefixesRDF.java
```diff
@@ -0,0 +1,220 @@
...
+
+            AtomicBoolean result = new AtomicBoolean(false);
+            dataset.executeRead(()->{
+                result.set(qExec.ask());
+            });
+
+            String update;
+            if (result.get())
+                update =
+                        "PREFIX prefixes: <http://jena.apache.org/prefixes#>\n" +
```
rvesse, line 141: Stylistic: Maybe use multi-line strings to make the queries and updates more readable in the code?

jena-fuseki2/jena-fuseki-main/src/test/java/org/apache/jena/fuseki/main/prefixes/TestPrefixesService.java
```diff
@@ -0,0 +1,584 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.apache.jena.http.HttpOp.*;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class TestPrefixesService {
```
rvesse, line 41: This appears to be a copy paste from `TestAbstractPrefixParam` with the only difference being the Fuseki setup is configuration rather than programmatically driven.

Could it not extend that class and simply `@Override` the `before()` and `afterSuite()` methods?

jena-ontapi/src/main/java/org/apache/jena/ontapi/OntModelFactory.java
rvesse, line 1: Unrelated changes from a merge conflict resolution?

jena-fuseki2/examples/config-prefixes.ttl
```diff
@@ -0,0 +1,24 @@
+## Licensed under the terms of http://www.apache.org/licenses/LICENSE-2.0
+
+PREFIX example:  <http://example.org/fuseki/>
+PREFIX fuseki:  <http://jena.apache.org/fuseki#>
+PREFIX rdf:     <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+PREFIX :        <http://telicent.io/config#>
+PREFIX ja:      <http://jena.hpl.hp.com/2005/11/Assembler#>
+PREFIX tdb2:    <http://jena.apache.org/2016/tdb#>
+
+PREFIX ex: <http://example.org/fuseki/operation/>
```
rvesse, line 10: This namespace does not appear to match that actually used later in the code so unclear this example is actually usable?

rvesse (changes requested):
Thanks for the contribution, this is a useful new feature and something that has been requested in the past e.g. #1793 

Comments wise my feedback boils down to the following issues:

- There's quite a bit of code duplication (both implementation and test) that looks like it could be reduced in places
- Parameter/field names should use constants
- Having prefix deletion be possible both via `POST` and `DELETE` seems unnecessary overlap of functionality
- Choice of namespace URI for the new operations needs to be a Jena controlled URI

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/prefixes/ActionProcPrefixes.java
rvesse: This still seems to be duplicating behaviour from `ActionPrefixesRW` but is now inconsistent because this is still using `removeprefix` and still allows a `POST` to both edit/remove a prefix

rvesse: It feels like if we made the `prefixes()` method from `ActionPrefixesBase` non-static this class could simply extend that class and override that method to return the configured storage i.e.

```java
public class ActionProcPrefixes extends ActionPrefixesRW {

    private final PrefixesAccess storage;

    public ActionProcPrefixes(PrefixesAccess storage) {
        this.storage = storage;
    }

    @Override
    protected PrefixesAccess prefixes(HttpAction action) {
        return this.storage;
    }
}
```

This might need some fixes to `HttpAction` because `ActionPrefixesRW` assumes a valid transaction which a plain action processor doesn't get (see https://github.com/apache/jena/pull/2563)

rvesse (changes requested):

afs:
Rebase to current Jena main and squash commits down to one commit.

This includes fixed for the issues identified in #2563.

Further changes to come to address dedup test code, to remove `removeprefix` and - new feature - use content negotiation for get-uri-by-prefix to support both consistent JSON responses and also the simple lookup form of returning the URI as `text/plain`.

rvesse (approved):
LGTM

PR #2563, opened 2024-07-03
HttpAction transaction fixes (GH-2543)

rvesse:
In looking into whether the prefixes service PR #2538 could be refactored to reduce code duplication I noticed that if an Action is registered directly as servicing a URL, rather than being associated with a Dataset, then if the Action tries to use any of the transaction handling of HttpAction it can result in NPEs because some access to the transactional context was not checking for non-null transactional objects.  This commit addresses that problem and should allow #2538 to be refactored.

----

----


kinow (approved):
Looks good to me, thanks @rvesse !

afs:
There is work going on elsewhere in response to the comments on #2538 that covers this and includes another change needed to `HttpAction`. It gets tested because all the tests in #2538 then run refactored and properly.

rvesse:
Per discussion, closing in favour of larger refactoring and fixes ongoing in #2538

PR #2581, opened 2024-07-14
GH-2580: Added parseCheck flag to query and update exec builders.

### LINKED ISSUE #2580: RDFLink: API for passing raw query strings to backend
### Version

5.2.0-SNAPSHOT

### What happened?

I was trying to send SPARQL queries with vendor extensions via `RDFConnection.query(queryString)` and it turned out in certain places in Jena's infrastructure, attempts to parse the queries are made which then fails.

For example, `RDFLinkModular.query(String)` is backed by the default method form RDFLink which parses the query. In this case, the following override missing in order to delegate the query string as-is to the underlying query connection:

```java
    @Override
    public QueryExec query(String queryString) {
        return queryConnection().query(queryString);
    }
```

_No response_

Yes

Aklakan:
GitHub issue resolved #2580 

Pull request Description: Added `parseCheck` flag to query and update exec builders in order to give users control over how to deal with strings passed to `QueryExecBuilder.query(String)` and `UpdateExecBuilder.update(String)`.

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/http/sys/ExecUpdateHTTPBuilder.java
+++ b/jena-arq/src/main/java/org/apache/jena/http/sys/ExecUpdateHTTPBuilder.java
@@ -20,12 +20,11 @@ package org.apache.jena.http.sys;
 
 import java.net.http.HttpClient;
 import java.util.*;
+import java.util.stream.Collectors;
 
-import org.apache.jena.atlas.lib.InternalErrorException;
 import org.apache.jena.graph.Node;
 import org.apache.jena.http.HttpEnv;
 import org.apache.jena.query.ARQ;
-import org.apache.jena.query.QueryException;
 import org.apache.jena.sparql.core.Var;
 import org.apache.jena.sparql.engine.binding.Binding;
 import org.apache.jena.sparql.exec.http.Params;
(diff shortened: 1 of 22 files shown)


Aklakan:
I changed to draft because I am currently investigating whether there are more places where the query string is being parsed rather than passed one.

Aklakan:
Converted back to draft because more discussion needed about which solution to pursue - see #2580

Aklakan:
The PR now proposes a boolean `parseCheck` flag on `QueryExecBuilder` and `UpdateExecBuilder`. Builders created from remote connections use the connection's parseCheck flag as the initial value.

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkFuseki.java
```diff
@@ -69,7 +69,7 @@ private static RDFLinkHTTPBuilder setupForFuseki(RDFLinkHTTPBuilder builder) {
             .acceptHeaderSelectQuery(ResultSetLang.RS_Thrift.getHeaderString())
             .acceptHeaderAskQuery(ResultSetLang.RS_JSON.getHeaderString())
             .acceptHeaderQuery(acceptHeaderSPARQL)
-            .parseCheckSPARQL(false)
+            .parseCheckSPARQL(true)
```
Aklakan, line 72: With `parseCheckSPARQL(false)` there is now one failing test:

`TestRDFLinkFuseki.query_build_02` assumes that `QueryExecBuilder.query(String)` parses the query even though the link was created with `RDFLink.parseCheckSPARQL(false)`.

Perhaps it is the test that should be adapted.

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTPBuilder.java
```diff
@@ -91,6 +91,7 @@ protected RDFLinkHTTPBuilder(RDFLinkHTTP base) {
         parseCheckUpdates   = base.parseCheckUpdates;
 
         querySendMode       = base.querySendMode;
+        updateSendMode      = base.updateSendMode;
```
Aklakan, line 94: I noticed that this line was forgotten in https://github.com/apache/jena/pull/2253/files#diff-bee294b1706b3bd7768eb35b8cbcbb286197873cd3421a82602e4daf15f90dfaR93

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkFuseki.java
rvesse: +1, sometimes the tests make assumptions about how APIs work that are no longer true when we fix bugs in the API/implementation so the affected tests should be adapted accordingly

rvesse (approved):
LGTM

jena-rdfconnection/src/main/java/org/apache/jena/rdfconnection/SparqlQueryConnection.java
```diff
@@ -88,7 +88,14 @@ public interface SparqlQueryConnection extends Transactional, AutoCloseable
      */
     public QueryExecution query(Query query);
 
-    /** Setup a SPARQL query execution.
+    /** Setup a SPARQL query execution from a query string.
+     *
+     *  The string is subject to immediate client-side parsing
+     *  so that (parse) errors can be detected right away.
+     *
+     *  For query strings that are not parseable by Jena, e.g. because of SPARQL extensions,
```
Aklakan, line 96: Ah, I still need to update the comment.

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkFuseki.java
Aklakan: Thinking about this issue, the code in question is:
```java
QueryExec qExec = link.newQuery().query("SELECT ?X { }")
    .substitution(binding)
    .build();
```

The problem is that the link has checkParse disabled, but substitution requires a query object.
I think it would be best if methods that require a object would attempt to parse a given query string when needed.

The alternative would be having to explicitly write `remoteLink.newQuery().checkParse(true).query("SELECT ?X { }")` which doesn't feel natural.

rvesse: Yeah, so maybe adjust the code and update the Javadoc on the `substitution` method to note that this requires parsing the query which means it can't be used with SPARQL extensions Jena doesn't understand

Aklakan:
There are mory nitty-gritty details:

The update methods in the UpdateExecBuilder are cumulative: Calling update multiple times with different strings actually appends the parsed update objects to the request.

https://www.w3.org/TR/sparql11-update/#updateLanguage states:

| A request is a sequence of operations and is terminated by EOF (End of File). Multiple operations are separated by a ';' (semicolon) character. 

So without parsing, it should be fine to just join the given strings on `;`. Still a bit of a change nonetheless.

Aklakan:
I updated `ExecUpdateHTTPBuilder` with additional internals: There is now an `UpdateElt` record for holding a single update object or a string, and there is an accumulator `UpdateEltAcc` class. The accumulator tracks whether all provided elements were objects, and supports building an overall UpdateRequest object or string from its state.

Aklakan:
I added some additional tests with multiple `update()` calls on the builder. I'd say I am done with the revision.

jena-arq/src/main/java/org/apache/jena/http/sys/ExecUpdateHTTPBuilder.java
```diff
@@ -200,28 +296,35 @@ private void ensureContext() {
...
-    }
-
     public X build() {
         Objects.requireNonNull(serviceURL, "No service URL");
-        if ( updateOperations == null && updateString == null )
-            throw new QueryException("No update for UpdateExecutionHTTP");
-        if ( updateOperations != null && updateString != null )
-            throw new InternalErrorException("UpdateRequest and update string");
+        if ( updateEltAcc.isEmpty() )
+            throw new UpdateException("No update for UpdateExecutionHTTP");
```
Aklakan, line 302: The original code used QueryException and I thought that it might have been a copy/paste issue. Then again, QueryParseException is also re-used for parse errors with update requests, so I am unsure what is the right choice her.

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkFuseki.java
Aklakan: I changed the Query/Update exec builders so that the build() method will attempt to parse any remaining strings if needed.

Issue #2665, opened 2024-08-28
Rename XMLLiteralType.theXMLLiteralType as XMLLiteralType.rdfXMLLiteral


afs:
### Change

More conventional naming.

The preferred constant for application use is still `RDF.dtXMLLiteral`.

### Are you interested in contributing a pull request for this task?

Yes

Issue #2664, opened 2024-08-28
Consolidate NullIterator and SingletonIterator around Iter functions


afs:
### Version

5.1.0

### Feature

Use Iter functions for creating null iterators and singleton iterators.

### Are you interested in contributing a solution yourself?

Yes

PR #2666, opened 2024-08-28
Misc

afs:
Resolved: #2664 
Resolved: #2665

A PR of small clean-ups and internal tidy-ups.

The number of files touched is to due to propagating renaming. Deprecated originals left in-place.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/system/G.java
+++ b/jena-arq/src/main/java/org/apache/jena/system/G.java
@@ -1047,11 +1047,12 @@ public class G {
         // Both nodes with language tags : compare languages case insensitively.
         String lex1 = match.getLiteralLexicalForm();
         String lex2 = data.getLiteralLexicalForm();
+        if ( ! lex1.equals(lex2) )
+            return false;
         String lang1 = match.getLiteralLanguage();
         String lang2 = data.getLiteralLanguage();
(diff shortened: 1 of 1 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterProjectMerge.java
```diff
@@ -37,68 +37,69 @@
...
-            //bind = new BindingProject(projectionVars, bind) ;
-            // return BindingUtils.merge(outerBinding, bind) ;
+            //bind = new BindingProject(projectionVars, bind);
+            // return BindingUtils.merge(outerBinding, bind);
         }
 
-        /** Merge two bindings, the outer and inner, projecting the inner with the give variables.
-         *  This is what happens in substitutIon execution, with a inner select-project.
+        /**
+         * Merge two bindings, the outer and inner, projecting the inner with the give variables.
```
kinow, line 79: s/give/given but it's not in this PR, so we can fix it later...

kinow (approved):

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterProjectMerge.java
afs: Done!

PR #2702, opened 2024-09-09
GH-2700: Configurable server content negotiation

### LINKED ISSUE #2700: Allow custom serialization formats in Fuseki
### Version

dev (d9bf473)

### Feature

I wrote a custom serialization format for RDF, along with the writer, parser, and all the registration shenanigans ([source in Scala](https://github.com/Jelly-RDF/jelly-jvm/blob/effb9ead5fb34a4ca0feee5c0147e093ba7d70d3/jena/src/main/scala/eu/ostrzyciel/jelly/convert/jena/riot/JellyLanguage.scala#L61)). It works great when loading RDF files into Fuseki UI, or when using the RIOT CLI utility. The problem is, I can't get it to work over content negotiation in Fuseki with SPARQL CONSTRUCT queries or with the graph store protocol. Fuseki always ignores my Accept header with the custom content type and returns Turtle anyway. I dug a bit into the code and found this:

- [`ResponseDataset.doResponseDataset`](https://github.com/apache/jena/blob/d9bf473405b4242215c626126d01f9463a27ccb4/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/ResponseDataset.java#L82) uses `DEF.constructOffer` as the list of accepted content types.
- [`DEF.constructOffer`](https://github.com/apache/jena/blob/d9bf473405b4242215c626126d01f9463a27ccb4/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/DEF.java#L35) is a public static final field, so it cannot be modified or extended in any way.

The `DEF` class also contains a few other lists of content types that serve similar purposes – they are also static final.

So, as I understand, there is currently no way of adding a new serialization format for Fuseki responses, besides forking the entire codebase or resorting to something like reflection (nooooo...).

Assuming this is correct, would it make sense to create some kind of a simple registration mechanism for supported content types in Fuseki, similar to [`RDFLanguages`](https://github.com/apache/jena/blob/d9bf473405b4242215c626126d01f9463a27ccb4/jena-arq/src/main/java/org/apache/jena/riot/RDFLanguages.java#L265) in RIOT?

Yes

afs:
GitHub issue resolved #2700

Pull request Description:

Allow the Fuskei server content negotiation choices to be altered on server start-up.

----

### DIFF
--- a/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/DEF.java
+++ b/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/DEF.java
@@ -22,74 +22,110 @@ import org.apache.jena.atlas.web.AcceptList;
 import org.apache.jena.atlas.web.MediaType;
 import static org.apache.jena.riot.WebContent.*;
 
+/**
+ * Content negotiation setup.
+ */
 public class DEF
 {
-    public static final MediaType acceptRDFXML        = MediaType.create(contentTypeRDFXML);
-    public static final MediaType acceptNQuads        = MediaType.create(contentTypeNQuads);
-    public static final MediaType acceptRSXML         = MediaType.create(contentTypeResultsXML);
-    public static final MediaType acceptJSON          = MediaType.create(contentTypeJSON);
-    public static final MediaType acceptTurtle        = MediaType.create(contentTypeTurtle);
+    // @formatter:off
 
-    public static final AcceptList jsonOffer          = AcceptList.create(contentTypeJSON);
+    // ---- Server configuration default media type
 
-    public static final AcceptList constructOffer     = AcceptList.create(contentTypeTurtle,
-                                                                          contentTypeTurtleAlt1,
-                                                                          contentTypeNTriples,
-                                                                          contentTypeNTriplesAlt,
-                                                                          contentTypeRDFXML,
-                                                                          contentTypeTriX,
-                                                                          contentTypeTriXxml,
-                                                                          contentTypeJSONLD,
-                                                                          contentTypeRDFJSON,
-                                                                          contentTypeRDFThrift,
+    public static MediaType acceptRDFXML        = acceptRDFXMLDefault();
+    public static MediaType acceptNQuads        = acceptNQuadsDefault();
+    public static MediaType acceptResultSetXML  = acceptResultSetXMLDefault();
+    public static MediaType acceptJSON          = acceptJSONDefault();
+    public static MediaType acceptTurtle        = acceptTurtleDefault();
 
-                                                                          contentTypeTriG,
-                                                                          contentTypeTriGAlt1,
-                                                                          contentTypeNQuads,
-                                                                          contentTypeNQuadsAlt1
-                                                                          );
+    // ---- Server configuration offers for content negotiation
 
-    public static final AcceptList rdfOffer           = AcceptList.create(contentTypeTurtle,
-                                                                          contentTypeTurtleAlt1,
-                                                                          contentTypeNTriples,
-                                                                          contentTypeNTriplesAlt,
-                                                                          contentTypeRDFXML,
-                                                                          contentTypeTriX,
-                                                                          contentTypeTriXxml,
-                                                                          contentTypeJSONLD,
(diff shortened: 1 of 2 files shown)


kinow (approved):

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/DEF.java
```diff
@@ -22,74 +22,110 @@
...
+    public static AcceptList rdfOffer           = rdfOfferDefault();
+
+    public static AcceptList quadsOffer         = quadsOfferDefault();
+
+    // Offer for SELECT
+    public static AcceptList rsOfferTable       = rsOfferTableDefault();
+
+    // Offer for ASK
+    public static AcceptList rsOfferBoolean     = rsOfferBooleanDefault();
+
```
rvesse, line 55: So this basically makes all these fields mutable so they can be modified within to the server operators preferred configuration?

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/DEF.java
```diff
@@ -22,74 +22,110 @@
...
+    public static final MediaType acceptNQuadsDefault()       { return MediaType.create(contentTypeNQuads); }
+    public static final MediaType acceptResultSetXMLDefault() { return MediaType.create(contentTypeResultsXML); }
+    public static final MediaType acceptJSONDefault()         { return MediaType.create(contentTypeJSON); }
+    public static final MediaType acceptTurtleDefault()       { return MediaType.create(contentTypeTurtle); }
+
+    // ---- Default offers for content negotiation.
+
+    public static final AcceptList jsonOfferDefault()         { return AcceptList.create(contentTypeJSON); }
+
+    public static final AcceptList constructOfferDefault()    { return AcceptList.create(contentTypeTurtle,
```
rvesse, line 69: Could the defaults be generated based on `RDFLanguages` or other suitable registry rather than hardcoding?

Then the OPs scenario in #2700 may just work without them needing any further code to modify the static fields

Obviously initialisation and registration order is going to be a thing so may need both approaches regardless, WDYT?

afs: > Could the defaults be generated based on `RDFLanguages` or other suitable registry rather than hardcoding?

1. The order in `Accept` and `offer` lists matters.
1. Maybe a user wants a server to offer a subset of what Jena can support.
1. `RDFLanguages` is a registry of constants. It does not guarantee a writer exists nor does it cover result sets.
1. Each offer type is context dependent. If it's a SELECT query, then `rsOfferTable` is used. If the result is a CONSTRUCT, `constructOffer`, if it is GSP, then it depends on graph vs dataset. It's not a choice. A registry would be an enum to offer with fixed choice of enum key. That's the same as asking for the right constant.

jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/DEF.java
afs: Yes. Jena has init mechanisms for external code.

Ostrzyciel:
This looks good to me, solves my issue entirely.

The only (probably insignificant) downside I see is that if hypothetically had two Jena extensions that both wanted to modify these fields, they could override each other... but this is a rare case I guess. You can always get around this by appending to the current list instead of overriding it.

+1

Ostrzyciel (approved):

afs:
> if hypothetically we had two Jena extensions that both wanted to modify these fields

:smile:

Best I can think of is to check the current value of a slot against the initial system setting. That detects possible conflict.
Testing object identity is good enough. 

Automatically resolving is problematic in the general case with adds and removes of the offer lists because the two subsystems may be have conflicting requirements. Additions could be merged reasonably safely.

The Jena initialization mechanisms (Jena System init, Fuseki modules) are both sequential so no concurrent overwrite problems.
They can be ordered by level if needed.

Issue #2739, opened 2024-09-27
how to debug LOADING_REMOTE_CONTEXT_FAILED problems?


VladimirAlexiev:
### Version

5.1.0

### Question

https://github.com/EKGF/dprod/issues/90:

Trying to convert https://github.com/EKGF/dprod/blob/main/examples/data-quality/example.json with riot causes error:
```
riot.bat -out=trig -syntax=jsonld example.json
There was a problem encountered loading a remote context [code=LOADING_REMOTE_CONTEXT_FAILED].
```
The context is https://ekgf.github.io/dprod/dprod.jsonld and:
- it's served as `content-type: application/ld+json` (reading through the JSONLD spec, I can't see whether it should be `json` or `ld+json` but all examples use `.jsonld` for the context URL)
- the certificate is valid
- example.json works ok on the JSONLD Playground: https://json-ld.org/playground/

So what may be the reason for the problem?
- https://stackoverflow.com/questions/41494678/how-can-i-override-the-context-when-loading-json-ld-in-jena correctly points out that the error is generic and may encompass a whole gamut of problems
- https://github.com/apache/jena/issues/1765 shows an extra reason, eg `message=Unexpected response code [403]` but that reason is not printed by `riot`
- https://github.com/apache/jena/commit/44b5535dc47b0159a74ff93e4d41e71e3c69c4ec says `removed the printStackTrace when "loading remote context failed"` but that's just in a test

afs:
Getting `example.json` returns an HTML document; the github display of the file.
Use the "raw", the processing is all inside Titanium JSON-LD.

Internally, there is a nested stacktrace that says:

```
    A keyword [@context] redefinition has been detected.
```

My guess is that it is because of the array but I'm not a JSON-LD expert.

VladimirAlexiev:
Thanks @afs !

The reason is context nested inside context (actually dprod.jsonld is not a context at all, it's a dumped ontology).
So the Playground is more lenient than Titanium (thus Riot). But really dprod.jsonld is to blame.

Thank for the PR you made!
I'll close this specific issue.

afs:
Looks like https://github.com/filip26/titanium-json-ld/issues/348.

PR #2742, opened 2024-09-28
Unwrap Titanium exceptions

afs:
While investigating #2739, I found that errors from Titanium can be wrapped exceptions with where the cause is also a Titanium exception and more meaningful.

This PR adds unwrapping the cause in that case.

----


rvesse (approved):

kinow (approved):

afs:
Thanks.

Could we stick to "rebase and merge" please? It keeps the log cleaner for the long term.

arne-bdt:
Sorry did not know.

afs:
NP.
I think the GH UI conveniently remembers the user's choice per repo.

PR #2752, opened 2024-10-02
GH-2741: Initialize LiteralLabel dtype correctly

### LINKED ISSUE #2741: Possible bug with normalize of dtype in LiteralLabel
### Version

5.2.0-SNAPSHOT

### What happened?

By accident, I stumbled across two calls in  `org.apache.jena.graph.impl.LiteralLabel` to the static function `normalize` , which set a local variable `dtype`, which is not used afterwards.
But there is a class variable `this.dtype` which maybe should have been set to the result of `normalize` instead, otherwise normalize would have no effect at all.
image

I tried to comment the two lines out but as expected the following tests were still sucessful:
- `org.apache.jena.graph.test.TestTypedLiterals`
- `org.apache.jena.graph.test.TestLiteralLabels`
- `org.apache.jena.graph.test.TestLiteralLabelSameValueAs`

Since I can´t find any example, showing what normalize should do, I don´t feel confident enough to add tests and provide an appropriate fix.

_No response_

Maybe

afs:
GitHub issue resolved #2741

Rename constructor argument to `datatype`.

----

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/graph/impl/LiteralLabel.java
+++ b/jena-core/src/main/java/org/apache/jena/graph/impl/LiteralLabel.java
@@ -51,31 +51,32 @@ final public class LiteralLabel {
     private String lexicalForm;
 
     /**
-     * The language tag. The empty string is not valid; {@code ""} is used to ensure this field is set.
+     * The language tag. The empty string is not a valid language tag;
+     * {@code ""} is used to ensure this field is set.
      */
-    private String lang;
+    private final String lang;
 
     /**
      * The initial text direction of a language literal.
      * Datatype rdf:dirLangSring.
      */
-    private TextDirection textDir;
+    private final TextDirection textDir;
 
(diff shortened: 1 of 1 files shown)


kinow (commented):
Change looks OK to me. Does it need a unit test too?

afs:
I don't have an example where the old situation shows a problem.

In #2741, commenting them out failed to narrow the datatype and did show up in the tests. 

So really we're relying on that fact that `LiteralLabel` is extensively tested already (and also via the Model API which depends on value processing).

kinow (approved):

arne-bdt (approved):
I tried to write a test but found no `String lex` where the normalized datatype differs from the given one. 
In all relevant cases, where there is a possible change in the datatype, the given datatype is used to parse `lex` so and there does not seem to be a string hat is accepted for parsing, where the type later can be narrowed.

PR #2754, opened 2024-10-02
GH-2753: Improve customising Fuseki args

### LINKED ISSUE #2753: Improve ability to support custom arguments for Fuseki Servers
### Version

5.1.0

### Feature

Yes

rvesse:
Improves how Fuseki arguments are customised, in particular makes the `ServerConfig` class public adding the ability to store extra configuration on it.

It also changes the previously `private static applyServerArgs()` method into a protected instance method so users can extend `FusekiMain` and override it to use that to customise the `FusekiServer.Builder` further.

Unit tests are added that validate that custom arguments can now be passed through CLI inputs and used to influence the Server being built.

GitHub issue resolved #2753

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-fuseki2/jena-fuseki-main/src/test/java/org/apache/jena/fuseki/main/TestFusekiMainCmdCustomArguments.java
+++ b/jena-fuseki2/jena-fuseki-main/src/test/java/org/apache/jena/fuseki/main/TestFusekiMainCmdCustomArguments.java
@@ -1,0 +1,94 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.jena.fuseki.main;
+
+import org.apache.jena.atlas.logging.LogCtl;
+import org.apache.jena.cmd.CmdException;
+import org.apache.jena.fuseki.Fuseki;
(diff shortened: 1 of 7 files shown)


afs:
General comment - when (if!) there is a switch to Fuseki Main + UI + Admin for the standalone server (`jena-fuseki-fulljar` and the jars file in the download), maybe there should be an extension point for this. 

This is not for this PR - but if it comes along, this would be fitted into a formal lifecycle for the command line.

Do you think adding to a new interface `FusekiModule` would fit your customization start flow? (`FusekiModule` collects several interfaces already)

jena-fuseki2/jena-fuseki-main/src/test/java/org/apache/jena/fuseki/main/TestFusekiMainCmdCustomArguments.java
```diff
@@ -0,0 +1,94 @@
...
+        try {
+            buildServer(buildCmdLineArguments(arguments));
+        } catch (Exception e) {
+            actual = e;
+        }
+        // then
+        assertNotNull(actual);
+        assertTrue("Expecting correct exception", (actual instanceof CmdException));
+        assertEquals("Expecting correct message", expectedMessage, actual.getMessage());
+    }
```
afs, line 80:
```suggestion
    private void testForCmdException(List<String> arguments, String expectedMessage) {
        CmdException ex = assertThrows(CmdException.class, ()->buildServer(buildCmdLineArguments(arguments)));
        assertEquals("Expecting correct message", expectedMessage, ex.getMessage());
    }
```
mixing Junit 4 and junit5.
Not that this function is actually called from anywhere!

jena-fuseki2/jena-fuseki-main/src/test/java/org/apache/jena/fuseki/main/TestFusekiMainCmdCustomArguments.java
```diff
@@ -0,0 +1,94 @@
...
+            LogCtl.setLevel(Fuseki.serverLog, level);
+    }
+
+    private FusekiServer server = null;
+    @After public void after() {
+        if ( server != null )
+            server.stop();
+    }
+
+    @Test
```
afs, line 56: This seems to be the only test and the rest of the code is taken from `TestFusekiCmdArguments`.
DRY?

afs:
Release 5.2.0 is approaching.

I have a discussion point https://github.com/apache/jena/issues/2753#issuecomment-2397485241 which is about whether this feature now will fit into future possible changes and some code comments here.

Would you prefer to wait or get into 5.2.0?

rvesse:
See larger discussion response on #2753, let's leave this out for now and look at other options as we can workaround this for now

Issue #2755, opened 2024-10-03
Fuseki 5.1.0 NullPointerException ("org.apache.jena.sparql.core.Transactional.transactionMode()" is null)


SamuelBoerlin:
### Version

5.1.0

### What happened?

We recently updated Fuseki from 5.0.0 to 5.1.0 and started experiencing a NullPointerException.
Once this this exception has occurred once it keeps happening for all subsequent queries until Fuseki is restarted.
While in this state Fuseki still successfully responds to `/$/ping`, but queries fail.

So far this has only happened on servers where we regularly do compaction, so maybe that's related somehow?

We also persistently ran into this exception while trying to do lots of update queries (i.e. several thousands) with intermittent compaction inbetween to keep the DB size down.

There aren't any other errors/warnings before the exception (except for that one `end called - Not in an action` warning), nor during the compaction.

We're running Fuseki 5.1.0 in a docker container based on the eclipse-temurin:21-jre-jammy image.

Configuration:

```
@prefix :           <http://base/#> .
@prefix fuseki:     <http://jena.apache.org/fuseki#> .
@prefix ja:         <http://jena.hpl.hp.com/2005/11/Assembler#> .
@prefix knora-base: <http://www.knora.org/ontology/knora-base#> .
@prefix rdf:        <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs:       <http://www.w3.org/2000/01/rdf-schema#> .
@prefix root:       <http://127.0.0.1:8888/$/datasets#> .
@prefix tdb2:       <http://jena.apache.org/2016/tdb#> .
@prefix text:       <http://jena.apache.org/text#> .

<http://jena.apache.org/rdf-patch#LoggedDataset>
        rdfs:subClassOf  ja:RDFDataset .

[ rdf:type         fuseki:Server ;
  fuseki:services  :service_tdb_all ;
  ja:loadClass     "org.apache.jena.query.text.TextQuery"
] .

ja:ModelRDFS  rdfs:subClassOf  ja:Model .

ja:RDFDatasetSink  rdfs:subClassOf  ja:RDFDataset .

<http://jena.hpl.hp.com/2008/tdb#DatasetTDB>
        rdfs:subClassOf  ja:RDFDataset .

tdb2:GraphTDB2  rdfs:subClassOf  ja:Model .

:indexLucene  rdf:type  text:TextIndexLucene ;
        text:analyzer   [ rdf:type        text:ConfigurableAnalyzer ;
[...]

PR #2760, opened 2024-10-08
GH-2755: Protect action cleanup

afs:
#2755  shows a stacktrace that happens after problem has been detected and logged. The trace does not add any information and makes the log confusing.

----

### DIFF
--- a/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/HttpAction.java
+++ b/jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/HttpAction.java
@@ -35,7 +35,6 @@ import java.util.zip.GZIPInputStream;
 import jakarta.servlet.ServletOutputStream;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
-
 import org.apache.jena.atlas.io.IO;
 import org.apache.jena.atlas.lib.NotImplemented;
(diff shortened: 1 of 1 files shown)


jena-fuseki2/jena-fuseki-core/src/main/java/org/apache/jena/fuseki/servlets/HttpAction.java
```diff
@@ -329,23 +329,31 @@ public void begin(TxnType txnType) {
...
+            return;
+        switch(rwMode) {
+            case READ -> {}
+            case WRITE -> {
+                // Write transactions must have explicitly called commit or abort.
+                FmtLog.warn(log, "[%d] Transaction still active - no commit or abort seen (forced abort)", this.id);
+                try {
+                    transactional.abort();
+                } catch (RuntimeException ex) {
+                    FmtLog.warn(log, "[%d] Exception in forced abort (trying to continue)", this.id, ex);
```
rvesse, line 351: Doesn't this still put the stack trace in?

afs: Not the one from #2755 - the except is `Cannot invoke "org.apache.jena.query.ReadWrite.ordinal()`
which is a null value of `transactional.transactionMode()` in`finishActionTxn()`.

```
        switch(transactional.transactionMode() )
```

and that says very little.

The `transactional.abort` exception may have some information (unlikely but possible) and would be hidden by the `switch` NPE.

PR #2773, opened 2024-10-14
GH-2753: Fuseki CLI arg customisation refactor

### LINKED ISSUE #2753: Improve ability to support custom arguments for Fuseki Servers

rvesse:
Refactors how the FusekiMain CLI can be customised to make it more flexible and interface driven.  CLI extensions have to be statically registered prior to invoking FusekiMain but are now able to both customise the arguments available and to directly affect the FusekiServer.Builder without needing to sub-class FusekiMain

Currently some tests are broken so still a Work in Progress!

GitHub issue resolved #2753

Pull request Description:

----

 - [ ] Tests are included.

----


jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/cmds/FusekiMain.java
```diff
@@ -299,32 +362,38 @@ protected void processModulesAndArgs() {
...
+            String jettyConfigFile = getValue(argJettyConfig);
+            if ( !FileOps.exists(jettyConfigFile)) {
+                throw new CmdException("Jetty config file not found: " + jettyConfigFile);
+            }
+            builder.jettyServerConfig(jettyConfigFile);
+        }
 
         // ---- Port
-        serverConfig.port = defaultPort;
+        builder.port(defaultPort);
```
rvesse, line 385: - [ ] Don't set this default if a Jetty config file present

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/sys/FusekiModule.java
```diff
@@ -56,7 +57,7 @@
  *     Modules must not rely on a call to {@code serverStopped} happening.</li>
  * </ul>
  */
-public interface FusekiModule extends FusekiBuildCycle, FusekiStartStop, FusekiActionCycle {
+public interface FusekiModule extends FusekiCliCustomiser, FusekiBuildCycle, FusekiStartStop, FusekiActionCycle {
```
rvesse, line 60: - [ ] Remove this, doesn't make sense given the CLI lifecycle

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/cmds/DSGSetup.java
afs, line 1: Maybe keep but return a record of three slots (DSG, label, update flag) rather than pass in ServerConfig.

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/sys/FusekiCliCustomiser.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public interface FusekiCliCustomiser {
+
+    /**
+     * Called prior to any CLI argument parsing starting, allows a Fuseki module to add custom arguments via
+     * {@link FusekiMain#addArg(String, String)} and its related overloads
+     *
+     * @param fuseki Fuseki Main
+     */
+    public default void customiseCli(FusekiMain fuseki) {
```
afs, line 35: Better to return a list of `ArgDecl`s (`ArgModuleGeneral`) to add? It reduces the exposed surface of `FusekiMain`.

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/sys/FusekiCliCustomiser.java
```diff
@@ -0,0 +1,52 @@
...
+     * <p>
+     * This happens prior to pretty much any other argument processing so can be used to do all sorts of customisation
+     * of the builder.  Note that Fuseki's default argument processing still applies so implementors need to be aware
+     * that default argument processing could overwrite some configuration.
+     * </p>
+     *
+     * @param fuseki  Fuseki Main
+     * @param builder Fuseki Builder
+     */
+    public default void processCliArgs(FusekiMain fuseki, FusekiServer.Builder builder) {
```
afs, line 50: Is the argument checking going to be done by the cusomization? (e.g. if arg1 , then arg2 is not allowed)

Someway to pass in the args that have been set by the earlier call?

afs: If the server is "everythign" then one case is a customizer to influence themoduels actually run. i.e there might be args that are a "switch off" e.g.`--nogeosparql` -- an arg to not run a GeoSPARQL module.

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/cmds/DSGSetup.java
rvesse: That code was only being called from FusekiMain so made more sense (to me at least) to inline it

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/sys/FusekiCliCustomiser.java
rvesse: > Is the argument checking going to be done by the cusomization? (e.g. if arg1 , then arg2 is not allowed)
> 
> Someway to pass in the args that have been set by the earlier call?

That was why I was passing `FusekiMain` in as you can call the various `contains()` and `getValue()` methods inherited from the rest of the Jena CLI machinery

rvesse: > If the server is "everythign" then one case is a customizer to influence themoduels actually run. i.e there might be args that are a "switch off" e.g.`--nogeosparql` -- an arg to not run a GeoSPARQL module.

Yeah I need to play around more with how modules interact and are configured here

afs: A simplifying assumption is that a set `FusekiCliCustomiser` and a set of modules.

`builder.addCliCustomisers` must be before the builder calls command line processing.

There is no need to add `FusekiCliCustomiser` to `FusekiModule` for all modules.
A module can still also have `FusekiCliCustomiser`.

afs: > That was why I was passing `FusekiMain` in as you can call the various `contains()` and `getValue()` methods inherited from the rest of the Jena CLI machinery

That seems a good argument to changing `FusekiMain` so it does not extend `CmdARQ`, and instead be a `FusekiMain` with a command line object as a member.

I'd find it helpful to have a written description of the the flow is expected to be.

If we can make `FusekiMain` smaller with some clearly defined functionality in separate classes, then when there are several modules/customization they'll get what they are allowed to modifiy.

PR #2796, opened 2024-10-26
GH-2795: Fix race condition in datatype registration

### LINKED ISSUE #2795: Race condition in TypeMapper.getSafeTypeByName
### Version

5.2.0

### What happened?

I've been chasing down the reasons for a flaky test in (you guessed it) a scalatest suite. The test was failing randomly and seemingly for no reason on isomorphism checks of two graphs that should be identical.

After some debugging, I've discovered the following:

- The root cause is comparing literals with non-standard datatypes (i.e., datatypes not preloaded on Jena startup).
- The literal comparison returns false for identical literals because the datatype comparison returns false: https://github.com/apache/jena/blob/04f377d8aefc0cb8faeb62022d402ef17a6e6e96/jena-core/src/main/java/org/apache/jena/graph/impl/LiteralLabel.java#L477-L479
- `LiteralLabel` assumes that the datatypes can be compared by reference, which makes sense, because Jena maintains a global hashmap of all used datatypes in `TypeMapper`.
- From the debugger I got the info that somehow there were two instances of `BaseDatatype` with the same IRI on the heap (`http://www.opengis.net/ont/geosparql#wktLiteral`): 
image
- Only one of these instances was registered in `TypeMapper.uriToDT`. I am constructing new datatypes only via the `NodeFactory`, which ALWAYS goes through the `TypeMapper`, so the other instance must have been overridden at some point. There is no other way it could have ended up on the heap.

I'm 99% sure that the issue is this part:

https://github.com/apache/jena/blob/04f377d8aefc0cb8faeb62022d402ef17a6e6e96/jena-core/src/main/java/org/apache/jena/datatypes/TypeMapper.java#L124-L140

If the datatype is not in the concurrent map, this is called:

https://github.com/apache/jena/blob/04f377d8aefc0cb8faeb62022d402ef17a6e6e96/jena-core/src/main/java/org/apache/jena/datatypes/TypeMapper.java#L187-L193

 The call to `uriToDT.put()` will override the previous value in the map, if it exists. This is absolutely possible, because between the call to get() and put() we have no synchronization, so another thread may add the same datatype to the map in the meantime.

I don't have a reproducible example, but I think this is a pretty clear case of a race condition... also relatively easy to fix, I think. I'll try doing that.

_No response_

Yes

Ostrzyciel:
GitHub issue resolved #2795

Pull request Description:

Fixed a race condition in user-defined datatype registration. This change only touches the user-friendly `getSafeTypeByName` and does not change anything from the user's perspective (other than fixing the race condition). The `registerDatatype` method is left as-is, to allow for overriding datatype registrations if someone really wants to.

I've added some notes to `registerDatatype` and `unregisterDatatype` that it they have good "shoot-yourself-in-the-foot-with-a-bazooka" potential.

----

 - [ ] Tests are included – *not sure how to test it... do we need tests here?*

----

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/datatypes/TypeMapper.java
+++ b/jena-core/src/main/java/org/apache/jena/datatypes/TypeMapper.java
@@ -126,18 +126,16 @@ public class TypeMapper {
             // Plain literal
             return null;
         }
-        RDFDatatype dtype = uriToDT.get(uri);
-        if (dtype == null) {
+        return uriToDT.computeIfAbsent(uri, u -> {
             // Unknown datatype
             if (JenaParameters.enableSilentAcceptanceOfUnknownDatatypes) {
-                dtype = new BaseDatatype(uri);
-                registerDatatype(dtype);
+                // No need to update classToDT because BaseDatatype.getJavaClass is always null
+                return new BaseDatatype(u);
             } else {
                 throw new DatatypeFormatException(
                     "Attempted to created typed literal using an unknown datatype - " + uri);
             }
-        }
-        return dtype;
+        });
     }
 
     /**
@@ -183,6 +181,9 @@ public class TypeMapper {
 
     /**
      * Register a new datatype
+     * This will overwrite any existing registration for this datatype IRI.
+     * Comparisons of literals with different datatype instances will fail, so be careful
+     * if you are using this outside of initialization code.
      */
     public void registerDatatype(final RDFDatatype type) {
         uriToDT.put(type.getURI(), type);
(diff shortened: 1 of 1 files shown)


arne-bdt:
I don´t think, this is the way to solve the issue.
--> discussion in https://github.com/apache/jena/issues/2795#issuecomment-2439653100

Ostrzyciel:
Updated the PR with a different approach as per the discussion in the issue #2795

afs (approved):

arne-bdt (approved):
@Ostrzyciel Thank you for that great solution and the additional helpful code comments.

PR #2798, opened 2024-10-26
GH-2797: Lazy init for ValidationState hash maps

### LINKED ISSUE #2797: Unneeded HashMap allocations in ValidationState
### Change

While profiling a piece of RDF parser code, I've noticed a whole lot of `HashMap` allocations on the heap – not HashMap entries – HashMaps. What could possibly be allocating tons of maps during RDF parsing?

It turns out to be the `ValidationState` class, an instance of which is allocated for each time we parse a datatype literal. It has these two fields: https://github.com/apache/jena/blob/04f377d8aefc0cb8faeb62022d402ef17a6e6e96/jena-core/src/main/java/org/apache/jena/ext/xerces/impl/validation/ValidationState.java#L53-L55

I find the note slightly funny.

The whole ID table tables thing is used only by `IDDV` and `IDREFDV` classes. I had to consult the RDF spec, because I didn't even know that such datatypes existed. It this XML thing: https://www.w3.org/TR/xmlschema11-2/#ID I think it's a very rarely used feature of RDF that certainly doesn't warrant allocating two hashmaps per literal.

Also: will the ID and IDREF validation even work if we allocate a new object for each literal parse? Then these hashmaps are not carried over... Did this even work?

## Solution

I'm submitting a PR that sets these hashmap fields to null by default, and only initializes them if needed (lazy initialization). Note that I'm not convinced that the ID/IDREF feature even works currently, but removing this outright would be a breaking change.

## Broader issue – for a separate thread

*(I will create a new issue with this, I'm leaving this here for context for now)*

Another thing is that I'm not convinced that we need to allocate a new `ValidationState` every time we parse a datatyped literal: 

https://github.com/apache/jena/blob/04f377d8aefc0cb8faeb62022d402ef17a6e6e96/jena-core/src/main/java/org/apache/jena/datatypes/xsd/XSDDatatype.java#L267-L272

This is a pretty heavy object to be allocating so often.

Many methods in ValidationState are unused, and there is a bunch of other dead code around it copied over from xerces, like the whole `ValidationManager` class. Removing this, however, would be a breaking change, so it's not in the scope of this issue.

Yes

Ostrzyciel:
GitHub issue resolved #2797

Pull request Description:

This is a conservative solution to #2797 – it does not change any behavior from the outside perspective, and only changes the initialization of these two hashmaps to be lazy.

This should improve RDF parsing performance.

I'm marking "tests are included" because I don't see anything to test here that is not already covered by existing tests.

----

 - [x] Tests are included.

----


kinow (commented):
Doesn't this allow for possible multithreading issues though? While with the `final` map that's mitigated...

Ostrzyciel:
> Doesn't this allow for possible multithreading issues though? While with the `final` map that's mitigated... maybe looking at other data structures/map implementations could solve it in a different way (there's a comment about ligher impl...)

The class is not thread-safe anyway, because `HashMap` is not thread-safe, so... it should be OK?

We would have to worry about it if we had a shared instance of `ValidationState` across multiple parsers, but we don't. Long-term, I think this would be a good idea, but for that to happen I think most of this random xerces bloat would have to be removed first, as it's mostly dead code anyway.

Ostrzyciel:
...I mean, I can make this thread-safe, but the real question is "what for" :laughing:

jena-core/src/main/java/org/apache/jena/ext/xerces/impl/validation/ValidationState.java
```diff
@@ -86,11 +85,25 @@ public void setSymbolTable(SymbolTable sTable) {
...
+            fIdRefTable = new HashMap();
+        }
+    }
+
     /**
      * return null if all IDREF values have a corresponding ID value;
      * otherwise return the first IDREF value without a matching ID value.
      */
     public String checkIDRefID () {
+        ensureIdTablesAreInitialized();
```
arne-bdt, line 106: Why init here? Instead:
`if(fIdRefTable == null) return null;`

and later in the while loop:

 if (fIdTable == null || !fIdTable.containsKey(key)) {

jena-core/src/main/java/org/apache/jena/ext/xerces/impl/validation/ValidationState.java
```diff
@@ -169,16 +182,21 @@ public boolean isEntityUnparsed (String name) {
...
     @Override
     public boolean isIdDeclared(String name) {
+        if (fIdTable == null) {
+            return false;
+        }
         return fIdTable.containsKey(name);
     }
     @Override
     public void addId(String name) {
+        ensureIdTablesAreInitialized();
```
arne-bdt, line 192: Only init fIdTable on demand here. 
idRefTable is used in oder methods, so it might be that only one of the maps is needed.
```java
if(fIdTable == null) {
    fIdTable = new HashMap();
}
```

jena-core/src/main/java/org/apache/jena/ext/xerces/impl/validation/ValidationState.java
```diff
@@ -169,16 +182,21 @@ public boolean isEntityUnparsed (String name) {
...
         fIdTable.put(name, fNullValue);
     }
 
     // idref
     @Override
     public void addIdRef(String name) {
+        ensureIdTablesAreInitialized();
```
arne-bdt, line 199: Only init idRefTableon demand here. 
fIdTable  is used in oder methods, so it might be that only one of the maps is needed.
```java
if(idRefTable== null) {
    idRefTable= new HashMap();
}
```

jena-core/src/main/java/org/apache/jena/ext/xerces/impl/validation/ValidationState.java
```diff
@@ -86,11 +85,25 @@ public void setSymbolTable(SymbolTable sTable) {
         fSymbolTable = sTable;
     }
 
+    /**
```
arne-bdt, line 88: This methos could be removed. The maps shoudl be individually created on demand, as they are used in differenct methods and not always togehter.

arne-bdt (changes requested):

arne-bdt:
> ...I mean, I can make this thread-safe, but the real question is "what for" 😆

Maybe add a comment to the javadoc header of the class with:
`* This class is not thread safe.`

Ostrzyciel:
> > ...I mean, I can make this thread-safe, but the real question is "what for" 😆
> 
> Maybe add a comment to the javadoc header of the class with: `* This class is not thread safe.`

Done.

afs (approved):

arne-bdt (approved):
@Ostrzyciel Thank you for this improvement!

Issue #2787, opened 2024-10-22
Deadlock in JenaSystem.init() ?


Ostrzyciel:
### Version

5.2.0

### What happened?

After upgrading to Jena 5.2.0, I've started noticing that some of my CI test runs were randomly just getting stuck. After a debugging session with sbt on my machine, I think this might be a deadlock in the initialization code. See the attached thread dump.

The first thread in the dump is within JenaSystem.init(), so it's doing something to initialize Jena. It's stuck on this line: https://github.com/apache/jena/blob/6443abda6e2717b95b05c45515817584e93ef244/jena-core/src/main/java/org/apache/jena/vocabulary/DC_11.java#L29

To run this line, JVM has to run the static initializer of ModelFactory, which includes a call to JenaSystem.init(): https://github.com/apache/jena/blob/6443abda6e2717b95b05c45515817584e93ef244/jena-core/src/main/java/org/apache/jena/rdf/model/ModelFactory.java#L47-L49

Oops?

The other thread in the dump is just waiting on the first one.

I'm not sure how the code got in this state, but I'm using sbt and scalatest, which absolutely love making lots of parallel calls to code that maybe shouldn't be called in parallel. Maybe that is the issue, but I had no problems with Jena 5.1.0.

I think this issue may be related to this change: https://github.com/apache/jena/commit/8654c07f23a892c03a7367981d814c3c8b094f62

### Relevant output and stacktrace

```shell
"pool-18-thread-17@31723" prio=5 tid=0x161 nid=NA runnable
  java.lang.Thread.State: RUNNABLE
	 blocks pool-18-thread-5@31711
	 blocks pool-18-thread-8@31714
	 blocks pool-18-thread-9@31715
	  at org.apache.jena.vocabulary.DC_11.<clinit>(DC_11.java:29)
	  at org.apache.jena.shared.PrefixMapping.<clinit>(PrefixMapping.java:230)
	  at org.apache.jena.sparql.ARQConstants.<clinit>(ARQConstants.java:120)
	  at org.apache.jena.query.ARQ.init(ARQ.java:652)
	  - locked <0x7bff> (a java.lang.Object)
	  at org.apache.jena.sparql.system.InitARQ.start(InitARQ.java:29)
	  at org.apache.jena.base.module.Subsystem.lambda$initialize$1(Subsystem.java:117)
	  at org.apache.jena.base.module.Subsystem$$Lambda$12363/0x00007fe8d9c5a6f0.accept(Unknown Source:-1)
[...]

PR #2807, opened 2024-10-29
GH-2787: Use lazy holder pattern for system initialization

afs:
Use the lazy holder pattern to perform system initialization.
Set `JenaSystem.initialized" flag early to avoid deadlocks.

----


arne-bdt (approved):
Deadlocks are still possible, but the probability of occurrence is massively reduced with this change.

PR #2810, opened 2024-10-29
GH-2808: Implemented Object#equals and #hashCode in BaseDatatype

### LINKED ISSUE #2808: Equality of RDFDatatypes is by reference only --> Should it stay that way?
### Version

5.3.0-SNAPSHOT

### Question

In [apache/jena#2795](https://github.com/apache/jena/issues/2795), I discovered that no implementation of `RDFDatatype` provides an override for `Object#equals` or `Object#hashCode`, resulting in all comparisons being performed by reference only.

For instance, these tests fail:

```java
@Test public void baseDataTypeEquality() {
    var rdfDataType1 = new BaseDatatype("urn:x-hp-dt:unknown");
    var rdfDataType2 = new BaseDatatype("urn:x-hp-dt:unknown");

    assertEquals(rdfDataType1, rdfDataType2);
}

@Test public void xsdDoubleEquality() {
    var rdfDataType1 = new XSDDouble("double", Double.class);
    var rdfDataType2 = XSDDatatype.XSDdouble;

    assertEquals(rdfDataType1, rdfDataType2);
}
```

To further illustrate the inconsistency, here’s a scenario where the first test fails, but the second one succeeds:

```java
@Test
public void testEqualityA() {
    var l1 = LiteralLabelFactory.create("foo", new BaseDatatype("urn:x-hp-dt:unknown"));
    var l2 = LiteralLabelFactory.create("foo", new BaseDatatype("urn:x-hp-dt:unknown"));

    assertEquals(l1, l2);
}

@Test
public void testEqualityB() {
    var m = ModelFactory.createDefaultModel();
    var l1 = m.createTypedLiteral("foo", "urn:x-hp-dt:unknown");
    var l2 = m.createTypedLiteral("foo", "urn:x-hp-dt:unknown");

    assertEquals(l1, l2);
}
```

I propose implementing `Object#equals` and `Object#hashCode` in `BaseDatatype` to compare instances by their class and URI, as shown below:

```java
@Override
public boolean equals(Object o) {
    if (this == o) return true;
    if (o == null || getClass() != o.getClass()) return false;
    final BaseDatatype that = (BaseDatatype) o;
    return Objects.equals(uri, that.uri);
}

@Override
public int hashCode() {
    return Objects.hashCode(uri);
}
```

I couldn’t find any subclass of `BaseDatatype` where uniqueness isn’t defined by the URI, so this approach should cover all cases without needing additional overrides.

arne-bdt:
GitHub issue resolved #2808

Pull request Description:

- Added implementations for `org.apache.jena.datatypes.BaseDatatype#equals` and `#hashCode` based on `RDFDatatype.getURI`, aligning with the principle: "Same URI, same datatype" for RDF datatypes.
-  Added corresponding tests in `org.apache.jena.datatypes.TestDatatypes` to verify equality and hash code behavior.

----

 - [x] Tests are included.
 - no Documentation change and no updates are needed for the [Apache Jena website](https://github.com/apache/jena-site/)

----

### DIFF
--- a/jena-core/src/test/java/org/apache/jena/datatypes/TestDatatypes.java
+++ b/jena-core/src/test/java/org/apache/jena/datatypes/TestDatatypes.java
@@ -20,6 +20,7 @@ package org.apache.jena.datatypes;
 
 import static org.junit.Assert.assertEquals ;
 import static org.junit.Assert.assertFalse ;
+import static org.junit.Assert.assertNotEquals ;
 import static org.junit.Assert.assertNotNull ;
 import static org.junit.Assert.assertTrue ;
 
@@ -27,6 +28,7 @@ import java.math.BigDecimal ;
 import java.util.UUID;
 
 import org.apache.jena.datatypes.xsd.XSDDatatype ;
+import org.apache.jena.datatypes.xsd.impl.XSDDouble ;
 import org.apache.jena.graph.Node ;
 import org.apache.jena.graph.NodeFactory ;
 import org.apache.jena.rdf.model.Resource ;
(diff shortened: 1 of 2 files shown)


jena-core/src/test/java/org/apache/jena/datatypes/TestDatatypes.java
```diff
@@ -197,6 +199,71 @@ public class TestDatatypes {
...
+        var rdfDataType2 = XSDDatatype.XSDdouble;
+        assertEquals(rdfDataType1, rdfDataType2);
+        assertEquals(rdfDataType2, rdfDataType1);
+
+        assertEquals(rdfDataType1.hashCode(), rdfDataType2.hashCode());
+    }
+
+    @Test public void baseDataTypeNotEquals() {
+        var rdfDataType1 = new BaseDatatype("urn:x-hp-dt:unknownA");
+        var rdfDataType2 = new BaseDatatype("urn:x-hp-dt:unknownB");
```
afs, line 248: FYI:

When this is merged, I'd update those as part of the IR3986 PR because "urn:x-" is now (RFC 8141) discouraged. I replaced usage with "urn:jena-dt:" where it is not visible to the application.

afs (approved):

PR #2822, opened 2024-11-05
GH-2821: Support for Timeouts on Updates

### LINKED ISSUE #2821: Cancel Signal unavailable without timeouts - breaks manual abort().
### Version

5.3.0-SNAPSHOT

### What happened?

The cancel signal in [QueryExecDataset](https://github.com/apache/jena/blob/510680cf290cc14d6edf594627c15f87569701b0/jena-arq/src/main/java/org/apache/jena/sparql/exec/QueryExecDataset.java#L482) is only placed into the context when a timeout is configured.

However, in the case of manual invocations to `QueryExec.abort()` the cancel signal is not placed into the context. 
This prevents functions to check the flag and possibly abort prematurely.

A use case is the query which uses the GeoSPARQL property function, which creates a cross product between all geometries (inside the property function) and returns all overlapping ones. @LorenzBuehmann has been working on a revised version that respects the cancel flag, however he noted that the flag is only available when timeouts have been configured, making it impossible to cancel externally using `QueryExec.abort();`

```sparql
PREFIX geo: <http://www.opengis.net/ont/geosparql#>
SELECT * WHERE {
  ?x geo:sfIntersects ?y
}
```

_No response_

Yes

Aklakan:
GitHub issue resolved #2821

Pull request Description:
* Fixed inconsistent context handling in the update builder machinery and made it consistent with that in the query builder machinery.
* The context of update requests is now forwarded to the query execution
* `UpdateEngineWorker` sets the remaining query timeout if the newly introduced `ARQ.updateTimeout` has been set. Note, that it is possible to set an initial timeout via the `ARQ.updateTimeout` from which internally an `ARQ.queryTimeout` setting is derived: An update request would abort if the corresponding query execution timed out on the initial timeout - so only updates with a `WHERE` clause are affected.
* I did not add `initialTimeout` methods to the update exec builders because perhaps initial timeout on updates are a bit to specific for the general interface - opinions?
* Added `UpdateProcessor.getContext()` for consistency with `QueryExec`. This is for discussion.
* Added `UpdateProcessor.abort()` which can abort the underlying query execution if it is based on `QueryIterator`.
* Added test cases to the `TestUpdateExecutionCancel` class.
* For HTTP-based updates, the overall update timeout is forwarded to the existing HTTP machinery.
* There is now a class `TestSPARQLProtocolTimeout` which tests setting the timeout on the update builder.
* `QueryIteratorBase.requestingCancel` was updated to check the `Thread.interrupted()` so that when Fuseki gets shut down (e.g. during unit testing) then any threads still busy executing queries can cancel execution in a timely manner.

* Added query/update tests that register a local function that checks for the availability of the cancel signal. Please also see the [test cases](https://github.com/apache/jena/pull/2822/files#diff-6463c230248848118801cb3a1a78ab22f0f9103de53bd812c87e916d1bfc1805R240) which register the function that checks for the cancel signal in three ways for both queries and updates:

    1. via query/update builder.set
    2. via query/update builder.context
    3. via query/update exec.getContext()

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/http/sys/ExecUpdateHTTPBuilder.java
+++ b/jena-arq/src/main/java/org/apache/jena/http/sys/ExecUpdateHTTPBuilder.java
@@ -20,6 +20,7 @@ package org.apache.jena.http.sys;
 
 import java.net.http.HttpClient;
 import java.util.*;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import org.apache.jena.graph.Node;
@@ -31,6 +32,7 @@ import org.apache.jena.sparql.exec.http.Params;
 import org.apache.jena.sparql.exec.http.UpdateSendMode;
 import org.apache.jena.sparql.syntax.syntaxtransform.UpdateTransformOps;
 import org.apache.jena.sparql.util.Context;
+import org.apache.jena.sparql.util.ContextAccumulator;
 import org.apache.jena.sparql.util.Symbol;
 import org.apache.jena.sys.JenaSystem;
 import org.apache.jena.update.Update;
@@ -56,8 +58,9 @@ public abstract class ExecUpdateHTTPBuilder<X, Y> {
 
     /** Accumulator for update elements. Can build an overall string or UpdateRequest from the elements. */
     private class UpdateEltAcc implements Iterable<UpdateElt> {
-        /** Delimiter for joining multiple SPARQL update strings into a single one. */
-        public static final String DELIMITER = ";\n";
+        /** Delimiter for joining multiple SPARQL update strings into a single one.
+          * The delimiter takes into account that the last line of a statement may be a single-line-comment. */
+        public static final String DELIMITER = "\n;\n";
 
         private List<UpdateElt> updateOperations = new ArrayList<>();
         private List<UpdateElt> updateOperationsView = Collections.unmodifiableList(updateOperations);
@@ -76,6 +79,7 @@ public abstract class ExecUpdateHTTPBuilder<X, Y> {
             add(new UpdateElt(update));
         }
 
+        /** Add a string by parsing it. */
         public void add(String updateRequestString) {
             UpdateRequest updateRequest = UpdateFactory.create(updateRequestString);
             add(updateRequest);
@@ -146,9 +150,11 @@ public abstract class ExecUpdateHTTPBuilder<X, Y> {
     protected UpdateSendMode sendMode = UpdateSendMode.systemDefault;
     protected List<String> usingGraphURIs = null;
     protected List<String> usingNamedGraphURIs = null;
-    protected Context context = null;
+    private ContextAccumulator contextAcc = ContextAccumulator.newBuilder(()->ARQ.getContext());
     // Uses query rewrite to replace variables by values.
     protected Map<Var, Node> substitutionMap     = new HashMap<>();
+    protected long timeout = -1;
+    protected TimeUnit timeoutUnit = null;
 
     protected ExecUpdateHTTPBuilder() {}
 
@@ -213,6 +219,12 @@ public abstract class ExecUpdateHTTPBuilder<X, Y> {
         return thisBuilder();
     }
 
+    public Y timeout(long timeout, TimeUnit timeoutUnit) {
+        this.timeout = timeout;
+        this.timeoutUnit = timeoutUnit;
+        return thisBuilder();
+    }
+
     public Y httpClient(HttpClient httpClient) {
         this.httpClient = Objects.requireNonNull(httpClient);
         return thisBuilder();
@@ -274,26 +286,18 @@ public abstract class ExecUpdateHTTPBuilder<X, Y> {
     public Y context(Context context) {
         if ( context == null )
             return thisBuilder();
-        ensureContext();
-        this.context.setAll(context);
+        this.contextAcc.context(context);
         return thisBuilder();
     }
 
     public Y set(Symbol symbol, Object value) {
-        ensureContext();
-        this.context.set(symbol, value);
+        this.contextAcc.set(symbol, value);
         return thisBuilder();
     }
 
     public Y set(Symbol symbol, boolean value) {
-        ensureContext();
-        this.context.set(symbol, value);
+        this.contextAcc.set(symbol, value);
         return thisBuilder();
-    }
-
-    private void ensureContext() {
-        if ( context == null )
-            context = Context.create();
     }
 
     public X build() {
@@ -322,7 +326,7 @@ public abstract class ExecUpdateHTTPBuilder<X, Y> {
         // If the UpdateRequest object wasn't built until now then build the string instead.
         String updateStringActual = updateActual == null ? updateEltAcc.buildString() : null;
 
-        Context cxt = (context!=null) ? context : ARQ.getContext().copy();
+        Context cxt = contextAcc.context();
         return buildX(hClient, updateActual, updateStringActual, cxt);
     }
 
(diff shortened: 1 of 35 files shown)


jena-arq/src/main/java/org/apache/jena/http/sys/ExecUpdateHTTPBuilder.java
```diff
@@ -275,25 +276,25 @@ public Y context(Context context) {
...
     }
 
     public Y set(Symbol symbol, boolean value) {
         ensureContext();
-        this.context.set(symbol, value);
+        this.contextAcc.set(symbol, value);
         return thisBuilder();
     }
 
     private void ensureContext() {
```
Aklakan, line 295: The empty ensureContext method is akin to [ExecHTTPBuilder](https://github.com/apache/jena/blob/c7eaf83cf328496cb4f8a24b42a483199ac216c0/jena-arq/src/main/java/org/apache/jena/http/sys/ExecHTTPBuilder.java#L304). It could be removed there and in this file (`ExecUpdateHTTPBuilder).

jena-arq/src/main/java/org/apache/jena/sparql/modify/UpdateEngineWorker.java
```diff
@@ -537,7 +537,7 @@ protected static Iterator<Binding> evalBindings(Query query, DatasetGraph dsg, B
         }
 
         // Not QueryExecDataset.dataset(...) because of initialBinding.
-        QueryExecDatasetBuilder builder = QueryExecDatasetBuilder.create().dataset(dsg).query(query);
+        QueryExecDatasetBuilder builder = QueryExecDatasetBuilder.create().dataset(dsg).query(query).context(context);
```
Aklakan, line 540: So far that context was never passed from the update engine to the query engine - so non-global custom settings (function registries, timeouts, etc) were lost here.

afs:
This PR is a good step forward.

This PR is really two things:
* The bug about manual query execution abort 
* A new feature - having timeouts on update.

The first is as little as move one line fix :-)

Having timeouts on update is good to have and the way it is done in this PR, timing out the WHERE clause of a modify `DELETE-INSERT-WHERE`, seems the best way (I'd played around with update timeout ... but stalled on that) because it protects against long-running `WHERE`.

There are various operations for updates and a general timeout on an update isn't a one change in one place. Update can be several operations all in one request so there could an overall timeout (caveat the issue of various operations).

The PR does impact on applications. This is not to say I think that it is  bad (or good) idea but to draw out a consequence.

If I've read the PR with it comes down to this one line: [pass the context to the WHERE evaluation](https://github.com/apache/jena/pull/2822/files/1db78a53978eefc808851dffe730492898f85671#diff-c57bb9dc6558a713203a51ec552b8d3f0e2ac2d76a693085664beaec44cde554R540). If a global timeout for queries is set, then that will be picked up by the update execution where it wasn't 5.2 and earlier.

For Fuseki, I would guess that timeout is usually per-endpoint; it is possible to set it database or server wide as well.

For an application using Jena as a library, an application wide timeout might be present and that would impact updates with this PR.

Aklakan:
>  If a global timeout for queries is set, then that will be picked up by the update execution where it wasn't 5.2 and earlier.

* One option would be to introduce `ARQ.updateTimeout` and have the UpdateEngineWorker (or perhaps already at `builder.build()`) do

```java
cxt.set(ARQ.queryTimeout, cxt.get(ARQ.updateTimeout))
```

* Furthermore, there could be an additional `ARQ.requestType` context symbol for query and update requests (and perhaps dataset ones; c.f. `LinkDataset`). And the appropriate timeout value is picked based on the context's request type which might be cleaner than having to mutate the context.

I'd say that the basic updateTimeout setting should affect the overall timeout of an update request - not per sub-update.
The suggestions should be backward compatible with the existing behavior (but may require a change for computing the effective timeout in the query exec).

jena-arq/src/main/java/org/apache/jena/sparql/exec/http/UpdateExecHTTP.java
```diff
@@ -130,13 +136,40 @@ private void executePostForm(Params thisParams) {
...
+    /** Best effort that tries to close an underlying HTTP connection.
+     *  May still hang waiting for the HTTP request to complete. */
+    @Override
+    public void abort() {
+        cancelSignal.set(true);
+        synchronized (cancelSignal) {
+            try {
+                InputStream in = retainedConnection;
+                if (in != null) {
+                    in.close();
```
Aklakan, line 167: I suppose this will just hang trying to consume any remaining data. Perhaps a test case could be created against the Fuseki server.

jena-arq/src/main/java/org/apache/jena/sparql/modify/UpdateEngineWorker.java
Aklakan: There is now an updateRemainingQueryTimeout that will set ARQ.queryTimeout based on the value of ARQ.updateTimeout and the time spent in the UpdateEngineWorker.

Aklakan:
I have consolidated the Timeout stuff (initial and overall) into the `Timeouts` class.
There is now ARQ.updateTimeout. The UpdateEngineWorker tracks its own processing time and updates its context with the remaining query time out before starting the underlying query execution.

The update timeouts are now only used to update ARQ.queryTimeout in the context.

Conversely, the update timeouts do not affect the sinks: In principle a sink could also refuse to accept changes once the timeout is reached - but if that's really needed it could be a future update.

jena-integration-tests/src/test/java/org/apache/jena/test/rdfconnection/TestRDFConnectionRemote.java
```diff
@@ -228,11 +230,27 @@ public void non_standard_syntax_update_remote_3c() {
...
+                .update("INSERT { <x> <y> <z> } WHERE { FILTER(?foo) }")
+                .substitution("foo", FALSE)
+                .build()
+                .execute();
+        }
+    }
+
+    /** Standard update syntax with substitution should work when comments are involved. */
+    @Test
+    public void standard_syntax_update_remote_2b() {
```
Aklakan, line 246: Strictly speaking, this additional test case could go into a separate commit. It tests for an oversight I introduced with #2580 where update strings were joined on the semicolon without a preceding newline. In such cases, joining update strings that ended on a comment would produce syntax errors.

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIteratorBase.java
```diff
@@ -70,7 +70,7 @@ protected QueryIteratorBase(AtomicBoolean cancelSignal) {
     }
 
     private boolean requestingCancel() {
-        return requestingCancel != null && requestingCancel.get() ;
+        return (requestingCancel != null && requestingCancel.get()) || Thread.interrupted() ;
```
Aklakan, line 73: The additional interrupted check is needed in order to cleanly stop the Fuseki server in `TestSPARQLProtocolTimeout` which extends `AbstractFusekiTest`: When shutting down the Fuseki server, Jetty attempts to stop running the remaining threads via interruption. However, so far this flag was not picked up by the `QueryIterator` machinery, causing jetty to abandon the thread and log the warning from `org.eclipse.jetty.util.thread.QueuedThreadPool`:

```
17:27:10 WARN  QueuedThreadPool :: Couldn't stop Thread[qtp1495161082-24,5,main]
```

Aklakan:
I updated this PR's summary in the first post. The main changes are the introduction of `ARQ.updateTimeout` and the additional cancel check based on ` Thread.interrupted()` in `QueryIteratorBase.requestingCancel()`.

Its' ready for another review.

afs (approved):
LGTM.
A couple of unused imports which I can sort out after the merge.

Aklakan:
I removed the two unused imports.

afs:
Thank you!

PR #2830, opened 2024-11-08
Profile to skip UI tests

afs:
This adds a profile that skips the UI tests.

Currently, the build fails on ASF Jenkins because many AFS Jenkins host nodes are older LTS Ubuntu and don't have the necessary dependencies installed + some dependencies also need updated system libraries.

I haven't made progress on a containerized build. 

The advantage of the ASF builds is that that have all the Java versions. There are changes  in the java world where it would be good to have builds running on the latest (non-LTS) java.

A hopefully temporary profile that builds the UI but does not run the tests would help keeping the development snapshot current and give builds using latest Java.

----

### DIFF
--- a/jena-fuseki2/jena-fuseki-ui/pom.xml
+++ b/jena-fuseki2/jena-fuseki-ui/pom.xml
@@ -32,7 +32,26 @@
         <node.version>v20.11.0</node.version>
         <yarn.version>v1.22.17</yarn.version>
         <automatic.module.name>org.apache.jena.fuseki.ui</automatic.module.name>
+
+        <!-- 
+             Allow for skipping the tests.
+             The test frameworks need various software compoents to 
+             be available on the host running maven. This isn't always 
+             practical. This property allows the UI to be built on such machines
+             by skipping the tests.
+             It is set "true" by profile "ui-skip-tests"
+        -->
+        <skip.ui.tests>false</skip.ui.tests>
+
     </properties>
+    <profiles>
+      <profile>
+        <id>ui-skip-tests</id>
+        <properties>
+          <skip.ui.tests>true</skip.ui.tests>
+        </properties>
+    </profile>
+    </profiles>
 
     <build>
         <plugins>
@@ -98,6 +117,7 @@
                         </goals>
                         <phase>test</phase>
                         <configuration>
+                            <skip>${skip.ui.tests}</skip>
                             <arguments>run test:unit</arguments>
                         </configuration>
                     </execution>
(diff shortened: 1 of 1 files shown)


kinow (approved):
>Currently, the build fails on ASF Jenkins because many AFS Jenkins host nodes are older LTS Ubuntu and don't have the necessary dependencies installed + some dependencies also need updated system libraries.

I believe once the build is containerized that will be fixed on ASF side.

>I haven't made progress on a containerized build.

:+1: 

>The advantage of the ASF builds is that that have all the Java versions. There are changes in the java world where it would be good to have builds running on the latest (non-LTS) java.

I think "Jena Development Deploy" is using JDK 21. GH Actions supports natively 

image

and with containers I think that issue should be gone (whether on ASF side or not)?

>A hopefully temporary profile that builds the UI but does not run the tests would help keeping the development snapshot current and give builds using latest Java.

Couldn't we either enable GH Actions just for the UI, or get Jenkins to notify GitHub later? This way the dependabot issues could be tested automatically there.

afs:
> Couldn't we either enable GH Actions just for the UI, or get Jenkins to notify GitHub later? This way the dependabot issues could be tested automatically there.

We could though an issue with GH actions is that people who fork the repo also get the actions and dependabot setup. I think this may have got better but in the past they got alters and any actions triggered ... on their own account allowances.

afs:
> I think "Jena Development Deploy" is using JDK 21.

Fixed.

kinow:
> > Couldn't we either enable GH Actions just for the UI, or get Jenkins to notify GitHub later? This way the dependabot issues could be tested automatically there.
> 
> We could though an issue with GH actions is that people who fork the repo also get the actions and dependabot setup. I think this may have got better but in the past they got alters and any actions triggered ... on their own account allowances.

I think in the GH Actions you can disable that with:

```yaml
name: test
on: [push, workflow_dispatch, etc...]
jobs:
  run_something:
    if: github.repository == 'apache/jena'
    runs-on: ubuntu-latest
    steps:
      - uses: action/checkout...
      - ...
```

e.g.:

- https://github.com/apache/arrow/blob/03c3f8ed41573b611c4c6dfaf1ec166a9b78decb/.github/workflows/java.yml#L101-L107
- https://github.com/apache/echarts/blob/18f8ea3aba11bb533885d2ed3b817d8da455602a/.github/workflows/stale.yml#L14
- https://github.com/apache/spark/blob/8f5d8d424644a62c326b0c430bd311c834236e2b/.github/workflows/build_coverage.yml#L32
- https://github.com/apache/camel/blob/281692df4f18434116663ea89664af2b90ce0d07/.github/workflows/pr-build-main.yml#L38

>and dependabot setup

This part I think depends on your repo settings. My `kinow/jena` doesn't have Dependabot enabled. I think I may have disabled it, but that's something users that forked the repo can do too. A bit annoying, but at least there's a way to disable it (could be worte :grimacing: )

afs:
> I think in the GH Actions you can disable that with:

PR? Let's try it out!

One of the things that PRs also need to checked for is warnings (java and javadoc). The build running succeeding isn't enough.

PR #2832, opened 2024-11-09
Javadoc fixes

afs:
Based on the output of the java23 build on Jenkins, there are some errors in the javadoc which this PR fixes.

The change in jena-iri3986 about maven-jar-plugin is to have both `automatic.module.name` and a module-info.java file ...
and avoid a javadoc error about mixed aggregation javadoc!

----

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/mem/SparseArraySpliterator.java
+++ b/jena-core/src/main/java/org/apache/jena/mem/SparseArraySpliterator.java
@@ -153,48 +153,46 @@ public class SparseArraySpliterator<E> implements Spliterator<E> {
     }
 
     /**
-     * Returns an estimate of the number of elements that would be
-     * encountered by a {@link #forEachRemaining} traversal, or returns {@link
-     * Long#MAX_VALUE} if infinite, unknown, or too expensive to compute.
+     * Returns an estimate of the number of elements that would be encountered by a
+     * {@link #forEachRemaining} traversal, or returns {@link Long#MAX_VALUE} if
+     * infinite, unknown, or too expensive to compute.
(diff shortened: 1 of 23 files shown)


jena-core/src/main/java/org/apache/jena/mem/SparseArraySpliterator.java
```diff
@@ -153,48 +153,46 @@ public Spliterator<E> trySplit() {
...
-     * For example, a sub-spliterator of an approximately balanced binary tree
-     * may return a value that estimates the number of elements to be half of
-     * that of its parent; if the root Spliterator does not maintain an
-     * accurate count, it could estimate size to be the power of two
-     * corresponding to its maximum depth.
+     *     For example, a sub-spliterator of an approximately balanced binary tree
+     *     may return a value that estimates the number of elements to be half of
+     *     that of its parent; if the root Spliterator does not maintain an accurate
+     *     count, it could estimate size to be the power of two corresponding to its
+     *     maximum depth.
```
kinow, line 174: Looks like indentation matters now for annotations, good to know!

kinow (approved):

jena-core/src/main/java/org/apache/jena/mem/SparseArraySpliterator.java
afs: I think it's a convention for readability - I used Eclipse to reformat. 

There's a similar unindented javadoc comment in that file that I didn't reformat and it displays correctly. The rule seems to be running until the next tag/end of comment. HTML will turn multiple spaces into one.

This isn't Python ... yet 😄

PR #2845, opened 2024-11-15
GH-2753: Customize Fuseki main arguments

### LINKED ISSUE #2753: Improve ability to support custom arguments for Fuseki Servers

afs:
GitHub issue resolved #2753

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/cmds/DSGSetup.java
+++ b/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/cmds/DSGSetup.java
@@ -20,12 +20,27 @@ package org.apache.jena.fuseki.main.cmds;
 
 import java.io.File;
 import java.nio.file.Path;
+import java.util.List;
 
+import arq.cmdline.ModDatasetAssembler;
 import org.apache.jena.atlas.io.IO;
+import org.apache.jena.atlas.lib.FileOps;
 import org.apache.jena.cmd.CmdException;
 import org.apache.jena.fuseki.system.spot.TDBOps;
+import org.apache.jena.query.Dataset;
+import org.apache.jena.riot.Lang;
+import org.apache.jena.riot.RDFDataMgr;
+import org.apache.jena.riot.RDFLanguages;
+import org.apache.jena.riot.RiotException;
+import org.apache.jena.sparql.core.DatasetGraphFactory;
+import org.apache.jena.system.Txn;
 import org.apache.jena.tdb1.TDB1Factory;
 import org.apache.jena.tdb2.DatabaseMgr;
+import org.slf4j.Logger;
+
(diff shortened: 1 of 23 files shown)


jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/cmds/DSGSetup.java
```diff
@@ -46,40 +61,78 @@
...
+        serverArgs.datasetDescription = "in-memory";
+        serverArgs.dsg = DatasetGraphFactory.createTxnMem();
+        serverArgs.allowUpdate = true;
+    }
+
+    /*package*/ static void setupFile(Logger log, List<String> filenames, ServerArgs serverArgs) {
+        serverArgs.datasetDescription = "in-memory, with files loaded";
+        serverArgs.dsg = DatasetGraphFactory.createTxnMem();
+
+        for(String filename : filenames ) {
```
kinow, line 110: s/for(/for (/ ?

kinow (approved):
Skimmed through, found nothing that seems wrong/broken. :+1: :clap:

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/sys/FusekiModules.java
```diff
@@ -23,16 +23,51 @@
...
  * @see FusekiAutoModules
  */
 public class FusekiModules {
 
+    private static FusekiModules autoLoadedFusekiModules = FusekiAutoModules.get();
+    // Never null, maybe empty
+    private static FusekiModules systemFusekiModules = autoLoadedFusekiModules;
+
+    /**
+     * There is a system wide set of modules used when o other are indicated.
```
rvesse, line 40:
```suggestion
     * There is a system wide set of modules used when no other are indicated.
```

rvesse (approved):
Looks good to me, and looking at the test cases this covers all the scenarios that motivated my original ticket #2753

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/cmds/DSGSetup.java
afs: Fixed.

afs:
> Looks good to me, and looking at the test cases this covers all the scenarios that motivated my original ticket #2753

There a few niggles in the command argument processing that could be improved and the overall system assumes a degree of cooperation between elements not to tread on each other. These are best dealt with from experience.

Please do let us know how using it goes.

afs:
Apologies - I forgot to push the fixes mentioned above. I'll include them in an args clearup PR.

PR #2846, opened 2024-11-17
GH-2828: Remove validation for XML-only datatypes

### LINKED ISSUE #2828: Clean up code in jena.ext.xerces
### Change

This is a continuation of #2797 – trimming out the bloat in datatyped literal validations.

That issue tackled the worst offender – allocating two hashmaps per each validated literal. This is however not the whole story. For every literal validation we still allocate a new `ValidationState`:

https://github.com/apache/jena/blob/c7eaf83cf328496cb4f8a24b42a483199ac216c0/jena-core/src/main/java/org/apache/jena/datatypes/xsd/XSDDatatype.java#L267-L271

This makes no sense, because that object is never really mutated (or, to be precise, does not need to be mutable). At the same time, it has quite a few fields. Although the JVM can probably figure out how to handle this efficiently with escape analysis, this is still unnecessary bloat.

The entire `org.apache.jena.ext.xerces` package contains a lot of unused code carried over from xerces. A lot of the infrastructure is not needed in Jena, because the more complex XML features make no sense in the context of RDF literals. For example, a large part of the original job of `ValidationState` was to check if ID and IDREF attributes are correct with respect to one another. This, along with a few other quirky XML thingies "SHOULD NOT" be used in RDF [according to the spec](https://www.w3.org/TR/rdf11-concepts/#xsd-datatypes), so I think we can safely remove this.

## The plan

- Remove code for special handling of datatypes: QName, ENTITY, ID, IDREF, NOTATION, IDREFS, ENTITIES, NMTOKENS – none of which are valid in RDF.
  - Note that although ID/IDREF validation is implemented in Jena, it currently does not work, because the `ValidationState` is allocated once per literal, not once per document. And what would a "document" even mean in RDF?
- Instantiate only one instance of `ValidationState` per Jena instance, or something to a similar effect.
- Remove all kinds of unused code from the xerces package to help maintainability, make the JARs a bit smaller etc.

## How do I do this?

What is the process for making breaking changes to Jena APIs? I assume that even if a public class is not used in the Jena codebase, it can be removed only in a MAJOR release. So, should I do something like:

- Make a PR deprecating all the different public methods and classes and marking them for removal in Jena 6
- Before Jena 6.0 release (but after last Jena 5.x release) make a PR actually removing the code

?

## Notes on unused code

- According to my IntelliJ, these methods of `ValidationState` are not used anywhere in the Jena codebase: 
  - All `set*` methods, except `setEntityState`, which is used only in `ValidationManager`, which is in turn an unused class.
  - `resetIDTables`, `reset`, `useNamespaces`
  - Of course, this needs to be double-checked.
- Other unused code:
  - `ConfigurableValidationState` class
  - `DatatypeValidator` interface
  - `EntityState` interface – it is only used in unused methods of `ValidationState` and `ValidationManager`
  - `SchemaDVFactory` methods: `createTypeRestriction`, `createTypeList`, `createTypeUnion` – this refers to some advanced XSD features... I think? Anyway, this doesn't seem to be used in RDF.
    - Implementations of these methods in child classes (`BaseSchemaDVFactory` and `BaseDVFactory`) are, by extension, also unused.
  - `XSDDatatype`: there is a huge comment block section that probably should be removed. The inner static class `XSDGenericType` is also unused.
  - `ValidatedInfo` methods: `stringValue`, `isComparable`, `copyFrom`
  - `NamespaceContext` fields: `XML_URI`, `XMLNS_URI` and methods `pushContext`, `popContext`, `declarePrefix`, `getDeclaredPrefixCount`, `getDeclaredPrefixAt`
    - Actually, the whole interface is effectively unused... there are no classes implementing it. There is some code passing around instances of `NamespaceContext`, but this must be always null.

Yes

Ostrzyciel:
GitHub issue resolved #2828

Pull request Description:

Remove the code for validating datatypes that according to the RDF 1.1 spec SHOULD NOT be used: QName, ENTITY, ID, IDREF, NOTATION, ENTITIES, NMTOKENS, IDREFS. Also remove code for validating XSD lists and unions. This should not change the behavior of Jena for users, unless they were relying on it to validate their weird XML data, which I don't think is an official functionality of Jena.

This is the first step to resolving #2828 – first I wanted to focus on untangling the unneeded the logic. In future PRs, I will remove all the dead code.

After applying these changes, the size of the jena-core JAR was reduced by 13 489 bytes. Not huge, but still something.

Details:

- `XSDDatatype`
    - Removed commented-out datatype definitions for ENTITIES, NMTOKENS, IDREFS.
    - Removed some other commented-out temporary code that was supposed to be cleaned up some time ago (I guess).
    - I left in the registered datatypes for QName, ENTITY, ID, IDREF, NOTATION, because they are not doing any harm here.
- `XSSimpleType`
    - Removed method `isIdType()` along with its implementations (was unused).
    - Removed method `getPrimitiveKind()` along with its implementations (was unused).
- `XSSimpleTypeDecl`
    - Removed the registration of type validators for: QName, ENTITY, ID, IDREF, NOTATION.
    - Removed validation code for these datatypes.
    - Removed all code related to XSD unions and lists (not needed in RDF). Notably, this left atomic types as the only possible types, which greatly reduced the number of branches.
    - *Note:* there was some weirdness in the code where anySimpleType was treated as NOTATION for some reason. I have untangled this. I think any changes to this won't matter anyway, because neither NOTATION nor xsd:anySimpleType make particular sense in RDF.
- `XSSimpleTypeDefinition`
    - Removed constants VARIETY_UNION, VARIETY_LIST.
    - Removed methods related to handling lists and unions.
- `BaseSchemaDVFactory` and `FullDVFactory`
    - Replaced the code for: QName, ENTITY, ID, IDREF, NOTATION, ENTITIES, NMTOKENS, IDREFS with dummy registrations. We still need to register something for these datatypes here, because this is used by `XSDDatatype` in a public interface.
    - Removed the code for XSD lists and unions.
- `BaseDVFactory` and `SchemaDVFactory`
    - Removed the code for XSD lists and unions.
- Removed validator classes (now unused): IDDV, IDREFDV, EntityDV, ListDV, UnionDV

----

 - [x] Tests are included.

----


jena-core/src/main/java/org/apache/jena/datatypes/xsd/XSDDatatype.java
```diff
@@ -124,9 +121,6 @@ public class XSDDatatype extends BaseDatatype {
     public static final XSDDatatype XSDName = new XSDBaseStringType("Name");
 
     /** Datatype representing xsd:QName */
-    // If you see this, remove commented lines.
-    // Merely temporary during switch over and testing.
-    // public static final XSDDatatype XSDQName = new XSDDatatype("QName");
     public static final XSDDatatype XSDQName = new XSDPlainType("QName");
```
afs, line 124: Lets' add `@deprecated(forRemoval = true)` and `/** @deprected Do not use */` to `XSDQname` and the other non-RDF XSD datatypes.

afs: Other than that, the rest looks great!

Ostrzyciel: Done. I've also added a note to remove these in Jena 6, together with the associated dummy code in BaseSchemaDVFactory.

afs (approved):

PR #2882, opened 2024-12-06
GH-2881: Mitigate cases of open iterators when cancelling queries.

### LINKED ISSUE #2881: Open iterators when cancelling queries
### Version

5.3.0-SNAPSHOT

### What happened?

Sometimes when cancelling queries warnings are logged such as:
```
Open iterator: QueryIterPeek/123
Open iterator: QueryIterSingleton/456
```

So far I noticed this to happen when BGPs and UNIONs are involved, but this may not be exhaustive.

It turns out that not all the iterator construction in ARQ is performed truly lazy, which causes some parts of the code to leave iterators open if an exception happens during the iterator construction.

After investigation it seems that the warnings should (usually) be harmless, because the `QueryIter`s are tracked in the `ExecutionContext` and thus eventually become closed without leaking resources.

However, seeing these warnings suggests that resources *might* be leaked - so this is an issue that IMO should be fixed.

See the PR for a test case and the proposed fixes.

_No response_

Yes

Aklakan:
GitHub issue resolved #2881 

Pull request Description:
- Added a test case that should emit warnings without the fixes (the tests do not fail otherwise)
- Adds exception handling to OpUnion and StageGeneratorGeneric, mainly to cope with concurrent QueryCancelledException.

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterPeek.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterPeek.java
@@ -25,23 +25,23 @@ import org.apache.jena.sparql.engine.binding.Binding ;
 
 public class QueryIterPeek extends QueryIter1
 {
-    private Binding binding = null ; 
+    private Binding binding = null ;
     private boolean closed = false ;
-    
+
     public static QueryIterPeek create(QueryIterator iterator, ExecutionContext cxt)
     {
         if ( iterator instanceof QueryIterPeek)
             return (QueryIterPeek)iterator ;
         return new QueryIterPeek(iterator, cxt) ;
     }
-    
+
     private QueryIterPeek(QueryIterator iterator, ExecutionContext cxt)
     {
         super(iterator, cxt) ;
     }
 
     /** Returns the next binding without moving on.  Returns "null" for no such element. */
-    public Binding peek() 
+    public Binding peek()
     {
         if ( closed ) return null ;
         if ( ! hasNextBinding() )
@@ -73,7 +73,7 @@ public class QueryIterPeek extends QueryIter1
     @Override
     protected void closeSubIterator()
     { closed = true ; }
-    
+
     @Override
     protected void requestSubCancel()
     { }
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/main/StageGeneratorGeneric.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/main/StageGeneratorGeneric.java
@@ -20,6 +20,7 @@ package org.apache.jena.sparql.engine.main;
 
 import org.apache.jena.atlas.lib.Lib ;
 import org.apache.jena.atlas.logging.Log ;
+import org.apache.jena.query.QueryCancelledException;
 import org.apache.jena.sparql.core.BasicPattern ;
 import org.apache.jena.sparql.core.Substitute ;
 import org.apache.jena.sparql.engine.ExecutionContext ;
@@ -49,6 +50,11 @@ public class StageGeneratorGeneric implements StageGenerator {
         return execute(pattern, reorder, input, execCxt) ;
     }
 
+    /**
+     * Attempts to construct an iterator that executes the input against the pattern.
+     * If the construction fails, such as due to {@link QueryCancelledException}, then the exception is passed on
+     * and the input iterator will be closed.
+     */
     protected QueryIterator execute(BasicPattern pattern, ReorderTransformation reorder,
                                     QueryIterator input, ExecutionContext execCxt) {
         Explain.explain(pattern, execCxt.getContext()) ;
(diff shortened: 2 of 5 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/engine/main/StageGeneratorGeneric.java
```diff
@@ -65,7 +86,7 @@ protected QueryIterator execute(BasicPattern pattern, ReorderTransformation reor
                 QueryIterPeek peek = QueryIterPeek.create(input, execCxt) ;
                 // And now use this one
                 input = peek ;
-                Binding b = peek.peek() ;
+                Binding b = closeOnException(peek::peek, peek) ;
```
afs, line 89: It looks to me like `QueryIterPeek` should be updated to handle cancellations rather than putting execution costs into the main path.

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/StageGeneratorGeneric.java
```diff
@@ -49,12 +53,29 @@ public QueryIterator execute(BasicPattern pattern, QueryIterator input, Executio
...
+     */
     protected QueryIterator execute(BasicPattern pattern, ReorderTransformation reorder,
                                     QueryIterator input, ExecutionContext execCxt) {
         Explain.explain(pattern, execCxt.getContext()) ;
 
-        if ( ! input.hasNext() )
-            return input ;
+        QueryIterator originalInput = input;
+        if ( closeOnException(() -> !originalInput.hasNext(), originalInput) )
+            return originalInput ;
```
afs, line 78: The `input` should be closed in a cancel. If it is not then something is missing elsewhere. 

See the issue #2881 for discussion about inverting the overall design to have explicit declaration and management of resources that have clear-up.

Aklakan: Right, I think this check is superfluous.

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/StageGeneratorGeneric.java
Aklakan: Mhm, an exception during `iter.peek()` could close `iter`.

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterPeek.java
```diff
@@ -25,27 +25,34 @@
...
-    public Binding peek() 
+    public Binding peek()
     {
         if ( closed ) return null ;
-        if ( ! hasNextBinding() )
-            return null ;
+        try {
+            if ( ! hasNextBinding() )
+                return null ;
+        } catch (Exception e) {
```
Aklakan, line 50: Perhaps having the try-catch block in `StageGeneratorGeneric.execute` is slightly better. Essentially, a method that returns a QueryIterator should close its own already allocated resources early in case of a failure. It feels not totally clean having peek() deal with an exception but not hasNextBinding / moveToNext.

The thing is, that even if an iterator's hasNextBinding / moveToNext raise an exception, then this iterator will usually connected to the 'root' iterator. The query execution will catch an exception and close the root iterator which will propagate to its children and only then a check for dangling resources is done.

The problem is that StageGeneratorGeneric creates an QueryIterPeek, then calls peek which fails, and then the already registered iterator is left dangling - so its not really an issue of peek().

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/StageGeneratorGeneric.java
Aklakan: Removed the check again.

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterPeek.java
Aklakan: I moved the try-catch block back to StageGeneratorGeneric because based on my previous argument it would be the appropriate place.

jena-arq/src/main/java/org/apache/jena/sparql/engine/main/iterator/QueryIterUnion.java
```diff
@@ -51,21 +50,51 @@ public QueryIterUnion(QueryIterator input,
...
+                    } else {
+                        if (subOpIt.hasNext()) {
+                            Op subOp = subOpIt.next();
+                            qIter = QC.execute(subOp, binding, getExecContext());
+                        } else {
+                            return false;
+                        }
+                    }
+                }
+            }
```
Aklakan, line 86: Making the nextStage() method of `QueryIterUnion` truly lazy removes the need for try-catch blocks: Sub-iterators are only opened when needed, and if a creation fails then there are no previously opened resources that need to be cleaned up.

afs (approved):

Issue #2833, opened 2024-11-10
Extract the Xerces implementation of regular expressions for SPARQL use.


afs:
### Change

As an option in ARQ, `REGEX` can be implemented using the Xerces regular expression engine. Formally, regular expressions in SPARQL are defined to be the [XPath/XQuery Functions and Operators regular expressions](https://www.w3.org/TR/xpath-functions-3/#regex-syntax) which are derived from [XML Schema Regular expressions](https://www.w3.org/TR/xmlschema11-2/#regexs).

ARQ uses the JDK by default - the differences are slight (only in regex flags supported).

This tasks is to take the Xerces derived code and produce a standalone regex engine solely for ARQ use.

See also #2828.

### Are you interested in contributing a pull request for this task?

Yes

PR #2906, opened 2024-12-23
GH-2828: Remove unused code in ext.xerces

### LINKED ISSUE #2828: Clean up code in jena.ext.xerces

Ostrzyciel:
GitHub issue resolved #2828 

Pull request Description:

This completes #2828, removing the vast majority of unused and unnecessary code in the ext.xerces namespace. Jena still compiles after this, and tests seem to be passing.

- In general, I went through the code iteratively, locating unreachable code branches and unused fields/methods/classes. I repeated this until there was nothing left to remove.
- The `ValidationState` class that I initially had my main issue with turned out to be completely unnecessary. As far as I understand, its only use was to store some state as the validator goes through the XML document. Because here we are always validating one literal (and not a document), this was not needed at all.
- After #2833, the only use of the `pattern` facet (for regex checks) was for `xsd:language`, where the pattern is very, very simple (`([a-zA-Z]{1,8})(-[a-zA-Z0-9]{1,8})*`). I have therefore removed the xerces regex implementation from core entirely and instead the code now uses Java `Pattern`. I'm 99.9% sure it will be fine. I've also added some tests to check this.
- Several other validation facets turned out to be unused in the code (for example: everything related to length). I have removed these.
- Many classes were used only for validation of custom XSD types, and therefore they were removed.
- The size of the `jena-core` JAR was reduced by ~165 KB (from 2_016_067 to 1_846_347 bytes).
- Datatype validation in Jena should be somewhat sped up by this, just because less objects are allocated and passed through the call stack. The leaner code size may also help with instruction caching on the CPU. I didn't do any benchmarks, though... I will check this and come back later with some numbers.

----

 - [x] Tests are included.

----


Ostrzyciel:
I did a quick JMH test to see if this makes anything faster (note: making datatype parsing/validation was **not** a goal of this PR, if anything is faster, it's just a happy accident). 

Benchmark code:

<details>

    package org.apache.jena.datatypes;
    
    import org.apache.jena.datatypes.xsd.XSDDatatype;
    import org.apache.jena.mem.graph.helper.JMHDefaultOptions;
    import org.junit.Assert;
    import org.junit.Test;
    import org.openjdk.jmh.annotations.*;
    import org.openjdk.jmh.infra.BenchmarkParams;
    import org.openjdk.jmh.infra.Blackhole;
    import org.openjdk.jmh.runner.Runner;
    
    import java.util.concurrent.TimeUnit;
    
    @State(Scope.Benchmark)
    public class TestDatatypeLiteralParse {
    
        @Param({
                "string",
                "1",
                "-340415",
                "0.00003",
                "PT7M",
                "pl-PL",
                "2021-07-01T12:00:00Z",
        })
        public String param0_LiteralValue;
    
        private XSDDatatype datatype;
    
        @Setup(Level.Trial)
        public void setup(BenchmarkParams params) {
            switch (param0_LiteralValue) {
                case "string":
                    datatype = XSDDatatype.XSDstring;
                    break;
                case "1":
                    datatype = XSDDatatype.XSDshort;
                    break;
                case "-340415":
                    datatype = XSDDatatype.XSDinteger;
                    break;
                case "0.00003":
                    datatype = XSDDatatype.XSDdecimal;
                    break;
                case "PT7M":
                    datatype = XSDDatatype.XSDduration;
                    break;
                case "pl-PL":
                    datatype = XSDDatatype.XSDlanguage;
                    break;
                case "2021-07-01T12:00:00Z":
                    datatype = XSDDatatype.XSDdateTime;
                    break;
            }
        }
    
        @Benchmark
        @OperationsPerInvocation(40_000_000)
        public void parse(Blackhole bh) {
            for (int i = 0; i < 40_000_000; i++)
                bh.consume(datatype.parse(param0_LiteralValue));
        }
    
        @Test
        public void benchmark() throws Exception {
            var opt = JMHDefaultOptions.getDefaults(this.getClass())
                    .timeUnit(TimeUnit.NANOSECONDS)
                    .warmupIterations(4)
                    .measurementIterations(4)
                    .build();
            var results = new Runner(opt).run();
            Assert.assertNotNull(results);
        }
    }
</details>

```
# JMH version: 1.37
# VM version: JDK 17.0.13-ea, OpenJDK 64-Bit Server VM, 17.0.13-ea+6-Debian-1
```

Code from this PR:

```
Benchmark                       (param0_LiteralValue)  Mode  Cnt   Score   Error  Units
TestDatatypeLiteralParse.parse                 string  avgt    4   4.688 ± 0.107  ns/op
TestDatatypeLiteralParse.parse                      1  avgt    4  13.556 ± 0.836  ns/op
TestDatatypeLiteralParse.parse                -340415  avgt    4  25.364 ± 0.111  ns/op
TestDatatypeLiteralParse.parse                0.00003  avgt    4  35.203 ± 0.335  ns/op
TestDatatypeLiteralParse.parse                   PT7M  avgt    4  16.747 ± 0.489  ns/op
TestDatatypeLiteralParse.parse                  pl-PL  avgt    4  98.848 ± 1.930  ns/op
TestDatatypeLiteralParse.parse   2021-07-01T12:00:00Z  avgt    4  65.588 ± 2.347  ns/op
```

Baseline (main 308b6e27):

```
Benchmark                       (param0_LiteralValue)  Mode  Cnt    Score   Error  Units
TestDatatypeLiteralParse.parse                 string  avgt    4    5.695 ± 0.078  ns/op
TestDatatypeLiteralParse.parse                      1  avgt    4   16.568 ± 0.222  ns/op
TestDatatypeLiteralParse.parse                -340415  avgt    4   26.902 ± 3.533  ns/op
TestDatatypeLiteralParse.parse                0.00003  avgt    4   35.531 ± 4.531  ns/op
TestDatatypeLiteralParse.parse                   PT7M  avgt    4   17.719 ± 0.265  ns/op
TestDatatypeLiteralParse.parse                  pl-PL  avgt    4  108.381 ± 4.348  ns/op
TestDatatypeLiteralParse.parse   2021-07-01T12:00:00Z  avgt    4   64.280 ± 0.884  ns/op
```

So the difference is marginal at best... but still something.

afs:
A quick look and it looks great.

"Less is more"

afs (approved):

Issue #2902, opened 2024-12-19
Switch fuseki-server.jar to being Fuseki main with admin functionality and UI


afs:
#### Fuseki main

The first step is to switch the jar file fuseki-server.jar in the Fuseki download to be Fuseki main with admin functionality and UI. This gives the same user functionality as the Jena 5.2.0 fuseki-server.jar.

* Create a Fuseki module for the server administration function. This manages the "run" area.
* Include the browser-side UI and serve the pages.
* Provide Shiro support.

There is functionality that can be extracted as modules, such as the Prometheus metrics, but later.

New functionality for fuseki-server.jar is not a goal for this issue.

### Are you interested in contributing a pull request for this task?

Yes

afs:
All the functionality of UI, administration, Shiro and Promethues metrics is in the `jena-fuseki-main` artifact.

The modules could be split out into a separate artifact. The exact nature of that artifact, and hence a good choice of name, depends on how new functionality is organised.   So, for now, the modules are in a separate package of `jena-fuskei-main`.

PR #2917, opened 2025-01-01
GH-2902: Fix: Work with Shiro-style resource names

afs:
It turns out that Shiro handles resources in its own way. A file name is "`file:<unencoded path>`", not an encoded path as in `file:` RFC IRI.

This showed up on the ASF Jenkins build because we have jobs with spaces in their names and the name is used to create the directory for the build.

----

 - [x] Tests are included.


afs:
The unit-test-and-build failure is because `jena-base` has changes and part of Fuseki depends on the change.

kinow (approved):

PR #2918, opened 2025-01-02
Suppress expected warnings

afs:
Pull request Description:

Build cleaning.

(Although Eclipse and Java disagree about suppressing "removal" warnings!)

----


afs:
The action failure seems to have been timing related in a test that I don't recall failing under JUnit4.

I am suspicious that switching to JUnit5, and the way it performs `@AfterAll` steps, is the cause. The consistent pattern is a test failing on a HTTP request with -1 (no such server) when said test fails intermittently.

PR #2925, opened 2025-01-07
GH-2924: Lateral - fixed injection for tables and enhanced QueryExec API for easier testing.

### LINKED ISSUE #2896: LATERAL: Unexpectedly unbound variable.
### Version

5.3.0-SNAPSHOT

### What happened?

I am not sure whether this is according to [SEP7](https://github.com/w3c/sparql-dev/blob/main/SEP/SEP-0007/sep-0007.md) or a bug in Jena:

I am trying to inject a value using lateral into a BIND in a subquery. The following works:
```sparql
SELECT * {
  BIND('x' AS ?xIn)
  LATERAL {
    { SELECT ?xIn ?xOut { BIND(?xIn AS ?xOut) } } # The algebraic no-op SELECT * gives the same result.
  }
}
```
The result is:

| ?xIn | ?xOut |
|------|-------|
| x    | x     |

However, if in the lateral block there is a Join, then it breaks:
```sparql
SELECT * {
  BIND('x' AS ?xIn)
  LATERAL {
    BIND("foo" AS ?foo)
    { SELECT ?xIn ?xOut { BIND(?xIn AS ?xOut) } } # SELECT * gives the same result.
  }
}
```

In the result, `?xOut` is (for me) unexpectedly unbound.
| ?xIn | ?foo | ?xOut |
|------|------|-------|
| x    | foo  |       |

I think this is a bug in Jena (but perhaps I am misunderstanding something about the scoping rules).

_No response_

Maybe

### LINKED ISSUE #2924: Issue with Nested Laterals
### Version

5.3.0-SNAPSHOT

### What happened?
The following query reveals a bug where an unexpected  exception is raised in `BindingBuilder.checkAdd`: `Attempt to reassign ?s from 1 to 1`. The issue is in `QueryIterLateral` which just adds the lhs and rhs binding to the builder although they may have variables in common.

```sparql
SELECT * {
    BIND(1 AS ?s)
    LATERAL {
        BIND(?s AS ?x)
        LATERAL {
            BIND(?s AS ?y)
        }
    }
}
```

My proposed solution in the PR is to add a check whether bindings are compatible on their common variables.

(Edit: I removed the report about the issues I encountered with var scoping so that this issue can be closed independently.)

_No response_

Yes

Aklakan:
GitHub issue resolved #2924

Pull request Description:

* Added a new test case with nested laterals.
* Fixed nested laterals by merging compatible bindings in `QueryIterLateral`
* While writing the tests, I added a few shorthands to the QueryExecBuilder API: Its now possible to write `QueryExec.dataset(...).query("SELECT ...").table()` to generally safely obtain a detached materialized in-memory table without having to worry about resource leaks or try-with-resources blocks. In the revised tests, the tables are used directly when asserting the equality of expected and actual results. In an attempt to allow for cleanly detaching BindingTDB instances, I added `Binding.detach()` with appropriate implementations. `Binding.detach` avoids copying of bindings that are already in-memory, and is thus more subtle than `BindingFactory.copy`.
* Added a `TableBuilder` class to build immutable tables.

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/algebra/TableFactory.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/algebra/TableFactory.java
@@ -22,36 +22,55 @@ import java.util.List ;
 
 import org.apache.jena.graph.Node ;
 import org.apache.jena.sparql.algebra.table.Table1 ;
+import org.apache.jena.sparql.algebra.table.TableBuilder;
 import org.apache.jena.sparql.algebra.table.TableEmpty ;
 import org.apache.jena.sparql.algebra.table.TableN ;
 import org.apache.jena.sparql.algebra.table.TableUnit ;
 import org.apache.jena.sparql.core.Var ;
 import org.apache.jena.sparql.engine.QueryIterator ;
+import org.apache.jena.sparql.engine.binding.Binding ;
+import org.apache.jena.sparql.exec.RowSet ;
 
 public class TableFactory
 {
     public static Table createUnit()
     { return new TableUnit() ; }
-    
+
     public static Table createEmpty()
     { return new TableEmpty() ; }
 
     public static Table create()
     { return new TableN() ; }
-    
+
     public static Table create(List<Var> vars)
     { return new TableN(vars) ; }
-    
+
     public static Table create(QueryIterator queryIterator)
-    { 
+    {
         if ( queryIterator.isJoinIdentity() ) {
             queryIterator.close();
             return createUnit() ;
         }
-        
-        return new TableN(queryIterator) ;
+
+        return builder().consumeRowsAndVars(queryIterator).build();
     }
 
     public static Table create(Var var, Node value)
     { return new Table1(var, value) ; }
+
+    /** Creates a table from the detached bindings of the row set. */
+    public static Table create(RowSet rs)
+    {
+        TableBuilder builder = builder();
+        builder.addVars(rs.getResultVars());
+        rs.forEach(row -> {
+            Binding b = row.detach();
+            builder.addRow(b);
+        });
+        return builder.build();
+    }
+
+    public static TableBuilder builder() {
+        return new TableBuilder();
+    }
 }
--- a/jena-arq/src/main/java/org/apache/jena/sparql/algebra/table/TableData.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/algebra/table/TableData.java
@@ -18,23 +18,21 @@
 
 package org.apache.jena.sparql.algebra.table ;
 
+import java.util.Collections;
 import java.util.List ;
 
 import org.apache.jena.sparql.ARQException ;
 import org.apache.jena.sparql.core.Var ;
 import org.apache.jena.sparql.engine.binding.Binding ;
 
+/** Immutable table. */
 public class TableData extends TableN {
     public TableData(List<Var> variables, List<Binding> rows) {
-        super(variables, rows) ;
+        super(Collections.unmodifiableList(variables), Collections.unmodifiableList(rows)) ;
     }
 
     @Override
     public void addBinding(Binding binding) {
         throw new ARQException("Can't add bindings to an existing data table") ;
-    }
-
-    public List<Binding> getRows() {
-        return rows ;
     }
 }
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/binding/BindingBase.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/binding/BindingBase.java
@@ -202,4 +202,19 @@ abstract public class BindingBase implements Binding
         }
         return hash;
     }
+
+    @Override
+    public Binding detach() {
+        Binding newParent = (parent == null) ? null : parent.detach();
+        Binding result = (newParent == parent)
+                ? detachWithOriginalParent()
+                : detachWithNewParent(newParent);
+        return result;
+    }
+
+    protected Binding detachWithOriginalParent() {
+        return this;
+    }
+
+    protected abstract Binding detachWithNewParent(Binding newParent);
 }
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterLateral.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterLateral.java
@@ -18,7 +18,6 @@
 
 package org.apache.jena.sparql.engine.iterator;
 
-import java.util.ArrayList;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
@@ -29,10 +28,11 @@ import org.apache.jena.graph.Node;
 import org.apache.jena.graph.Triple;
 import org.apache.jena.sparql.algebra.Op;
 import org.apache.jena.sparql.algebra.Table;
+import org.apache.jena.sparql.algebra.TableFactory;
 import org.apache.jena.sparql.algebra.TransformCopy;
 import org.apache.jena.sparql.algebra.op.*;
 import org.apache.jena.sparql.algebra.table.Table1;
-import org.apache.jena.sparql.algebra.table.TableN;
+import org.apache.jena.sparql.algebra.table.TableBuilder;
 import org.apache.jena.sparql.core.*;
 import org.apache.jena.sparql.engine.ExecutionContext;
 import org.apache.jena.sparql.engine.QueryIterator;
@@ -291,18 +291,25 @@ public class QueryIterLateral extends QueryIterRepeatApply {
 
             // By the assignment restriction, the binding only needs to be added to each row of the table.
             Table table = opTable.getTable();
-            // Table vars.
-            List<Var> vars = new ArrayList<>(table.getVars());
-            binding.vars().forEachRemaining(vars::add);
-            TableN table2 = new TableN(vars);
+
+            TableBuilder tableBuilder = TableFactory.builder();
+            tableBuilder.addVars(table.getVars());
+            tableBuilder.addVarsFromRow(binding);
+
             BindingBuilder builder = BindingFactory.builder();
-            table.iterator(null).forEachRemaining(row->{
+            table.iterator(null).forEachRemaining(row -> {
                 builder.reset();
                 builder.addAll(row);
-                builder.addAll(binding);
-                table2.addBinding(builder.build());
+
+                // Forcibly add the input binding - this may reassign variables.
+                // The restriction imposed by SyntaxVarScope.checkLATERAL prevents
+                // reassignment of a variable to a _different_ value.
+                binding.forEach(builder::set);
+
+                tableBuilder.addRow(builder.build());
             });
-            return OpTable.create(table2);
+            Table newTable = tableBuilder.build();
+            return OpTable.create(newTable);
         }
 
         private Triple applyReplacement(Triple triple, Function<Var, Node> replacement) {
--- a/jena-arq/src/main/java/org/apache/jena/sparql/exec/QueryExecBuilder.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/exec/QueryExecBuilder.java
@@ -25,6 +25,9 @@ import org.apache.jena.graph.Node;
 import org.apache.jena.query.ARQ;
 import org.apache.jena.query.Query;
 import org.apache.jena.query.Syntax;
+import org.apache.jena.riot.rowset.RowSetOnClose;
+import org.apache.jena.sparql.algebra.Table;
+import org.apache.jena.sparql.algebra.TableFactory;
 import org.apache.jena.sparql.core.Var;
 import org.apache.jena.sparql.engine.binding.Binding;
 import org.apache.jena.sparql.util.Context;
@@ -81,9 +84,16 @@ public interface QueryExecBuilder extends QueryExecMod {
 
     // build-and-use short cuts
 
-    /** Build and execute as a SELECT query. */
+    /**
+     * Build and execute as a SELECT query.
+     * The caller must eventually close the returned RowSet
+     * in order to free any associated resources.
+     * Use {@link #table()} to obtain an independent in-memory copy of the row set.
+     */
     public default RowSet select() {
-        return build().select();
+        QueryExec qExec = build();
+        RowSet core = qExec.select();
+        return new RowSetOnClose(core, qExec::close);
     }
 
     /** Build and execute as a CONSTRUCT query. */
@@ -105,5 +115,19 @@ public interface QueryExecBuilder extends QueryExecMod {
         try ( QueryExec qExec = build() ) {
             return qExec.ask();
         }
+    }
+
+    /**
+     * Build and execute as a SELECT query.
+     * Creates and returns an independent in-memory table by materializing the underlying row set.
+     * Subsequently, {@link Table#toRowSet()} can be used to obtain a fresh row set view over the table.
+     */
+    public default Table table() {
+        Table result;
+        try (QueryExec qExec = build()) {
+            RowSet rowSet = qExec.select();
+            result = TableFactory.create(rowSet);
+        }
+        return result;
     }
 }
--- a/jena-arq/src/test/java/org/apache/jena/sparql/graph/GraphsTests.java
+++ b/jena-arq/src/test/java/org/apache/jena/sparql/graph/GraphsTests.java
@@ -28,8 +28,12 @@ import org.apache.jena.graph.Triple ;
 import org.apache.jena.query.* ;
 import org.apache.jena.rdf.model.Model ;
 import org.apache.jena.rdf.model.ModelFactory ;
+import org.apache.jena.sparql.algebra.Table ;
 import org.apache.jena.sparql.core.Quad ;
+import org.apache.jena.sparql.exec.QueryExec ;
+import org.apache.jena.sparql.exec.QueryExecBuilder ;
 import org.apache.jena.sparql.sse.SSE ;
+import org.apache.jena.system.Txn ;
 import org.junit.Test ;
 
 /** Test API use of models, including some union graph cases : see also DatasetGraphTests */
@@ -40,12 +44,12 @@ public abstract class GraphsTests
     protected static final String graph1 = "http://example/g1" ;
     protected static final String graph2 = "http://example/g2" ;
     protected static final String graph3 = "http://example/g3" ;
-    
+
     private Dataset dataset ;
     private Model calcUnion = ModelFactory.createDefaultModel() ;
 
     protected abstract Dataset createDataset() ;
-    
+
     protected Dataset getDataset()
     {
         if ( dataset == null )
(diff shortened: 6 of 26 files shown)


jena-arq/src/main/java/org/apache/jena/query/Query.java
```diff
@@ -98,7 +98,7 @@ public class Query extends Prologue implements Cloneable, Printable
     public static final int ORDER_UNKNOW              = -3 ;
 
     // VALUES trailing clause
-    protected TableData valuesDataBlock = null ;
+    protected TableN valuesDataBlock = null ;
```
Aklakan, line 101: Should the valuesDataBlock be mutable or not? One test case written by me some years ago assumed it to be mutable - but the implementation of TableData suggests to be immutable as the addBinding method is overridden to reject updates.

As I see it: TableN -> mutable, TableData -> immutable - much like guava's List and ImmutableList.

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterLateral.java
```diff
@@ -292,17 +307,33 @@ public Op transform(OpTable opTable) {
...
+                }
+            });
+
+            List<Binding> bindings = new ArrayList<>();
             BindingBuilder builder = BindingFactory.builder();
             table.iterator(null).forEachRemaining(row->{
-                builder.reset();
-                builder.addAll(row);
-                builder.addAll(binding);
-                table2.addBinding(builder.build());
```
Aklakan, line 303: addBinding checks whether the table's variables need to be updated; this check is not needed here.

jena-arq/src/main/java/org/apache/jena/sparql/algebra/table/TableData.java
```diff
@@ -18,23 +18,21 @@
...
 
 import org.apache.jena.sparql.ARQException ;
 import org.apache.jena.sparql.core.Var ;
 import org.apache.jena.sparql.engine.binding.Binding ;
 
+/** Immutable table. */
 public class TableData extends TableN {
     public TableData(List<Var> variables, List<Binding> rows) {
-        super(variables, rows) ;
+        super(Collections.unmodifiableList(variables), Collections.unmodifiableList(rows)) ;
```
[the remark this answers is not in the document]
afs: `Collections.unmodifiableList` only puts a wrapper around the base list. The base `variables` and `rows` can be modified.

jena-arq/src/main/java/org/apache/jena/query/ResultSet.java
```diff
@@ -103,4 +103,8 @@ public default ResultSet materialise() {
     }
 
     public void close();
+
+    default RowSet toRowSet() {
```
afs, line 107: I prefer `asRowSet` -- `as*` when it is the  same thing, different presentation, and `to*` when it is a created conversion.

c.f. `Dataset::asDatasetGraph`

jena-arq/src/main/java/org/apache/jena/sparql/algebra/TableFactory.java
```diff
@@ -27,31 +28,45 @@
...
-        
+
         return new TableN(queryIterator) ;
     }
 
     public static Table create(Var var, Node value)
     { return new Table1(var, value) ; }
+
+    /** Creates a mutable table from the detached bindings of the row set. */
+    public static Table create(RowSet rs)
```
afs, line 62: I don't understand what going on. Could you explain this please? 

This seems to duplicate the intent of `RowSet.materialize` and `RowSet.rewindable`.

jena-arq/src/main/java/org/apache/jena/sparql/engine/binding/BindingBase.java
```diff
@@ -202,4 +202,19 @@ public static int hashCode(Binding bind) {
         }
         return hash;
     }
+
+    @Override
+    public Binding detach() {
+        Binding newParent = parent == null ? null : parent.detach();
+        Binding result = newParent == parent
+                ? detachWithOriginalParent()
+                : detachWithNewParent(newParent);
```
afs, line 211:
```suggestion
        Binding newParent = (parent == null) ? null : parent.detach();
        Binding result = (newParent == parent)
                ? detachWithOriginalParent()
                : detachWithNewParent(newParent);
```

jena-arq/src/main/java/org/apache/jena/sparql/engine/binding/BindingBase.java
```diff
@@ -202,4 +202,19 @@ public static int hashCode(Binding bind) {
...
+        return result;
+    }
+
+    protected Binding detachWithOriginalParent() {
+        return this;
+    }
+
+    protected abstract Binding detachWithNewParent(Binding newParent);
```
afs, line 219: This would be nicer in `BindingFactory` ... if we were using Java21 and could use switch patterns. Oh well, another time.

jena-arq/src/main/java/org/apache/jena/sparql/exec/QueryExec.java
```diff
@@ -54,6 +54,12 @@ public static QueryExecBuilder graph(Graph graph) {
         return QueryExecDatasetBuilder.create().graph(graph);
     }
 
+    /** Create a {@link QueryExecBuilder} initialized with an empty dataset. */
+    public static QueryExecBuilder emptyDataset() {
```
afs, line 58: This is only used from `TestQueryExecution`. Please can the code go there so it is not in the runtime API.? 

It can use`DatasetGraphFactory.empty()`.

```java
return QueryExec.dataset(DatasetGraphFactory.empty()`);
```

jena-arq/src/main/java/org/apache/jena/sparql/exec/QueryExecBuilder.java
```diff
@@ -106,4 +116,18 @@ public default boolean ask() {
             return qExec.ask();
         }
     }
+
+    /**
+     * Build and execute as a SELECT query.
+     * Creates and returns an independent in-memory table by materializing the underlying row set.
+     * Subsequently, {@link Table#toRowSet()} can be used to obtain a fresh row set view over the table.
+     */
+    public default Table table() {
```
afs, line 125: See the question about "RowSet.matrialize/rewindable" above.

jena-arq/src/main/java/org/apache/jena/sparql/exec/RowSet.java
```diff
@@ -93,4 +93,8 @@ public default Stream<Binding> stream() {
     public static RowSet create(QueryIterator qIter, List<Var> vars) {
         return RowSetStream.create(vars, qIter);
     }
+
+    default ResultSet toResultSet() {
```
afs, line 97:
```suggestion
    default ResultSet asResultSet() {
```

jena-arq/src/test/java/org/apache/jena/sparql/graph/GraphsTests.java
```diff
@@ -106,56 +110,80 @@ protected void fillDataset(Dataset dataset) {
...
+                Txn.executeWrite(ds, () -> fillDataset(ds)) ;
+                actual = Txn.calculateRead(ds, () -> QueryExec.dataset(ds.asDatasetGraph()).query(queryString).table()) ;
+            } else {
+                fillDataset(ds) ;
+                actual = QueryExec.dataset(ds.asDatasetGraph()).query(queryString).table() ;
+            }
+        } finally {
+            ds.close() ;
+        }
+        assertEquals(expected, actual) ;
```
afs, line 148: See `ResultSetCompare.exactEquals` (which should probably be in `RowSetOps`).

afs:
First part of a review - trying to understand the machinery changes. There seem to be a lot of them and I'm trying to understand what are necessary.

jena-arq/src/main/java/org/apache/jena/sparql/algebra/TableFactory.java
Aklakan: The difference (to my understanding) is that a Table acts as a Collection and a RowSet as an Iterator.
A Table can thus be seen as a factory for RowSets via `Table.toRowSet()` - each obtained RowSet is an independent iterator over the Table.
My qualms with RowSet.materialize/rewindable is that these methods operate on the iterator level; it seemed clean to me to add Table as a collection-level view over a list of bindings.

jena-arq/src/main/java/org/apache/jena/sparql/exec/QueryExec.java
Aklakan: It was just a suggestion for a shortcut - I'll remove it.

jena-arq/src/main/java/org/apache/jena/query/Query.java
Aklakan: Ok, so I'll change back to TableData and update the test case to expect failure when trying to modify the rows of a queries valuesDataBlock directly.

jena-arq/src/main/java/org/apache/jena/sparql/algebra/table/TableData.java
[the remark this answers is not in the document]
Aklakan: The constructor already existed. The wrappers prevent mutation via `TableN.getRows().add()` which wasn't the case before. Its not perfect, but from your comment

> It should be "immutable once built"

its a step closer towards `TableData` being immutable.

jena-arq/src/main/java/org/apache/jena/sparql/engine/binding/BindingBase.java
Aklakan: I was also thinking that BindingFactory could take care of making bindings independent. But then again, why should BindingFactory have special logic for BindingTDB instances? In principle, `BindingTDB.detach` could be implemented as just setting filling the already present (but unused) cache - so no copy would be needed then.

jena-arq/src/test/java/org/apache/jena/sparql/graph/GraphsTests.java
Aklakan: As mentioned, I view Table as a collection and RowSet as an Iterator.
If I am not mistaken, then table's `equals()` method naturally and succinctly performs an exact comparison (using term equality). For working with iterators (RowSet/ResultSet) one should of course use ResultSetCompare / RowSetOps methods.
In this case, one could also write a bit more verbose: `assertTrue(ResultSetCompare.exactEquals(expected.toRowSet(), actual.toRowSet());`

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterLateral.java
```diff
@@ -292,17 +295,33 @@ public Op transform(OpTable opTable) {
...
+            List<Var> tableVars = table.getVars();
+            List<Var> vars = new ArrayList<>(tableVars);
+
+            // Track variables that appear both in the table and the binding.
+            List<Var> commonVars = new ArrayList<>();
+
+            // Index variables in a set if there are more than a few of them.
+            Collection<Var> tableVarsIndex = tableVars.size() > 4 ? new HashSet<>(tableVars) : tableVars;
+            binding.vars().forEachRemaining(v -> {
+                if (tableVarsIndex.contains(v)) {
```
afs, line 307: The table vars can't contain a binding var. That case is caught by the 

https://github.com/w3c/sparql-dev/blob/main/SEP/SEP-0007/sep-0007.md  
"Disallow syntactic forms that set variables that may already be present in the current row."

It's tested in `SyntaxVarScope.checkLATEAL`.

(I removed the `if`, left the `else` and the second test passes). 

If so, it simplifies to `vars.add(v);`
And hence "vars = tableVars" (copy not needed).

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterLateral.java
```diff
@@ -292,17 +295,33 @@ public Op transform(OpTable opTable) {
...
+            });
+
+            List<Binding> bindings = new ArrayList<>(table.size());
             BindingBuilder builder = BindingFactory.builder();
             table.iterator(null).forEachRemaining(row->{
-                builder.reset();
-                builder.addAll(row);
-                builder.addAll(binding);
-                table2.addBinding(builder.build());
+                if (Algebra.compatible(row, binding, commonVars.iterator())) {
```
afs, line 317: If there is a variable in common from substitution it must be the same term.

So this is a contains relationship. (Given the comments above there may be a simpler way.)

jena-arq/src/main/java/org/apache/jena/query/Query.java
Aklakan: Query has been reverted and the test case has been updated.

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterLateral.java
Aklakan: There is potentially another can of worms I was about to open w.r.t. sep7:

Perhaps its worthy of opening an issue and discussion on sparql-dev.  The reason I added the compatibility check was because I had the following query in mind, which for reasons I do not understand is currently prohibited:

```sparql
SELECT * { VALUES ?department {
    <urn:dept1>
    <urn:dept2>
  }
  LATERAL { SELECT * { VALUES (?department ?employee) {
        ( <urn:dept1> <urn:person1> )
        ( <urn:dept1> <urn:person2> )
        ( <urn:dept2> <urn:person1> )
      }
    } ORDER BY ?employee LIMIT 1
  }
}
```

| department  | employee      |
|--------------------|---------|
| <urn:dept1> | <urn:person1> |
| <urn:dept2> | <urn:person1> |
-------------------------------

Removing the restrictions `SyntaxVarScope.checkLATERAL` on `Element_Bind` and `Element_Data`) and adding the check for compatible bindings (as done in the PR) naturally makes the query work.

jena-arq/src/main/java/org/apache/jena/query/ResultSet.java
Aklakan: Renamed to `asRowSet`.

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterLateral.java
afs: There must be a problem with `SyntaxVarScope.checkLATERAL` because the `SELECT *` is effectively `SELECT ?employee` which works.

Aklakan: > `SELECT *` is effectively `SELECT ?employee` which works.

Hm, depending on the viewpoint it is the effective query *after* substitution, but before that, its `SELECT ?department ?employee` as to make `?department` in-scope such that the substitution could affect the table and discard its incompatible rows.

I think the scope error is in-line with the SEP7 statement:

> "Disallow syntactic forms that set variables that may already be present in the current row."

But I think that this is too restrictive, because it prevents inlining data using VALUES blocks. IMO if substitution leads to incompatible bindings then they should just be discarded as usual - no? Perhaps you have corner cases in mind where under this logic the substitution could become ambiguous and that this might be the reason for the restrictions.

Removing the restrictions on `Element_Bind` and `Element_Data` restrictions in `checkLATERAL` in order to make the above "department-employee" query work causes the following test cases to fail:

 syntax-lateral-bad-{[04](https://github.com/apache/jena/blob/main/jena-arq/testing/ARQ/Syntax/Syntax-Lateral/syntax-lateral-bad-04), [05](https://github.com/apache/jena/blob/main/jena-arq/testing/ARQ/Syntax/Syntax-Lateral/syntax-lateral-bad-05), [06](https://github.com/apache/jena/blob/main/jena-arq/testing/ARQ/Syntax/Syntax-Lateral/syntax-lateral-bad-06), [08](https://github.com/apache/jena/blob/main/jena-arq/testing/ARQ/Syntax/Syntax-Lateral/syntax-lateral-bad-08)}.arq

For example, `syntax-lateral-bad-04`:
```sparql
SELECT * {
   ?s ?p ?o
   LATERAL { BIND( 123 AS ?o) }
}
```

This is certainly not an efficient way to write a query, but I don't understand why the query should be syntactically illegal in the first place. To me it looks like it would only retain bindings where ?o = 123. So it would be the task of an optimizer to rewrite the `Lateral/Bind` to a plain `FILTER(?o = 123)`.

When disabling the SyntaxVarScope check for ElementBind, each ?o gets injected into the unit table beneath `(extend ((?o 123))  (table unit))` such as `(extend ((?o 123))  (table (row (?o 456))))` and the evaluation of OpExtend discards the incompatible binding - which seems like the natural way it should be to me.

afs: > Hm, depending on the viewpoint it is the effective query _after_ substitution,

My bad - I was thinking about `GROUP BY`.

For `ORDER BY`, the query is illegal by the assignment restriction.

> To me it looks like it would only retain bindings where ?o = 123.

That would turn `BIND` into a filter. (ARQ's extension `LET` does that.)

We have to have a restriction of some kind. The SEP-0007 design makes it a compile-time condition (test once per query).

In general, some system have optimizations that on the static analysis of the query. Making the RHS of a LATERAL have runtime variance would interfere.

afs: Sorry, I have to get Jena 5.3.0 out without this.

This issue is important and something that is, I hope, going to get more importance in the SPARQL community.

[WG sparql-query PR 177](https://github.com/w3c/sparql-query/pull/177) at least gets the essential machinery into the resp with the WG contribution licnesing. A small step, but a step.

Aklakan: > In general, some system have optimizations that on the static analysis of the query.
Well, the tradeoff to what extend existing implementations need to drive restrictions of new features in a standard.

For the sake of closing the issue, should I modify/simplify the logic in QueryIterLateral or should I leave the slightly more general approach as it is for now? Essentially, the var scope check should prevent non-empty sets of commonVars.

Aklakan: > Sorry, I have to get Jena 5.3.0 out without this.

No worries, whenever it's ready.

Aklakan:
Just to summarize: The main open point for this PR is whether to remove the `Algebra.compatible` check on the `commonVars` under the assumption that the bindings have to be compatible anyway (because of `SyntaxVarScope.checkLATERAL`).

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterLateral.java
```diff
@@ -291,18 +291,25 @@ public Op transform(OpTable opTable) {
 
             // By the assignment restriction, the binding only needs to be added to each row of the table.
             Table table = opTable.getTable();
-            // Table vars.
-            List<Var> vars = new ArrayList<>(table.getVars());
-            binding.vars().forEachRemaining(vars::add);
-            TableN table2 = new TableN(vars);
+
+            TableBuilder tableBuilder = TableFactory.builder();
```
Aklakan, line 295: Based on the discussion, I now added a TableBuilder class for building immutable Tables.

Aklakan:
I added a TableBuilder class for building immutable tables.
In accordance with the discussion, in `QueryIterLateral` I removed the algebra compatibility check and updated the table creation to use the new TableBuilder. Migrating existing table creation would be a future endeavor.

afs:
I'd mostly finished reviewing - does the TableBuilder make a difference?

Let me know when this is ready.

The main issues are to do with the formal description of LATERAL, and not the core in this PR which is faithful to it the formal description. A deeper reworking of execution could so better but it would also change other places.

Functionality before optimization!

Aklakan:
> I'd mostly finished reviewing - does the TableBuilder make a difference?

It's not a big difference - It exists now because you suggested that this might be the better approach - and I used it in the fixed code.

Other than that its finished.

afs (approved):

PR #2970, opened 2025-01-22
cypress-run to depend on a maven build

afs:
This PR does two things:

* `cyrpress-run` is made to run after the build job. It appears that the jobs run in parallel at the moment - I see the Cypress runs finish before the build step has. The Cypress job is quick.

* Change the build step to do a "dev" build. Currently, it does not build from the start of the Jena build. If there has been a change in earlier dependency, then the Fuseki steps fail.

If this become smooth, we should consider combining into a PR workflow.

----

### DIFF
--- a/.github/workflows/ui.yml
+++ b/.github/workflows/ui.yml
@@ -17,7 +17,7 @@ jobs:
     if: github.repository == 'apache/jena'
 
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 5
+    timeout-minutes: 15
 
     strategy:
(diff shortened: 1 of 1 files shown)


afs:
Bumping the timeout on the maven step ...

kinow (approved):
Good idea to run cypress after unit/build, that should avoid some unnecessary e2e tests. +1 :tada:

afs:
The fun thing is that it is the updated workflow in the PR is the one run for the PR.

PR #2971, opened 2025-01-22
Format string error in LangTags

hoijui:
`%c` requires a `char`, but a `String` was supplied.

For more details, see here:
<https://docs.oracle.com/javase/8/docs/api/java/util/Formatter.html#syntax>

----

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/graph/langtag/LangTags.java
+++ b/jena-core/src/main/java/org/apache/jena/graph/langtag/LangTags.java
@@ -113,7 +113,7 @@ public class LangTags {
                 sb.setLength(0);
                 continue;
             }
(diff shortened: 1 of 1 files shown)


afs:
Hi @hoijui 

The function `str(char)` takes the `char` and returns a string of the quoted character together with its Unicode codepoint.

```java
  private static String str(char ch) {
        return String.format("'%s' U+%04X", Character.valueOf(ch), (int)ch);
    }
```
so the bug is that the `%c` should be `%s`.

hoijui:
Ahh thank you!
I was wondering which of the two ways to change this would be right. should have had a look at `str(char)`.
Opened [a new PR](https://github.com/apache/jena/pull/2972).

afs:
No problem - I had to check  the code to work out how it got in that state. Putting in a formatted string with codepoint, not `%c` is a late addition.

PR #3010, opened 2025-02-14
Cleanup node warnings in build

OyvindLGjesdal:
GitHub issue resolved #

Pull request Description:

Going through node warnings from the GH-action build.

Changing legacy node-sass api to modern and adding peer dependencies solves the warnings. while the other changes just silences them (1250kb as a bundle limit sounds huge to me). Since the largest number of SASS warnings are inherited from bootstrap, I thought it could make sense to hide them. I am more unsure if it makes sense to up the bundle limit for less noise in the build output (since it looks like an issue).

----

 - [ ] Tests are included.

----

### DIFF
--- a/jena-fuseki2/jena-fuseki-ui/package.json
+++ b/jena-fuseki2/jena-fuseki-ui/package.json
@@ -39,6 +39,8 @@
     "vue-upload-component": "^3.1.8"
   },
   "devDependencies": {
+    "@babel/core": "^7.0.1",
+    "@babel/preset-env": "^7.0.0",
     "@cypress/code-coverage": "^3.12.4",
     "@cypress/vue": "^6.0.0",
     "@eslint/js": "^9.2.0",
@@ -48,6 +50,7 @@
     "@vue/compiler-sfc": "^3.3.6",
     "@vue/eslint-config-standard": "^8.0.1",
     "@vue/test-utils": "^2.4.1",
+    "babel-loader": "^9",
     "concurrently": "^9.0.0",
     "cross-env": "^7.0.3",
     "cypress": "^14.0.0",
@@ -70,7 +73,8 @@
     "vite": "^6.0.11",
     "vite-plugin-istanbul": "^7.0.0",
     "vitest": "^3.0.5",
-    "wait-on": "^8.0.0"
+    "wait-on": "^8.0.0",
+    "webpack": "^5"
   },
   "resolutions": {},
   "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
--- a/jena-fuseki2/jena-fuseki-ui/package.json
+++ b/jena-fuseki2/jena-fuseki-ui/package.json
@@ -65,14 +65,13 @@
     "nanoid": "^5.0.9",
     "nodemon": "^3.0.1",
     "run-script-os": "^1.1.6",
-    "sass": "^1.69.4",
+    "sass-embedded": "^1.85.0",
     "sinon": "^19.0.2",
     "vite": "^6.0.11",
     "vite-plugin-istanbul": "^7.0.0",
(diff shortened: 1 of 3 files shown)


afs (approved):

jena-fuseki2/jena-fuseki-ui/package.json
```diff
@@ -65,14 +68,14 @@
     "nanoid": "^5.0.9",
     "nodemon": "^3.0.1",
     "run-script-os": "^1.1.6",
-    "sass": "^1.69.4",
+    "sass-embedded": "^1.85.0",
```
kinow, line 71: :+1:

jena-fuseki2/jena-fuseki-ui/vite.config.js
```diff
@@ -48,8 +56,8 @@ export default defineConfig({
     },
   },
   build: {
-    // Our largest chunk: dist/assets/yasqe.min-ec8f4984.js 508.16 kB │ gzip: 130.97 kB
-    chunkSizeWarningLimit: 550,
+    // Our largest chunk: target/webapp/static/Query-CakHSd_3.js  1,172.48 kB │ gzip: 350.95 kB
+    chunkSizeWarningLimit: 1250,
```
kinow, line 60: I haven't looked at the code in a while, but I remember the warnings from s/css and chunk size, but I never had time to check each and see if we had a fix or if we had to simply suppress.

Can't tell whether it's better to simply suppress or leave the warnings in the build log. I'd leave them so I don't forget to fix it later, but if that is an issue for others actively maintaining it then it's probably better to suppress these.

Thanks!

kinow (approved):

kinow:
And about the deps, I think sass-embedded is a good choice, and the other ones (IIRC) were transitive dependencies and somewhere in the log it would say that we were using it but not declaring in `package.json`. Thanks for fixing those.

jena-fuseki2/jena-fuseki-ui/vite.config.js
OyvindLGjesdal: Hi @kinow and thanks for the review

> . I'd leave them so I don't forget to fix it later, but if that is an issue for others actively maintaining it then it's probably better to suppress these.

I think there are two different issues:
* The chunk size warning I think is probably better to leave in the logs, so it is kept visible.
* the scss (44/60 warnings in the build) warnings I *believe* are mostly inherited from Bootstrap. There are some scss errors which are local to the Jena Fuseki UI, but most of the warnings are for rules within node_modules which are native to Bootstrap. Waiting (creating an issue?) for the Bootstrap patch to complete and then (fixing the Jena warnings/attempting migration) also make senses, but I silenced it due to it mostly being external and because of the large number of warnings.

Attempting to use the helper for migration (https://sass-lang.com/documentation/breaking-changes/import/) complains about bootstrap, and exits before changing anything.

Hopefully this will be fixed upstream:

```
# after minor fixes to import paths in index.scss to relative, and adding a css file extension:
sass-migrator --migrate-deps --load-path . --dry-run module --migrate-deps src/styles/index.scss --dry-run
Error: The migrator wants to rename a member in node_modules/bootstrap/scss/_functions.scss, but it is not being migrated. You should re-run the migrator with --migrate-deps or with node_modules/bootstrap/scss/_functions.scss as one of your entrypoints.
```

I'm ok with updating to either, I am unsure of if the warning silencings are helpful or not, but got a bit caught in the moment and tried to get towards 0 warnings in the build 😎

afs: I think dart-sccs / Bootstrap is targetted for Bootstrap6 ... which is not out yet.
https://github.com/twbs/bootstrap/issues/40849#issuecomment-2367274006

Much as I dislike hiding warnings, this may be the better way for jena-fuseki-ui because there are so many lines of output.

kinow: No objections for that then., and maybe +1 for the chunk size warning (although from time to time I used to use that vue/vite embedded helper app to visualize metrics for the app, and there was a tab/view that shows that too -- so no objection if we suppress that too).

PR #3019, opened 2025-02-18
GH-3018: Report the offending object in UniqueLangConstraint

### LINKED ISSUE #3018: Missing Value Node in `sh:uniqueLang` Constraint Validation Report
### Version

5.3.0

#### **Issue Description**
When the `sh:uniqueLang` constraint is violated, the validation report does not include the duplicated value node. This makes it difficult to identify which specific values caused the violation.

#### **Expected Behavior**
The validation report should provide the duplicated value node, allowing users to pinpoint the exact issue in their data.

#### **Steps to Reproduce**
1. Define a SHACL shape with `sh:uniqueLang true`.
2. Validate data that contains multiple values with the same language tag for a property.
3. Observe the validation report.

#### **Actual Behavior**
- The validation report states that the constraint was violated but does not indicate the specific duplicated value node.

#### **Expected Behavior**
- The validation report should include the exact value nodes that caused the violation.

##### **SHACL Shape**
```turtle
ex:BuildingShape a sh:NodeShape ;
    sh:targetClass ex:Building ;
    sh:property [
        sh:path ex:label ;
        sh:datatype rdf:langString ;
        sh:uniqueLang true ;
        sh:message "Each building label should have only one value per language." ;
    ] .
```

##### **Invalid Data**
```turtle
ex:Building1 a ex:Building ;
    ex:label "Building"@en , "Edifice"@en , "Gebäude"@de .
```

```shell

```

Yes

aliariff:
GitHub issue resolved #3018

Pull request Description:

Report the offending object in `UniqueLangConstraint`

----

 - [ ] Tests are included.

----

### DIFF
--- a/jena-shacl/src/test/files/std/core/property/uniqueLang-001.ttl
+++ b/jena-shacl/src/test/files/std/core/property/uniqueLang-001.ttl
@@ -64,6 +64,7 @@ ex:ValidInstance1
           sh:resultSeverity sh:Violation ;
           sh:sourceConstraintComponent sh:UniqueLangConstraintComponent ;
           sh:sourceShape ex:TestShape-testProperty ;
+          sh:value "Myself"@en ;
         ] ;
       sh:result [
           rdf:type sh:ValidationResult ;
@@ -72,6 +73,7 @@ ex:ValidInstance1
           sh:resultSeverity sh:Violation ;
           sh:sourceConstraintComponent sh:UniqueLangConstraintComponent ;
           sh:sourceShape ex:TestShape-testProperty ;
+          sh:value "I"@en ;
         ] ;
       sh:result [
           rdf:type sh:ValidationResult ;
@@ -80,6 +82,7 @@ ex:ValidInstance1
           sh:resultSeverity sh:Violation ;
           sh:sourceConstraintComponent sh:UniqueLangConstraintComponent ;
           sh:sourceShape ex:TestShape-testProperty ;
+          sh:value "Ich"@de ;
         ] ;
     ] ;
   mf:status sht:approved ;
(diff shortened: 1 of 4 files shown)


jena-shacl/src/test/files/std/core/property/uniqueLang-001.ttl
afs, line 1: There's a problem with this test because the validation report entry is about the first failure - the second node encountered with the same language tag. But there is no definite order from the "I", or "Me" nodes being tested and that means different validation report entries

I get about 50% failures on github CI.

I'm not sure what to do about this. In the test runner, there is way to choose which properties to test of equivalent of validation reports and that includes sh:values but that is useful in other cases.

NB `files/std/core/*` are direct copies from the SHACL test suite.

afs:
The only way I can think of is to have a "test mode" flag./ The flag would be an enum of two elements: "NORMAL" and "TEST". Horrible but changing the system for this one case is also horrible.

The flag could be a static in `ShaclSystem`. When set to TEST, the old way (to pass a null) is used and when not set (normal mode), the `obj`. 

The changes to be test from the WG test suite aren't needed.

aliariff:
> The only way I can think of is to have a "test mode" flag./ The flag would be an enum of two elements: "NORMAL" and "TEST". Horrible but changing the system for this one case is also horrible.
> 
> The flag could be a static in `ShaclSystem`. When set to TEST, the old way (to pass a null) is used and when not set (normal mode), the `obj`.
> 
> The changes to be test from the WG test suite aren't needed.

Hi @afs 

I have updated the PR as you suggested. Please take a look.

afs:
Thank you!

afs (approved):

PR #3027, opened 2025-02-21
GH-3026: spatial index per graph and kryo serialization

### LINKED ISSUE #3026: Spatial Index improvements (index-per-graph + kryo)
### Version

5.4.0-SNAPSHOT

### Feature

This proposal is to enhance the spatial index with support for index-per-graph as well as to improve its serialization using kryo - via Apache Sedona's kryo/jts implementation.

This is an incremental improvement of the existing in-memory implementation - its not a complete overhaul such as a disk-based incrementally updated transaction-aware R-tree (if someone contributed that than this issue's PR could be discarded 😄 ).

The impact of this work have been evaluated and presented at the GeoLD workshop last year [proceedings](https://i3mainz.github.io/GeoLD2024/#proceedings):

Simon Bin, Claus Stadler, Lorenz Bühmann, and Michael Martin
[Getting practical with GeoSPARQL and Apache Jena](http://ceur-ws.org/Vol-3743/paper2.pdf)
[Slides](https://i3mainz.github.io/GeoLD2024/presentations/paper2_presentation.pdf)

The essence is presented on the following slides:

Using a named graph per index (unsurprisingly) boots the performance when multiple graphs have geometries and only a subset is queried (slide 15):

Image

As for serialization performance (slide 16), while index building became a bit slower, this is outweighed by near-instant loading of the spatial index:

Image

A new `geosparql:indexPerGraph` option (boolean) is added to the `geosparql:GeosparqlDataset` assembler.

The implementation has been mainly done by @LorenzBuehmann.

As for compatibility, I need to check for whether it is backward compatible but I think due to the change of the serializer, existing spatial indexes would have to be rebuilt.

Yes

Aklakan:
GitHub issue resolved #3026

Pull request Description: Adds support for spatial index per graph and kryo serialization.

- Updated `jena-geosparql` with spatial index-per-graph `(STRtreePerGraph.java)` structure and kryo serialization.
- The upgrade should be backward compatible: If reading of a spatial index file fails then the old (slow) deserialization method will be attempted.
- Added `jena-fuseki-mod-geosparql` module under `jena-fuseki2`. This module comes with a simple HTML page where users can pick which graphs to update in the index:
  image

- Added `jena-fuseki-mod-geosparql` to `jena-fuseki-server`. This bundles the fmod (and geosparql) with Fuseki, but `--modules=true` must still be set to enable it!

----

 - [x] Additional tests for the `GeoSPARQL` module are included.
 - [x] Tests for the `spatial-indexer fuseki module` and its UI are included. Note, that the test uses Selenium to click a couple of buttons in the browser which requires a local Chrome. Therefore the test is set to ignore by default.
 - [x] Benchmark setup is included.
 - [x] Documentation change and updates are provided for the [Apache Jena website](https://github.com/apache/jena-site/) (PR: https://github.com/apache/jena-site/pull/213)

----


jena-geosparql/pom.xml
```diff
@@ -98,10 +147,12 @@
 
     <dependency>
       <groupId>org.apache.logging.log4j</groupId>
-      <artifactId>log4j-slf4j2-impl</artifactId>
+      <artifactId>log4j-slf4j-impl</artifactId>
```
afs, line 150: `log4j-slf4j2-impl` is for slf4j v2.x.x, `log4j-slf4j-impl` is for slf4j v1.x.x

Jena uses slf4j v2.x.x

afs:
Build failure: https://github.com/apache/jena/actions/runs/13534795897/job/37839302002

Aklakan:
It should build now. Its about 90% done. One issue is, that @LorenzBuehmann used `ShapeSerde` from `sedona-spark` which actually only does JTS geometry de-/serializer. Probably it is faster then other approaches (otherwise it wouldn't exist) - but causes enforcer conflicts and draws in quite a lot of dependencies for what it does.

Right now I switched to JTS native WKB serialization which makes the build work normally.
I am waiting for input from Lorenz about the performance differences between the serializers.

I also added a little compatibility layer for the old index IO which used java serialization. So old index files should continue to load. Saving indexes will use the now serializer.

Aklakan:
About performance differences, @SimonBin found this article: https://kontinuation.one/posts/2022/12/improving-geometry-serde-performance-of-apache-sedona/

image

This suggests that Sedona's ShapeSerde is generally a lot faster. Not sure if the current implementation is perhaps already even one of the optimized variants.

Aklakan:
I am a bit worried that a hard dependency on sedona-spark may cause too many issues, so I am trying to decouple the geometry serde (serializer/deserializer) from the rest of the spatial-index serde.

For this, I added a plain text JSON header field to the spatial index format with a field for the geometry serde. This field can be read and used to configure the remaining kryo serde. The index reader makes a lookup with Class.forName() and tries to create an instance via the default constructor.

With this approach it would be possible to use the slower JTS-based geometry serialization for now and have a compatible upgrade path to a faster approach at a later point.

This snippet is how the header is set up:

```java
// SpatialIndexIoKryo.java

    public static void writeToOutputStream(OutputStream os, SpatialIndexPerGraph index) {
        // geometrySerde could later be switched to sedona's ShapeSerde.
        // (Need to check whether I can readily reuse an interface from JTS/Sedona instead of my own GeometrySerdeAdapter)
        GeometrySerdeAdapter geometrySerde = new GeometrySerdeAdapterJtsWkb(); 

        JsonObject header = new JsonObject();
        header.addProperty("type", "jena-spatial-index");
        header.addProperty("version", "2.0.0");
        header.addProperty("geometrySerde", geometrySerde.getClass().getName());

        Kryo kryo = new Kryo();
        JtsKryoRegistrator.registerClasses(kryo, geometrySerde);
        try (Output output = new Output(os)) {
            writeJson(output, header);
            writeIndex(kryo, output, index);
            output.flush();
        }
```

jena-geosparql/src/main/java/org/locationtech/jts/index/strtree/CustomIndexSerdeV2.java
```diff
@@ -0,0 +1,156 @@
...
+
+import org.apache.jena.geosparql.spatial.index.v2.serde.sedona.CustomGeometrySerde;
+import org.locationtech.jts.geom.Envelope;
+
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+
+/* This file is an adapted copy of org.locationtech.jts.index.strtree.IndexSerde. */
+/* The change is, that geometry read/write is delegated to a GeometrySerdeAdapter instance. */
```
Aklakan, line 32: This class is copy&adapt form sedona-common 1.7.0 (latest)- they actually used this package hack to access non-visible stuff. in the long run, jts should fix this design.

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/index/v2/serde/sedona/CustomGeometrySerde.java
```diff
@@ -0,0 +1,209 @@
...
+import org.locationtech.jts.geom.Polygon;
+
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.Registration;
+import com.esotericsoftware.kryo.Serializer;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+
+/* This file is an adapted copy of apache-sedona org.apache.sedona.common.geometrySerde.GeometrySerde */
+/* The change is, that geometry read/write is delegated to a GeometrySerdeAdapter instance. */
```
Aklakan, line 38: This file is a copy from sedona-common.

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/index/v2/serde/sedona/CustomSpatialIndexSerde.java
```diff
@@ -0,0 +1,119 @@
...
+import org.locationtech.jts.index.strtree.STRtree;
+
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.Serializer;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+
+/* This file is an adapter copy of apache-sedona org.apache.sedona.common.geometrySerde.SpatialIndexSerde */
+/* The change is, that geometry read/write is delegated to a GeometrySerdeAdapter instance. */
+
```
Aklakan, line 33: This file is a copy&adapt from sedona-common.

Aklakan:
So I had to copy and adapt a total of 3 classes from sedona-common:

* [IndexSerde.java](https://github.com/apache/sedona/blob/sedona-1.7.0/common/src/main/java/org/locationtech/jts/index/strtree/IndexSerde.java)
* [GeometrySerde.java](https://github.com/apache/sedona/blob/sedona-1.7.0/common/src/main/java/org/apache/sedona/common/geometrySerde/GeometrySerde.java)
* [SpatialIndexSerde.java](https://github.com/apache/sedona/blob/sedona-1.7.0/common/src/main/java/org/apache/sedona/common/geometrySerde/SpatialIndexSerde.java)

Aklakan:
The spatial index does not story any geometries - just their envelopes - so geometry serialization performance is not an issue in the first place :man_facepalming:  - but good having talked about it :grinning: 

* I could remove sedona as a dependency alltogether - the copied classes are needed though.
* I added tests for the corner case where the same feature exists with different geometries in different graphs.
For the original spatial index implementation, some results are incorrect - for the new one they are correct.
* I need to write some tests for the kryo serializer for RDF1.2 Nodes.

Aklakan:
I now also added the benchmark that evaluates loading a spatial index from disk - so w.r.t. index loading during server startup, the kryo version is significantly faster.

```
Benchmark                   (param0_jenaVersion)  (param1_geometryMixes)  Mode  Cnt  Score   Error  Units
BenchmarkSpatialIndex.load               current                    1000  avgt    5  0.002 ± 0.001   s/op
BenchmarkSpatialIndex.load               current                   10000  avgt    5  0.006 ± 0.001   s/op
BenchmarkSpatialIndex.load               current                  100000  avgt    5  0.064 ± 0.004   s/op

BenchmarkSpatialIndex.load                 5.1.0                    1000  avgt    5  0.062 ± 0.009   s/op
BenchmarkSpatialIndex.load                 5.1.0                   10000  avgt    5  0.565 ± 0.471   s/op
BenchmarkSpatialIndex.load                 5.1.0                  100000  avgt    5  7.439 ± 0.442   s/op
```
A geometry mix is comprises 8 different geometry types (point, linestring, linearring, polygon, multipoint, multilinestring, multipolygon, geometrycollection). So the actual number of envelopes in the spatial index is `8 x geometryMixes`.

Aklakan:
@afs What would be the best way to add a Fuseki module for re-building the index for selected graphs? In an ideal world, the spatial index would be synchronized with the datasat, but for the time being we created a little Fuseki-Mod, where one can select the graphs for which to rebuild the index. Back then we forked your `afs/fuseki-mods` git repo - but this no longer exists. At first glance, it seems they were moved to `jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/mod/*`. IMO the geosparql support should be just a fuseki mod - perhaps with a bundle release for convenience, but not as a fork.

This is what the spatial index rebuilder looks like. The html is a single self-contained file (ai generated).

image

jena-arq/src/main/java/org/apache/jena/web/HttpSC.java
```diff
@@ -306,7 +306,7 @@ public boolean isServerError()
       */
      public static Code getCode(int code)
      {
-         if (code <= MAX_CODE)
+         if (code >= 0 && code <= MAX_CODE)
```
Aklakan, line 309: Needed to have server side events (SSE) working in the spatial indexer fmod.

jena-fuseki2/jena-fuseki-mod-geosparql/src/test/java/org/apache/jena/fuseki/mod/geosparql/TestFMod_SpatialIndexer.java
```diff
@@ -0,0 +1,147 @@
...
+
+        // Generate geometries into the default graph and a named graph
+        GeometryGenerator.generateGraph(dsg.getDefaultGraph(), envelope, conf);
+
+        conf.put(GeometryType.POLYGON, 1);
+        GeometryGenerator.generateGraph(dsg.getGraph(graphName1), envelope, conf);
+    }
+
+    @Test
+    @Ignore // Ignored for new because this test requires a local Chrome browser.
```
Aklakan, line 107: Remove @Ignore to run this test; but requires a local chrome. This test could be improved e.g. by starting chrome in docker via testcontainers. Perhaps it should be migrated to the cypress setup (if possible).

jena-fuseki2/jena-fuseki-mod-geosparql/src/test/java/org/apache/jena/fuseki/mod/geosparql/TestFMod_SpatialIndexer.java
```diff
@@ -0,0 +1,147 @@
...
+        clearLastEvent();
+
+        // Remove the named graph and update the index.
+        dsg.removeGraph(graphName1);
+        WebElement cleanButton = driver.findElement(By.id("clean-action"));
+        cleanButton.click();
+        awaitEvent();
+        try (Stream<?> s = Iter.asStream(dsg.listGraphNodes())) {
+            long actualNamedGraphCount = s.count();
+            Assert.assertEquals(0, actualNamedGraphCount);
```
Aklakan, line 124: Whoops, this should test whether the spatial index (not dsg) only has one graph.

jena-fuseki2/jena-fuseki-mod-geosparql/src/main/java/org/apache/jena/fuseki/mod/geosparql/FMod_SpatialIndexer.java
```diff
@@ -0,0 +1,148 @@
...
+        // This does not appear to do anything - datasetNames may be empty but data access points may exist.
+        // datasetNames.forEach(name -> builder.addEndpoint(name, "spatial", op));
+        // datasetNames.forEach(name -> builder.addEndpoint(name, "spatial-events", op));
+    }
+
+    /**
+     * The indexer is only created if there is a SPARQL update endpoint. The update endpoint's auth policy is inherited.
+     */
+    @Override
+    public void configured(FusekiServer.Builder serverBuilder, DataAccessPointRegistry dapRegistry, Model configModel) {
```
Aklakan, line 81: One important point is how to handle authentication to the spatial indexer servlet as to prevent service disruption by indexer recomputation requests. The logic below only registers the indexer servlet for a data access point if there are update endpoints with at most 1 auth policy. The auth policy is then inherited for the spatial indexer.

In other words, multiple update endpoints with multiple auth policies prevent spatial indexer registration. In this case, the indexer would have to be manually configured in the fuseki config. Alternatively, for each update endpoint there could be a updateName-spatial endpoint which inherits that update's endpoint auth policy.

jena-arq/src/main/java/org/apache/jena/sparql/util/Context.java
```diff
@@ -387,6 +389,11 @@ public void clear() {
         context.clear();
     }
 
+    /** Atomic compute. */
+    public <V> Object compute(Symbol key, BiFunction<Symbol, Object, V> remappingFunction) {
```
afs, line 393: This does mot seem to be used anywhere.

jena-geosparql/src/main/java/org/apache/jena/geosparql/kryo/NodeSerializer.java
```diff
@@ -0,0 +1,346 @@
...
+                        throw new RuntimeException("Unknown literal sub-type: " + subTypeVal);
+                }
+                break;
+            case TYPE_BNODE:
+                v1 = input.readString();
+                result = NodeFactory.createBlankNode(v1);
+                break;
+            case TYPE_TRIPLE:
+                t = kryo.readObject(input, Triple.class);
+                result = NodeFactory.createTripleNode(t);
```
afs, line 320:
```suggestion
                result = NodeFactory.createTripleTerm(t);
```
`createTripleNode` is deprecated.

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/SpatialIndex.java
```diff
@@ -17,530 +17,45 @@
...
+     * The default graph can be referenced with null or any value for which {@link Quad#isDefaultGraph()} returns true.
+     * The union graph (= all named graphs but not the default graph) can be referenced with {@link Quad#unionGraph}.
      *
-     * @param spatialIndexFile
-     * @param spatialIndexItems
-     * @param srsURI
-     * @throws SpatialIndexException
+     * @param searchEnvelope
+     * @param graph
+     * @return
```
afs, line 55: Incomplete javadoc. `@return` needs a type.

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/SpatialIndexFindUtils.java
```diff
@@ -0,0 +1,189 @@
...
+public class SpatialIndexFindUtils {
+    private static final Logger LOGGER = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+    /**
+     * Find Spatial Index Items from all graphs in Dataset.<br>
+     *
+     * @param datasetGraph
+     * @param srsURI
+     * @return SpatialIndexItems found.
+     * @throws SpatialIndexException
```
afs, line 50: This causes a javadoc warning because there isn't a `throws`  for `SpatialIndexException`  (3 occurrences of same situation).

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/SpatialIndexFindUtils.java
```diff
@@ -0,0 +1,189 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.jena.geosparql.spatial;
+
+import java.lang.invoke.MethodHandles;
+import java.util.Iterator;
+import java.util.List;
```
afs, line 22: Not used?

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/index/v1/SpatialIndexV1.java
```diff
@@ -0,0 +1,577 @@
...
+
+/**
+ * SpatialIndex for testing bounding box collisions between geometries within a
+ * Dataset.<br>
+ * Queries must be performed using the same SRS URI as the SpatialIndex.<br>
+ * The SpatialIndex is added to the Dataset Context when it is built.<br>
+ * QueryRewriteIndex is also stored in the SpatialIndex as its content is
+ * Dataset specific.
+ */
+@Deprecated /** Superseded by {@link SpatialIndex} and {@link SpatialIndexPerGraph}. */
```
afs, line 80: There isn't a `SpatialIndexPerGraph`.

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/index/v1/SpatialIndexAdapterV1.java
```diff
@@ -0,0 +1,83 @@
...
+
+import org.apache.jena.geosparql.implementation.SRSInfo;
+import org.apache.jena.geosparql.spatial.SpatialIndex;
+import org.apache.jena.graph.Node;
+import org.apache.jena.rdf.model.RDFNode;
+import org.apache.jena.rdf.model.Resource;
+import org.locationtech.jts.geom.Envelope;
+
+/** Adapter class for spatial index v1. */
+public class SpatialIndexAdapterV1
```
afs, line 35: Several deprecation warnings.

Maybe add `@SuppressWarnings("deprecation")` to the class.

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/index/v1/SpatialIndexV1.java
afs: Should this be `@Deprecated(forRemoval = true)`?

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/index/v2/STRtreePerGraphSerializer.java
```diff
@@ -0,0 +1,49 @@
...
+{
+    @Override
+    public void write(Kryo kryo, Output output, STRtreePerGraph index) {
+        kryo.writeClassAndObject(output, index.getInternalTreeMap());
+        output.writeBoolean(index.isBuilt());
+    }
+
+    @Override
+    public STRtreePerGraph read(Kryo kryo, Input input, Class<STRtreePerGraph> type) {
+        Map<Node, STRtree> treeMap = (Map<Node, STRtree>)kryo.readClassAndObject(input);
```
afs, line 41: Add `@SuppressWarnings("unchecked")`

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/task/AbortableThread.java
```diff
@@ -0,0 +1,200 @@
...
+
+    protected void setStatusMessage(String statusMessage) {
+        this.statusMessage = statusMessage;
+    }
+
+    public String getStatusMessage() {
+        return statusMessage;
+    }
+
+    public final void run() {
```
afs, line 106:
```suggestion
    @Override
    public final void run() {
```

jena-geosparql/src/test/java/org/apache/jena/geosparql/spatial/SpatialIndexTestData.java
```diff
@@ -60,21 +71,28 @@ public class SpatialIndexTestData {
...
     public static final SRSInfo OSGB_SRS_INFO = new SRSInfo(SRS_URI.OSGB36_CRS);
 
-    private static SpatialIndex TEST_SPATIAL_INDEX = null;
+    private static SpatialIndexPerGraph TEST_SPATIAL_INDEX = null;
     private static Dataset TEST_DATASET = null;
 
-    public static final SpatialIndex createTestIndex() {
+    public static final List<SpatialIndexItem> getTestItems() {
+        List<SpatialIndexItem> items = List.of(
+                SpatialIndexItem.of(LONDON_GEOMETRY_WRAPPER.getEnvelope(), LONDON_FEATURE),
```
afs, line 79: 5 deprecated functions here. Add `.asNode()` to second argument.

afs:
The review is not all warnings I was getting - and sometimes new ones appear as old ones are addressed.

Aklakan:
Right, I'll look into cleaning up those warnings; especially I haven't checked the javadoc much yet. As for deprecation, I recall that adding forRemoval=true caused even more warnings, so I removed it again. As for the deprecation warnings related to SpatialIndexV1, I wanted to keep the implementation for compatibility and also comparison reasons, although it should be considered deprecated.

Not sure what do best do there; for the time being perhaps disable deprecation warnings for the module (if possible) or remove the deprecations for now.

afs:
No need to use `forRemoval=true` if it does not work for you.

(I find it a bit erratic as in javadoc across Eclispe and the maven javadoc plugin - it seems to be considered a different warning, sometimes)

jena-geosparql/src/main/java/org/locationtech/jts/index/strtree/STRtreeSerializer.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.List;
+
+import org.locationtech.jts.geom.Envelope;
+
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.Serializer;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+
+/* This file is an adapted copy of org.locationtech.jts.index.strtree.IndexSerde. */
```
afs, line 31: Is this via sedona?
What's the license of the code?

I think it is better to put the class in a org.apache.jena package. Jena is a library so maybe classed from the locationtech package are on the classpath another route.

As it is a single, modified class, maybe like HttpSC, which is from Jetty and copied into Jena.

Aklakan: This class for serializing JTS 'STRtree' structures is from Sedona but it requires access to non-public parts of `STRtree`. So unfortunately it must reside in the `org.locationtech.jts.index.strtree` package.

jena-arq/src/main/java/org/apache/jena/sparql/util/Context.java
Aklakan: It is used from the module `jena-fuseki-mod-geosparql` - class `SpatialIndexComputeService.scheduleTask`.
This method uses `Context.compute` to atomically check for a currently running indexing task if none exists creates a one.

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/index/v1/SpatialIndexAdapterV1.java
Aklakan: > Maybe add `@SuppressWarnings("deprecation")` to the class.

:+1:

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/SpatialIndexFindUtils.java
Aklakan: An oversight when I changed the original logic to streaming via `Iter`. The SpatialIndex exception is no longer thrown at the time when searching for spatial features.

jena-geosparql/src/test/java/org/apache/jena/geosparql/spatial/SpatialIndexTestData.java
Aklakan: Added `.asNode()` as suggested.

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/SpatialIndexFindUtils.java
Aklakan: Removed unused imports.

jena-fuseki2/jena-fuseki-mod-geosparql/src/main/java/org/apache/jena/fuseki/mod/geosparql/FMod_SpatialIndexer.java
Aklakan: I changed to logic of the spatial indexer fuseki mod such that it registers for each update endpoint a corresponding `<updateEndpointName>-spatial` spatial indexer endpoint which inherits the update's endpoint auth policy.

Aklakan: This needs one more round of update to the logic: It should be possible to specify in the global config model whether to auto-register the spatial indexer endpoint for all update endpoints. After all, if someone is allowed to update the data, then I think it is reasonable to also allow for updating the spatial index which depends on the data.
If the global option disabled (default), then spatial-indexer operations and endpoints would have to be manually configured.

afs:
Here are the Java warnings I am seeing in Eclipse:

<details>
  <summary>Eclipse Warnings</summary>

```
The method load(File) from the type SpatialIndexV1 is deprecated
SpatialIndexIo.java
/jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/index/compat
line 51

The type SpatialIndexV1 is deprecated
SpatialIndexIo.java
/jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/index/compat
line 51

The type SpatialIndexV1 is deprecated
SpatialIndexIo.java
/jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/index/compat
line 51

Type safety: Unchecked cast from Object to Map<Node,STRtree>
STRtreePerGraphSerializer.java
/jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/index/v2
line 41

The method write(Kryo, Output, STRtree) of type STRtreeSerializer should be tagged with @Override since it actually overrides a superclass method
STRtreeSerializer.java
/jena-geosparql/src/main/java/org/locationtech/jts/index/strtree
line 68

Javadoc: Parameter includeWgs84 is not declared
QueryRewriteTestData.java
/jena-geosparql/src/test/java/org/apache/jena/geosparql/geo/topological
line 60

The import org.apache.jena.rdf.model.ModelFactory is never used
QueryRewriteTestData.java
/jena-geosparql/src/test/java/org/apache/jena/geosparql/geo/topological
line 27

The method buildSpatialIndex(Dataset, String) from the type SpatialIndexV1 is deprecated
TestSpatialIndexGraphLookupV1.java
/jena-geosparql/src/test/java/org/apache/jena/geosparql/spatial
line 32

The type SpatialIndexV1 is deprecated
TestSpatialIndexGraphLookupV1.java
/jena-geosparql/src/test/java/org/apache/jena/geosparql/spatial
line 32

The type SpatialIndexV1 is deprecated
TestSpatialIndexGraphLookupV1.java
/jena-geosparql/src/test/java/org/apache/jena/geosparql/spatial
line 32

The method buildSpatialIndex(Dataset, File) from the type SpatialIndexV1 is deprecated
SpatialIndexTest.java
/jena-geosparql/src/test/java/org/apache/jena/geosparql/spatial/index/v2
line 130

The method createTripleNode(Triple) from the type NodeFactory is deprecated
TestNodeSerializer.java
/jena-geosparql/src/test/java/org/apache/jena/geosparql/spatial/index/v2
line 74

The method save(File, Collection<SpatialIndexItem>, String) from the type SpatialIndexV1 is deprecated
SpatialIndexTest.java
/jena-geosparql/src/test/java/org/apache/jena/geosparql/spatial/index/v2
line 58

The type SpatialIndexV1 is deprecated
SpatialIndexTest.java
/jena-geosparql/src/test/java/org/apache/jena/geosparql/spatial/index/v2
line 58

The type SpatialIndexV1 is deprecated
SpatialIndexTest.java
/jena-geosparql/src/test/java/org/apache/jena/geosparql/spatial/index/v2
line 130
```

</details>

jena-geosparql/src/main/java/org/locationtech/jts/index/strtree/STRtreeSerializer.java
afs: Could you put that it comes from Sedona in the comment please so that it is clear is not derived from JTS code?
And a note about why it is in the package it is in would be good.

Whether having classes in the same package spread across jars will continue working is anyone's guess, although shading crushes the distinction.

afs:
jena-fuseki-mod-geosparql is shaded.

Making a dependency of jena-fuseki-server produces a lot (!!) of overlap warnings.

Either:
1. jena-fuseki-mod-geosparql could go in as module in `lib/` in the apache-jena-fuseki packaging
2. Not be shaded.

Aklakan:
> No need to use forRemoval=true if it does not work for you.

I realized that when `forRemoal=true` is used then suppression requires `@SuppressWarnings("removal")` rather than `@SuppressWarnings("deprecation")`.

All your reported warnings (and some more) should now be cleaned up - I updated my Eclipse error/warning level configuration accordingly.

jena-geosparql/src/main/java/org/apache/jena/geosparql/spatial/index/v1/SpatialIndexV1.java
Aklakan: Yes - updated.

Aklakan:
> jena-fuseki-mod-geosparql is shaded.
> 
> Making a dependency of jena-fuseki-server produces a lot (!!) of overlap warnings.
> 
> Either:
> 
>     1. jena-fuseki-mod-geosparql could go in as module in `lib/` in the apache-jena-fuseki packaging
> 
>     2. Not be shaded.

My intention was to go with 2 - but I had overlooked the shade configuration for the `package` goal. I moved the shading into a `bundle` profile akin to the setup in the [service enhancer pom.xml](https://github.com/apache/jena/blob/ff27e3df401ca5cbff414906d64ad9d026e16d7f/jena-extras/jena-serviceenhancer/pom.xml#L88-L136).

The bundle produced by the setup should be suitable for the /lib folder - but I need to test that.

jena-geosparql/src/main/java/org/apache/jena/geosparql/InitGeoSPARQL.java
```diff
@@ -42,7 +42,7 @@ public static void init() {
         if ( initialized )
             return ;
         synchronized (initLock) {
-            if ( initialized ) {
+            if ( initialized || System.getProperty("jena.geosparql.skip", "false").equalsIgnoreCase("true") ) {
```
afs, line 45: That's going to need documenting!

jena-arq/src/main/java/org/apache/jena/system/Txn.java
```diff
@@ -137,4 +139,46 @@ private static <T extends Transactional> void onThrowable(Throwable th, T txn) {
...
+     */
+    public static AutoTxn begin(Transactional txn, ReadWrite readWrite) {
+        return begin(txn, TxnType.convert(readWrite));
+    }
+
+    /**
+     * Begins a transaction and returns a transaction control instance suitable
+     * for use with try-with-resources blocks.
+     * Allows for raising checked exceptions in an idiomatic way.
+     * Closing the TxnCtl instance will abort the transaction unless it
```
afs, line 156: `TxnCtl` is an old name (the tests need updating as well).

jena-arq/src/main/java/org/apache/jena/system/Txn.java
```diff
@@ -137,4 +139,46 @@ private static <T extends Transactional> void onThrowable(Throwable th, T txn) {
...
+     * has been manually committed.
+     * <p>
+     *
+     * Usage example:
+     * <pre>
+     * public void myMethod() throws IOException {
+     *   try (AutoTxn txn = Txn.begin(dataset, TxnType.WRITE)) {
```
afs, line 163: `begin` is rather overloaded. This "begin" isn't visibly paired with abort/commit in the application code.

`autoTxn` is a better name and works with static imports.

jena-arq/src/main/java/org/apache/jena/system/Txn.java
```diff
@@ -137,4 +139,46 @@ private static <T extends Transactional> void onThrowable(Throwable th, T txn) {
...
+     *     // Do work
+     *     if (someError) {
+     *         throw new IOException();
+     *     }
+     *     // Must manually call commit on success.
+     *     txn.commit();
```
afs, line 169: That's a different style to other `Txn` which commits on normal termination of the runnable and aborts on an exception, though I don't immediately see how to do the exception capture here.

jena-arq/src/test/java/org/apache/jena/system/TestTxn.java
```diff
@@ -378,41 +379,67 @@ public void txn_threaded_01() {
...
-            a.run(); 
+            a.run();
             // Does not block promotion.
             boolean b = tx.promote();
             assertTrue(b);
             assertEquals(ReadWrite.WRITE, tx.transactionMode());
         });
     }
 
+    @Test public void txn_ctl_write_01() {
```
afs, line 417: `txn_ctl` is an old name?

afs:
> My intention was to go with 2
(jena-fuseki-mod-geosparql is not shared)

I agree this is the way to go.

afs:
Where does this leave `jena-fuseki-geosparql` long term?  If `jena-fuseki-mod-geosparql` integrated into `jena-fuseki-server` is a replacement, we could drop `jena-fuseki-geosparql` at Jena6 (late this year, early next).

That looses the command line setup - we could ask users about this.

jena-fuseki2/jena-fuseki-mod-geosparql/src/main/java/org/apache/jena/fuseki/mod/geosparql/SpatialIndexComputeService.java
```diff
@@ -0,0 +1,579 @@
...
+            }
+            return;
+        } else {
+            boolean isReplaceMode = isReplaceMode(action);
+            int threadCount = getThreadCount(action);
+
+            // Only SpatialIndexPerGraph can be updated.
+            // Check if the index can be updated. If not then raise an exception
+            // that informs that only replace mode can be used in this situation.
+            if (!isReplaceMode && !(index instanceof SpatialIndexPerGraph)) {
```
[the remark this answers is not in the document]
Aklakan: My intention is that the servlet can upgrade any legacy index (such as SpatialIndexV1) to the latest implementation version. I need to revise the code for that.

Aklakan:
According to  the [jena-fuseki-geosparql documentation](https://jena.apache.org/documentation/geosparql/geosparql-fuseki.html) it is a bit of a jena-geo toolkit:

- it adds its own cli options to fuseki, such as setting a spatial index file or whether to enable geosparql query rewriting. There the question is whether it would be possible to write plugins that extend fuseki's CLI. I think the clean/reliable way would be to use the assembler, but having a quick way to get a fuseki running with these features is still useful.

- some commands add data processing that could also be accomplished with sparql queries. These could be perhaps moved to a `geo` cli tool akin to `arq` or `riot`. Also, there could be an fmod that adds a little geo workbench html site with buttons to carry out these tasks (Validate Geometry Literals, Convert Geo predicates, ...)

- it adds support for loading data based on a file-to-graph mapping file ("Load Tabular file into dataset"). This could be of interest for the arq/sparql (or similar) tool which executes it as a set of LOAD statements or a general fuseki-client-cli which may use fuseki specific APIs for performance/security reasons. Not sure right now whether sparql LOAD without a custom Locator could access file:// URLs.

So I think in the long run there should be only one Fuseki, and the features of jena-fuseki-geosparql would be either moved to mods or separate cli tools. But before that happens I guess things just stay as they are?

jena-arq/src/main/java/org/apache/jena/system/Txn.java
Aklakan: When `AutoTxn.close()` is called by the try-with-resources block, then there seems to be no standard way for close to detect whether it is called after normal or exceptional block exit. Therefore the logic is that always exceptional block exit is assumed (-> rollback) and commit() has to be called manually.

jena-arq/src/main/java/org/apache/jena/system/Txn.java
Aklakan: Alternatively, the AutoTxn-related begin methods could be moved to the AutoTxn class.

`try (AutoTxn txn = AutoTxn.begin(transactional, TxnType.WRITE)) {`

Aklakan: So in summary I tend towards to `Txn.autoTxn` solution.

jena-fuseki2/jena-fuseki-mod-geosparql/pom.xml
```diff
@@ -0,0 +1,167 @@
...
+    <profile>
+      <id>bundle</id>
+      <dependencies>
+        <dependency>
+          <groupId>org.apache.jena</groupId>
+          <artifactId>jena-fuseki-main</artifactId>
+          <version>5.5.0-SNAPSHOT</version>
+          <scope>provided</scope>
+        </dependency>
+        <dependency>
```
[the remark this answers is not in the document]
Aklakan: jena-geosparql (compile) in this pom transitively depends on jena-arq - so jena-arq needs to be excluded as well. Conversely, without setting jena-arq to provided, it is listed as compile in `mvn -Pbundle dependency:tree`.

jena-fuseki2/jena-fuseki-mod-geosparql/pom.xml
```diff
@@ -0,0 +1,167 @@
...
+    <dependency>
+      <groupId>org.seleniumhq.selenium</groupId>
+      <artifactId>selenium-java</artifactId>
+      <version>4.18.1</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>io.github.bonigarcia</groupId>
+      <artifactId>webdrivermanager</artifactId>
```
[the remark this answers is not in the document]
Aklakan: done.

jena-geosparql/src/main/java/org/apache/jena/geosparql/InitGeoSPARQL.java
Aklakan: @LorenzBuehmann Do you recall the intention of this line?

One reason I can think of is that that jena-geosparql takes long to initialize. IIRC due to setting up a derby database with SIS data. So any command line jar that uses jena-geosparql takes long to load.
The loading time of own java cli tool increased from 200ms to 1.000ms just by adding jena-geosparql.

So a future issue is to change parts of jena-geosparql's eager init to lazy init on first use.

jena-fuseki2/jena-fuseki-mod-geosparql/src/main/java/org/apache/jena/fuseki/mod/geosparql/SpatialIndexComputeService.java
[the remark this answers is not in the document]
Aklakan: Logic should be fixed now.

jena-fuseki2/jena-fuseki-mod-geosparql/src/main/java/org/apache/jena/fuseki/mod/geosparql/SpatialIndexComputeService.java
```diff
@@ -0,0 +1,579 @@
...
+                        SpatialIndexIoKryo.save(targetFile, newIndex);
+                    }
+                    String statusMsg = String.format("Updated spatial index with %d graphs.", graphCount);
+                    setStatusMessage(statusMsg);
+                    if (logger.isInfoEnabled()) {
+                        logger.info("Indexing of {} graphs completed successfully.", graphCount);
+                    }
+                }
+
+                public void requestCancel() {
```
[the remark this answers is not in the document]
Aklakan: fixed.

jena-fuseki2/jena-fuseki-mod-geosparql/src/test/java/org/apache/jena/fuseki/mod/geosparql/TestFMod_SpatialIndexer.java
Aklakan: updated the test.

afs:
> So I think in the long run there should be only one Fuseki, and the features of jena-fuseki-geosparql would be either moved to mods or separate cli tools. But before that happens I guess things just stay as they are(?)

Fair enough.

Is the geosparql assembler sufficient?

Aklakan:
Two aspects I still need to address:

- A security issue: "Remove absent graphs from index": This needs to consider DynamicDatasets . An authenticated user must not be able to remove graphs from the index which for which he is not authorized. However, right now if a user clicks on this button it will remove all the index entries for any graph which he can't see. I think this can be solved by unwrapping the DynamicDataset to get the underlying complete graph list and prevent removal of any graphs in there.

- A small one for which I should write a test case: JTS STRtree has an 'isBuilt' flag; and for this reason my STRtreePerGraph structure has such a flag too: If `STRtreePerGraph.isBuilt` is true on the then any added STRtree's should be built on insert. I should check whether serialization w.r.t. flag works properly; would be ugly if the index format broke for this reason.

> "TODO Raise some HTTP error"

I also had a look into the existing Fuseki code (e.g. `SPARQL_QueryGeneral`) to see how to do the error handling properly - I should update this part to something like `ServletOps.errorBadRequest("Failed to load URL " + uri);`

So I should be able to resolve this issues this weekend.

Aklakan:
I completed the final changes (I used a second commit for those):

- Renamed `SpatialIndexUtils` to `SpatialIndexLib`
- Moved symbols to `SpatialIndexConstants` (akin to `ARQConstants`)
- Fixed a bug where server side events (SSE) about index/clean tasks were sent to listeners across all endpoints
- All task logic is now disentangled from the servlet and resides in SpatialIndexLib. A `TaskListener` indirection is used to advertise task life cycle state changes (e.g. running, aborting) as SSE events. I still think this can be improved but it is much better than before.
- Disabled auto-registration of spatial indexer endpoints for sparql update endpoints. Due to security concerns I think its better and more idiomatic to require a usual Fuseki endpoint declaration as below.
- Added additional tests for index updates, including the case when access is restricted to a subset of the available graphs. I only then learned that Fuseki denies modding endpoints for graph based access controlled datasets. The spatial indexer implementation would be ready for it.
- Added warnings if SRS and/or spatial index file location are not configured in the assembler.
- Cleaned up logging out


This config snippet needs to be documented:
```turtle
<#ep-si> fuseki:name "spatial-indexer" ;
  fuseki:operation fuseki:spatial-indexer ;
  fuseki:allowedUsers "test" . # Optional access control.
```

afs:
> 1. This will need an entry in jena-fuseki2/apache-jena-fuseki/dist/LICENSE
>    That is a task on me.

PR #3246.

PR #3029, opened 2025-02-22
GH-3028: Use OpExtend (BIND)

### LINKED ISSUE #3028: LATERAL with SERVICE sends query with LET
### Version

5.4.0-SNAPSHOT

### What happened?

`LATERAL` with `SERVICE` sends non-standard SPARQL making use of `LET` to the remote endpoint.
Not sure whether this is a conceptual issue of [SEP-0007](https://github.com/w3c/sparql-dev/blob/main/SEP/SEP-0007/sep-0007.md) or an implementation issue in Jena.

The problem arose for me when I tried to sample a remote endpoint for instances of certain classes.

```java
    // TestQueryExecution.java

    @Test public void lateral_with_service() {

        // Register a custom service handler to capture the
        // algebra of what is being executed.
        ServiceExecutorRegistry reg = new ServiceExecutorRegistry();
        Op[] executedOpHolder = {null};
        reg.addSingleLink((opExecute, opOriginal, binding, execCxt, chain) -> {
            executedOpHolder[0] = opExecute;
            return QueryIterRoot.create(execCxt);
        });

        String qsReport = """
            PREFIX eg: <http://www.example.org/>
            SELECT * {
              BIND(eg:MyClass AS ?t)
              LATERAL {
                SERVICE <urn:my:service> {
                  { SELECT * { ?s a ?t } LIMIT 2 }
                }
              }
            }
            """;

        QueryExec.dataset(DatasetGraphFactory.empty())
                .query(qsReport)
                .set(ARQConstants.registryServiceExecutors, reg)
                .table();

        Op executedOp = executedOpHolder[0];
        Query executedQuery = OpAsQuery.asQuery(executedOp);
        System.out.println("Query:");
        System.out.println(executedQuery);
        System.out.println();
        System.out.println("Algebra:");
        System.out.println(executedOp);
    }
```

## Output
Query:
```sparql
SELECT  *
WHERE
  { SERVICE <urn:my:service>
      { SELECT  *
        WHERE
          { ?s  a  <http://www.example.org/MyClass>
            LET (?t := <http://www.example.org/MyClass>)
          }
        LIMIT   2
      }
  }
```

Algebra:
```lisp
(service <urn:my:service>
  (slice _ 2
    (assign ((?t <http://www.example.org/MyClass>))
      (bgp (triple ?s <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.example.org/MyClass>)))))
```

```shell

```

None

afs:
GitHub issue resolved #3028

Pull request Description:
Use OpExtend (BIND) rather than OpAssign (LET).
All uses go via a single function for switchability.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterLateral.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterLateral.java
@@ -54,10 +54,20 @@ public class QueryIterLateral extends QueryIterRepeatApply {
         this.isUnit = isJoinIdentity(subOp);
     }
 
-    private boolean isJoinIdentity(Op op) {
+    private static boolean isJoinIdentity(Op op) {
         if( ! ( op instanceof OpTable table ) )
             return false;
         return table.isJoinIdentity();
+    }
+
+    /** Choice of variable setting */
+    private static Op assignments(Op op, VarExprList varExprs) {
+        // Assign (LET) is safer (it is not an error to assign to variable if already
+        // set to the same term) but this operation is not portable.
+        //return OpAssign.create(op, varExprs);
+
+        // BIND
+        return OpExtend.create(op, varExprs);
     }
 
     @Override
@@ -114,7 +124,7 @@ public class QueryIterLateral extends QueryIterRepeatApply {
             Op opExec = substitute
                     ? Substitute.substitute(opBGP, substitutions)
                     : opBGP;
-            opExec = OpAssign.create(opExec, assigns);
+            opExec = OpExtend.create(opExec, assigns);
             return opExec;
         }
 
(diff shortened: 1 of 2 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterLateral.java
```diff
@@ -114,7 +124,7 @@ public Op transform(OpBGP opBGP) {
             Op opExec = substitute
                     ? Substitute.substitute(opBGP, substitutions)
                     : opBGP;
-            opExec = OpAssign.create(opExec, assigns);
+            opExec = OpExtend.create(opExec, assigns);
```
Aklakan, line 127: For consistency also call `assignments(opExec, assigns)` instead?

afs: Thanks for catching that. It was change from "assign" to "extend" in the investigation phase of the report!

PR #3033, opened 2025-02-24
Support for JSON query type in QueryTransformOps.

Aklakan:
Pull request Description: Minor fix for handling JSON in QueryTransformOps. One of our Linked Data viewers actually used this query type - and some of my rewrite plugins broke with it due to the missing transformation support.

----

 - (Key commit messages start with the issue number (GH-xxxx))

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/query/Query.java
+++ b/jena-arq/src/main/java/org/apache/jena/query/Query.java
@@ -503,6 +503,17 @@ public class Query extends Prologue implements Cloneable, Printable
         return Collections.unmodifiableMap(jsonMapping);
     }
 
+    /**
+     * Overwrite all prior JSON mappings with new ones.
+     *
+     * @param newJsonMapping The new JSON mappings. Must not be null.
+     */
+    public void setJsonMapping(Map<String, Node> newJsonMapping) {
+        Objects.requireNonNull(newJsonMapping);
+        jsonMapping.clear();
+        jsonMapping.putAll(newJsonMapping);
+    }
+
     // ---- Aggregates
 
     // Record allocated aggregations.
--- a/jena-arq/src/main/java/org/apache/jena/sparql/syntax/syntaxtransform/QueryTransformOps.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/syntax/syntaxtransform/QueryTransformOps.java
@@ -21,9 +21,12 @@ package org.apache.jena.sparql.syntax.syntaxtransform;
 import static org.apache.jena.sparql.syntax.syntaxtransform.QuerySyntaxSubstituteScope.scopeCheck;
 
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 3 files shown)


rvesse (approved):

afs:
Test case?

Aklakan:
> Test case?

Good point - the problem is that the generated JSON is simply empty - which silently worked but is incorrect.

Trying to add a test case reveals a rat-tail of issues - for example `JSON { "x": FOO }` does not support IRI terms such as `<BAR>` in place of FOO - so variable substitution breaks. Using literals raises scoping errors.

At first it seemed easier to add the missing transformation support to Jena - but now seems its more reliable to just adapt our ld viewer to use SELECT with `application/sparql-results+json` (with all this ugly unnesting of json sparql results and JSON number conversion depending on the datatype ).

I set this issue to draft for now - but I seems I'll withdraw this PR.

afs:
> I set this issue to draft for now - but it seems I'll withdraw this PR.

I'd like to put the PR in even though standards based `application/sparql-results+json`, rather than an ARQ extension, is better when possible.

Thank you for adding the test case.

Aklakan:
Well the ARQ extension does have its merits, even though the JSON mapping capabilities are somewhat limited. 

Maybe this is an opportunity to mention some other current work of mine related to GraphQL:
There aren't many mentions of GraphQL on Jena's issue tracker - one mention is from yourself when the `JSON` query form was added: https://github.com/apache/jena/pull/114#issuecomment-371460985

* I am aware of https://github.com/telicent-oss/graphql-jena which operates on the DatasetGraph level.
* I have also been working on a GraphQL-to-SPARQL rewriter which allows use of arbitrary SPARQL graph patterns to traverse the RDF data. I have a  couple of [online demos](https://smartdataanalytics.github.io/RdfProcessingToolkit/graphql/demo). In the demos, the graph patterns are annotations in the GraphQL query, but since recently it is also possible to add the annotations to the GraphQL schema. Also, the graph patterns can be connected by multiple variables that form composite keys. For example, statistical observations might be blank nodes but identified by (city, country) attributes as in this  [test case](https://github.com/Scaseco/jenax/blob/d606cb4526ba921355ef1868a7c6102a829e1fbb/jenax-graphql-parent/jenax-graphql-v2/src/test/java/org/aksw/jenax/graphql/sparql/TestGraphQlToSparqlMultiVars.java#L47).

So this is something I have in the pipeline and could contribute sometime after the spatial index as a more general approach to querying RDF data as JSON.


From the [Movie Browser Demo](https://smartdataanalytics.github.io/RdfProcessingToolkit/demos/movies/):
graphql

afs:
@Aklakan Would you mind setting this to "Ready to review"? I'll take it from there but, for good form, I'd prefer to take a PR being made.

GraphQL for RDF is very interesting.

Aklakan:
Set to ready for review.

afs:
#3045

Aklakan:
> maybe this could go in a discussion? It'll not get seen here.

Yes, will open one soon.

PR #3047, opened 2025-03-05
GH-3044: HashJoin iterator construction no longer eagerly builds hash probe table.

### LINKED ISSUE #3044: timeout parameter is not always respected by fuseki server
### Version

5.3.0

### What happened?

Using the Python _requests_ client, I am running sequentially a list of diverse SPARQL queries on a fuseki endpoint at localhost:3030, using POST request and a timeout parameter set to 10 s. For the few first lengthy query in the list, the timeout works as expected, i.e. the request terminates after approx 10 s and with a 503 status code. However at some point the timeout of subsequent queries is ignored, i.e.  the queries executing correctly and completely. 

As I suspect that this problem might be linked to the fuseki HTTP server, I have not investigated the content of the queries that are not timing out. Should I?

Thanks in advance,

  Marco

```shell

```

None

Aklakan:
GitHub issue resolved #3044 
Pull request Description: Timeouts in QueryExecDataset were postponed until a so-far eager hash probe table construction completed because of locking. This PR proposes to initialize the hash probe table lazily such that the iterator construction becomes cheap and the lock is released quickly.

I added a test case that fails without the proposed fix.

----

 - [x] Tests are included.
 - (Documentation change and updates are provided for the [Apache Jena website](https://github.com/apache/jena-site/))

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterFilterExpr.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterFilterExpr.java
@@ -21,6 +21,7 @@ package org.apache.jena.sparql.engine.iterator;
 import org.apache.jena.atlas.io.IndentedWriter;
 import org.apache.jena.atlas.lib.Lib;
 import org.apache.jena.atlas.logging.Log;
+import org.apache.jena.query.QueryCancelledException;
 import org.apache.jena.sparql.engine.ExecutionContext;
 import org.apache.jena.sparql.engine.QueryIterator;
 import org.apache.jena.sparql.engine.binding.Binding;
@@ -48,6 +49,9 @@ public class QueryIterFilterExpr extends QueryIterProcessBinding {
             if ( expr.isSatisfied(binding, super.getExecContext()) )
                 return binding;
             return null;
+        } catch (QueryCancelledException ex) {
+            ex.addSuppressed(new RuntimeException("Query cancelled exception."));
+            throw ex;
         } catch (ExprException ex) {
             // Some evaluation exception: should not happen.
             Log.warn(this, "Expression Exception in " + expr, ex);
--- a/jena-arq/src/main/java/org/apache/jena/sparql/exec/RowSetOps.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/exec/RowSetOps.java
@@ -30,9 +30,9 @@ import org.apache.jena.sparql.ARQConstants;
 import org.apache.jena.sparql.core.Prologue;
 import org.apache.jena.sparql.resultset.ResultsWriter;
 
-/** 
+/**
  * RowSetOps - Convenience ways to call the various output formatters.
- * 
+ *
  * @see ResultSetMgr
  */
 
@@ -51,7 +51,7 @@ public class RowSetOps {
      * This operation consumes the RowSet.
      */
     public static long count(RowSet rowSet)
-    { return rowSet.rewindable().size(); }
+    { return rowSet.stream().count(); }
 
     /**
      * Output a result set in a text format.  The result set is consumed.
--- a/jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/StageMatchTuple.java
+++ b/jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/StageMatchTuple.java
@@ -20,6 +20,7 @@ package org.apache.jena.tdb2.solver;
 
 import java.util.Iterator;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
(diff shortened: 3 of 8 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/exec/RowSetOps.java
```diff
@@ -51,7 +51,7 @@ public static void consume(RowSet rowSet)
      * This operation consumes the RowSet.
      */
     public static long count(RowSet rowSet)
-    { return rowSet.rewindable().size(); }
+    { long c[] = {0}; rowSet.forEach(b -> ++c[0]); return c[0]; }
```
Aklakan, line 54: Intent: Avoid memory copy.

rvesse: Or maybe `rowSet.stream().count()` would be cleaner and still avoid the memory copy (though it has conversion to `Stream` overhead?

Aklakan: good point!

jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/StageMatchTuple.java
```diff
@@ -68,6 +70,17 @@ private static Iterator<BindingNodeId> access(NodeTupleTable nodeTupleTable, Bin
             iterMatches = x.iterator();
         }
 
+        // Add cancel check.
```
Aklakan, line 73: A check is needed here for the following query on TDB2:
```
time curl http://localhost:3030/ds --data-urlencode 'query=PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#> SELECT ?x { ?x rdfs:label ?y . ?y rdfs:label ?z }' --data-urlencode 'timeout=1'
```

Against this preloaded [TDB2 DBpedia dataset](https://maven.aksw.org/archiva/#artifact-details-download-content~internal/tdb2.org.aksw.data.text2sparql.2025/dbpedia/1.0.0) (~30GB)

jena-arq/src/main/java/org/apache/jena/sparql/exec/RowSetOps.java
afs: Maybe it does not matter with thhis distructive operation but the cost of `stream()` is not trivial. 
`forEach` does not use `stream`.

afs: > Intent: Avoid memory copy.

`RowSetMem` rewind should not be a memory copy. It shares the rows array.

A rewindable RowSet (e.g. Mem) goes back to the beginning where as `forEach` continues the iteration.

Aklakan:
Set to draft because I am still investigating this query on in memory datasets.

```sparql
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
SELECT ?x { ?x rdfs:label ?y . ?y rdfs:label ?z }
```

jena-arq/src/main/java/org/apache/jena/sparql/exec/RowSetOps.java
Aklakan: Behavior so far:
* For non-RowSetMem instances counting would create an intermediate in-memory copy.
* The count for a RowSetMem instances would be the number of bindings in the row set - rather than the remaining bindings.

I added a the method `RowSetStream.forEachRemaining` - the stack trace for `RowSetOps.count` then becomes a delegate to the forEachRemaining method (with some boilerplate):

image

Aklakan:
To summarize:

* Aborting the now lazy HashJoin has a dedicated test.
* For the empty join case, there is a dedicated test for `StageMatchTriple` based on a custom `GraphBase.find()` implementation that returns an infinite stream of matches.
* But: not sure how I to mock a similar test for `StageMatchTuple` in TDB1/2, because it requires some large dataset or mocking NodeIds. Here is a brief screencapture of the effect of this PR's code:

[CancelStageMatchTuple.webm](https://github.com/user-attachments/assets/8ef59741-6db3-472f-a392-cac811393423)

Aklakan:
Hm, the long stack trace is not ideal - probably some additional exception catching is needed in the bridge between the `QueryIter(ator)` and plain iterator levels.

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterFilterExpr.java
```diff
@@ -48,6 +49,9 @@ public Binding accept(Binding binding) {
             if ( expr.isSatisfied(binding, super.getExecContext()) )
                 return binding;
             return null;
+        } catch (QueryCancelledException ex) {
+            ex.addSuppressed(new RuntimeException("Query cancelled exception."));
+            throw ex;
```
Aklakan, line 54: Should mitigate the case of ugly logging reported by @osma .

Issue #3008, opened 2025-02-12
Test sometimes fails on GH-Actions


OyvindLGjesdal:
### Version

main

### What happened?

When I manually start github actions on linux it sometimes gives the following error:

> [INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.070 s -- in org.apache.jena.fuseki.mod.TS_FusekiMods
[INFO] 
[INFO] Results:
[INFO] 
Error:  Failures: 
Error:    TestFusekiStdReadOnlySetup.stdSetup_endpoint_bad_2:158 POST http://localhost:45875/ds/query :: ==> expected: <415> but was: <-1>
[INFO] 
Error:  Tests run: 580, Failures: 1, Errors: 0, Skipped: 0

Other times it passes.

Haven't seen it apply to the java 17 job.

To me it looks unrelated to the commits where it happens, e.g https://github.com/OyvindLGjesdal/jena/commit/8cf9c8b50a849fe7c510fedd5760f0857d025c0e

### Relevant output and stacktrace

```shell
Error:  Tests run: 19, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.088 s <<< FAILURE! -- in org.apache.jena.fuseki.main.TestFusekiStdReadOnlySetup
Error:  org.apache.jena.fuseki.main.TestFusekiStdReadOnlySetup.stdSetup_endpoint_bad_2 -- Time elapsed: 0.009 s <<< FAILURE!
org.opentest4j.AssertionFailedError: POST http://localhost:45875/ds/query :: ==> expected: <415> but was: <-1>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:563)
	at org.apache.jena.fuseki.test.HttpTest.execWithHttpException(HttpTest.java:78)
	at org.apache.jena.fuseki.test.HttpTest.expect415(HttpTest.java:66)
[...]

PR #3063, opened 2025-03-10
Debug ci failures

OyvindLGjesdal:
GitHub issue resolved #

Pull request Description:

* Add temporarily more logging to httpClient to check what errors are cause CI failures in #3008
* I think one error suggests that one problem could be places where QueryExecution isn't closed. I have now naively added it to try-with for all occasions of conn.Query() which wasn't a try-with, like the javadocs suggest. The error shows up in the verbose log, but continues the run, so could be unrelated.

Expanded logs when running are not readable from the GUI (becomes laggy) because of size, but can be downloaded from "Download log archive".

Skjermbilde 2025-03-10 kl  21 38 43

Alot of errors show up and then the CI job continues, like:
```
2025-03-10T19:22:41.9792413Z Mar 10, 2025 7:22:41 PM jdk.internal.net.http.Http1AsyncReceiver onReadError
2025-03-10T19:22:41.9794722Z INFO: ERROR: HTTP/1 read subscriber recorded error: http://localhost:45971/ds-dataset?query=PREFIX+afn:+%3Chttp://jena.apache.org/ARQ/function%23%3E%0APREFIX+fuseki:+%3Chttp://jena.apache.org/fuseki%23%3E%0APREFIX+rdf:+%3Chttp://www.w3.org/1999/02/22-rdf-syntax-ns%23%3E%0APREFIX+rdfs:+%3Chttp://www.w3.org/2000/01/rdf-schema%23%3E%0A%0ASELECT+?V+%7B+BIND%28afn:context%28%27CONTEXT:DATASET%27%29+AS+?V%29+%7D GET - java.io.IOException: subscription cancelled
2025-03-10T19:22:41.9795050Z Mar 10, 2025 7:22:41 PM jdk.internal.net.http.Http1AsyncReceiver onReadError
```

There are still errors and failing CI jobs.

 - [ ] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/http/HttpLib.java
+++ b/jena-arq/src/main/java/org/apache/jena/http/HttpLib.java
@@ -183,7 +183,10 @@ public class HttpLib {
         else if ( inRange(httpStatusCode, 200, 299) ) {
             // Success. Continue processing.
         }
-        else if ( inRange(httpStatusCode, 300, 399) ) {
+        // update end of range to 599 to cover exception for larger range
+        // to debug if replacing exception(response, httpStatusCode) removes
+        // CI failure
+        else if ( inRange(httpStatusCode, 300, 599) ) {
             // We had follow redirects on (default client) so it's http->https,
             // or the application passed on a HttpClient with redirects off.
             // Either way, we should not continue processing.
@@ -270,7 +273,8 @@ public class HttpLib {
      * This consumes the response body.
      */
     static HttpException exception(HttpResponse<InputStream> response, int httpStatusCode) {
-        URI uri = response.request().uri();
+        // URI not used
+       // URI uri = response.request().uri();
         InputStream in = response.body();
         if ( in == null )
             return new HttpException(httpStatusCode, HttpSC.getMessage(httpStatusCode));
@@ -281,6 +285,7 @@ public class HttpLib {
                 if ( msg.isBlank())
                     msg = null;
             } catch (RuntimeIOException e) {
+                LOG.warn("HttpException exception(hr<is> r, int hsc) msg:{} stack trace: {}", e.getMessage(), e.getStackTrace().toString());
                 msg = null;
             }
             return new HttpException(httpStatusCode, HttpSC.getMessage(httpStatusCode), msg);
--- a/jena-fuseki2/jena-fuseki-main/pom.xml
+++ b/jena-fuseki2/jena-fuseki-main/pom.xml
@@ -215,6 +215,7 @@
           <includes>
             <include>**/TS_*.java</include>
           </includes>
+            <argLine>-Dfuseki.loglogging=true -Djdk.httpclient.HttpClient.log=errors,requests,headers,frames[:control:data:window:all..],content,ssl,trace,channel</argLine>
         </configuration>
       </plugin>
 
--- a/jena-fuseki2/jena-fuseki-main/src/test/resources/log4j2-test.properties
+++ b/jena-fuseki2/jena-fuseki-main/src/test/resources/log4j2-test.properties
@@ -13,7 +13,7 @@ appender.console.layout.type = PatternLayout
 appender.console.layout.pattern = %d{HH:mm:ss} %-5p %-10c{1} :: %m%n
 #appender.console.layout.pattern = [%d{yyyy-MM-dd HH:mm:ss}] %-5p %-10c{1} :: %m%n
 
-rootLogger.level                  = INFO
+rootLogger.level                  = DEBUG
 rootLogger.appenderRef.stdout.ref = OUT
 
 logger.jena.name  = org.apache.jena
@@ -23,7 +23,7 @@ logger.arq-exec.name  = org.apache.jena.arq.exec
 logger.arq-exec.level = INFO
 
 logger.fuseki.name  = org.apache.jena.fuseki
-logger.fuseki.level = WARN
+logger.fuseki.level = INFO
 
 ## Some tests correctly log warnings. TS_PrefixesService
 logger.fuseki-fuseki.name  = org.apache.jena.fuseki.Fuseki
@@ -33,7 +33,7 @@ logger.fuseki-autoload.name  = org.apache.jena.fuseki.main.sys.FusekiAutoModules
 logger.fuseki-autoload.level = ERROR
 
 logger.http.name  = org.apache.jena.http
-logger.http.level = INFO
+logger.http.level = DEBUG
 
 logger.riot.name  = org.apache.jena.riot
 logger.riot.level = INFO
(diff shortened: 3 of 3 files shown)


jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkAdapter.java
```diff
@@ -89,10 +84,18 @@ public DatasetGraph getDataset() {
     }
 
     @Override
-    public QueryExec query(Query query) { return QueryExec.adapt(conn.query(query)); }
+    public QueryExec query(Query query) {
+        try (QueryExecution qExec = conn.query(query)) {
+        return QueryExec.adapt(qExec);
```
afs, line 89: The adapter implements `close()` by calling `close()` on the wrapped QueryExecution object.

This change calls close() on the QueryExecution before leaving this method.

(same below)

Aklakan: IMHO this doesn't look right - this would return an adapter backed by a closed QueryExecution object.

(same below)

OyvindLGjesdal: Removed. I think I don't have the background and skill for creating a fix here, but hopefully I can help to debug.

afs:
Unrelated but relevant:
The PR #3077 seems to improved the situation for the UI step in the build. 🤞🏻
It does at least mean the UI code can be updated to use the latest cypress.

(There is a codemirror update in backlog which seems to need some code fixes.)

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkAdapter.java
afs: > Removed. I think I don't have the background and skill for creating a fix here, but hopefully I can help to debug.

Personally, I don't what the problem is! Fixing it would be easier otherwise. 😞 

1. It is only on github actions
1. It is almost always one of a few tests that fail e.g. `TestFusekiShaclValidation.shacl_no_data_graph`
1. Very occasional a different test will fail.
1. One server per test seems more reliable than one server per test suite.
1. Not stopping any servers and just leaving it lying around does not fix the problem.

At the moment, I have a trial system where the problem tests are isolated in their own test suite.  I'm runnign it on my personal GH account so as not to drown the ASF project.  https://github.com/afs/jena/tree/tests-experiment

`TestFusekiShaclValidation.shacl_no_data_graph` will intermittently fail when run in its own test suite with its own server setup. But the test is single threaded and the server side operation does execute and does log returning the expected 404. But the test does not receive the response ("0 bytes"). Even if the server is never closed!

(I'm likely to run out of GH free-use resources this accounting period.)

PS The UI steps can fail because of "yarn install" failure. This may be because I'm using theGH mirror registry too much.

afs:
Today:
20 runs, each of both java 17 and java 21, so 40 builds, on Linux.

2 failures of Java17/ TestFusekiShaclValidation2.shacl_no_data_graph (the isolated form of this test.
1 yarn install failure.

```
[INFO] error An unexpected error occurred: "https://registry.yarnpkg.com/lru-cache/-/lru-cache-10.4.3.tgz: Request failed \"500 Internal Server Error\"".
```

afs:
I added the logging from the PR - thank you for the suggestions @OyvindLGjesdal.

50 odd runs later - it seems that the addition of java.net.http logging makes the test TestFusekiShaclValidation (the more common failure) more reliable.

afs:
Hi @OyvindLGjesdal,

What does the situation look like to you? Are you seeing what I see?

The test files in this PR aren't one I have found to be problematic so I'm wondering if there is some external factor that changes the situation.

OyvindLGjesdal:
Hi @afs 

I think you are right that the tests I added are not related to the failures.

It looks to me that (one of) the errors is that HttpLib.exception(httpResponse, httpStatusCode) can sometimes cause failures.

My understanding that I’m not confident in:

The `TestFusekiShaclValidation.shacl_no_data_graph` failure  seems to go away by removing usage of  the method `HttpException exception(httpResponse, httpStatusCode)` in `  `void handleHttpStatusCode(HttpResponse<InputStream> response)` in the HttpLib class, and replacing it with the exception handling and `void finish(httpResponse<InputStream>)` that is also used for the case of 300-399 return codes. This removes use of the httpResponse body in the returned exception message, so this is  for experimenting, and no fix. I also see other errors sometimes that may not have been there before. The exception method is only used in these cases for 400-499 and  500-599. Could consume 

Most notable difference to me is that `finish(hr)->finish->(in)->consume(final in)` does not close the `in` and  the `exception method does (using finally { IO.close(in); }`).

I *think* this is the steps in `shacl_no_data_graph` when the test fails: 

* `expect404(…)` passes.
* `conn.update("CLEAR ALL»);` then tries to run, but the connection is already closed (sometimes). 

## After the change
There are also some other new(?) CI errors  that then show up, like in TestFusekiStdReadOnlySetup.stdSetup_endpoint_7, which returns:

```
 TestFusekiStdReadOnlySetup.stdSetup_endpoint_7:112 PUT http://localhost:41455/ds/data :: ==> expected: <405> but was: <-1>
```

## Other things 

 * It looks like the function copied from commons-io for consume has new code, I tried it, but it made no difference. Have reverted.
 * There was a similar error message in selenium https://github.com/SeleniumHQ/selenium/issues/11798  caused by a bug in the HttpClient which was turned into a JDK bug and fixed in Java22 (https://bugs.java.com/bugdatabase/view_bug?bug_id=8304701). Using the workaround of setting Redirect.NONE in the default http client had no effect afaik. Could still test if current main runs without issues on 22 in github-actions.

afs:
Hi @OyvindLGjesdal 

We're seeing the same sort of thing.

I think you may be on to something about `conn.update("CLEAR ALL")`.

I can't explain why its always the same test that fails but one server per test + protected "CLEAR ALL" is working (caveat - its now a quieter time of day).

I've got some changes that can go to a PR for `main` when cleaned up.

The most annoying failure is `TestFusekiShaclValidation.shacl_no_data_graph` because it is in `jena-fuseki-main`, whereas the other typical failures are in `jena-integration-tests` which nothing depends. Running "fail at end" would at least continue most code checking.

* reinforced the exception handling path to always consume a potential message and always close the connection,
* Made tests run one server per test (code-change configurable to one per suite) - this reduces the failure rate to 1 in 10
* Protected `conn.update("CLEAR ALL")` - the error rate on java test failures is zero
* Move most of the typically failing tests into their own classes for convenience.

These rates are job failures, runs of both Java17 and Java21, so the per-build twice as good (it is very rare to have both fail).
The UI issue looks to be an unrelated random failure.

afs:
I've looked through this PR and made sure everything is in #3129, except one log message (unexpected IOException) which may occur for different reasons (erratic remote server or proxy, networking goes bad, ... ) and which might then flood the logs.

PR #3203, opened 2025-05-18
GH-2809: Integration tests for JenaSystem#init

### LINKED ISSUE #2809: Integration tests for JenaSystem#init
### Version

5.3.0-SNAPSHOT

### Feature

Implementing a JUnit integration test for `JenaSystem#init` is straightforward, but it must be run as a single-method test to ensure a fresh JVM. Only then can we effectively debug and test the execution of static initializers.

```java
    @Test
    public void init() {
        assertDoesNotThrow(
                () -> JenaSystem.init());
    }
```

A regression test for [apache/jena#2675](https://github.com/apache/jena/issues/2675) would look like this:

```java
    @Test
    public void initRDFConnectionFuseki() {
        try (RDFConnection conn = RDFConnectionFuseki.service("http://localhost:3030/ds").build()) {
            assertTrue(true);
        }
    }
```

For [apache/jena#2787](https://github.com/apache/jena/issues/2787), the regression test is as follows:

```java
    @Test
    public void initParallel() {
        var pool = Executors.newFixedThreadPool(2);
        try {
            var futures = IntStream.range(1, 3)
                    .mapToObj(i -> pool.submit(() -> {
                        if (i % 2 == 0) {
                            ModelFactory.createDefaultModel();
                        } else {
                            JenaSystem.init();
                        }

                        return i;
                    }))
                    .toList();
            var intSet = new HashSet<Integer>();
            assertTimeoutPreemptively(
                    Duration.of(4, ChronoUnit.SECONDS),
                    () -> {
                        for (var future : futures) {
                            intSet.add(future.get());
                        }
                    });
            assertEquals(2, intSet.size());
        } catch (Exception e) {
            throw new RuntimeException(e);
        } finally {
            pool.shutdown();
        }
    }
```

When running multiple tests within a suite or a single file, static initializers may be triggered unpredictably, as any test could initialize them, and they cannot be re-triggered within the same JVM instance (or class loader). I attempted various JUnit annotations and methods to make these tests work within a Maven build, but found no JUnit-based solution. However, I found a workaround using the JMH benchmark framework, which runs each benchmark in a freshly started JVM.

JMH uses code generation, which needs to be triggered whenever the benchmark code is changed. For JetBrains IDEA, there's a [plugin](https://plugins.jetbrains.com/plugin/7529-jmh-java-microbenchmark-harness) that automates code generation in the background. Unfortunately, for other IDEs like Eclipse, this process has to be triggered manually by running `mvn clean install`.

I would like to submit a PR to extend the integration tests. However, I’d like to confirm whether introducing JMH into the `jena-integration-tests` is accepted by the Jena developer community before proceeding further.

Yes

arne-bdt:
The test code triggered a "Deadlock in JenaSystem.init()" (GH-2787 https://github.com/apache/jena/issues/2787) in Jena 5.2.0. 
--> It seems to be stable now.

GitHub issue resolved #2809

Pull request Description:
The tests in https://github.com/apache/jena/issues/2809 reliably reproduced the Deadlock in JenaSystem.init() from GH-2787.
--> Currently these tests don´t seem to trigger any deadlock. They should be useful to keep it that way.

Since the output of JMH is set to SILENT, there is no verbose output, if the tests run successfully.

Downside: These tests add about 7 seconds to the test execution. 

----

 - [x] Tests are included.
 - [x] Key commit messages start with the issue number (GH-2809)

----

### DIFF
--- a/jena-integration-tests/pom.xml
+++ b/jena-integration-tests/pom.xml
@@ -152,6 +152,18 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-generator-annprocess</artifactId>
+      <scope>test</scope>
+    </dependency>
+
   </dependencies>
 
   <build>
@@ -186,6 +198,11 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <compilerArgs>
+            <arg>-proc:full</arg>
+          </compilerArgs>
+        </configuration>
       </plugin>
 
       <plugin>
--- a/jena-integration-tests/src/test/java/org/apache/jena/sys/TS_Sys.java
+++ b/jena-integration-tests/src/test/java/org/apache/jena/sys/TS_Sys.java
@@ -1,0 +1,30 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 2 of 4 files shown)


afs:
> Downside: These tests add about 7 seconds to the test execution.

7 seconds isn't long.

We can at least add these to the test suite - we may then decide not to run them everytime.

afs:
> Unfortunately, I cannot reproduce the error on my machine.
> 
> I can't see, how or why this might have something to do with my PR...
> 
> Does anyone have an idea?

This is a github-only error. The GH build servers can be assumed to be heavily loaded VMs. We have seen pause times of 60 seconds and longer in code that isn't even doing any OS calls.

While we made some changes so that commonly recurring github specific errors went away, it wasn't a deep understanding of the root cause. It is impossible to reproduce and I only have so many GH credits per month to run 20+  jobs in parallel repeatedly.

The failing tests involve networking but the tests are deterministic start server, test, stop server - so how a "no bytes" message can come back when the test server when the test hasn't run the code to shutdown the server is bizarre. It's not impossible that the issue is in the JDK/OS (runs out of some resource due to delayed cleanup - TCP cleanup is partly async) and/or the way JUnit fires the post-test clean-up. The situation of many servers start-stop is not realistic usage.

Ignore the failure - rerun the job. I did - it passed.

There has been a recent (in the last few days) recurrence of this after a quiet spell.

afs (approved):

PR #3239, opened 2025-06-03
Fixed ExprUtils.eval raising exception because of null dataset.

Aklakan:
This PR changes a couple of `ExecutionContext.create(null, context)` instances to `ExecutionContext.create(context)` because the former now raises an exception because of a null dataset.

I noticed this issue because in my code I used `ExprUtils.eval(expr, binding)` which suddenly stopped working.


afs:
Sorry about that.

Aklakan:
Thanks, no worries.

PR #3264, opened 2025-06-18
Make Node and ExprTransform affect OpPropFunc arguments

Aklakan:
As the title says. (If desired I can create an issue). The issue occurred to me when I was trying to post process variable names in Ops after the optimizer had already run. OpPropFunc arguments were not affected then.
One detail question is whether the NodeTransform should affect the property function URI.
My argument against it are:
* It's probably desired to prevent a node transform from mapping a  property function name to a blank node or literal.
* Renaming property function names in the BGPs could be done before the optimizer runs.
* In the worst case, it would still be possible to subclass TransformCopy just for that case.

----

 - [x] Tests are included.
 - (Key commit messages start with the issue number (GH-xxxx))

----


afs:
> * It's probably desired to prevent a node transform from mapping a  property function name to a blank node or literal.

I agree.
Changing the URI is changing the whole functionality. A transform can do that by replacing the whole unit. The replacement may not even be a property function.

afs (approved):

PR #3266, opened 2025-06-18
GH-3258: Support for bzip2 encoding in upload via GSP and SPARQL LOAD.

### LINKED ISSUE #3258: Support for uploading compressed data with GSP and SPARQL
### Version

5.5.0-SNAPSHOT

### Feature

Trying to upload compressed files via the Fuseki UI (uses graph store protocol) and also via SPARQL LOAD fails with compressed data, which is cumbersome for large files:

Image

It would be great if (compression) encoded uploads would work out of the box with GSP and SPARQL LOAD.

Perhaps?

Aklakan:
GitHub issue resolved #3258 

Pull request Description: Proposal to add bzip2 as a recognized encoding to `HttpLib` and `DataUploader`.

Note: The Web server from which one attempts to upload a file needs to be configured to serve content type and encoding separately:

```
curl -I http://myserver/file.nt.bz2

Content-Type: application/n-triples
Content-Encoding: bzip2
```

For example, by default, the Apache Web server would serve .bz2 files as `Content-Type: application/x-bzip2`. This way a client wouldn't know what actual content is being encoded.

----

 - (No extra tests are so far included because the change turned out to be trivial)

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/http/HttpLib.java
+++ b/jena-arq/src/main/java/org/apache/jena/http/HttpLib.java
@@ -43,6 +43,7 @@ import java.util.function.Function;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.InflaterInputStream;
 
+import org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream;
 import org.apache.jena.atlas.RuntimeIOException;
 import org.apache.jena.atlas.io.IO;
 import org.apache.jena.atlas.lib.IRILib;
@@ -150,6 +151,8 @@ public class HttpLib {
                     return responseInput;
                 case "gzip" :
                     return new GZIPInputStream(responseInput, 2*1024);
+                case "bzip2" :
+                    return new BZip2CompressorInputStream(responseInput, true);
                 case "inflate" :
                     return new InflaterInputStream(responseInput);
                 case "br" : // RFC7932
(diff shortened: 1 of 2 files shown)


Aklakan:
The PR turned out to be straight-forward - gz was already supported and I added bzip2 next to it.
In the future a registry system could be used to support all formats of commons-compress - but that'd be a new issue.

image

Aklakan:
As a side note, I am aware that `bzip2` is not an official encoding value of [rfc9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-content-codings), but bzip2 is relevant because there are tools that can perform parallel processing directly on bzip2 compressed files, such as Apache Spark or lbzip2.

afs:
> In the future a registry system 

There are very few encodings - and this isn't particularly open to protocol negotiation.

One reason maybe that a one-time compress-protocol-decompress cycle is not much gain (only on long distance, i.e. slower, connections) and sometimes worse than "just send it" for high-bandwidth connections.

(having a compress-decompress co-processor subsystem would change that - but that is often applied transparently)

If the file is already compressed, it can make sense.

Does the UI need to change?

Aklakan:
The use case for this PR was exactly an upload of local bzip2-compressed data via a slow connection.

No UI change is needed - the screenshot in my last message with the green status bar for the bz2 file is taken from a fuseki built with this PR.

PR #3295, opened 2025-07-04
Fix for E_Function.equals() and hashCode() to also consider the functionIRI

Aklakan:
Pull request Description: E_Function instances compared to equals regardless of their functionIRI field. This causes caching issues such as with the service enhancer, where the same cache hit is produced for queries that should actually be different.

In our case, two queries that differed only by the geosparql filter functions geof:sfIntersects and geof:sfContains were incorrectly considered to be equal.

----
 - [x] Tests are included.

----


afs (approved):

afs:
Thanks!

PR #3347, opened 2025-07-25
GH-3322: [jena-ontapi] add assemblers

### LINKED ISSUE #3322: New ontapi and Jena assembers.
Issue created from [a comment on #3152](https://github.com/apache/jena/issues/3152#issuecomment-3070046838)

@afs
> [@sszuev](https://github.com/sszuev) - jena-ontapi question - does it wire itself in the assemblers for the old API?

@sszuev 
> "new" model doesn't work with `Assembler` and `FileManger` classes directly, it contains its own interface `GraphRepository`, which is open and can be extended.
> Having the specification, I can extend document management in the "new" model, now this management is very simple, but it seems to me sufficient in most cases.

sszuev:
GitHub issue resolved #3322

Pull request Description:

Due to a conflict (`assembler.ttl`) I had to copy-paste some components from jena-arq (`ReasonerFactoryAssembler`, etc), so it may not be possible to merge this MR right now and we need to wait until the old API is completely removed.

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-ontapi/src/main/java/org/apache/jena/ontapi/assemblers/DocumentGraphRepositoryAssembler.java
+++ b/jena-ontapi/src/main/java/org/apache/jena/ontapi/assemblers/DocumentGraphRepositoryAssembler.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.jena.ontapi.assemblers;
+
+import org.apache.jena.assembler.Assembler;
+import org.apache.jena.assembler.Mode;
+import org.apache.jena.assembler.assemblers.AssemblerBase;
+import org.apache.jena.ontapi.impl.repositories.DocumentGraphRepository;
+import org.apache.jena.rdf.model.Model;
+import org.apache.jena.rdf.model.RDFNode;
+import org.apache.jena.rdf.model.Resource;
+
+public class DocumentGraphRepositoryAssembler extends AssemblerBase {
+
+    /**
+     * example:
+     * <pre>{@code
+     * @prefix : <http://ex.com#> .
+     * @prefix oa: <https://jena.apache.org/ontapi/Assembler#> .
+     * @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+     *
+     * :repo a oa:DocumentGraphRepository ;
+     *      oa:graph :g1, :g2 .
+     *
+     * :g1 a oa:Graph ;
+     *      oa:graphIRI "vocab1" ;
+     *      oa:graphLocation "file:ontologies/vocab1.ttl" .
+     *
+     * :g2 a oa:Graph ;
+     *      oa:graphIRI "vocab2" ;
+     *      oa:graphLocation "file:ontologies/vocab2.ttl" .
+     * }</pre>
+     */
+    @Override
+    public Object open(Assembler a, Resource root, Mode mode) {
+        checkType(root, OA.DocumentGraphRepository);
+        DocumentGraphRepository repo = new DocumentGraphRepository();
+
+        Model model = root.getModel();
+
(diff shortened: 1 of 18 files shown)


jena-ontapi/src/main/java/org/apache/jena/ontapi/assemblers/DocumentGraphRepositoryAssembler.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.jena.rdf.model.Model;
+import org.apache.jena.rdf.model.RDFNode;
+import org.apache.jena.rdf.model.Resource;
+
+public class DocumentGraphRepositoryAssembler extends AssemblerBase {
+
+    /**
+     * example:
+     * <pre>{@code
+     * @prefix : <http://ex.com#> .
```
afs, line 34: Comment: using `PREFIX`, rather than `@prefix` - and no trailing DOT, is a more modern style.

jena-ontapi/src/main/java/org/apache/jena/ontapi/assemblers/DocumentGraphRepositoryAssembler.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.jena.ontapi.impl.repositories.DocumentGraphRepository;
+import org.apache.jena.rdf.model.Model;
+import org.apache.jena.rdf.model.RDFNode;
+import org.apache.jena.rdf.model.Resource;
+
+public class DocumentGraphRepositoryAssembler extends AssemblerBase {
+
+    /**
+     * example:
+     * <pre>{@code
```
afs, line 33: FYI only:

Eclipse IDE makes a complete mess of the javadoc

It also messes up javdoc that links to inner classes.

In both cases, it's the IDE. When I look at the generated HTML it looks OK.

It does rather make the case strongly for markdown in javadoc!

afs:
Is there anything on the website to change? (not necessary for merging this PR)

sszuev:
> Is there anything on the website to change? (not necessary for merging this PR)

The documentation needs to be updated. 

But I consider this pull request as a draft. Due to a conflict, I had to copy-paste a couple of classes from jena-core (e.g.ReasonerFactoryAssembler), in the final version they need to be replaced with the original ones. But this is possible only after/during completely removing the old model. At least I don't see a better way.

afs:
I took a look at removing jena-core org.apache.jena.ontology.

There are two references outside jena-core (`schemagen` in jena-cmds, and some old code in TDB1 for old non-transactional uses cases). That's all good.

Inside `jena-core` the situation is not too bad but not trivial.

A few factories have ontology operations. Mostly, they need to be deprecated and point to `jena-ontapi`.

The two big areas are the assembler subsystem and the reasoner support for owl reasoning.  These look more complicated to disentangle. 

It is not that the reasoner code uses org.apache.jena.ontology (mostly, it doesn't in 'main', some in 'test') but it needs the OWL vocabulary file).

So I'm happy to merge this PR - including the noted caveat about copied code. OK?

Doing changes in steps is more manageable; and byJena6 we will definitely be in a better place even if the time needed to do the whole job doesn't line up. 

We can signal the intent and not be morally bound to a "Jena7" for the final step.

sszuev:
> I took a look at removing jena-core org.apache.jena.ontology.
> 
> There are two references outside jena-core (`schemagen` in jena-cmds, and some old code in TDB1 for old non-transactional uses cases). That's all good.
> 
> Inside `jena-core` the situation is not too bad but not trivial.
> 
> A few factories have ontology operations. Mostly, they need to be deprecated and point to `jena-ontapi`.
> 
> The two big areas are the assembler subsystem and the reasoner support for owl reasoning. These look more complicated to disentangle.
> 
> It is not that the reasoner code uses org.apache.jena.ontology (mostly, it doesn't in 'main', some in 'test') but it needs the OWL vocabulary file).
> 
> So I'm happy to merge this PR - including the noted caveat about copied code. OK?
> 
> Doing changes in steps is more manageable; and byJena6 we will definitely be in a better place even if the time needed to do the whole job doesn't line up.
> 
> We can signal the intent and not be morally bound to a "Jena7" for the final step.

Not many people use jena-ont-api and nobody uses assemblers yet. I think reasoner assemblers and ont-api assemblers are two different things. Currently (in this PR) ont-api assemblers use a cut down version of reasoner assemblers with `oa:` prefix, while they should use full-fetched reasoner assemblers with `ja:` prefix. The reason why I can't use ja-assemblers is that there is a conflict declarations in assembler.ttl.

But, with this in mind we can take a risk and merge this pull request. Any bugs can be fixed later, and the transition to ja-resoners can be done at the time of removing the old ont-api.

Another minor point: in this PR I use literals for constants (like spec), while old onr-api-assemblers use uri-resources. Not sure is this an issue.

afs:
The interaction with reasoners is that there are rule sets for subsets of OWL.

I don't use ontologies much and in the communities I'm active in, OWL is considered useful for describing data, less emphasis on OWL-inference , possibly because in an application that is serious about it, it would use a complete 3rd party reasoner.

> Not many people use jena-ont-api

I'm working on a PR that will deprecate jena-core `OntModel`.

The project can also send out a survey email asking who uses Jena for ontology work.

I would have though the availability of OWL2 in jena-ontapi would be a strong positive to migrate.

PR #3425, opened 2025-09-05
kryo5 upgrade

Aklakan:
GitHub issue resolved (no issue created yet).

Pull request Description: Upgrade to kryo5. ~~Need yet to test whether spatial indexes created with kryo4 can be loaded with kryo5. Also, there was an inconsistent use of `ConcurrentHashMap` and `LinkedHashMap` that was revealed when a test case failed with kryo5.~~
In my first experiments I was not able to load a spatial index created with kryo4 using kryo5.

----

 - ~~[ ] Tests are included.~~

----

### DIFF
--- a/jena-geosparql/pom.xml
+++ b/jena-geosparql/pom.xml
@@ -65,8 +65,8 @@
     </dependency>
 
     <dependency>
-      <groupId>com.esotericsoftware</groupId>
-      <artifactId>kryo-shaded</artifactId>
+      <groupId>com.esotericsoftware.kryo</groupId>
+      <artifactId>kryo5</artifactId>
     </dependency>
 
     <!-- Non-free; testing only -->
--- a/jena-geosparql/src/main/java/org/apache/jena/geosparql/kryo/EnvelopeSerializer.java
+++ b/jena-geosparql/src/main/java/org/apache/jena/geosparql/kryo/EnvelopeSerializer.java
@@ -20,10 +20,10 @@ package org.apache.jena.geosparql.kryo;
 
 import org.locationtech.jts.geom.Envelope;
 
-import com.esotericsoftware.kryo.Kryo;
-import com.esotericsoftware.kryo.Serializer;
-import com.esotericsoftware.kryo.io.Input;
-import com.esotericsoftware.kryo.io.Output;
+import com.esotericsoftware.kryo.kryo5.Kryo;
+import com.esotericsoftware.kryo.kryo5.Serializer;
+import com.esotericsoftware.kryo.kryo5.io.Input;
+import com.esotericsoftware.kryo.kryo5.io.Output;
 
 public class EnvelopeSerializer extends Serializer<Envelope> {
     @Override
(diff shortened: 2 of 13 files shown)


Aklakan:
I was not able to load a spatial-index created with kryo4 using kryo5 - it fails to deserialize the ConcurrentHashMap whose registration was missing. Kryo4 auto-registered classes in that case, whereas kryo5 by default requires registration. There is a flag `kryo5.setRegistrationRequired(false);` but this does not make loading work.

The question is whether to schedule a kryo5 upgrade for Jena6.

afs:
> The question is whether to schedule a kryo5 upgrade for Jena6.

Probably, yes. It is safer.

A major version change to Jena comes every 2 years when the supported java version changes so the opportunity is now.

Running both, while possible, is a maintenance cost.

Even if Kryo4 works now, bug fixes and security fixes will likely be to kryo5 first. I don't see much development around kryo4.

Choosing Kryo5 does not impact applications using Kryo4 elsewhere in their codebase.

afs:
No rush - trying to get all the "bumps" done early in the 6. cycle,; this one is isolated so not getting in the way of anything outside of geosparql - can this merged now or are there other changes necessary?

Aklakan:
There are a couple of small QoL improvements that I should add to avoid confusion.

I think it is still possible to read the metadata (from kryo's perspective a plain string) of spatial indexes created with kryo4 using kryo5.
- If so, then i can add a check whether the version field is `2.0.0` and raise an exception that this format is no longer supported. Deleting the spatial index file should re-create it on Fuseki start up.
- I should use a different version value for kyro5 in the metadata, such as `3.0.0`.

Aklakan:
I added the version check - with an old index it now outputs:

```
org.apache.jena.geosparql.spatial.SpatialIndexException: Spatial index version 2.0.0 is no longer supported.
Move or delete this file to allow for creation of a new index in its place: /foo/bar/spatial.index
	at org.apache.jena.geosparql.spatial.index.v2.SpatialIndexIoKryo.load(SpatialIndexIoKryo.java:219)
```

afs:
Thanks for updating this.
I was doing the same at the same time!

Is there anything needed for the Jena6 announcement?
Is there anything else for geosparql that is needed for the Jena6 announcement?

Aklakan:
:laughing: I just noticed that this PR hasn't been updated while in a yearly recap telco.

> Is there anything needed for the Jena6 announcement?

So the kryo4->kryo5 upgrade breaks existing spatial indexes which need to be regenerated.
With Fuseki, deleting the spatial index file should usually regenerate it on startup.

afs:
> So the kryo4->kryo5 upgrade breaks existing spatial indexes which need to be regenerated.
> With Fuseki, deleting the spatial index file should usually regenerate it on startup.

Could you write a paragraph I can copy-and-paste please?

Aklakan:
Like this?

* `jena-geosparql`: Jena 6 requires re-creation of spatial index files that were generated with earlier versions. The reason is a bump to the latest major version of the kryo serialization framework dependency. 
In general, **it is recommended to back up existing databases** including the spatial indexes prior to engaging in the migration process. The [GeoSPARQL Assembler](https://jena.apache.org/documentation/geosparql/geosparql-assembler.html) will automatically re-create all referenced absent spatial index files. Hence, moving or deleting existing spatial indexes is usually sufficient. The Fuseki SPARQL server runs spatial index creation of GeoSPARQL assemblers during server start. This process may take a while depending on the sizes of the involved datasets.

afs:
Thank you for the ANN material.

PR #3441, opened 2025-09-15
Choose HTTP headers for non-parsed queries based on API usage.

Aklakan:
GitHub issue resolved (Can create one if desired)

Pull request Description: Headers configured on `RDFConnectionRemoteBuilder` were not passed on to the created `QueryExecHTTPBuilder` instances when parseCheck was disabled. For example, setting the select-header to thrift on a connection to DBpedia would unexpectedly work - although DBpedia (so far) does not support this content type.

* Added "set-acceptHeader" methods to `QueryExecHTTPBuilder`. Headers are on response format level - e.g. one graph header for describe, construct and GSP GET. If desired, fine grained setters for each aspect could be added.
*  `RDFConnectionRemoteBuilder` initializes created `QueryExecHTTPBuilder` instances with its headers.
* Added `QueryExecHTTP.getAcceptHeader` methods to inspect the headers for the query types.

Unrelated minor fix: Quad.isConcrete() would raise NPE if the graph was null  ("true triple in quad").

----

 - [x] Tests are included.

----


jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
```diff
@@ -316,53 +312,54 @@ protected void setQuery(Query query, String queryStr) {
             queryType = query == null ? QueryType.UNKNOWN : query.queryType();
         }
 
+        /**
+         * For parsed queries: Derive and return the appropriate accept header.
+         * For non-parsed queries: This method returns null. The header is then chosen
+         * based on the used method of the query exec. For example,
+         * {@link QueryExecHTTP#select()} will use the configured select header.
+         */
+        private String deriveAcceptHeaderFromQueryType() {
```
Aklakan: I think this method can now be removed: QueryExecHTTP now knows all headers and can choose among the configured ones when `QueryExecHTTP.[queryType]` is called.

jena-examples/src/main/java/rdfconnection/examples/RDFConnectionExampleDBpedia01.java
```diff
@@ -0,0 +1,55 @@
...
+ * <pre>
+ * Trying to query with content type: application/sparql-results+thrift
+ * Request failed: Not Acceptable
+ *
+ * Trying to query with content type: application/sparql-results+json
+ * -------------------------
+ * | l                     |
+ * =========================
+ * | "Jena (Framework)"@de |
+ * -------------------------<
```
afs:
```suggestion
 * -------------------------
```

afs:
> (Can create one if desired)

Yes, please. Separating the report from the solution makes for a better long term record.

jena-rdfconnection/src/test/java/org/apache/jena/rdfconnection/TestRDFConnectionHTTP.java
```diff
@@ -20,80 +20,176 @@
...
+                .acceptHeader("o") // Override header
+                .build() ) {
+            QueryExecHTTP qe = (QueryExecHTTP)QueryExec.adapt(qExec);
+            assertEquals("o", qe.getAppProvidedAcceptHeader());
+            assertEquals("o", qe.getAcceptHeaderSelect());
+            assertEquals("o", qe.getAcceptHeaderAsk());
+            assertEquals("o", qe.getAcceptHeaderDescribe());
+            assertEquals("o", qe.getAcceptHeaderConstructGraph());
+            assertEquals("o", qe.getAcceptHeaderConstructDataset());
+        }
```
Aklakan: I think that when setting the override header on the builder, then the concrete QueryExecHTTP instance should report that header for all query types.

jena-arq/src/main/java/org/apache/jena/sparql/exec/http/QueryExecHTTP.java
```diff
@@ -133,11 +149,15 @@ public QueryExecHTTP(String serviceURL, Query query, String queryString, int url
...
+        this.selectAcceptHeader = selectAcceptheader;
+        this.askAcceptHeader = askAcceptHeader;
+        this.graphAcceptHeader = graphAcceptHeader;
+        this.datasetAcceptHeader = datasetAcceptHeader;
         // Important - handled as special case because the defaults vary by query type.
         if ( httpHeaders.containsKey(HttpNames.hAccept) ) {
-            if ( this.appProvidedAcceptHeader != null )
-                this.appProvidedAcceptHeader = httpHeaders.get(HttpNames.hAccept);
+            if ( this.overrideAcceptHeader != null )
+                this.overrideAcceptHeader = httpHeaders.get(HttpNames.hAccept);
```
Aklakan: I think that this overrideAcceptHeader should eventually be handled exclusively in the builder - but since this feature is part of the public API I better not touch it.

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
Aklakan: Removed.

Aklakan:
I have completed my revisions.

afs:
Please rebase to current main.
I tried to compare using GH and there are too many files!

Aklakan:
Should be rebased - I had removed a couple of whitespaces just before your message :smile:

afs:
> Should be rebased - I had removed a couple of whitespaces just before your message 😄

(claimed) 7000+ different files down to 12 files.

jena-rdfconnection/src/test/java/org/apache/jena/rdfconnection/TestRDFConnectionHTTPHeaders.java
afs, line 1: Add to `TS_RDFConnection`.

jena-rdfconnection/src/test/java/org/apache/jena/rdfconnection/TestRDFConnectionHTTP.java
```diff
@@ -1,99 +0,0 @@
-/*
```
afs, line 1: Where did the tests from this class go to?
(It should have been added to `TS_RDFConnection` when added.)

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
```diff
@@ -221,163 +220,54 @@ public Graph queryDescribe(String queryString) {
...
         return
             Txn.calculateRead(this, ()->{
-                try ( QueryExec qExec = query(queryString, QueryType.ASK) ) {
+                try ( QueryExec qExec = query(queryString) ) {
                     return qExec.ask();
                 }
             } );
     }
 
-    /**
```
afs, line 230: Could you say which this change is in the PR? The `QueryType` may not be particulaly useful in this impl but when inherited it might be useful.

jena-rdfconnection/src/test/java/org/apache/jena/rdfconnection/TestRDFConnectionHTTPHeaders.java
```diff
@@ -0,0 +1,121 @@
...
+            QueryExecHTTP qe = (QueryExecHTTP)QueryExec.adapt(qExec);
+            assertEquals("o", qe.getAppProvidedAcceptHeader());
+            assertEquals("o", qe.getAcceptHeaderSelect());
+            assertEquals("o", qe.getAcceptHeaderAsk());
+            assertEquals("o", qe.getAcceptHeaderDescribe());
+            assertEquals("o", qe.getAcceptHeaderConstructGraph());
+            assertEquals("o", qe.getAcceptHeaderConstructDataset());
+        }
+    }
+}
```
afs, line 153: There is a missing test:

```
        .acceptHeader("o")
        .acceptHeaderAskQuery("a")
```
which I would expect to set just the ASK query type to "a" and leave everything else as "o"

jena-rdfconnection/src/test/java/org/apache/jena/rdflink/TestRDFLinkHTTPHeaders.java
```diff
@@ -0,0 +1,97 @@
...
+
+package org.apache.jena.rdflink;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.apache.jena.sparql.exec.http.QueryExecHTTP;
+import org.apache.jena.sparql.exec.http.QueryExecHTTPBuilder;
+import org.junit.jupiter.api.Test;
+
+public class TestRDFLinkHTTPHeaders {
```
afs, line 27: I don't see any tests for `acceptHeaderQuery`.

afs, line 27: Add to `TS_RDFLink` please.

jena-arq/src/main/java/org/apache/jena/sparql/exec/http/QueryExecHTTP.java
```diff
@@ -119,10 +117,28 @@ public static QueryExecHTTPBuilder service(String serviceURL) {
     private HttpClient httpClient = HttpEnv.getDftHttpClient();
     private Map<String, String> httpHeaders;
 
+    public QueryExecHTTP(String serviceURL, Query query, String queryString, int urlLimit,
```
afs, line 126: `public` does not like it is necessary

So Jena6, shall we make it `protected` and have only one constructor.

If so, add `@Deprecated` with a javadoc comment saying the constructor will become protected on both constructors.

afs:
> * Minor unrelated fix: Quad.isConcrete() would raise NPE if the graph was null  ("true triple in quad").

I created issue #3475 for this.
Could it be a separate commit in this PR please?

Aklakan:
> I created issue #3475 for this.
> Could it be a separate commit in this PR please?

Done with https://github.com/apache/jena/pull/3476

jena-rdfconnection/src/main/java/org/apache/jena/rdflink/RDFLinkHTTP.java
Aklakan: > The QueryType may not be particularly useful in this impl

The main reason is that I though that `QueryExecBuilder` should be the single point of truth, and logic outside of it should be avoided. `QueryExecBuilder` does not have an API to which the query type can be passed, and in these specific cases, the `QueryExec` created by the `QueryExecBuilder` attempts to execute the string with a certain query type just one line below - so it felt redundant.

Of course, one could argue that e.g. in the future `QueryExecBuilder` could receive e.g. `query(queryString, queryType)` method variants. Then the query type here could eventually be passed to the extended builders. But as these specific methods are shortcuts with consistent use - the appropriate exec method, such as `.ask()` is just one line below -  it wouldn't add anything here.

So I couldn't find good arguments for why to have this extra complexity in the shortcuts.

jena-rdfconnection/src/test/java/org/apache/jena/rdfconnection/TestRDFConnectionHTTPHeaders.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Test void test_headers_non_parsed_fallback() {
+        try (RDFConnection conn = RDFConnectionRemote.service("urn:dummy-service")
+            .parseCheckSPARQL(false)
+            .acceptHeaderQuery("f") // fallback
+            .acceptHeaderSelectQuery(null)
+            .acceptHeaderAskQuery(null)
+            .acceptHeaderGraph(null)
+            .acceptHeaderDataset(null)
```
Aklakan, line 63: There are `acceptHeaderQuery` tests are in RDF**Connection**HTTPHeaders - if desired, they could be duplicated for the RDFLink level.

Note: The comment on the field `acceptSparqlResults` originally read:

>  // All purpose SPARQL results header used if above specific cases do not apply.

So I interpreted this as a fallback. An alternative interpretation is that this sets *all* SPARQL query headers similar to `QueryExecHTTPBuilder.acceptHeader(headerString)`. In that case, separate SPARQL-only fields should be introduced that are not shared with GSP. `acceptHeaderQuery` would then set the fields for sparql graph results (construct and describe) but not for GSP retrieval.

jena-rdfconnection/src/test/java/org/apache/jena/rdfconnection/TestRDFConnectionHTTP.java
Aklakan: These tests used the `getAppProvidedAcceptHeader()` method, which I introduced only for testing in a previous PR and which was also stated in the javadoc ("For testing only."). In hindsight, a better solution would have been a package private method together with the test case in the `org.apache.jena.sparql.exec.http` package within jena-rdfconnection :/

The field and the method are now marked as deprecated - and the behavior for `getAppProvidedAcceptHeader` has changed  - so the tests for this property no longer apply.
In contrast, the new `QueryExecHTTP.getAcceptHeader*()` methods for the different query forms should be future proof.

The value of the `appProvidedAcceptHeader` used to rely on the QueryType - but since this wouldn't work generically for strings, this PR removed this logic. The header is now chosen based on the exec call such as `.ask()`.

jena-rdfconnection/src/test/java/org/apache/jena/rdfconnection/TestRDFConnectionHTTPHeaders.java
Aklakan: Added as `TestRDFConnectionHTTPHeaders.test_override_header_{02, 03}`.
Note, that order matters:

```
.acceptHeaderAskQuery("a")
.acceptHeader("o") // Sets everything to "o" -> overrides setting for "a".
```

The alternative would be to only compute the effective headers during build based on the explicitly set properties. What's your preference?

jena-rdfconnection/src/test/java/org/apache/jena/rdflink/TestRDFLinkHTTPHeaders.java
Aklakan: done.

Aklakan: There are tests in `RDFConnectionHTTPHeaders` for this which however are not duplicated on the RDFLink level. See my corresponding comment there about how to interpret this method.

jena-arq/src/main/java/org/apache/jena/sparql/exec/http/QueryExecHTTP.java
Aklakan: Done. The new constructor no longer uses the now deprecated override header, whereas the original one still supports it.

jena-rdfconnection/src/test/java/org/apache/jena/rdfconnection/TestRDFConnectionHTTPHeaders.java
Aklakan: done.

jena-arq/src/main/java/org/apache/jena/sparql/exec/http/QueryExecHTTP.java
Aklakan: Added handling of the overrideAcceptHeader to the builder - the new constructor of QueryExecHTTP no longer uses the parameter whereas the old constructor still supports it. This should resolve the issue.

jena-arq/src/main/java/org/apache/jena/http/sys/ExecHTTPBuilder.java
```diff
@@ -253,22 +262,58 @@ public Y substitution(Var var, Node value) {
...
+    /** Setting this header overrides any other header. */
     public Y acceptHeader(String acceptHeader) {
         Objects.requireNonNull(acceptHeader);
         this.appAcceptHeader = acceptHeader;
+        this.selectAcceptHeader = acceptHeader;
+        this.askAcceptHeader = acceptHeader;
+        this.graphAcceptHeader = acceptHeader;
+        this.datasetAcceptHeader = acceptHeader;
+        return thisBuilder();
+    }
```
Aklakan, line 274: This method now sets all header fields to the given value. Subsequent calls to e.g. `acceptHeaderSelectQuery()` can set specific headers again. The `appAcceptHeader` field is not used anymore.

jena-rdfconnection/src/test/java/org/apache/jena/rdfconnection/TestRDFConnectionHTTPHeaders.java
afs: When passing half-built builders around (uncommon) the later options should override.

PR #3464, opened 2025-09-23
Async HTTP

### LINKED ISSUE #3471: Mitigate cases where QueryExecHTTP.abort() hangs.
### Version

5.6.0-SNAPSHOT

### Feature

Aborting an HTTP query execution waits until the input stream of the HTTP response becomes available - which can take a while if e.g. the connection request hangs or the time to the first result row is high. Arguably, the expected behavior of `abort()` is to return quickly even in these situations. The underlying reason is, that the synchronous API of the HTTP client is used, whereas the async one would allow one to cancel the java future of the ongoing request. 

This issue was encountered while testing the execution tracker UI of #3184 , where clicking the abort button would hang.

Yes

Aklakan:
GitHub issue resolved #3471

This is the async HTTP part factored out from #3184 .

Pull request Description:
* Update HTTP(s) machinery to use async requests. This allows for `QueryExecHTTP.abort()` to immediately return even if the server has not yet responded to the connection request.
* Fixes resource leak in `Service.java`
* Adds (nullable) `getUpdateString` and `getUpdateRequest` methods to UpdateProcessor so that the underlying request may be exposed.

----

 - ~~[ ] Tests are included~~ (no additional tests added)

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/http/AsyncHttpRDF.java
+++ b/jena-arq/src/main/java/org/apache/jena/http/AsyncHttpRDF.java
@@ -160,26 +160,50 @@ public class AsyncHttpRDF {
      * This operation extracts RuntimeException from the {@code CompletableFuture}.
      */
     public static <T> T getOrElseThrow(CompletableFuture<T> cf) {
+        return getOrElseThrow(cf, null);
+    }
+
+    /**
+     * Get the value of a {@link CompletableFuture} that executes of an HTTP request.
+     * In case on any error, an {@link HttpException} is thrown.
+     *
+     * @param <T> The type of the value being computed.
+     * @param cf The completable future.
+     * @param httpRequest An optional HttpRequest for improving feedback in case of exceptions.
+     * @return The value computed by the completable future.
+     */
+    public static <T> T getOrElseThrow(CompletableFuture<T> cf, HttpRequest httpRequest) {
         Objects.requireNonNull(cf);
         try {
             return cf.join();
         //} catch (CancellationException ex1) { // Let this pass out.
         } catch (CompletionException ex) {
-            if ( ex.getCause() != null ) {
-                Throwable cause = ex.getCause();
-                if ( cause instanceof RuntimeException )
-                    throw (RuntimeException)cause;
-                if ( cause instanceof IOException ) {
-                    IOException iox = (IOException)cause;
-                    // Rather than an HTTP exception, bad authentication becomes IOException("too many authentication attempts");
-                    if ( iox.getMessage().contains("too many authentication attempts") ||
-                            iox.getMessage().contains("No credentials provided") ) {
-                        throw new HttpException(401, HttpSC.getMessage(401));
-                    }
-                    IO.exception((IOException)cause);
+            Throwable cause = ex.getCause();
+            if ( cause != null ) {
+
+                // Pass on our own HttpException instances such as 401 Unauthorized.
+                if ( cause instanceof HttpException httpEx ) {
+                    throw new HttpException(httpEx.getStatusCode(), httpEx.getStatusLine(), httpEx.getResponse(), cause);
                 }
+
+                final String msg = cause.getMessage();
+
+                if ( cause instanceof IOException ) {
+                    // Rather than an HTTP exception, bad authentication becomes IOException("too many authentication attempts");
+                    if ( msg != null &&
+                            ( msg.contains("too many authentication attempts") ||
+                              msg.contains("No credentials provided") ) ) {
+                        throw new HttpException(401, HttpSC.getMessage(401), null, cause);
+                    }
+                    if (httpRequest != null) {
+                        throw new HttpException(httpRequest.method()+" "+httpRequest.uri().toString(), cause);
+                    }
+                }
+
+                throw new HttpException(msg, cause);
             }
-            throw ex;
+            // Note: CompletionException without cause should never happen.
+            throw new HttpException(ex);
         }
     }
 
--- a/jena-arq/src/main/java/org/apache/jena/http/HttpLib.java
+++ b/jena-arq/src/main/java/org/apache/jena/http/HttpLib.java
@@ -37,6 +37,7 @@ import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -560,7 +561,30 @@ public class HttpLib {
      * @return HttpResponse
      */
     public static HttpResponse<InputStream> execute(HttpClient httpClient, HttpRequest httpRequest) {
-        return execute(httpClient, httpRequest, BodyHandlers.ofInputStream());
+        return AsyncHttpRDF.getOrElseThrow(executeAsync(httpClient, httpRequest, BodyHandlers.ofInputStream()), httpRequest);
+    }
+
+    /**
+     * Execute a request, return a {@code HttpResponse<X>} which
+     * can be passed to {@link #handleHttpStatusCode(HttpResponse)} which will
+     * convert non-2xx status code to {@link HttpException HttpExceptions}.
+     * <p>
+     * This function applies the HTTP authentication challenge support
+     * and will repeat the request if necessary with added authentication.
+     * <p>
+     * See {@link AuthEnv} for authentication registration.
+     * <br/>
+     * See {@link #executeJDK} to execute exactly once without challenge response handling.
+     *
+     * @see AuthEnv AuthEnv for authentic registration
+     * @see #executeJDK executeJDK to execute exacly once.
+     *
+     * @param httpClient
+     * @param httpRequest
+     * @return HttpResponse
+     */
+    public static CompletableFuture<HttpResponse<InputStream>> executeAsync(HttpClient httpClient, HttpRequest httpRequest) {
+        return executeAsync(httpClient, httpRequest, BodyHandlers.ofInputStream());
     }
 
     /**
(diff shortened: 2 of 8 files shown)


jena-arq/src/main/java/org/apache/jena/http/AsyncHttpRDF.java
```diff
@@ -167,8 +167,10 @@ public static <T> T getOrElseThrow(CompletableFuture<T> cf) {
         } catch (CompletionException ex) {
             if ( ex.getCause() != null ) {
                 Throwable cause = ex.getCause();
-                if ( cause instanceof RuntimeException )
+                if ( cause instanceof RuntimeException ) {
+                    cause.addSuppressed(new RuntimeException("Passed through here."));
```
rvesse: Leftover debug statement?

jena-arq/src/main/java/org/apache/jena/http/AsyncHttpRDF.java
```diff
@@ -179,10 +181,12 @@ public static <T> T getOrElseThrow(CompletableFuture<T> cf) {
                     IO.exception((IOException)cause);
                 }
             }
+            ex.addSuppressed(new RuntimeException("Passed through here."));
```
rvesse: Again seems like leftover debugging code?

jena-arq/src/main/java/org/apache/jena/http/HttpLib.java
```diff
@@ -640,25 +650,53 @@ public static HttpResponse<InputStream> executeJDK(HttpClient httpClient, HttpRe
...
+                        // or IOException("No credentials provided") if the authenticator decides to return null.
+                        if ( ex.getMessage().contains("too many authentication attempts") ||
+                          ex.getMessage().contains("No credentials provided") ) {
+                            throw new HttpException(401, HttpSC.getMessage(401));
+                        }
+                    }
+                    // Note: Can't reuse AsyncHttpRDF.handleRuntimeException because of this HttpException.
+                    throw new HttpException(httpRequest.method()+" "+httpRequest.uri().toString(), cause);
+                } else if (cause instanceof RuntimeException re) {
+                    re.addSuppressed(new RuntimeException("Passed through here."));
```
rvesse: Again debug statement?

jena-arq/src/main/java/org/apache/jena/update/UpdateProcessor.java
```diff
@@ -26,6 +26,19 @@
...
 {
+    /**
+     * The update request associated with this update execution. May be null.
+     */
+    default public UpdateRequest getUpdateRequest() { return null; }
+
+    /**
+     * The update request as a string. May be null.
+     * The string may contain syntax extensions that can not be parsed by Jena.
+     * If {@code getUpdateRequest()} is not null then the string returned by this method
```
rvesse:
```suggestion
     * If {@link #getUpdateRequest()} is not null then the string returned by this method
```

Aklakan:
> [...] the `HttpLib` changes in particular may have some unexpected side effects for downstream users.

The main change should be only that the original sync methods now internally wait for the completion of a CompletableFuture. My feeling is that maintaining an additional non-async implementation based on the sync java.net.http API might eventually cause issues as it duplicates code that may diverge. The original test cases pass that way, but I am not sure how well tested this part is. Perhaps your previous use of the API could be added as (integration) tests?

rvesse:
> > [...] the `HttpLib` changes in particular may have some unexpected side effects for downstream users.
> 
> The main change should be only that the original sync methods now internally wait for the completion of a CompletableFuture. My feeling is that maintaining an additional non-async implementation based on the sync java.net.http API might eventually cause issues as it duplicates code that may diverge. The original test cases pass that way, but I am not sure how well tested this part is. 

Totally agree, and no problems with this general approach

> Perhaps your previous use of the API could be added as (integration) tests?

That was many years ago now and against a much older Jena version, the code in question has bit-rotted horribly.  These days I use other APIs for HTTP stuff rather than leaning on Jena's APIs.

This might be a non-issue, just something to highlight and make sure we communicate to users in the release this eventually lands in.

afs:
It is already async!

The JDK implementation of `HttpClient.send` (`HttpClientImpl.send`) is a call to `HttpClientImpl.sendAsync` then `CompletableFuture.get` with some exception management around that.

General use of HTTP/2 and HTTP/3 makes it necessary because they both multiplex the underlying network connection.

Aklakan:
> It is already async!

Right but that's the JDK internals. For clarification: Jena's `HttpLib.executeJDK` relied on `HttpClient.send` and this PR changes this and `HttpLib`'s surrounding logic to `HttpClient.sendAsync`.

I removed the breadcrumb exceptions (the clean way to retain the breadcrumbs seems to be wrapping with custom ~~checked~~ exceptions).

jena-arq/src/main/java/org/apache/jena/http/HttpLib.java
```diff
@@ -640,25 +650,52 @@ public static HttpResponse<InputStream> executeJDK(HttpClient httpClient, HttpRe
...
+                        // Rather than an HTTP exception, bad authentication becomes IOException("too many authentication attempts");
+                        // or IOException("No credentials provided") if the authenticator decides to return null.
+                        if ( ex.getMessage().contains("too many authentication attempts") ||
+                          ex.getMessage().contains("No credentials provided") ) {
+                            throw new HttpException(401, HttpSC.getMessage(401));
+                        }
+                    }
+                    // Note: Can't reuse AsyncHttpRDF.handleRuntimeException because of this HttpException.
+                    throw new HttpException(httpRequest.method()+" "+httpRequest.uri().toString(), cause);
+                } else if (cause instanceof RuntimeException re) {
```
afs: This looks like it is a change to exception handling in the sync case. (I hope I read the code correctly.) Does this matter?

`AsyncHttpRDF.getOrElseThrow` passes up runtime exceptions untouched. That means the app gets the exception from the async thread and a stacktrace does not indicate where in the application code their synchronous call was made.

`HttpClientImpl.send` builds an exception with the other threads exception as the cause. The app gets an exception with the app stack. When printed, the stacktrace identifies the app code that makes the sync call.

The `HttpClientImpl.send` exception rebuild is a bit yuk - it has to enumerate the possible classes.

We could have a `HttpException`, even `HttpAsyncException` < `HttpException`, as a carrier to the app call stack making it robust against any future exception introduced by the JDK library. The change would be here, although in `AsyncHttpRDF.getOrElseThrow` is a possibility as well.

Aklakan: > That means the app gets the exception from the async thread and a stacktrace does not indicate where in the application code their synchronous call was made.

Yes, that's the issue.

I think your suggestion to wrap an exception with the already existing unchecked `HttpException` is the clean way forward. Consolidating this exception handling in `AsyncHttpRDF.getOrElseThrow` seems reasonable.

A variant of `getOrElseThrow` with an additional `HttpRequest` argument could be used to add the request URIs to the exceptions (which is what happens in `HttpLib`):

```java
throw new HttpException(httpRequest.method()+" "+httpRequest.uri().toString(), cause);
```

jena-arq/src/main/java/org/apache/jena/sparql/exec/http/QueryExecHTTP.java
```diff
@@ -624,22 +649,71 @@ private static void logQuery(String queryString, HttpRequest request) {}
...
+     */
+    private InputStream registerInputStream(InputStream in) {
+        synchronized (abortLock) {
+            this.retainedConnection = in;
+            // Note: Used ProxyInputStream because the ctor of CloseShieldInputStream is deprecated.
+            this.retainedConnectionView = new ProxyInputStream(in) {
+                @Override
+                protected void beforeRead(int n) throws IOException {
+                    if (isAborted) {
+                        throw new QueryCancelledException();
```
Aklakan: Still not the nicest solution because it pulls the rug from under the parsers' feet - but the original solution also just closed the physical HTTP input stream.
Now `abort()` just closes the ProxyInputStream - whereas only `close()` closes the physical stream.

Aklakan:
I have completed my revisions:
* There is now a unified `AsyncHttpRDF.getOrElseThrow(CompletableFuture<T> cf, HttpRequest httpRequest)` method that wraps any exception as an HTTPException and in some cases also adds information from the optional httpRequest argument. Existing references were updated to use this method.
* `QueryExecHTTP.abort()` no longer closes the physical HTTP InputStream but just a wrapper. Only the close() method (assumed to be called from the thread that started the query exec) will close the physical input stream and thereby consume any remaining data.

jena-arq/src/main/java/org/apache/jena/sparql/exec/http/QueryExecHTTP.java
Aklakan: Oops, snippet was modified during cleanup - resolving here.

jena-arq/src/main/java/org/apache/jena/sparql/exec/http/QueryExecHTTP.java
```diff
@@ -624,22 +649,69 @@ private static void logQuery(String queryString, HttpRequest request) {}
...
+    private InputStream registerInputStream(InputStream in) {
+        synchronized (abortLock) {
+            this.retainedConnection = in;
+            // Note: Used ProxyInputStream because the ctor of CloseShieldInputStream is deprecated.
+            this.retainedConnectionView = new ProxyInputStream(in) {
+                @Override
+                protected void beforeRead(int n) throws IOException {
+                    checkNotAborted();
+                    super.beforeRead(n);
+                }
```
Aklakan: Just as a note: This is still not the nicest solution because it pulls the rug from under the parsers' feet - but the original solution also just closed the physical HTTP input stream.
Now `abort()` just sets a flag (instant action) that causes ProxyInputStream's next read to fail - whereas only `close()` closes the physical stream (may take time to consume remaining data).
If really needed, further revisions could go into future PRs.

jena-arq/src/main/java/org/apache/jena/http/HttpLib.java
Aklakan: Code was revised and is ready for review.

afs:
> * `QueryExecHTTP.abort()` no longer closes the physical HTTP InputStream but just a wrapper. Only the close() method (assumed to be called from the thread that started the query exec) will close the physical input stream and thereby consume any remaining data.

As long as someone :tm: does an exhaustive close to the HTTP connection.

afs:
One conflict otherwise looks good.

Aklakan:
Conflict resolved.

PR #3480, opened 2025-09-30
GH-3473: Identify spatial objects by properties instead of types.

### LINKED ISSUE #3473: jena-geosparql: issues when building spatial index from assembler
### Version

5.5.0, 5.6.0-SNAPSHOT

### What happened?

What works: Loading an existing `spatial.index` file and rebuilding the spatial index using the Fuseki  Mod UI/API.

What fails: If the `spatial.index` file is absent, it should be created on Fuseki server start - via the assembler configuration - but there is weird behavior going on:

* According to https://www.mail-archive.com/users@jena.apache.org/msg21183.html the spatial index is created but not active on the dataset.
* ~~When trying to reproduce the issue on my setup, the spatial index setup created 64 spatial index files with name pattern `spatial.index-{0..63} ` with varying sizes until a final ordinary `spatial.index` file was created. Unclear whether no index or a random one eventually becomes the active one.~~ It turned out that this was an issue of my own docker setup and unrelated to Jena.

Saving a spatial index creates a temporary `spatial-index-{number}` file which only on success should overwrite the final `spatial.index` file - this is aimed to protect against (a) corrupted saves such as when shutting down a docker container while a write is active (b) concurrent recreation of the spatial index (should not really happen because only one indexing task can be active per dataset context).

~~So temporary spatial index files are not unexpected but it seems that there is a bugged loop on some level involved - could be a mix of endpoints, CPUs or (named) graphs.~~

```shell

```

Yes

Aklakan:
GitHub issue resolved #3473 

Pull request Description: In `jena-geosparql`, replaced all lookups of spatial objects by type with lookups by the corresponding relevant properties. E.g. instead of searching for instances of type `geo:Geometry`, instances with `geo:asWKT`, `geo:asGML` and `geo:hasSerialization` properties are looked for.

* Consolidated all vocabulary-based graph access into `AccessGeoSPARQL` and `AccessWGS84` classes.
* Added benchmark. No performance issues revealed. Detailed results in this thread.
* Spatial Indexer UI: Replace mode with no selected graphs clears the index

* Added `computeIfAbsent/Present` methods to `Context`.
* Fixed missing override of `DatasetGraphRDFS.find()`
* Fixed possible resource leaks by changing `forEachRemaining` to `forEach` which closes the iterator after use. Credits to @SimonBin for spotting this.

----

 - [x] `GenericPropertyFunction` tests are covered by existing test suite
 - [x] Added new test cases based on the reported issue.
 - (I tested the UI change manually, but did not add an automated test)

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/rdfs/DatasetGraphRDFS.java
+++ b/jena-arq/src/main/java/org/apache/jena/rdfs/DatasetGraphRDFS.java
@@ -67,6 +67,10 @@ public class DatasetGraphRDFS extends DatasetGraphWrapper implements DatasetGrap
         return new GraphRDFS(base, setup);
     }
 
+    @Override
+    public Iterator<Quad> find()
+    { return find(Node.ANY, Node.ANY, Node.ANY, Node.ANY); }
+
     // Quad-centric access
     @Override
     public Iterator<Quad> find(Quad quad) {
--- a/jena-arq/src/main/java/org/apache/jena/sparql/util/Context.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/util/Context.java
@@ -24,6 +24,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
+import java.util.function.Function;
 
 import org.apache.jena.atlas.lib.Lib;
 import org.apache.jena.atlas.logging.Log;
@@ -389,12 +390,6 @@ public class Context {
         context.clear();
     }
 
-    /** Atomic compute. */
-    @SuppressWarnings("unchecked")
-    public <V> V compute(Symbol key, BiFunction<Symbol, Object, ? extends V> remappingFunction) {
-        return (V)context.compute(key, remappingFunction);
-    }
-
     @Override
     public String toString() {
         String x = "";
@@ -439,13 +434,36 @@ public class Context {
         }
     }
 
+    /** Atomic compute. */
+    @SuppressWarnings("unchecked")
+    public <V> V compute(Symbol key, BiFunction<Symbol, Object, ? extends V> remappingFunction) {
+        Object obj = context.compute(key, remappingFunction);
+        return (V)obj;
+    }
+
+    /** Atomic computeIfAbsent. */
+    @SuppressWarnings("unchecked")
+    public <V> V computeIfAbsent(Symbol key, Function<Symbol, ? extends V> mappingFunction) {
+        Object obj = context.computeIfAbsent(key, mappingFunction);
+        return (V)obj;
+    }
+
+    /** Atomic computeIfPresent. */
+    @SuppressWarnings("unchecked")
+    public <V> V computeIfPresent(Symbol key, BiFunction<Symbol, Object, V> remappingFunction) {
+        Object obj = context.computeIfPresent(key, remappingFunction);
+        return (V)obj;
+    }
+
+    /** Get the context's cancel signal. Create and set one if needed. Context must not be null. */
     public static AtomicBoolean getOrSetCancelSignal(Context context) {
-        AtomicBoolean cancelSignal = getCancelSignal(context);
-        if (cancelSignal == null) {
-            cancelSignal = new AtomicBoolean(false);
-            context.set(ARQConstants.symCancelQuery, cancelSignal);
+        try {
+            AtomicBoolean result = context.computeIfAbsent(ARQConstants.symCancelQuery, sym -> new AtomicBoolean(false));
+            return result;
+        } catch (ClassCastException ex) {
+            Log.error(Context.class, "Class cast exception: Expected AtomicBoolean for cancel control: "+ex.getMessage());
+            return null;
         }
-        return cancelSignal;
     }
 
     /** Merge an outer (defaults to the system global context)
--- a/jena-geosparql/src/main/java/org/apache/jena/geosparql/geo/topological/GenericPropertyFunction.java
+++ b/jena-geosparql/src/main/java/org/apache/jena/geosparql/geo/topological/GenericPropertyFunction.java
@@ -23,11 +23,10 @@ import java.util.List;
 import org.apache.jena.atlas.iterator.Iter;
 import org.apache.jena.geosparql.geof.topological.GenericFilterFunction;
 import org.apache.jena.geosparql.implementation.GeometryWrapper;
+import org.apache.jena.geosparql.implementation.access.AccessGeoSPARQL;
+import org.apache.jena.geosparql.implementation.access.AccessWGS84;
 import org.apache.jena.geosparql.implementation.index.QueryRewriteIndex;
-import org.apache.jena.geosparql.implementation.vocabulary.Geo;
-import org.apache.jena.geosparql.implementation.vocabulary.SpatialExtension;
 import org.apache.jena.geosparql.spatial.SpatialIndex;
-import org.apache.jena.geosparql.spatial.SpatialIndexException;
 import org.apache.jena.geosparql.spatial.index.v2.SpatialIndexLib;
 import org.apache.jena.graph.Graph;
 import org.apache.jena.graph.Node;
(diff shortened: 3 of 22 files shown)


Aklakan:
Converted to draft because there are more instances in GenericPropertyFunction that rely on explicit types, such as:
https://github.com/apache/jena/blob/c2d1a89991e35ab04f7cbce171dea9e8e7cf20af/jena-geosparql/src/main/java/org/apache/jena/geosparql/geo/topological/GenericPropertyFunction.java#L147-L151

I am looking into consolidating this into a separate set of methods that check whether resources are features or geometries and then retrieve the appropriate literal. Perhaps even an interface that let's one choose whether to classify resources by types or by properties.

jena-geosparql/src/main/java/org/apache/jena/geosparql/geo/topological/GenericPropertyFunction.java
```diff
@@ -189,18 +183,7 @@ private QueryIterator findAll(Graph graph, Node boundNode, Node unboundNode, Bin
...
-            ExtendedIterator<Triple> featureTriples = graph.find(null, RDF.type.asNode(), Geo.FEATURE_NODE);
-            ExtendedIterator<Triple> geometryTriples = graph.find(null, RDF.type.asNode(), Geo.GEOMETRY_NODE);
-            spatialTriples = featureTriples.andThen(geometryTriples);
-        } else {
-            //Check for Geo Predicate Features in the Graph if no GeometryLiterals found.
-            spatialTriples = graph.find(null, SpatialExtension.GEO_LAT_NODE, null);
-        }
-        return spatialTriples;
+        return SpatialObjectAccess.findSpatialTriplesByProperties(graph, null);
     }
```
Aklakan, line 204: TLDR: Probably the lookup strategy should become a config option for the time being.

The code above is quite some opinionated/reasoning-centric logic - e.g. the presence of a single `foo a SpatialObject` triple would mean that all non-SpatialObject-resources of e.g. type `Geometry` would be ignored. Also, as reported in the issue, geometry resources with e.g. a WKT serialization but without the `Geometry` type are also ignored here.

My alternative implementation concats several iterators and adds filters that drop duplicates, but for fully materialized datasets this is performance-wise also far from ideal. Merge-scans of sorted iterators would help greatly here to efficiently iterate large amounts of resources that have co-occurrent properties, such as `geo:hasDefaultGeometry` and `geo:hasGeometry` - but we don't have them. AFAIK the TDB2 internals would be capable of this but not the general Graph API.
The `DatasetGraphRDFS` wrapper implementation doesn't solve this. In an ideal world, data access could go uniformly through an (virtual) RDFS reasoning layer - i.e. in the code above `?s a geo:SpatialObject` would be sufficient, and the details would be handled by other layers.

Aklakan:
So benchmark results with this PR on a test dataset with 80K geometries are pretty much consistent with what was there before.

```
(p2_queryId)  (p3_inferences)  (p4_index)  (p5_jenaVersion)  Mode  Cnt  Score   Error  Units
          q1              off       false           current  avgt    5  2.460 ± 1.095   s/op
          q1              off       false             5.5.0  avgt    5  2.365 ± 0.259   s/op
          q1              off        true           current  avgt    5  0.250 ± 0.118   s/op
          q1              off        true             5.5.0  avgt    5  0.265 ± 0.035   s/op
          q1          virtual       false           current  avgt    5  2.947 ± 0.134   s/op
          q1          virtual       false             5.5.0  avgt    5  4.454 ± 0.820   s/op
          q1          virtual        true           current  avgt    5  0.310 ± 0.013   s/op
          q1          virtual        true             5.5.0  avgt    5  0.405 ± 0.012   s/op
          q1     materialized       false           current  avgt    5  2.168 ± 1.352   s/op
          q1     materialized       false             5.5.0  avgt    5  2.200 ± 1.186   s/op
          q1     materialized        true           current  avgt    5  0.149 ± 0.142   s/op
          q1     materialized        true             5.5.0  avgt    5  0.180 ± 0.129   s/op
          q2              off       false           current  avgt    5  0.079 ± 0.071   s/op
          q2              off       false             5.5.0  avgt    5  0.081 ± 0.012   s/op
          q2              off        true           current  avgt    5  0.070 ± 0.015   s/op
          q2              off        true             5.5.0  avgt    5  0.090 ± 0.007   s/op
          q2          virtual       false           current  avgt    5  0.158 ± 0.022   s/op
          q2          virtual       false             5.5.0  avgt    5  0.310 ± 0.021   s/op
          q2          virtual        true           current  avgt    5  0.159 ± 0.008   s/op
          q2          virtual        true             5.5.0  avgt    5  0.307 ± 0.010   s/op
          q2     materialized       false           current  avgt    5  0.047 ± 0.028   s/op
          q2     materialized       false             5.5.0  avgt    5  0.042 ± 0.007   s/op
          q2     materialized        true           current  avgt    5  0.069 ± 0.008   s/op
          q2     materialized        true             5.5.0  avgt    5  0.043 ± 0.004   s/op
```

A slight performance increase is seen on virtual inferences - most likely because because lookups by type (e.g. geo:Geometry) in addition would also have to scan all triples by which that type could be inferred. The new code just checks the properties directly.
```
          q1          virtual       false           current  avgt    5  2.947 ± 0.134   s/op
          q1          virtual       false             5.5.0  avgt    5  4.454 ± 0.820   s/op
```

The benchmark is compatible with jena-5.5.0 and thus adds type Geometry triples which with the now no core are longer needed. Plain geo:asWKT, geo:asGML or geo:hasSerialization is now sufficient.

Aklakan:
The code itself ready for review. I will do a round of testing on our deployment for whether this reveals any issues and report back.

[Update] looked fine.

Aklakan:
Hm, similar access patterns (e.g. the check for Geometry type triples) are used in SpatialIndexUtils, GenericSpatialPropertyFunction, SpatialObjectAccess. While at it, I'll consolidate this in one place.

Aklakan:
The code is now consolidated.

jena-geosparql/src/main/java/org/apache/jena/geosparql/geo/topological/GenericPropertyFunction.java
Aklakan: Configurable lookup strategies are overkill at this point. This pr already consolidates all graph access in one place so future enhancements in that area should be easy to add if needed.

jena-arq/src/main/java/org/apache/jena/rdfs/DatasetGraphRDFS.java
```diff
@@ -67,6 +67,10 @@ public Graph getGraph(Node graphNode) {
         return new GraphRDFS(base, setup);
     }
 
+    @Override
```
afs, line 70: As I read the code on main:

`find()` ->  
`find()` in DatasetGraphWrapper ->  
default method DatasetGraph.find() ->  
find(Node.ANY, Node.ANY, Node.ANY, Node.ANY);

so this is not necessary.

And `Node.ANY` is better than null :smile:

jena-arq/src/main/java/org/apache/jena/sparql/util/Context.java
```diff
@@ -439,13 +434,36 @@ public static AtomicBoolean getCancelSignal(Context context) {
         }
     }
 
+    /** Atomic compute. */
```
afs, line 437: Haven't I seen these Context changes in another of your PRs?

jena-arq/src/main/java/org/apache/jena/rdfs/DatasetGraphRDFS.java
Aklakan: The default find() method of DatasetGraphWrapper goes to the delegated dataset which does not have the inferences - so DatasetGraphRDFS does have to override it to use findInf. I'll add a test case.

https://github.com/apache/jena/blob/29e8a5601da695741a342059ab9810794f65f7ca/jena-arq/src/main/java/org/apache/jena/sparql/core/DatasetGraphWrapper.java#L195-L197

> `Node.ANY`

Agreed.

jena-arq/src/main/java/org/apache/jena/sparql/util/Context.java
Aklakan: It's part of the ExecTracker PR #3184 - but it also fits here where it is easier to review due to the size of the PR :)

jena-arq/src/main/java/org/apache/jena/rdfs/DatasetGraphRDFS.java
Aklakan: Added test case (fails without the fix) and switched to `Node.ANY`.

afs (approved):

PR #3502, opened 2025-10-08
GH-3501: Updates for Titanium 1.7.0

### LINKED ISSUE #3501: Update to Titanium JSON-LD 1.7.0
This needs code changes to adapt to the new Titanium internals, and deprecation of the old world.

afs:
GitHub issue resolved #3501

This replaces #3494.

Titanium 1.7.0 has introduced a new way to go from 3rd party RDF system to JSON with `QuadsToJsonld`.

----


jena-arq/src/main/java/org/apache/jena/riot/writer/JsonLD11Writer.java
```diff
@@ -116,29 +116,85 @@ public Lang getLang() {
...
+                String dir = null;
+                if ( obj.getLiteralBaseDirection() != null )
+                    dir = obj.getLiteralBaseDirection().toString();
+                try {
+                    consumer.quad(s, p, lex, datatype, lang, dir, g);
+                } catch (RdfConsumerException e) {
+                    e.printStackTrace();
+                }
+            } else if  ( obj.isTripleTerm() ) {
+                throw new JenaTitaniumException("Encountered a triple term");
```
rvesse: Maybe something a little clearer for the end user:

```suggestion
                throw new JenaTitaniumException("Triple terms not supported in JSON-LD");
```

I don't follow JSON-LD specs so don't know if they plan to support it in future iterations of the spec in which case the wording might want to be "not currently supported"

jena-arq/src/main/java/org/apache/jena/riot/writer/JsonLD11Writer.java
```diff
@@ -116,29 +116,85 @@ public Lang getLang() {
...
+                String lang = obj.getLiteralLanguage();
+                if ( lang.isEmpty() )
+                    lang = null;
+                String dir = null;
+                if ( obj.getLiteralBaseDirection() != null )
+                    dir = obj.getLiteralBaseDirection().toString();
+                try {
+                    consumer.quad(s, p, lex, datatype, lang, dir, g);
+                } catch (RdfConsumerException e) {
+                    e.printStackTrace();
```
rvesse: Probably ought to wrap this into a Jena specific error?

jena-arq/src/main/java/org/apache/jena/riot/writer/JsonLD11Writer.java
afs: They (JSON-LD WG) do have a design for triple terms. Titanium JSON-LD does not support them, only incompatible "RDF*", which is not RDF 1.2.

PR #3517, opened 2025-10-17
GH-3516: Fix NPE in AbstractIterHashJoin.close()

### LINKED ISSUE #3516: SPARQL execution error: this.hashTable is null
### Version

5.7.0-SNAPSHOT

### What happened?

From https://lists.apache.org/thread/rpo126f0t706ngn9v77ok6d3b4pckhdc

Sparql query execution may give this error:
```
Cannot invoke
"org.apache.jena.sparql.engine.join.MultiHashProbeTable.clear()" because
"this.hashTable" is null
```

Reason is an oversight from https://github.com/apache/jena/pull/3047 - This PR initialized the hash table only lazily - but if the iterator is closed without use (call to next/hasNext) then the close method raises an NPE.

```shell

```

Yes

Aklakan:
GitHub issue resolved #3516 

Pull request Description: Adds an NPE check for the hash table to the close method of AbstractIterHashJoin.

Perhaps there is a suggestion for a better place for the test case?

----

 - [x] Tests are included.
 - ~[ ] Documentation change and updates are provided for the [Apache Jena website](https://github.com/apache/jena-site/)~

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/join/AbstractIterHashJoin.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/join/AbstractIterHashJoin.java
@@ -44,14 +44,16 @@ public abstract class AbstractIterHashJoin extends QueryIter2 {
     // See also stats in the probe table.
 
     protected JoinKey                   joinKey ;
-    protected MultiHashProbeTable       hashTable ;
+
+    // HashTable is initialized lazily.
+    protected MultiHashProbeTable       hashTable   = null ;
 
     private QueryIterator               iterStream ;
-    private Binding                     rowStream       = null ;
+    private Binding                     rowStream   = null ;
     private Iterator<Binding>           iterCurrent ;
     private boolean                     yielded ;       // Flag to note when current probe causes a result.
     // Hanlde any "post join" additions.
-    private Iterator<Binding>           iterTail        = null ;
+    private Iterator<Binding>           iterTail    = null ;
 
     enum Phase { INIT, HASH , STREAM, TRAILER, DONE }
     Phase state = Phase.INIT ;
@@ -234,13 +236,15 @@ public abstract class AbstractIterHashJoin extends QueryIter2 {
         if ( JoinLib.JOIN_EXPLAIN ) {
             String x = String.format(
                          "HashJoin: LHS=%d RHS=%d Results=%d Table=%s",
-                         s_countProbe, s_countScan, s_countResults,
-                         hashTable.toString()) ;
+                         s_countProbe, s_countScan, s_countResults, hashTable) ;
             System.out.println(x) ;
         }
         // In case it's a peek iterator.
         iterStream.close() ;
-        hashTable.clear();
+        if ( hashTable != null ) {
+            hashTable.clear() ;
+            hashTable = null ;
+        }
     }
 
     @Override
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/join/MultiHashProbeTable.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/join/MultiHashProbeTable.java
@@ -69,7 +69,7 @@ class MultiHashProbeTable
     private final JoinIndex   initialIndex;
 
     /** Tables for hash lookups. Will be created dynamically during {@link #getCandidates(Binding)}. */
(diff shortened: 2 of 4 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/engine/join/AbstractIterHashJoin.java
```diff
@@ -234,13 +234,11 @@ protected void closeSubIterator() {
...
             String x = String.format(
                          "HashJoin: LHS=%d RHS=%d Results=%d Table=%s",
-                         s_countProbe, s_countScan, s_countResults,
-                         hashTable.toString()) ;
+                         s_countProbe, s_countScan, s_countResults, hashTable) ;
             System.out.println(x) ;
         }
         // In case it's a peek iterator.
         iterStream.close() ;
-        hashTable.clear();
```
Aklakan, line 243: The hashTable does not manage any resources - so there is no need to do the work to clear it. It can just be gc'd.

afs: The heap subsystem is more than just GC'ing. Clearing objects can release memory earlier (minor GC) or (unlikely here) th eheap sees short lived objects and does effectively a malloc free.

It is hard to tell if it makes a real difference, but it isn't complicated to do here.

And `hashTable = null;` It would also be an internal consistency check to make sure it isn't mistakenly reused.

This is single thread code so a simple solution 

```java
        if ( hashTable != null ) {
            hashTable.clear();
            hashTable = null;
        }
```
either harmless or may be an advantage.

jena-arq/src/test/java/org/apache/jena/sparql/engine/join/TestHashJoin.java
```diff
@@ -19,12 +19,27 @@
...
+import org.junit.jupiter.api.Test;
 
 public class TestHashJoin extends AbstractTestInnerJoin {
     @Override
     public QueryIterator join(JoinKey joinKey, Table left, Table right, ExprList conditions) {
         return Join.hashJoin(joinKey, left.iterator(null), right.iterator(null), null);
     }
+
+    @Test
+    public void testCloseOfUnusedIterator() {
```
afs: Can this be pushed up into `AbstractTestInnerJoin`?

It needs to call `join(joinKey,...)` or have a new abstract method.

jena-arq/src/main/java/org/apache/jena/sparql/engine/join/AbstractIterHashJoin.java
Aklakan: updated.

jena-arq/src/test/java/org/apache/jena/sparql/engine/join/TestHashJoin.java
Aklakan: done - it could be pushed into AbstractTestInnerJoin.

jena-arq/src/main/java/org/apache/jena/sparql/engine/join/MultiHashProbeTable.java
```diff
@@ -180,9 +176,13 @@ public Iterator<Binding> getCandidates(Binding row) {
...
     void doFinalize() {
-        // Note: We need to stick with the variable order provided in the initial index -> don't sort!
+        // If an initial join key is given then its variables are prepended to seen vars
+        // so that those correspond to the first bits of the bit keys!
+        // This means we need to stick with the variable order provided in the initial index
+        // -> don't sort!
         // Arrays.sort(seenVars, (a, b) -> a.getName().compareTo(b.getName()));
-        seenVarsJoinKey = JoinKey.create(seenVarSet);
+        seenVarsJoinKey = JoinKey.newBuilder()
+            .addAll(initialIndex.getSuperJoinKey()).addAll(seenVarSet).build();
```
Aklakan, line 185: Separated the set of initial variables from the set of variables encountered in the build phase - so that clear() can just clear the latter.

To recap: The initial variables are obtained from peeking the first binding of each of lhs and rhs. All bindings in the build phase are initially only indexed by those variables. Later indexes are only created on-demand in the streaming phase based on which variables are unbound in the streamed bindings.

Aklakan:
Revision completed.

afs (approved):

Issue #3535, opened 2025-10-23
query timeout being ignored


lalewis1:
### Version

5.5.0

## issue description

Some queries will ignore the arq:queryTimeout value and run until completion.
this can cause compounding performance problems if multiple such queries are run.

This has been an issue for us as we have some large datasets 300Gb + where these erroneous
queries can take hours to complete and if you get a few running at the same time, the 
CPUs will be consumed with these tasks for multiple hours making the system unresponsive /
seriously degrading performance during this time.

I've done my best to narrow down the simplest forms of the queries that are causing the
problem, but I suspect there will be more cases I am yet to identify.

## Reproducing it

Below is a server I set up specifically to test this issue. But I have found the same problem
on a few machines of different sizes and across datasets of 300gb, 5gb, and 300mb.

I have also tested using podman and docker with a dockerized jena-fuseki-server.

And I have tested without the text index and the issue is still there.

### VM

Azure VM
Linux (rhel 9.4)
Standard D8ls v6 (8 vcpus, 16 GiB memory)

128Gb SSD (5000 Max IOPS) with xfs filesystem mounted at /etc/fuseki

### fuseki

jena-fuseki-server 5.5.0
run as a systemd service using [this unit file](https://github.com/apache/jena/blob/main/jena-fuseki2/apache-jena-fuseki/fuseki.service)
Java 21 openjdk
nginx as reverse_proxy

### data

300mb of ntriples data

```bash
#!/bin/bash

# adapted from https://github.com/qlever-dev/qlever-control/blob/main/src/qlever/Qleverfiles/Qleverfile.freebase

# results in a dataset of ~ 300mb

rm /etc/fuseki/rdf/olympics.nt
wget -nc "https://github.com/wallscope/olympics-rdf/raw/master/data/olympics-nt-nodup.zip" -O /etc/fuseki/rdf/olympics.zip
unzip -qo /etc/fuseki/rdf/olympics.zip -d /etc/fuseki/rdf
rm /etc/fuseki/rdf/olympics.zip
```

### loading

tdb2.tdbloader and jena.textindexer used to create a text indexed TDB2 dataset

specifically, created using this command.

```bash
#!/bin/bash

rm -rf /etc/fuseki/databases/ds
time podman run \
  -v "/etc/fuseki/rdf:/rdf:z" \
  -v "/etc/fuseki/databases:/etc/fuseki/databases:z" \
  -v "/etc/fuseki/configuration/ds.ttl:/config.ttl:z" \
[...]

PR #3603, opened 2025-11-19
GH-3535: Improved cancellation, more testing, possible fix for DatasetText.

Aklakan:
~GitHub issue resolved: ( perhaps #3535 )~

GitHub issue resolved: #3535

This PR addresses query timeout issues that I had in mind for my work on `jena-geosparql` and that might be related to #3535 - but it yet needs to be assessed whether that is actually the case.

Pull request Description:

* Created the TestQueryExecutionCancel class to `jena-integration-tests` and refactored it for use with different datasets.
    * Wired up datasets: mem, TDB2 and Text+TDB2 (aimed for investigating #3535)
* Added `G.find(cancelSignal, graph, s, p, o)`
* Updated several query engine call sites of `graph.find` with `G.find(cancelSignal, graph, s, p, o)`
* Updated `IterAbortable` to accept the cancel signal from the ExecCxt.
* Updated code in `jena-geosparql` to use cancellable find.
* There is an unreported issue from my side: Aborting a query on a DatasetText may break it until server restart. This PR contains a possible fix for concurrent abort with DatasetText: It seems that synchronizing `dsText.abort()` with the write transaction lock fixes race conditions that such as: (1) NPEs in the transaction system and (2)  concurrent read access to the lucene index while a rollback is in progress.
* Added `QueryIterFailed` to wrap `QueryIterPeek` should an eager peek fail. Reason: `QueryIter{Failed,Peek}` are both tracked in the exec cxt. Bailing out with exception either (1) has to leave iterators dangling (warning message) or (2) close the input iterator although that is not the responsibility of the callee. With this approach, a failed construction does return an iterator that can be cleanly closed.  Downside: exception is raised only when accessing an item of that iterator.

----

 - [ ] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/IterAbortable.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/IterAbortable.java
@@ -19,6 +19,8 @@
 package org.apache.jena.sparql.engine.iterator;
 
 import java.util.Iterator;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
 
 import org.apache.jena.atlas.iterator.IteratorCloseable;
 import org.apache.jena.atlas.iterator.IteratorWrapper;
@@ -29,51 +31,69 @@ import org.apache.jena.query.QueryCancelledException;
  * Iterator that adds an abort operation which can be called at any time,
  * including from another thread, and causes the iterator to throw an exception
  * when next touched (hasNext, next).
+ *
+ * The abort signal may be shared across a set of components.
+ * Hence, an abort on one component will be visible to all shared components.
  */
 public class IterAbortable<T> extends IteratorWrapper<T> implements Abortable, IteratorCloseable<T> {
-    private volatile boolean abortFlag = false;
-    private boolean haveAborted = false;
+    /**
+     * The cancel signal. Thread safe. Never null.
+     * May be set by external callers or from this instance.
+     */
+    private AtomicBoolean cancelSignal;
 
     public IterAbortable(Iterator<T> iterator) {
+        this(iterator, null);
+    }
+
+    protected IterAbortable(Iterator<T> iterator, AtomicBoolean cancelSignal) {
         super(iterator);
+        this.cancelSignal = (cancelSignal != null) ? cancelSignal : new AtomicBoolean(false);
+    }
+
+    public static <T> IterAbortable<T> wrap(Iterator<T> it, AtomicBoolean cancelSignal) {
+        return new IterAbortable<>(it, cancelSignal);
+    }
+
+    private boolean isAborted() {
+        return cancelSignal.get() || Thread.currentThread().isInterrupted();
+    }
+
+    private void checkAbort() {
+        if ( isAborted() ) {
+            throw new QueryCancelledException();
+        }
     }
 
     /** Can call asynchronously at any time */
     @Override
     public void abort() {
-        abortFlag = true;
-    }
-
-    private void execAbort() {
-        if ( ! haveAborted )
-            close();
-        haveAborted = true;
+        cancelSignal.set(true);
     }
 
     @Override
     public boolean hasNext() {
-        if ( abortFlag ) {
-            execAbort();
-            throw new QueryCancelledException();
-        }
-        return iterator.hasNext();
+        checkAbort();
+        return get().hasNext();
     }
 
     @Override
     public T next() {
-        if ( abortFlag ) {
-            execAbort();
-            throw new QueryCancelledException();
-        }
-        return iterator.next();
+        checkAbort();
+        return get().next();
     }
 
     @Override
     public void remove() {
-        if ( abortFlag ) {
-            execAbort();
-            throw new QueryCancelledException();
-        }
-        iterator.remove();
+        checkAbort();
+        get().remove();
+    }
+
+    @Override
+    public void forEachRemaining(Consumer<? super T> action) {
+        get().forEachRemaining(item -> {
+            checkAbort();
+            action.accept(item);
+        });
     }
 }
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterFailed.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterFailed.java
@@ -1,0 +1,65 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.jena.sparql.engine.iterator;
+
+import org.apache.jena.query.QueryCancelledException;
+import org.apache.jena.query.QueryExecException;
+import org.apache.jena.sparql.engine.ExecutionContext;
+import org.apache.jena.sparql.engine.QueryIterator;
+import org.apache.jena.sparql.engine.binding.Binding;
+
+/**
+ * A helper iterator for when an intermediate step in the construction of a compound iterator fails.
+ */
+public class QueryIterFailed
+    extends QueryIter1
+{
+    private Throwable cause;
+
+    public QueryIterFailed(QueryIterator input, ExecutionContext execCxt, Throwable cause) {
+        super(input, execCxt);
+        this.cause = cause;
+    }
+
+    @Override
+    protected void requestSubCancel() {
+    }
+
+    @Override
+    protected void closeSubIterator() {
+    }
+
+    @Override
+    protected boolean hasNextBinding() {
+        throw buildException();
+    }
+
+    @Override
+    protected Binding moveToNextBinding() {
+        throw buildException();
+    }
+
+    protected QueryExecException buildException() {
+        if (cause instanceof QueryCancelledException e) {
+            return new QueryCancelledException(e);
+        } else {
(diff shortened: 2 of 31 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/exec/QueryExecDataset.java
```diff
@@ -378,7 +378,7 @@ private static boolean isTimeoutSet(long x) {
     class TimeoutCallback implements Runnable {
         @Override
         public void run() {
-            synchronized (lockTimeout) {
+            // synchronized (lockTimeout) {
```
afs: But the code elsewhere has a read-set (compute) `getOrSetCancelSignal` which sets `cancelSignal` (probably safely at the moment but its an implicit contract) so the test here and the action should be atomic.

We could switch to `cancelSignal` is never null.

`expectedCallback` may also be changing.

Aklakan: I am/was more concerned about the concurrent callback switch related to `expectedCallback.get()`:

`timeout-thread`: calls callback
`main-iterator-thread`: concurrently sets callback to a different instance
`timeout-thread`: callback tests whether it is the expected one -> nope because of the concurrent switch. a call to abort missed. but probably its harmless because it just means that query timeout1 completed just in time and dodged the bullet of abort due to the initial timeout. (if my mental model is correct)
`main-iterator-thread`: sets a new callback and schedules a new timeout (if there is an overall timeout).

The cancel signal should already be safe and never be null: It's always created with `QueryExecDataset`. It can only be null if there is a type error in the context (highly unlikely). Null handling could be changed to raise an exception. Note, that the cancel signal may originate from an update exec that passed it down to a query exec.

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterFailed.java
```diff
@@ -0,0 +1,65 @@
...
+        throw buildException();
+    }
+
+    protected RuntimeException buildException() {
+        if (cause instanceof QueryCancelledException e) {
+            return new QueryCancelledException(e);
+        } else {
+            return new QueryExecException(cause);
+        }
+    }
```
afs: Do `QueryCancelledException` and `QueryExecException` need rewrapping?

```suggestion
    protected QueryExecException buildException() {
        return switch(cause) {
            // Covered by next branch: case QueryCancelledException ex -> ex;
            case QueryExecException ex -> ex;
            default -> new QueryExecException(cause);
        };
    }
```

jena-arq/src/test/java/org/apache/jena/sparql/api/TestQueryExecutionCancel.java
```diff
@@ -60,6 +56,9 @@
 import org.apache.jena.sparql.sse.SSE;
 import org.apache.jena.sparql.util.Context;
 import org.apache.jena.sparql.util.Symbol;
+import org.junit.jupiter.api.BeforeAll;
```
afs: Would it be possible to keep the import order?

(In Eclispe, I have the imports grouped to have the order `java`, `javax`, `com`, `org.junit` then all others. A group has a blank line after it. This arrangement tends to put jena ones in a block.)

jena-arq/src/main/java/org/apache/jena/query/QueryCancelledException.java
```diff
@@ -22,4 +22,5 @@
 public class QueryCancelledException extends QueryExecException
 {
     public QueryCancelledException() {}
+    public QueryCancelledException(Throwable cause) { super(cause) ; }
```
afs, line 25: This can be removed.

The two uses are double wrapping a `QueryCancelledException`.

jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/access/AccessGeoSPARQL.java
```diff
@@ -229,4 +231,11 @@ private static boolean containsGeoLiterals(Graph graph, Node node) {
             graph.contains(node, Geo.AS_GML_NODE, null);
         return result;
     }
+
+    public static RuntimeException buildException(RuntimeException e) {
+        if (e instanceof QueryCancelledException e2) {
+            return new QueryCancelledException(e2);
+        }
+        return new RuntimeException(e);
+    }
```
afs, line 240: See `QueryIterFailed`:

```suggestion
    protected static QueryExecException buildException(RuntimeException cause) {
        return switch(cause) {
            // Covered by next branch. case QueryCancelledException ex -> ex;
            case QueryExecException ex -> ex;
            default -> new QueryExecException(cause);
        };
    }
```

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/IterAbortable.java
```diff
@@ -31,49 +33,61 @@
  * when next touched (hasNext, next).
  */
 public class IterAbortable<T> extends IteratorWrapper<T> implements Abortable, IteratorCloseable<T> {
-    private volatile boolean abortFlag = false;
-    private boolean haveAborted = false;
+    private AtomicBoolean cancelSignal;
+    private volatile boolean abortedHere = false;
```
afs: This can be removed - just use `cancelSignal`.

```suggestion
```

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/IterAbortable.java
```diff
@@ -31,49 +33,61 @@
...
 
     public IterAbortable(Iterator<T> iterator) {
+        this(iterator, null);
+    }
+
+    protected IterAbortable(Iterator<T> iterator, AtomicBoolean cancelSignal) {
         super(iterator);
+        this.cancelSignal = cancelSignal;
```
afs:
```suggestion
        this.cancelSignal = (cancelSignal != null) ? cancelSignal : new AtomicBoolean(false);
```

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/IterAbortable.java
```diff
@@ -31,49 +33,61 @@
...
+    }
+
+    public static <T> IterAbortable<T> wrap(Iterator<T> it, AtomicBoolean cancelSignal) {
+        return new IterAbortable<>(it, cancelSignal);
+    }
+
+    private boolean isAborted() {
+        return abortedHere || (cancelSignal != null && cancelSignal.get()) || Thread.interrupted();
```
afs:
```suggestion
        return cancelSignal.get() || Thread.interrupted();
```

afs: I'm not sure `Thread.interruped()` is a good idea here.

It returns the current flag state and clears it - it is then lost. An application using Jena as a library may have other uses for Thread  interrupt.

Is this for a specific, known usage?

Maybe `Thread.currentThread().isInterrupted()` which is a read-only.

afs:
> GitHub issue resolved: ( perhaps #3535 )

Running this PR,with the suggested changes, on the configuration of #3535, and also in memory dataset, addresses the report.

The crux is the change to TDB2 triple block processing.

jena-text/src/main/java/org/apache/jena/query/text/DatasetGraphText.java
```diff
@@ -186,9 +186,11 @@ public void commit() {
...
     @Override
     public void abort() {
-        super.getMonitor().finish();
-        abortAction.run();
-        readWriteMode.set(null);
+        synchronized (txnExitLock) {
+            super.getMonitor().finish();
+            abortAction.run();
+            readWriteMode.set(null);
+        }
```
Aklakan: Abort calls textIndex.rollback() which needs to re-open the lucene index writer (rollback closes it). this must be synchronized so that no other concurrent abort interfers. There can only be a single index writer - so afaik DatasetText effectively has to employ MRSW locking.

jena-tdb1/src/main/java/org/apache/jena/tdb1/solver/SolverLibTDB.java
```diff
@@ -148,16 +148,13 @@ static QueryIterator graphNames(DatasetGraphTDB ds, Node graphNode, QueryIterato
         List<Abortable> killList = new ArrayList<>();
         Iterator<Tuple<NodeId>> iter1 = ds.getQuadTable().getNodeTupleTable().find(NodeId.NodeIdAny, NodeId.NodeIdAny,
                                                                                    NodeId.NodeIdAny, NodeId.NodeIdAny);
+        iter1 = makeAbortable(iter1, killList, execCxt.getCancelSignal());
```
Aklakan: I added abort wrapper here as early as possible (before any filters or distinct operations).
Should make the double check (before and after distinct ) unnecessary.

Aklakan:
I cleaned up the PR.

The second commit updates places where QueryIterAbortable is used (should this break anything it can be reverted).

Actually, QueryIterAbortable should no longer be needed because the abort signal is pushed beneath it.
So another round of cleanup could even remove that.
I kept my original exception wrapping in order to preserve the full stack traces.

I don't have your tests - if it works with your setup then maybe this additional cleanup could be done with later PRs.

afs:
> I cleaned up the PR.

Thanks!

(( Not for this PR ))

It would be good if `QueryIterAbortable` could be just a `QueryIterPlainWrapper`.

Another clear-up that looks possible is `ExecutionContext.checkForCancel` to replace capturing the `cancelSignal` in several places. Code in query itertators would then just be:

```
    execCxt.checkForCancel();
```

jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/QueryIterFailed.java
```diff
@@ -0,0 +1,65 @@
...
+    protected boolean hasNextBinding() {
+        throw buildException();
+    }
+
+    @Override
+    protected Binding moveToNextBinding() {
+        throw buildException();
+    }
+
+    protected RuntimeException buildException() {
```
afs: Any reason not to make this a `QueryExecException`?

Aklakan: Changed to QueryExecException.

afs:
> I kept my original exception wrapping in order to preserve the full stack traces.

I don't understand this esp the `QueryCancelledException`, which is a routine (expected) exception, because the `cause` has the stack trace.  However, it behaves correctly functions so not enough to delay this PR.

(+ I'm a convert to switch statements with class pattern matching!)

PR #3706, opened 2026-01-19
ServiceEnhancer Improvements

Aklakan:
GitHub issue resolved (TDB)

This is a long standing revision of the service enhancer module. It can go into a dot release after jena 6. There are a couple of places that seem odd to me and which I want to give another look.

Pull request Description:

Code maintenance:
* Migrated to Junit 5 / Jupiter
* Migrated to Caffeine Cache (instead of Guava Cache - The Guava dependency is still needed for RangeMap though)
* Refactored the core cache execution class `RequestExecutor` to support read-transaction-aware concurrent execution.

Bug Fixes:
* The Service Enhancer assembler now supports the standard `ja:dataset` property instead of the custom `ja:baseDataset`. Use of the latter still works but now raises a warning.
* Fixed a bug with the plugin that caused colons at the end of URLS to become removed (`SERVICE <https://.../foo:>` became `SERVICE <https://.../foo>`)
* Fixed a bug where result sets could become cut off due to concurrent cache eviction. Typically this bug occurred when using `SELECT ... LIMIT X` where X was slightly larger than the maximum cache size.
* Introduced context constants (that match that of the assembler) to configure cache sizes. As a side effect, Service Enhancer Cache initialization is now lazy in order to have a chance to read the config from the context. Once created, the settings are fixed.
* Iterators that cache data should now honor the abort signal.

New Features:
* Added `scoped` option for loop (`SERVICE <loop+scoped:>`) to mimic the behavior of LATERAL - when `scoped` is set, then only in-scope variables become substituted instead of all variables with the same base name.
* Added experimental concurrent feature. There is an example in the class `ServiceConcurrentExample`. Essentially, it allows one to do `SERVICE <loop+scoped:concurrent+2-10:bulk+10:cache:https://example.org/sparql>`: Use 2 additional threads with 10 bindings each, to build bulk requests of 10 bindings, caching the responses to query the specified sparql endpoint. Concurrent requests try to preserve the order of bindings. Of course, one must be cautious not to flood public endpoints. The feature is intended as a tool to improve query federation performance over endpoints that one controls.

Open TODOs:
- [ ] More stress testing with concurrent requests that fail or get cancelled - my impression is that there may be a resource leak.
- [ ] AI suggests that the ClaimingCache class can be greatly simplified by relying on Caffeine callbacks. Need to investigate.
```java
this.cache = Caffeine.newBuilder()
    .scheduler(Scheduler.systemScheduler())
    .expireAfter(new Expiry<K, ResourceEntry<V>>() {
        @Override public long expireAfterCreate(K k, ResourceEntry<V> h, long t) {
            return h.state.get() > 0 ? Long.MAX_VALUE : nanos;
        }
        @Override public long expireAfterUpdate(K k, ResourceEntry<V> h, long t, long d) {
            return h.state.get() > 0 ? Long.MAX_VALUE : nanos;
        }
        @Override public long expireAfterRead(K k, ResourceEntry<V> h, long t, long d) { return d; }
    })
    .removalListener((k, holder, cause) -> { closeEntry(k, holder); })
    .build();
```

----

 - [x] Tests are included.

----


afs:
Hi!

Thanks for keeping this PR up-to-date with `main`.

PR #3818, opened 2026-03-26
ExecTracking with Dispatchers Proposal

Aklakan:
GitHub issue resolved (TDB) 

Follow up to #3184 and #3511

Pull request Description: This is a distill of classes from both aforementioned PRs. It is primarily a stripped down version of the original "dispatcher" proposal of #3184. This PR no longer escalates to the exec builder abstractions anymore, and it no longer tries to pass on non-parsed query/update strings to the execution system (as it was done in #3511). 

Core Changes:
* The `dispatch` package contains the newly introduced dispatcher infrastructure (4 classes for each query/update)
* There is change to `QueryExecDatasetBuilder` and `UpdateExecDatasetBuilder` in order to delegate to the dispatchers.
* The changes to the core are this time aimed to be minimally invasive and yet understandable to make things review-able.

Tracker Implementation:
This is the much larger part of this PR and it is included here to "proof" that the core changes accomplish the goal of execution tracking. The whole tracking part could be moved into a separate PR as it only relies on the introduced dispatcher system. This would reduce the size of this PR to ~15 changed files in the core.
If the dispatcher approach is acceptable, then I could move all exec tracking into a separate PR (perhaps even just the fuseki mod module for the time being).

So this is a bit of a chicken & egg problem:
- Without the tracker, the core changes are not justified.
- With the tracker the PR quickly gets quite large
- The tracking could be implemented without further change against the QueryEngineRegistry, but on this level − effectively − only the "SELECT" query form can be intercepted, but not the ASK, DESCRIBE or CONSTRUCT forms.

Notes on the tracker packages:
* The `tracker.core` package implements a basic exec tracking with started, finished normally, and finished exceptionally events. It is currently in ARQ so that all tests (especially security sensitive ones such as `TS_Access`) can be run against it. 
* The `tracker.system` package adds event sources and sinks and time metadata (creation, started, aborted, finished).

The `tracker.system` package could be removed from the PR and `tracker.core` would still provide a simple demo.

Running `mvn install` will print out all query|update exec start/finish events.

The main question is, whether this approach via dispatchers is worthwhile, or whether the exec tracking should be done in a more integrated way (as it was suggested in yesterday's call).
The advantage of the dispatcher system for exec tracking:
- the event system can evolve independently
- any existing code can be retrofitted with exec tracking without change
- the tracking wrapping can be made opt-in (the dispatcher registration can be omitted and the dispatcher could support a context symbol for whether to do any wrapping)

The disadvantages are:
- The wrapping only detects exceptions in the query/update execution machinery. It would not see an exception during e.g. result set serialization in Fuseki.
- Requires the dispatcher infrastructure (core change)

There are two main methods:
* `ExampleExecTrackerCore.java`: Running queries and updates prints out started/finished events - including any raised exception.
* `ExampleExecTrackerSystem.java`: Shows a history of the run queries and any abnormal termination.

----

 - [ ] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/ARQConstants.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/ARQConstants.java
@@ -326,6 +326,13 @@ public class ARQConstants
     public static final Symbol registryExtensions =
         SystemARQ.allocSymbol("registryExtensions") ;
 
-    public static void init() {}
+    /** The query dispatcher registry key */
+    public static final Symbol registryQueryDispatchers =
+        SystemARQ.allocSymbol("registryQueryDispatchers") ;
 
+    /** The update dispatcher registry key */
+    public static final Symbol registryUpdateDispatchers =
+        SystemARQ.allocSymbol("registryUpdateDispatchers") ;
+
+    public static void init() {}
 }
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/Timeouts.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/Timeouts.java
@@ -236,10 +236,20 @@ public class Timeouts {
        return result;
     }
 
-    // Set times from context if not set directly. e..g Context provides default values.
-    // Contrast with SPARQLQueryProcessor where the context is limiting values of the protocol parameter.
+    /**
+     * Update unset values in the builder with values from the context.
+     *
+     * Set times from context if not set directly, i.e. context provides default values.
+     * Contrast with SPARQLQueryProcessor where the context is limiting values of the protocol parameter.
+     */
     public static void applyDefaultQueryTimeoutFromContext(TimeoutBuilderImpl builder, Context cxt) {
         Timeout queryTimeout = extractQueryTimeout(cxt);
+        applyDefaultTimeout(builder, queryTimeout);
+    }
+
+    /** Update unset values in the builder with values from the context. */
+    public static void applyDefaultUpdateTimeoutFromContext(TimeoutBuilderImpl builder, Context cxt) {
+        Timeout queryTimeout = extractUpdateTimeout(cxt);
         applyDefaultTimeout(builder, queryTimeout);
     }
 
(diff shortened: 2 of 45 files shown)


jena-arq/src/main/java/org/apache/jena/http/sys/ExecUpdateHTTPBuilder.java
```diff
@@ -40,106 +44,10 @@
 import org.apache.jena.sys.JenaSystem;
 import org.apache.jena.update.Update;
 import org.apache.jena.update.UpdateException;
-import org.apache.jena.update.UpdateFactory;
 import org.apache.jena.update.UpdateRequest;
 
 public abstract class ExecUpdateHTTPBuilder<X, Y> {
 
-    /** Update element. Either an Update object or a string. */
-    private record UpdateElt(Update update, String updateString) {
```
Aklakan, line 49: Extracted as class for reuse in deferred builder.

jena-arq/src/main/java/org/apache/jena/http/sys/ExecUpdateHTTPBuilder.java
```diff
@@ -40,106 +44,10 @@
...
-        @Override
-        public String toString() {
-            return isParsed()
-                    ? new UpdateRequest(update()).toString() // Reuse UpdateRequest's serialization approach
-                    : updateString();
-        }
-    }
-
-    /** Accumulator for update elements. Can build an overall string or UpdateRequest from the elements. */
-    private class UpdateEltAcc implements Iterable<UpdateElt> {
```
Aklakan, line 63: Extracted as class for reuse in deferred builder.

jena-arq/src/main/java/org/apache/jena/http/sys/ExecUpdateHTTPBuilder.java
Aklakan: file no longer changed.

Aklakan:
For the time being, I factored the execution tracking code out into its own Fuseki plugin JAR for Jena 6+. It has limitations on the information in can show, but it sufficient to list the running executions and cancel them manually:  https://github.com/Scaseco/jena-exectracker

Feedback welcome.

Improving the display-able information would require some changes to the core. Roughly, the two options are:
- the dispatchers as proposed in this PR
- dedicated hooks with execution start/abort events
(perhaps the latter could build upon the former)

In any case, I am closing this PR, because I feel it's better to start this topic over at another time with fresh momentum, perhaps with some community feedback by then.

PR #3846, opened 2026-04-08
GH-3844: Using JEP 394: Pattern Matching for instanceof for #equals in Node and Triple implementations

### LINKED ISSUE #3844: Using JEP 394: Pattern Matching for instanceof for #equals in Node and Triple implementations
### Change

Using JEP 394: Pattern Matching for instanceof for `#equals` in `Node` and `Triple` implementations

Background: 
While working with benchmarks Jena 6 seemed slower than Jena 5.6 although the measured code was almost identical. The profiler showed more time spent in those `#equals` methods, so the OpenJDK 21 optimizer seemed to have skipped those. Maybe I was just hunting ghosts but nevertheless, the code is more concise now and the benchmarks 'magically' now show the expected comparable results for both jena releases.

Yes

arne-bdt:
GitHub issue resolved #3844

Pull request Description:
- Using JEP 394: Pattern Matching for instanceof for #equals in Node and Triple implementations

----

----

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/graph/Node_ANY.java
+++ b/jena-core/src/main/java/org/apache/jena/graph/Node_ANY.java
@@ -45,9 +45,7 @@ public class Node_ANY extends Node {
     @Override
     public boolean equals(Object other) {
         // There is only one such object.
-        if ( this == other )
-            return true;
-        return false;
+        return this == other;
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


afs:
FWIW The semantics changed; probably doesn't matter.

`instanceof` takes account of subclasses, `getClass() != obj.getClass()` does not.

We could make the main RDF term / Node classes final (not this PR!).

afs (approved):

arne-bdt:
@afs Should I have waited for a second approval?

afs:
No. Like all open source, resource is limited.

If we were to wait, for a fixed number of approvals, we'd have a long backlog which causes its own problems and costs.

We are also async - merging is where sync happens.

In short - use your judgement.

PR #3847, opened 2026-04-08
Fix canceled federated SERVICE queries wedging dataset (GH-3837)

utafrali:
Fixes #3837

When a federated SERVICE query gets canceled, the old code still tried to drain the response body on close. This caused the HTTP client to wait for the full response even though the query was aborted, which could wedge the connection and lock up the target dataset.

The fix adds an `isAborted` check in `closeRetainedConnection()`. If the query was canceled, we just cancel the future and close the connection without trying to read the response. Otherwise, we use the original logic to drain the body before reusing the connection.

Added a test case to verify that aborted queries don't hang on close.


utafrali:
gonna close this for now, doesn't seem like it's needed anymore

PR #3869, opened 2026-04-19
GH-3868: Output illegal surrogates as \uFFFD

### LINKED ISSUE #3868: Print Unicode replacement char (U+FFFD) when outputting bad surrogates.
Bad surrogates (i.e. a lone surrogate, a low-high or high-high pair) are illegal. Only valid surrogate pairs are accepts by the Turtle family of parsers.

An "RDF String" does not allow surrogates.

For Java strings, supplementary characters are stored as two `char` but when writing them out to UTF-8, they are one codepoint, 3 or more bytes. Any other use of surrogates is illegal.

When printing, the JDK uses `?` as the replacement character if it encounters a bad surrogate. (This is inside class java library class `UTF_8.Encoder`.)

It would be better to print the Unicode "replacement character" `U+FFFD`. This is what is read in if a raw surrogate is found (encoded) in a UTF-8 stream. (Surrogates are illegal in UTF-8.)

Print in escaped form for clarity -- `\uFFFD` -- and not rely on a font with a defined glyph for U+FFFD.

The alternative of printing the surrogate in `\u` form. However, reading it back in is not legal.

afs:
GitHub issue resolved #3868

Pull request Description:
#3868 and some unrelated improvements while looking at the code.

----

 - [x] Tests are included.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/Abortable.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/engine/iterator/Abortable.java
@@ -21,4 +21,13 @@
 
 package org.apache.jena.sparql.engine.iterator;
 
-public interface Abortable { public void abort(); }
+public interface Abortable {
+    /**
+     * Signal an abort.
+     * This operation is thread-safe can be called asynchorously.
+     * <p>
+     * Abort does not mean the implementation immediately stops processing,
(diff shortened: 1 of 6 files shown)


afs:
NB JSON-LD (titanium-java) prints `?` - it uses the standard UTF-8.Encoder.

PR #3909, opened 2026-05-10
GH-3628: riot exit code

### LINKED ISSUE #3628: riot returns a non-zero exit code when only printing warnings
### Version

5.6.0

### What happened?

Running `riot --validate myfile.ttl` should exit with a non-zero code in the case of errors, which it does. However, it also exits with `1` when only `WARN` messages are printed. I tested a different operation not focused on validation, and found the same result: `riot --count myfile.ttl`.

I think if these warnings are considered errors and will fail the command, then they should be logged as errors. Alternatively, if they are really warnings, then the command should not fail. The non-zero exit code disrupts workflows which depend on a zero return code to continue.

One of the warnings I'm seeing which is leading to a `1` exit code is:

```
11:55:29 WARN  riot            :: [line: 8, col: 21] Lexical form '2021-01-18T10:30:00' not valid for datatype XSD date
```

(I do understand that the message is correct)

Here is a test file:

```turtle
@prefix prov: <http://www.w3.org/ns/prov#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix sosa: <http://www.w3.org/ns/sosa/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
@prefix ex: <http://example.org/> .

ex:AAL210144001R.20210112.1763231 a ex:Observation ;
    sosa:resultTime "2021-01-18T10:30:00"^^xsd:date .
```

```shell

```

Maybe

### LINKED ISSUE #3908: riot to flush parser output when logging a warning or error.
Currently, up to Jena 6.0.0, a warning or error is output much earlier than the offending triple or quad.

This is because parser output stream/writer is buffered.

Change `riot` to flush the output so that the log message for warning or error is next to the place it occurs.

For a parse error, the error message will be the last line output before process exit.

afs:
GitHub issue resolved #3628
GitHub issue resolved #3908

Pull request Description: See issues

----

### DIFF
--- a/jena-core/src/main/java/org/apache/jena/rdfxml/xmloutput/impl/Unparser.java
+++ b/jena-core/src/main/java/org/apache/jena/rdfxml/xmloutput/impl/Unparser.java
@@ -181,7 +181,9 @@ class Unparser {
             while (ss.hasNext()) {
                 Statement s = ss.nextStatement();
                 RDFNode obj = s.getObject();
-                processRDFNode(obj, false);
+                boolean foundTripleTerm = processRDFNode(obj, false);
+                if ( foundTripleTerm )
+                    hasTripleTerms = true;
             }
         } finally {
             ss.close();
@@ -228,7 +230,8 @@ class Unparser {
         }
     }
 
-    private void processRDFNode(RDFNode obj, boolean inTripleTerm) {
+    // Return true for "RDF 1.2 triple term found"
+    private boolean processRDFNode(RDFNode obj, boolean inTripleTerm) {
         if ( obj.isResource() ) {
             Resource r = obj.asResource();
             increaseObjectCount(r);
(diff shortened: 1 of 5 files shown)


jena-core/src/main/java/org/apache/jena/rdfxml/xmloutput/impl/Unparser.java
```diff
@@ -271,7 +278,7 @@ private void processTripleTermsOneLevel(Statement stmt) {
         RDFNode ttSubj = stmt.getSubject();
         if ( ttSubj.isAnon() ) {
             Resource r = ttSubj.asResource();
-                // XXX !!! twice to make sure it is printed and not compacted.
+                // !!! increateObjectCount twice to make sure it is printed and not compacted.
```
rvesse:
```suggestion
                // !!! increaseObjectCount twice to make sure it is printed and not compacted.
```

rvesse (approved):

PR #3954, opened 2026-05-27
Update version of node

afs:
This is a precursor for PR #3949
Specifically `"@babel/generator@^8.0.0-rc.4"`

However, updating the node version may break some CI (some jenkins nodes) because the local platform is node 20 . 
The proper fix is a container build, which I'd need some help setting up.

----

### DIFF
--- a/jena-fuseki2/jena-fuseki-ui/pom.xml
+++ b/jena-fuseki2/jena-fuseki-ui/pom.xml
@@ -32,17 +32,18 @@
         <version>6.2.0-SNAPSHOT</version>
     </parent>
 
-    <properties>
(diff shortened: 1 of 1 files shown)


afs:
A _partial_ quick-fix is to add exclusions to the build with `mvn -fae -pl -jena-fuseki2/jena-fuseki-ui ...`.

It's partial because it does not work on first build of a version. 

A build picks up an existing UI jar in `jena-fuseki-main` scope=test to run module tests. The tests do not require the UI.
Making the dependency optional=true does not work (apparently, "optional" applies to downstream, not at the point of use).

`jena-fuseki-server` does depend on jena-fuseki-ui.

PR #3972, opened 2026-06-04
Bump node from v22.18.0 to v24.16.0 LTS

afs:
v24 is LTS

----


afs:
`com.github.eirslett:frontend-maven-plugin` seems to download `node`. It used not to, and only use the local install.

If this is true, Jena is a little more isolated from the CI server environment.

PR #3986, opened 2026-06-10
Enable having JUnit6 tests for jena-core

afs:
Steps towards updating the jena-core tests.

This makes it possible to have JUnit6 tests in jena-core.

1. Split into two sets of tests -- JenaCoreTestAll_JU6 and JenaCoreTestAll_JU4
2. Convert some of the packages to JUnit6, especially those that were JUnit4, not JUnit3.
3. Tidy up and rationalize; de-duplicate.

The jena-core POM now names these two classes. 

I was hoping to use JUnit4 directly, and not junit-vintage-engine, which is deprecated and may be removed.

This would mean two test phases, different surefire providers, but I haven't managed to get surefire to work with this.

It has to use the vintage engine. If the POM depends on JUnit4, not junit-vintage-engine, then one or other of the test suite isn't found.

----

 - [x] Tests are included.


afs:
The conversion is guided AI and human checking.

@arne-bdt -- I'm going to add converting `jena-core/src/test/java/org/apache/jena/mem/` from JUnit4 to JUnit6 to this PR.

That's much easier (= much more reliable in copilot) than JUnit3 conversion. It's almost applying certain fixed conversions. Only parameterization is different.

JUnit3 conversion is made complicated by the amount of local infrastructure. It shows the JUnit3 framework was quite basic. The local JUnit3-related code around the tests is providing what in effect is parameterization, shared tests, and the way tests are grouped together, all of which are quite different in JUnit nowadays.

afs:
Done! as a separate commit. The bulk was automatic but this time a lot of manual ... because I ran out of quota.

But JUnit4 to JUnit6 is much less of a bump:
1. Change imports
2. Check for assert order if the assert has a message
3. Formulaic rewrite `@Parameterized` to `@ParamterizedClass`

The test count in Eclipse for `mem.GraphMemIndexedSetTest` is different by 5 (the number of strategies). JUnit4 and JUnit6 count `AbstractGraphMemTest.testDeleteAll` and it's overrided method differently.

Difference between runtime and static counting maybe.
JUnit4 counts the test and the override by class scanning, whereas JUnit6 counts one run.

PR #3995, opened 2026-06-16
GH-3994: Improved TDB2 index access for single pattern queries

### LINKED ISSUE #3994: TDB2 Skip scan for single pattern queries.
### Version

6.2.0-SNAPSHOT

### Feature

@rvesse once created https://github.com/apache/jena/pull/1655 which makes `GRAPH ?g { }` return the set of distinct graph names from the TDB2 index.
So far this work was not leveraged for arbitrary single tuple patterns, such as:

* `SELECT DISTINCT ?p { ?s ?p ?o }`
* `SELECT ?g (COUNT(DISTINCT ?p) AS ?c) { GRAPH ?g { ?s ?p ?o } } GROUP BY ?g`
* `...`

The proposal is to extend `OpExecutorTDB2` to make `OpDistinct` and `OpGroup` leverage indexes

Yes

Aklakan:
GitHub issue resolved #3994 

Pull request Description: Adds skip scan evaluation for single-pattern queries to TDB2. Added special code paths to `OpExecutorTDB2` for `OpDistinct` and `OpGroupBy`.

Execution time to find all distinct predicates on Wikidata Truthy (8B triples) becomes between 1 and 100 seconds (warm-cold caches)
Also works for simple group by patterns such as
`SELECT ?p (COUNT(DISTINCT ?g) AS ?c) { GRAPH ?g { ?s ?p ?o } } GROUP BY ?p`

Need yet to add tests - such as create permutations of patterns and compare results with a reference engine (eg. ARQ).

----

 - [x] Tests are included.
 - ~~[ ] Documentation change and updates are provided for the [Apache Jena website](https://github.com/apache/jena-site/)~~

----

### DIFF
--- a/jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/OpExecutorTDB2.java
+++ b/jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/OpExecutorTDB2.java
@@ -47,9 +47,13 @@ import org.apache.jena.sparql.engine.optimizer.reorder.ReorderProc;
 import org.apache.jena.sparql.engine.optimizer.reorder.ReorderTransformation;
 import org.apache.jena.sparql.expr.ExprList;
 import org.apache.jena.sparql.mgt.Explain;
+import org.apache.jena.tdb2.solver.skipscan.OpExecutorTDB2SkipScan;
+import org.apache.jena.tdb2.solver.skipscan.OpExtSkipScan;
+import org.apache.jena.tdb2.solver.skipscan.SkipScanRewrite;
 import org.apache.jena.tdb2.store.DatasetGraphTDB;
 import org.apache.jena.tdb2.store.GraphTDB;
 import org.apache.jena.tdb2.store.NodeId;
+import org.apache.jena.tdb2.sys.SystemTDB;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -92,11 +96,38 @@ public class OpExecutorTDB2 extends OpExecutor
 
     // Retrieving nodes isn't so bad because they will be needed anyway.
     // And if their duplicates, likely to be cached.
-    // Need to work with SolverLib which wraps the NodeId bindgins with a converter.
+    // Need to work with SolverLib which wraps the NodeId bindings with a converter.
 
     @Override
     protected QueryIterator execute(OpDistinct opDistinct, QueryIterator input) {
+        if ( isForTDB ) {
+            boolean isSkipScanEnabled = execCxt.getContext().isTrueOrUndef(SystemTDB.symSkipScan);
+            if ( isSkipScanEnabled ) {
+                Op rewritten = SkipScanRewrite.tryRewriteAsSkipScan(opDistinct);
+                if (rewritten != null && !opDistinct.equals(rewritten)) {
+                    OpExtSkipScan.setOpExecutorIfAbsent(execCxt.getContext(), this);
+                    return exec(rewritten, input);
+                }
+            }
+        }
+
         return super.execute(opDistinct, input);
+    }
+
+    @Override
+    protected QueryIterator execute(OpGroup opGroup, QueryIterator input) {
+        if ( isForTDB ) {
+            boolean isSkipScanEnabled = execCxt.getContext().isTrueOrUndef(SystemTDB.symSkipScan);
+            if ( isSkipScanEnabled ) {
+                Op rewritten = SkipScanRewrite.tryRewriteAsSkipScan(opGroup);
+                if (rewritten != null) {
+                    OpExtSkipScan.setOpExecutorIfAbsent(execCxt.getContext(), this);
+                    return exec(rewritten, input);
+                }
+            }
+        }
+
+        return super.execute(opGroup, input);
     }
 
     @Override
--- a/jena-tdb2/src/test/java/org/apache/jena/tdb2/solver/skipscan/TestOpExecutorTDB2SkipScan.java
+++ b/jena-tdb2/src/test/java/org/apache/jena/tdb2/solver/skipscan/TestOpExecutorTDB2SkipScan.java
@@ -1,0 +1,350 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ *   SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.apache.jena.tdb2.solver.skipscan;
+
+import static org.junit.jupiter.api.DynamicContainer.dynamicContainer;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Stream;
+
+import org.junit.jupiter.api.DynamicNode;
+import org.junit.jupiter.api.DynamicTest;
+import org.junit.jupiter.api.TestFactory;
+
+import org.apache.commons.collections4.iterators.PermutationIterator;
+import org.apache.jena.atlas.iterator.Iter;
+import org.apache.jena.graph.Node;
+import org.apache.jena.graph.NodeFactory;
+import org.apache.jena.graph.Triple;
+import org.apache.jena.query.Query;
+import org.apache.jena.query.ReadWrite;
+import org.apache.jena.sparql.core.BasicPattern;
+import org.apache.jena.sparql.core.DatasetGraph;
+import org.apache.jena.sparql.core.DatasetGraphFactory;
+import org.apache.jena.sparql.core.Quad;
+import org.apache.jena.sparql.core.Var;
+import org.apache.jena.sparql.core.Vars;
+import org.apache.jena.sparql.exec.QueryExec;
+import org.apache.jena.sparql.expr.Expr;
+import org.apache.jena.sparql.expr.ExprVar;
+import org.apache.jena.sparql.expr.aggregate.AggCountDistinct;
+import org.apache.jena.sparql.expr.aggregate.AggCountVarDistinct;
+import org.apache.jena.sparql.sse.SSE;
+import org.apache.jena.sparql.syntax.ElementNamedGraph;
+import org.apache.jena.sparql.syntax.ElementTriplesBlock;
+import org.apache.jena.system.AutoTxn;
+import org.apache.jena.system.Txn;
+import org.apache.jena.tdb2.TDB2Factory;
+import org.apache.jena.tdb2.sys.CopyDSG;
+import org.apache.jena.tdb2.sys.TDBInternal;
+import org.apache.jena.vocabulary.RDF;
+import org.apache.jena.vocabulary.RDFS;
+import org.jspecify.annotations.NonNull;
+
+public class TestOpExecutorTDB2SkipScan
+    extends AbstractDatasetGraphCompare
+{
+    @FunctionalInterface
+    interface TestGenerator {
+        List<DynamicTest> createTests(DatasetGraph referenceDsg, DatasetGraph testDsg, DatasetGraph dataDsg);
+    }
+
+    public TestOpExecutorTDB2SkipScan() {
+        super("");
+    }
+
+    @TestFactory
+    public List<DynamicNode> staticTest01() {
+        DatasetGraph referenceDsg = SSE.parseDatasetGraph(
+        """
+        (dataset
+            (graph
+              (:s :p :oa)
+              (:s :p :ob)
+            )
+            (graph :g1
+              (:s :p :oa)
+              (:s :p :ob)
+            )
+            (graph :g2
+              (:s :p :oa)
+              (:s :p :ob)
+            )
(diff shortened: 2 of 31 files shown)


Aklakan:
The code is ready for review.


* The most critical changes to existing code are those in `BPTreeDistinctKeyPrefixIterator` which are meant to restrict skip scans / index scans to custom sub-ranges.
\
When no custom range is specified, then the range used to be `bptRootNode.{minRecord, maxRecord}`, which indicates that maxRecord is inclusive for page-level iteration. However, when doing record-level interation using `records.getRecordBuffer().iterator(minRecord, maxRecord)` then maxRecord is exclusive. So this is something @rvesse and/or @afs might want to look at whether this is correct and how it should be.
* The class `TestOpExecutorTDB2SkipScan` generates `SELECT DISTINCT ...` and `SELECT ... (COUNT(DISTINCT ?v) AS ?c) { ... } GROUP BY ...` queries with different combinations and permutations.
* There is the `SystemTDB.symSkipScan` symbol to disable the optimization.

jena-base/src/main/java/org/apache/jena/atlas/iterator/Iter.java
```diff
@@ -607,15 +607,7 @@ public static <T> Iterator<T> takeUntil(Iterator<T> iter, Predicate<T> predicate
      *  returned iterator.
      */
     public static <T> Iterator<T> dropWhile(Iterator<T> iter, Predicate<T> predicate) {
-        PeekIterator<T> iter2 = new PeekIterator<>(iter);
-        for(;;) {
-            T elt = iter2.peek();
```
Aklakan: Endless loop because iterator is never advanced.
This method was used during development before settling on
`iter = records.getRecordBuffer().iterator(minRecord, maxRecordPlusOne);` in `BPTreeDistinctKeyPrefixIterator`.

jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/index/OpExecutorTDB2SkipScan.java
```diff
@@ -0,0 +1,557 @@
...
+        // Any constant not matched by HEAD is considered projected and filtered on.
+        int j; // additional needed value slots
+        for (j = keyLen; j < idxLen && uncoveredSlotsMask != 0; ++j) {
+            int tupleSlot = tm.mapIdx(j);
+            Node node = quad[tupleSlot];
+            if (node.isVariable()) {
+                Var v = Var.alloc(node);
+                uncoveredSlotsMask &= ~(1 << tupleSlot);
+
+                // Consider: SELECT DISTINCT ?g ?p { ?g ?s :p ?s }
```
rvesse:
```suggestion
                // Consider: SELECT DISTINCT ?g ?p { GRAPH ?g { ?s :p ?s } }
```

jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/index/OpExecutorTDB2SkipScan.java
```diff
@@ -0,0 +1,557 @@
...
+                // skip
+                return null;
+            }
+        }
+
+        if (lookup.equalityLinks() != null) {
+            mappedLinks = mapEqualityLinks(lookup.equalityLinks(), lookup.equalityLinks().length, tm);
+        }
+
+        // TODO The projection is all columns - need to restrict to the actual projection vars.
```
rvesse: Quite a few leftover TODOs in the code still?

This one would appear to be done since IIUC the `mapProjection()` method handles this?

jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/index/OpExecutorTDB2SkipScan.java
```diff
@@ -0,0 +1,557 @@
...
+                    }
+                }
+            }
+        }
+        return null;
+    }
+
+    private static Var distinctVarOrNull(Aggregator agg) {
+        Var v = null;
+        if (agg instanceof AggCountVarDistinct acvd) {
```
rvesse: There are other aggregates, e.g. `SUM()`, `AVG()`, `MAX()`, `MIN()`, that have a single value expression so should probably cover those cases as well.

AFAICT having the distinct applied first in the scan shouldn't change the results of those `DISTINCT` aggregates

jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/index/OpExecutorTDB2SkipScan.java
```diff
@@ -0,0 +1,557 @@
...
+        return v;
+    }
+
+    private static ExprAggregator convertToNonDistinct(ExprAggregator eAgg) {
+        Aggregator newAgg = convertToNonDistinct(eAgg.getAggregator());
+        return new ExprAggregator(eAgg.getVar(), newAgg);
+    }
+
+    private static Aggregator convertToNonDistinct(Aggregator agg) {
+        if (agg instanceof AggCountVarDistinct acd) {
```
rvesse: Other distinct aggregators should be handled as well

jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/index/OpExecutorTDB2SkipScan.java
```diff
@@ -0,0 +1,557 @@
...
+
+        Iterator<Binding> bindingIt = Iter.map(bindingIdIt, bid -> new BindingTDB(bid, nodeTable));
+        QueryIterator qIter = QueryIterPlainWrapper.create(bindingIt, execCxt);
+
+        // Apply filters if needed.
+        // Presence of filters also implies allocated variables that need projecting away.
+        if (filterExpr != null) {
+            qIter = new QueryIterFilterExpr(qIter, filterExpr, execCxt);
+
+            // XXX QueryIterDistinguishedVars?
```
rvesse: Future TODO/optimisation point?

jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/OpExecutorTDB2.java
```diff
@@ -92,13 +95,40 @@ protected QueryIterator exec(Op op, QueryIterator input) {
...
+            boolean isSkipScanEnabled = execCxt.getContext().isTrueOrUndef(SystemTDB.symSkipScan);
+            if ( isSkipScanEnabled ) {
+                QueryIterator qIter = OpExecutorTDB2SkipScan.tryExec(opGroup, input, execCxt);
+                if ( qIter != null ) {
+                    return qIter;
+                }
+            }
+        }
+
+        QueryIterator qIter = exec(opGroup.getSubOp(), input);
```
rvesse: Can this last part of the method not just be `super.execute(opGroup, input)`?

jena-tdb2/src/test/java/org/apache/jena/tdb2/solver/index/TestOpExecutorTDB2SkipScan.java
```diff
@@ -0,0 +1,264 @@
...
+    /**
+     * Creates queries such as:
+     * <pre>{@code
+     * SELECT ?s ?p (COUNT(DISTINCT ?g) AS ?c) {
+     *   GRAPH ?g { ?s ?p :o }
+     * }
+     * GROUP BY ?s ?p
+     * }</pre>
+     *
+     * <p>If {@code projVars} is empty then the query becomes:</p>
```
rvesse:
```suggestion
     * <p>If {@code vars} is empty then the query becomes:</p>
```

jena-tdb2/src/test/java/org/apache/jena/tdb2/solver/skipscan/TestOpExecutorTDB2SkipScan.java
```diff
@@ -0,0 +1,264 @@
...
+                      (:s :p :ob)
+                    )
+                    (graph :g2
+                      (:s :p :oa)
+                      (:s :p :ob)
+                    )
+                )
+                """);
+        return referenceDsg;
+    }
```
rvesse: This is a pretty trivial dataset for testing an optimisation that has the potential to break query execution semantics, IMO 6 quads is far too trivial to properly exercise and validate a feature like this.

There's several improvements I'd like to see with tests around this optimisation:

1. The tests should actively compare executing the queries with the new optimisation on vs with the new optimisation off to verify that there aren't any regression cases.  This is in line with how we test most other query execution optimisations across Jena
2. There should be tests that generate a reasonable size test dataset with some known levels of distinctness that are used to regression test as well.  This doesn't have to be massive, for example could be 10k quads with 1k unique subjects, 20 unique predicates and 100 unique objects over the dataset (basically pick some constant values that yield differing levels of distinctness for different queries)
3. As noted in earlier comments it seems like this optimisation should also be valid for `GROUP BY` queries that use other aggregators, e.g. `MIN(DISTINCT ?var)`, and I'd like to see that validated as well

rvesse (changes requested):
Hey @Aklakan this looks like a really great potential improvement. I had always thought what I'd done previously for prefix scanning had potential to be applied to other kinds of queries but never any chance to explore that further so thanks for taking this on

I have various comments across the PR, these covers three main things (plus the odd typo):

1. Whether this can/should handle other aggregates like `MAX(DISTNCT ?var)` since it would seem applying the distinct in the scan first wouldn't change their outputs either.
2. Improvements to testing, particularly around regression and scale
3. Leftover TODOs which may/may not remain relevant

jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/index/OpExecutorTDB2SkipScan.java
Aklakan: whoops, yes, this todo has already been resolved.

jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/index/OpExecutorTDB2SkipScan.java
Aklakan: For MIN and MAX this can be added as an incremental improvement. An even better solution would be if these aggregators would not scan all distinct values but just pick the lowest/highest value from the index by trying the NodeIdTypes (decimal, double, int, ptr) in the right order.

jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/index/OpExecutorTDB2SkipScan.java
Aklakan: Added handling for other aggregators.

jena-tdb2/src/main/java/org/apache/jena/tdb2/solver/index/OpExecutorTDB2SkipScan.java
Aklakan: removed the note; using an explicit projection should be fine.

jena-tdb2/src/test/java/org/apache/jena/tdb2/solver/skipscan/TestOpExecutorTDB2SkipScan.java
Aklakan: 1. TestOpExecutorTDB2SkipScan does this comparison. It now also tests whether skip scans were used in the execution using two debug context symbols `OpExecutorTDB2SkipScan.symSkipScanOkCounter` and `OpExecutorTDB2SkipScan.symSkipScanFailoverCounter`.

2. TestOpExecutorTDB2SkipScan now supports configuring different dataset sizes via
```
    private List<DatasetSpec> dynamicDatasetSpecs() {
        return List.of(new DatasetSpec("dataset1", 3, 5, 100.0f, 5, 0.5f));
    }
```

3. Support for further aggregate functions has been added. One limitation is, that skip scan is only considered if there is exactly 1 aggregator. ~~The current logic does not optimize `SELECT ?s (min(?o) AS ?min) (max(?o) AS ?max) { ... } GROUP BY ?s`. This could be added with another PR.~~ I revised the logic that if all aggregators use distinct on the same variable then skip scan can be used.

Aklakan:
I have completed my revision. The code is now more modular:

* OpExecutorTDB2 may try to perform skip scan execution for OpDistinct and OpGroup. It is aimed to work for simple queries such as
  * `SELECT (min(?o) AS ?min) (max(?o) AS ?max) { ?s ?p ?o }` and
  * `SELECT DISTINCT ?p { ?s ?p ?o }`
* `SkipScanRewrite.isSkipScanCandidate` restricts the op-structures eligible to skip scan execution is applied. The restriction could be relaxed in the future.
* `TransformDistinctPlacement` implements a transformer that pushes DISTINCT operations down - ideally adjacent to a single the triple pattern / quad. This transform could eventually go to ARQ as a general tool.
* `OpExtSkipScan` is a custom operator that effectively encapsulates `(distinct (project (vars) (triple-or-quad)))`
* `TestOpExecutorTDB2SkipScan` generates data of configurable size and all permutations of single-{triple/quad} access patterns and compares the result set against a reference in-memory dataset.

rvesse:
> I have completed my revision. The code is now more modular:
> 
> * OpExecutorTDB2 may try to perform skip scan execution for OpDistinct and OpGroup. It is aimed to work for simple queries such as
>   
>   * `SELECT (min(?o) AS ?min) (max(?o) AS ?max) { ?s ?p ?o }` and
>   * `SELECT DISTINCT ?p { ?s ?p ?o }`
> * `SkipScanRewrite.isSkipScanCandidate` restricts the op-structures eligible to skip scan execution is applied. The restriction could be relaxed in the future.
> * `TransformDistinctPlacement` implements a transformer that pushes DISTINCT operations down - ideally adjacent to a single the triple pattern / quad. This transform could eventually go to ARQ as a general tool.
> * `OpExtSkipScan` is a custom operator that effectively encapsulates `(distinct (project (vars) (triple-or-quad)))`
> * `TestOpExecutorTDB2SkipScan` generates data of configurable size and all permutations of single-{triple/quad} access patterns and compares the result set against a reference in-memory dataset.

Thanks @Aklakan, I will aim to review this again later in the week

PR #4032, opened 2026-07-01
GH 1868 zero-length paths must not match absent nodes

### LINKED ISSUE #1868: Optimization error for "
### Version

4.7.0

### What happened?

The following query, on an empty dataset should return no rows because `{0}` only applies to nodes in the graph. But optimization creates `1 <http://example.com/p>{0} 1` which does have a result. 

```
SELECT * WHERE {
   VALUES ?v { 1 }
   ?v <http://example.com/p>{0} ?v
}
```
`{0}` isn't in SPARQL 1.1 but this is the base for `?` operator for property paths which is in SPARQL.

_No response_

None

ThomasThelen:
GitHub issue resolved #1868

Pull request Description:

This is a bit of an experimental path taken. Initially I had done this with a longer walking of the AST and realized that the performance hit would be too big because `evalGroundedPath` gets called on each binding.

Instead, the approach that I took here is to hardcode the zero length component state as part of the various path classes, culminating in the dispatch to [evalGroundedPath](https://github.com/apache/jena/compare/main...ThomasThelen:jena:issue_1868?expand=1#diff-ae05f150796221eeb88c0c8bd337eec22ced6a1cff4c2d28e2f80d1509ab2305R135). This should get called _once_ rather than the _n_ times for each row in my previous AST implementation. 

Happy to provide performance tests, address feedback, or close this if it's a bit far fetched.

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/path/Path.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/path/Path.java
@@ -29,4 +29,7 @@ public interface Path
     public void visit(PathVisitor visitor) ;
     public boolean equalTo(Path path2, NodeIsomorphismMap isoMap) ;
     public String toString(Prologue prolog) ;
+
+    // Return true if this path can match in zero steps
+    public default boolean hasZeroLengthComponent() { return false; }
 }
--- a/jena-arq/src/main/java/org/apache/jena/sparql/path/P_Alt.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/path/P_Alt.java
@@ -24,9 +24,15 @@ package org.apache.jena.sparql.path;
 import org.apache.jena.sparql.util.NodeIsomorphismMap;
 
 public class P_Alt extends P_Path2 {
+    private final boolean hasZeroLength;
+
     public P_Alt(Path p1, Path p2) {
         super(p1, p2);
+        this.hasZeroLength = p1.hasZeroLengthComponent() || p2.hasZeroLengthComponent();
     }
+
+    @Override
+    public boolean hasZeroLengthComponent() { return hasZeroLength; }
 
     @Override
     public void visit(PathVisitor visitor) {
--- a/jena-arq/src/main/java/org/apache/jena/sparql/path/P_FixedLength.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/path/P_FixedLength.java
@@ -43,6 +43,9 @@ public class P_FixedLength extends P_Path1 {
     }
 
     @Override
+    public boolean hasZeroLengthComponent() { return count == 0; }
+
+    @Override
     public int hashCode() {
         return hashFixedLength ^ (int)count ^ getSubPath().hashCode();
     }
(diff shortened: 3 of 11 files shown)


rvesse (approved):
Fix looks quite straightforward and elegant to me, it's been a while since I've been delving in the SPARQL engine internals so I will leave this open for now to give others with fresher eyes chance to review before merging

afs:
I'd like time to explore all the possibilities with complex paths to convince myself that pulling the test up to the evaluation is always correct. (Finding clear "think" time is hard.)

jena-arq/src/main/java/org/apache/jena/sparql/path/PathBase.java
```diff
@@ -46,6 +46,9 @@ public abstract class PathBase implements Path
     protected static final int hashShortest     = 0x205 ;
 
 
+    // Return true if this path can match in zero steps
```
afs: This isn't necessary because it is a default method in the interface.
(and it generates a warning because there is no `@Override`.)

jena-arq/src/main/java/org/apache/jena/sparql/path/Path.java
```diff
@@ -29,4 +29,7 @@ public interface Path
     public void visit(PathVisitor visitor) ;
     public boolean equalTo(Path path2, NodeIsomorphismMap isoMap) ;
     public String toString(Prologue prolog) ;
+
+    // Return true if this path can match in zero steps
+    default boolean hasZeroLengthComponent() { return false; }
```
afs:
```suggestion
    public default boolean hasZeroLengthComponent() { return false; }
```
Consistency - the interface is using public on other interface methods.

ThomasThelen:
Thanks for the feedback here! For more context here's this gist of what my original implementation was like - but it felt way too heavy for how often this gets called, hence hardcoding the true/false values into the classes themselves.

In PathLib.java I had,
```
private static boolean pathHasZeroLengthComponent(Path path) {
    if ( path instanceof P_ZeroOrOne )   return true;
    if ( path instanceof P_ZeroOrMore1 ) return true;
    if ( path instanceof P_ZeroOrMoreN ) return true;
    if ( path instanceof P_FixedLength p ) return p.getCount() == 0;
    if ( path instanceof P_Mod p ) return p.getMin() == 0;
    if ( path instanceof P_Alt p )
        return pathHasZeroLengthComponent(p.getLeft())
            || pathHasZeroLengthComponent(p.getRight());
    if ( path instanceof P_Seq p )
        return pathHasZeroLengthComponent(p.getLeft())
            && pathHasZeroLengthComponent(p.getRight());
    if ( path instanceof P_Inverse p )
        return pathHasZeroLengthComponent(p.getSubPath());
    return false;
}
```

jena-arq/src/main/java/org/apache/jena/sparql/path/PathBase.java
ThomasThelen: Thanks for catching this - I made the change and squashed it into https://github.com/apache/jena/pull/4032/changes/90b64bff9d2a099b8341d7210168ac8d3dbd8e26.

afs:
Sorry for the delay.

Looks good - I've approved it.

Could you please squash the PR to one commit? Thanks.

afs (approved):

ThomasThelen:
> Sorry for the delay.
> 
> Looks good - I've approved it.
> 
> Could you please squash the PR to one commit? Thanks.

Absolutely no rush on this - I appreciate you both. Commits should now be merged into the single one.

afs:
Thank you!

PR #4051, opened 2026-07-08
GH-4043: fix(dboe): don't poison shared buffer on block slice

### LINKED ISSUE #4043: TDB2 queries fail persistently with "newPosition > limit" from BlockAccessMapped after a transient error
### Version

6.1.0

### What happened?

A long-running Fuseki instance with a TDB2 dataset started returning HTTP 500 for every query touching one B+Tree index, and never recovered until the JVM was restarted. (See attached stack trace) 

[Claude Code](https://claude.ai/code) performed an analysis of this issue and gave the below findings.

This signature is usually read as on-disk index truncation/corruption from an unclean shutdown. In this case the database files were intact — the failure is in-memory state corruption inside `BlockAccessMapped`, and a restart fully clears it. 

## Analysis

`BlockAccessMapped.getByteBuffer` slices a block out of the shared per-segment `MappedByteBuffer` by temporarily mutating the shared buffer's state:

```java
segBuffer.position(segOff);
segBuffer.limit(segOff + blockSize);   // shrinks the SHARED buffer's limit
ByteBuffer dst = segBuffer.slice();
segBuffer.limit(segBuffer.capacity()); // restore -- but NOT in a finally
```

If any throwable escapes between the limit-shrink and the restore — e.g. an `OutOfMemoryError` raised at the `slice()` allocation under heap pressure — the shared segment buffer is left with a shrunken limit. Every subsequent access to a higher block in that segment then throws `IllegalArgumentException: newPosition > limit` from `Buffer.position(int)`, and the failure is self-sustaining: the poisoned call throws *before* reaching the line that would restore the limit.

Note this poisoning is not confined to writes. There is one `MappedByteBuffer` per 8 MiB segment, shared by every block access to that segment. A read never changes the mapped bytes, but the old slicing technique mutates the shared buffer's *cursors* (`position`/`limit`) to steer `slice()` onto the target block, because the classic `slice()` always carves out `[position, limit)` and had no absolute offset/length overload before Java 13. So `read` — the hot path taken by every query, via `BPTreeNode.get → BlockAccessMapped.read → getByteBuffer` — borrows and restores shared mutable state on each call, and is equally able to both create and hit the poisoned limit.

Supporting evidence from the observed failure:

- The reported limit `2588672` is not segment-aligned (segments are 8 MiB, mapped in full via `FileChannel.map(READ_WRITE, offset, SegmentSize)`, so a mapped buffer's capacity is always 8388608). `2588672 = 315 * 8192 + 8192` is exactly a leftover per-block slice limit — in-memory state, not a file-length artifact.
- All failures in the incident showed the same `Seg` and the same stuck limit, consistent with a single poisoned segment buffer.
- A short physical file cannot produce this exception at all: mapping a segment extends the file, so a truncated file yields zero-filled blocks (a parse error in `BPTreeNode`), never `newPosition > limit`.

The window is small, but on a heavily loaded server (large UPDATEs, high heap pressure, weeks of uptime) it is eventually hit — and one hit permanently poisons the segment. The in-code comment acknowledges the assumption: `// Shouldn't (ha!) happen because the second "limit" resets`.

## Reproduction

The trigger (an asynchronous throwable inside the 3-line window) cannot be provoked deterministically, but the state it leaves behind can be injected: shrink `segments[seg]`'s limit via reflection, then `read()` any block in that segment with a byte offset above the limit. This reproduces the exception verbatim, deterministically:

```
java.lang.IllegalArgumentException: newPosition > limit: (448 > 64)
    at org.apache.jena.dboe.base.file.BlockAccessMapped.getByteBuffer(...)
```

A test doing exactly this is included in the linked PR (`TestBlockAccessMappedSegmentState`).

```shell
[ERROR] 2026-07-03T09:40:00,154 [... | jena.dboe.base.file.BlockAccessMapped] Id: 218892
[ERROR] ... Seg=213
[ERROR] ... Segoff=6389760
[ERROR] ... newPosition > limit: (6389760 > 2588672)
java.lang.IllegalArgumentException: newPosition > limit: (6389760 > 2588672)
    at org.apache.jena.dboe.base.file.BlockAccessMapped.getByteBuffer(BlockAccessMapped.java:155)
    at org.apache.jena.dboe.base.file.BlockAccessMapped.read(BlockAccessMapped.java:93)
    ...
    at org.apache.jena.dboe.trans.bplustree.BPTreeNode.get(BPTreeNode.java:163)
```

Yes

m3l0n-sq:
GitHub issue resolved #4043 

Pull request Description:

## Problem

`BlockAccessMapped.getByteBuffer` slices a block out of the shared per-segment `MappedByteBuffer` by mutating that buffer's state:

```java
segBuffer.position(segOff);
segBuffer.limit(segOff + blockSize);   // shrinks the SHARED buffer's limit
ByteBuffer dst = segBuffer.slice();
segBuffer.limit(segBuffer.capacity()); // restore -- but NOT in a finally
```

The restore is not exception-safe. If any throwable escapes between the limit-shrink and the restore — realistically an `OutOfMemoryError` raised at the `slice()` allocation under heap pressure — the segment buffer is left with a shrunken limit. From then on, **every access to a higher block in that segment fails**:

```
java.lang.IllegalArgumentException: newPosition > limit: (6389760 > 2588672)
    at org.apache.jena.dboe.base.file.BlockAccessMapped.getByteBuffer(BlockAccessMapped.java:155)
    at org.apache.jena.dboe.base.file.BlockAccessMapped.read(BlockAccessMapped.java:93)
    ...
    at org.apache.jena.dboe.trans.bplustree.BPTreeNode.get(BPTreeNode.java:163)
```

The failure is self-sustaining: the poisoned `position(segOff)` call throws *before* reaching the line that would restore the limit. (An access to a block *below* the stuck limit happens to heal the buffer, but a hot B+Tree node above it keeps failing indefinitely.)

Observed in production as a TDB2 dataset returning HTTP 500 for every query touching one B+Tree index, persisting until JVM restart. The reported limit was not segment-aligned (`2588672 = 315 * 8192 + 8192`) — exactly a leftover slice limit, not a file-length artifact. This failure presents exactly like on-disk index corruption, but nothing on disk is wrong — which matters, because the standard advice for this signature (dump/reload the database) is unnecessary here; a restart clears it.

The existing in-code comment anticipated the hazard:

```java
// Shouldn't (ha!) happen because the second "limit" resets
```

## Fix

Slice through a cleared duplicate, using the absolute `slice(index, length)`:

```java
ByteBuffer dst = segBuffer.duplicate().clear().slice(segOff, blockSize);
```

This neither mutates nor depends on the shared buffer's position/limit, so the poisoned state can no longer be created — and reads keep working even if the buffer state were corrupted by other means.

Why not plain `slice(index, length)` on the shared buffer? Because `slice(index, length)` validates against the buffer's *current limit*, not its capacity — it would prevent creating the poison but still fail on an already-poisoned buffer. The cleared duplicate makes the access path fully state-independent.

Semantics of the returned slice are unchanged: position 0, limit and capacity = `blockSize`, big-endian byte order, view of the same backing region. Cost is one extra view object per block access (the previous code already allocated one view per access).

The `catch (IllegalArgumentException ...)` diagnostic block is widened to `RuntimeException`, since the absolute slice reports range violations as `IndexOutOfBoundsException`.

The identical code exists in TDB1 (`org.apache.jena.tdb1.base.file.BlockAccessMapped`) and is fixed the same way.

### Files changed

- `jena-db/jena-dboe-base/src/main/java/org/apache/jena/dboe/base/file/BlockAccessMapped.java`
- `jena-tdb1/src/main/java/org/apache/jena/tdb1/base/file/BlockAccessMapped.java`
- `jena-db/jena-dboe-base/src/test/java/org/apache/jena/dboe/base/file/TestBlockAccessMappedSegmentState.java` (new)
- `jena-db/jena-dboe-base/src/test/java/org/apache/jena/dboe/base/file/TS_File.java` (register new test class)

## Testing

The trigger (an asynchronous throwable inside a 3-line window) cannot be provoked deterministically in a test, so `TestBlockAccessMappedSegmentState` injects the state it leaves behind and verifies behavior at that boundary:

- `poisonedSegmentLimit_readsStillWork` — reflectively shrinks the shared segment buffer's limit (the state an escaped throwable left behind), then reads all higher blocks. Against the previous implementation this reproduces the production exception verbatim: `IllegalArgumentException: newPosition > limit: (448 > 64)`. With the fix, all blocks read back correctly.
- `sharedSegmentBufferStateUntouchedByAccess` — locks in the invariant that block access never mutates the shared buffer's position/limit. Fails against the previous implementation (which left `position` at the last-accessed segment offset).
- `returnedSlicesAreIndependent` — returned slices do not interfere with each other or with the shared buffer.
- `blockRoundTripAfterMixedAccess` — content round-trip through the new slice path.

Red/green verified: with the fix reverted, the new tests fail with the exact production signature; with the fix applied, they pass.

Full test suites pass with the change:

| Module | Tests | Result |
|---|---|---|
| jena-dboe-base | 160 | pass |
| jena-dboe-transaction | 117 | pass |
| jena-dboe-trans-data | 232 | pass |
| jena-tdb2 | 734 | pass |
| jena-tdb1 | 551 | pass (5 pre-existing skips) |

## Compatibility notes

- `ByteBuffer.slice(int, int)` requires JDK 13+; fine for the current Java baseline (21). A backport to older branches would need `((ByteBuffer)segBuffer.duplicate().clear()).position(segOff).limit(segOff+blockSize).slice()` on a duplicate instead.
- No public API change; `getByteBuffer` is private and the returned slice is indistinguishable from before.

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-db/jena-dboe-base/src/main/java/org/apache/jena/dboe/base/file/BlockAccessMapped.java
+++ b/jena-db/jena-dboe-base/src/main/java/org/apache/jena/dboe/base/file/BlockAccessMapped.java
@@ -151,13 +151,12 @@ public class BlockAccessMapped extends BlockAccessBase
             try {
                 // Need to put the alloc AND the slice inside a sync.
                 ByteBuffer segBuffer = allocSegment(seg);
-                // Slice from a cleared duplicate: never touches the shared
-                // segment buffer's position/limit, and does not depend on them
-                // either. The old position/limit/slice/reset sequence could
-                // leave a shrunken limit on the shared buffer if anything threw
-                // before the reset, poisoning every later access to a higher
-                // block in the segment ("newPosition > limit").
-                ByteBuffer dst = segBuffer.duplicate().clear().slice(segOff, blockSize);
+                // Absolute slice: never mutates the shared segment buffer's
+                // position/limit. The old position/limit/slice/reset sequence
+                // shrank the shared buffer's limit; if anything threw before the
+                // reset, the limit stayed shrunk and every later access to a
+                // higher block in the segment failed ("newPosition > limit").
+                ByteBuffer dst = segBuffer.slice(segOff, blockSize);
                 // Extend block count when we allocate above end.
                 numFileBlocks = Math.max(numFileBlocks, id+1);
                 return dst;
--- a/jena-db/jena-dboe-base/src/main/java/org/apache/jena/dboe/base/file/BlockAccessMapped.java
+++ b/jena-db/jena-dboe-base/src/main/java/org/apache/jena/dboe/base/file/BlockAccessMapped.java
@@ -149,20 +149,19 @@ public class BlockAccessMapped extends BlockAccessBase
 
         synchronized (this) {
             try {
-                // Need to put the alloc AND the slice/reset inside a sync.
+                // Need to put the alloc AND the slice inside a sync.
                 ByteBuffer segBuffer = allocSegment(seg);
-                // Now slice the buffer to get the ByteBuffer to return
-                segBuffer.position(segOff);
-                segBuffer.limit(segOff+blockSize);
-                ByteBuffer dst = segBuffer.slice();
-
-                // And then reset limit to max for segment.
-                segBuffer.limit(segBuffer.capacity());
+                // Slice from a cleared duplicate: never touches the shared
+                // segment buffer's position/limit, and does not depend on them
+                // either. The old position/limit/slice/reset sequence could
+                // leave a shrunken limit on the shared buffer if anything threw
+                // before the reset, poisoning every later access to a higher
+                // block in the segment ("newPosition > limit").
+                ByteBuffer dst = segBuffer.duplicate().clear().slice(segOff, blockSize);
                 // Extend block count when we allocate above end.
                 numFileBlocks = Math.max(numFileBlocks, id+1);
                 return dst;
-            } catch (IllegalArgumentException ex) {
-                // Shouldn't (ha!) happen because the second "limit" resets
+            } catch (RuntimeException ex) {
                 log.error("Id: "+id);
                 log.error("Seg="+seg);
                 log.error("Segoff="+segOff);
--- a/jena-db/jena-dboe-base/src/test/java/org/apache/jena/dboe/base/file/TestBlockAccessMappedSegmentState.java
+++ b/jena-db/jena-dboe-base/src/test/java/org/apache/jena/dboe/base/file/TestBlockAccessMappedSegmentState.java
@@ -1,0 +1,196 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ *   SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.apache.jena.dboe.base.file;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+import java.lang.reflect.Field;
+import java.nio.MappedByteBuffer;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import org.apache.jena.atlas.lib.FileOps;
+import org.apache.jena.dboe.ConfigTestDBOE;
+import org.apache.jena.dboe.base.block.Block;
+
+/**
+ * The shared per-segment {@link MappedByteBuffer} in {@link BlockAccessMapped}
+ * must never have its position/limit mutated by block access.
+ *
+ * <p>The previous implementation sliced blocks with a
+ * position/limit/slice/reset-limit sequence on the shared segment buffer. If
+ * anything threw between the limit-shrink and the reset (e.g. an
+ * {@link OutOfMemoryError} during {@code slice()}), the segment buffer was left
+ * with a shrunken limit and every later access to a higher block in that
+ * segment failed with {@code IllegalArgumentException: newPosition &gt; limit}
+ * from {@code Buffer.position(int)} - observed in production as persistent 500s
+ * on reads of a B+Tree index until restart. The fix slices with the absolute
+ * {@code slice(index, length)}, which never touches the shared buffer's state.
+ * These tests lock in that invariant.</p>
+ */
+public class TestBlockAccessMappedSegmentState {
+
+    private static final int BlockSize = 64;
+    private static int counter = 0;
+
+    private String filename;
+    private BlockAccessMapped file;
+
+    @BeforeEach public void before() {
+        filename = ConfigTestDBOE.getTestingDir() + "/test-segment-state-" + (counter++);
+        FileOps.deleteSilent(filename);
+        file = new BlockAccessMapped(filename, BlockSize);
+    }
+
+    @AfterEach public void after() {
+        file.close();
+        FileOps.deleteSilent(filename);
+    }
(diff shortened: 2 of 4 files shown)


jena-db/jena-dboe-base/src/test/java/org/apache/jena/dboe/base/file/TestBlockAccessMappedSegmentState.java
```diff
@@ -0,0 +1,196 @@
...
+        // End on a high block: the old implementation left position at that
+        // block's segment offset.
+        for ( int i = 0; i < numBlocks; i++ )
+            file.read(blocks[i].getId().longValue());
+        file.read(blocks[0].getId().longValue());
+        file.read(blocks[numBlocks - 1].getId().longValue());
+
+        MappedByteBuffer seg0 = segmentBuffer(0);
+        assertNotNull(seg0, "segment 0 should be mapped after block access");
+        System.out.printf("segment 0 after %d writes + %d reads: position=%d limit=%d capacity=%d%n",
```
afs: Tests shouldn't write to `System.out`.

These will need to be removed when finalizing the PR.

jena-db/jena-dboe-base/src/main/java/org/apache/jena/dboe/base/file/BlockAccessMapped.java
```diff
@@ -149,20 +149,19 @@ private ByteBuffer getByteBuffer(long _id) {
...
-
-                // And then reset limit to max for segment.
-                segBuffer.limit(segBuffer.capacity());
+                // Slice from a cleared duplicate: never touches the shared
+                // segment buffer's position/limit, and does not depend on them
+                // either. The old position/limit/slice/reset sequence could
+                // leave a shrunken limit on the shared buffer if anything threw
+                // before the reset, poisoning every later access to a higher
+                // block in the segment ("newPosition > limit").
+                ByteBuffer dst = segBuffer.duplicate().clear().slice(segOff, blockSize);
```
afs: Both `duplicate` and `slice(segOff, blockSize)` create a new independent ByteBuffer objects with a call to DirectByteBuffer. So that's two objects.

The root cause is mutating `segBuffer` to pass information to `slice()` (method with 0 arguments),

i.e. set to size 64, reset to 8388608. For a short while, `segBuffer` limit is 64.

This can be done with a single object creation `segBuffer.slice(segOff, blockSize)`.
This does not change `segBuffer` at any point.

Test `poisonedSegmentLimit_readsStillWork` fails in the check reading limit() of block(0)
but the test the simulates the effect observed, not how it happened.

poisonedSegmentLimit_readsStillWork:
```
        MappedByteBuffer seg0 = segmentBuffer(0);
        seg0.limit(BlockSize);
```

`seg0` is the returned fresh MappedByteBuffer `dst` and block 0 is always size 64. It is not misread as 8388608 because with `segBuffer.slice(segOff, blockSize)`, the limit is always the correct 64.

jena-db/jena-dboe-base/src/test/java/org/apache/jena/dboe/base/file/TestBlockAccessMappedSegmentState.java
```diff
@@ -0,0 +1,196 @@
...
+                    System.out.printf("MISMATCH block=%d byte=%d expected=%02x actual=%02x%n",
+                                      i, j, expected, actual);
+                assertEquals(expected, actual,
+                             String.format("content mismatch: block=%d byte=%d", i, j));
+            }
+        }
+    }
+
+    /**
+     * Deterministic reproduction of the production failure.
```
afs: Better to refer to the GH issue. 
No need to take the description out of context.
This code isn't Fuseki specific.

jena-db/jena-dboe-base/src/test/java/org/apache/jena/dboe/base/file/TestBlockAccessMappedSegmentState.java
m3l0n-sq: removed

jena-db/jena-dboe-base/src/main/java/org/apache/jena/dboe/base/file/BlockAccessMapped.java
m3l0n-sq: You're right that `slice(segOff, blockSize)` does not modify `segBuffer`. Simplified this down to just using `slice(segOff, blockSize)` and removing `duplicate`

jena-db/jena-dboe-base/src/test/java/org/apache/jena/dboe/base/file/TestBlockAccessMappedSegmentState.java
m3l0n-sq: Added the comment

m3l0n-sq:
@afs thanks for the feedback. I've pushed a new commit addressing them.

afs (approved):

afs:
Thank you!

PR #4054, opened 2026-07-08
GH-1998: Support more locales in the Fuseki docker image

### LINKED ISSUE #1998: Support more locales in Apache Jena Docker image
### Version

main

### Feature

The docker image provided with Apache Jena code uses a multi-stage build, where the last layer (used to run it) uses Alpine Linux.

In Skosmos we are upgrading the Docker setup to switch from stian's Jena image to the one included in Apache Jena code, and also to run CICD tests with the Apache Jena image.

We ran into [an issue](https://github.com/NatLibFi/Skosmos/pull/1496#discussion_r1306959366) now where we are not able to use multiple locales in PHP due to the missing `locale` command. Looking at the locales available for Java and Jena [^1], I get:

```bash
/ # java Available

en_US_POSIX
en
en_US_#Latn
en_US
```

I believe we will be able to install extra locales in the Skosmos container, but perhaps that should be included in the Apache Jena Docker image as well, allowing users that rely on other locales (for ARQ query collation, for example) to use the container with the desired locale.

If that sounds like a good idea, we can send a pull request upstream with our solution for the locales.

As for implementation details, it could be i) a fixed list of common locales, ii) just English but with the necessary tools and configuration to easily add more locales, or iii) something more elaborate like trying to download all locales but enable selectively via some env var (no idea if that's actually doable). I think the simplest would be ii), and then in Skosmos we would add an extra step to install finnish, swedish, german, etc.

[^1] Compiled the program from [this oracle docs page](https://docs.oracle.com/javase/tutorial/i18n/locale/identify.html) with `-source 8 -target 8` inside the Alpine container

Yes

ThomasThelen:
GitHub issue resolved #1998 

Pull request Description:

This PR addresses the issue by going with @osma's proposed solution at the end of the issue thread. The only tricky part was including the locale information, but luckily the folks in the issue thread pinpointed how to do this. The gist is that `jdk.localedata` is now explicitly put in the jlink module list.

The main concern here is image size _because the default behavior is to include all locales_.

Here are two image builds: the first being the default case of including all locales vs he second which only has english.
```
     Name                                                              Disk Usage                   Image Size
fuseki:5.3.0-all                                   997c0b2b9c4f        318MB                          140MB
fuseki:5.3.0-en                                    14718521a9b4        297MB                          130MB
```

Demo

Here's an example of using the de locale and doing proper ordering
Screenshot 2026-07-08 at 7 47 29 AM

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-fuseki2/jena-fuseki-docker/Dockerfile
+++ b/jena-fuseki2/jena-fuseki-docker/Dockerfile
@@ -61,13 +61,21 @@ RUN ./download.sh --chksum sha1 "$JAR_URL"
 
 ## -- Make reduced Java JDK
 
-ARG JDEPS_EXTRA="jdk.crypto.cryptoki,jdk.crypto.ec"
+## Provides locale data. Jdeps does not detect it because of SPI, so add it here.
+ARG JDEPS_EXTRA="jdk.crypto.cryptoki,jdk.crypto.ec,jdk.localedata"
+
+## Optional: comma-separated list of locale language tags to keep from jdk.localedata,
+## e.g. "en,fr,de,fi". Leave empty to include all locales.
+ARG JDK_LOCALES=""
+
 RUN \
   JDEPS="$(jdeps --multi-release base --print-module-deps --ignore-missing-deps ${FUSEKI_JAR})"  && \
+  if [ -n "${JDK_LOCALES}" ] ; then LOCALE_OPT="--include-locales=${JDK_LOCALES}" ; else LOCALE_OPT="" ; fi && \
   jlink \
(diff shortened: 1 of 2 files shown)


afs:
> The main concern here is image size because the default behavior is to include all locales.

I agree with the approach in this PR - default to including locales and document how to build with selected ones on the principle that default things should "just work" (unless there is a reason why not).

afs (approved):

osma:
Oh wow! Thanks a lot @ThomasThelen for this fix, and @afs for the shepherding! Much appreciated.

Issue #4084, opened 2026-07-22
Fuseki Memory Churn When Serving HTTP Responses


AniRamadoss:
### Version

Up to version 6.1

### What happened?

Jena Fuseki: 1 MB output buffers are never pooled by the default Jetty ByteBufferPool, causing DirectMemory churn under load

## Summary

Jena Fuseki (jena-fuseki-main) builds its Jetty `ServerConnector` without configuring a `ByteBufferPool`, so Jetty falls back to a default `ArrayByteBufferPool` whose maximum pooled buffer capacity is 65536 bytes (64 KB). At the same time, Fuseki configures the connector's `HttpConfiguration` with a 1 MB `outputBufferSize`. Because 1 MB is larger than the pool's 64 KB maximum capacity, every output buffer that Jetty acquires for an `HttpConnection` is allocated fresh and then dropped on release rather than being retained for reuse. The result is continuous allocation and garbage collection of 1 MB direct buffers with no pooling benefit. Under sustained high request throughput this drives elevated DirectMemory usage and GC pressure. This report describes how the pieces connect, the root cause, and options the maintainers may want to weigh for a fix on the Jena side.

## Environment

- Jena Fuseki: jena-fuseki-main 4.10.0
- Jetty: 10.0.17 (the version Jena 4.10.0 declares via the `jetty-bom` import in its root `pom.xml`, see [`pom.xml` line 71](https://github.com/apache/jena/blob/jena-4.10.0/pom.xml#L71))
- Java: Java 17 (Amazon Corretto 17.0.19)
- OS: Amazon Linux 2

## Background: how the pieces fit together

Three separate configuration points combine to produce this behavior.

First, Fuseki provisions the Jetty connector in [`JettyHttps.java` around line 101](https://github.com/apache/jena/blob/jena4/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/JettyHttps.java#L101):

```java
ServerConnector plainConnector = new ServerConnector(server, new HttpConnectionFactory(http_config));
```

[...]

PR #4088, opened 2026-07-23
GH-4084: Fix ByteBufferPool sizing in Fuseki's Jetty Server

divyankshah:
Fuseki builds its Jetty `Server` with `new Server(threadPool)`, which wires in a
default `ArrayByteBufferPool` capped at 64KB. Fuseki configures a 5MB
`outputBufferSize`, so every output buffer exceeds the pool's max capacity and gets
discarded on release instead of reused, causing continuous DirectMemory churn under
load. Full analysis in #4084.

This fixes it in `JettyServer.jettyServer(...)`, the shared factory both
`JettyHttps.java` and `FusekiServer.java` call, so neither file needs changes.
Constructs an `ArrayByteBufferPool` sized to
`FusekiSystemConstants.jettyOutputBufferSize` and passes it into
`Server(ThreadPool, Scheduler, ByteBufferPool)` at construction time (adding the pool
via `addBean()` afterward doesn't work, since the default pool is already wired in
during construction).

Added a test asserting the built server's pool max capacity covers the configured
output buffer size. Full `jena-fuseki-main` suite passes (799/799).

### DIFF
--- a/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/JettyServer.java
+++ b/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/JettyServer.java
@@ -50,6 +50,8 @@ import org.eclipse.jetty.ee11.servlet.FilterHolder;
 import org.eclipse.jetty.ee11.servlet.ServletContextHandler;
 import org.eclipse.jetty.ee11.servlet.ServletHolder;
 import org.eclipse.jetty.http.MimeTypes;
+import org.eclipse.jetty.io.ArrayByteBufferPool;
+import org.eclipse.jetty.io.ByteBufferPool;
 import org.eclipse.jetty.security.SecurityHandler;
 import org.eclipse.jetty.server.*;
 import org.eclipse.jetty.server.handler.ErrorHandler;
@@ -473,8 +475,16 @@ public class JettyServer {
         maxThreads = Math.max(minThreads, maxThreads);
         // Args reversed: Jetty uses (max,min)
         threadPool = new QueuedThreadPool(maxThreads, minThreads);
-        Server server = new Server(threadPool);
+        // Server(ThreadPool) alone installs a default 64KB ArrayByteBufferPool; pass ours explicitly.
+        Server server = new Server(threadPool, null, newByteBufferPool());
         return server;
+    }
+
+    /** ByteBufferPool sized so its maximum pooled buffer capacity is at least
+     * {@link FusekiSystemConstants#jettyOutputBufferSize}. */
+    private static ByteBufferPool newByteBufferPool() {
+        int maxCapacity = FusekiSystemConstants.jettyOutputBufferSize;
+        return new ArrayByteBufferPool(0, 2048, maxCapacity, -1, -1, -1);
     }
 
     private static void serverAddConnectors(Server server, int port,  boolean loopback) {
--- a/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/JettyServer.java
+++ b/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/JettyServer.java
@@ -58,6 +58,7 @@ import org.eclipse.jetty.server.handler.ErrorHandler;
 import org.eclipse.jetty.util.Callback;
 import org.eclipse.jetty.util.resource.Resource;
 import org.eclipse.jetty.util.thread.QueuedThreadPool;
+import org.eclipse.jetty.util.thread.Scheduler;
 import org.eclipse.jetty.util.thread.ThreadPool;
 import org.eclipse.jetty.xml.XmlConfiguration;
 import org.slf4j.Logger;
--- a/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/JettyServer.java
+++ b/jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/JettyServer.java
@@ -476,7 +476,7 @@ public class JettyServer {
         // Args reversed: Jetty uses (max,min)
         threadPool = new QueuedThreadPool(maxThreads, minThreads);
         // Server(ThreadPool) alone installs a default 64KB ArrayByteBufferPool; pass ours explicitly.
-        Server server = new Server(threadPool, null, newByteBufferPool());
+        Server server = new Server(threadPool, (Scheduler)null, newByteBufferPool());
         return server;
     }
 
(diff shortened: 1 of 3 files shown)


jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/JettyServer.java
AniRamadoss, line 469: Did we also want to address the other route that Andy mentioned on https://github.com/apache/jena/issues/4084#issuecomment-5055567310 ?

divyankshah: Hi @AniRamadoss, Thank you for asking - yes, it's covered. FusekiServer.java#L1907 calls
JettyServer.jettyServer(minThreads, maxThreads), the same shared factory I fixed. Since
the pool construction now lives inside that factory, both JettyHttps.java and
FusekiServer.java inherit the fix automatically also neither file needed direct changes,
it is dependent on JettyServer.java.

I confirmed this by tracing both call sites before submitting, also happy to paste the
exact call chain here if it's useful for review.

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/JettyServer.java
```diff
@@ -473,10 +475,18 @@ public static Server jettyServer(int minThreads, int maxThreads) {
...
+        // Server(ThreadPool) alone installs a default 64KB ArrayByteBufferPool; pass ours explicitly.
+        Server server = new Server(threadPool, null, newByteBufferPool());
         return server;
     }
 
+    /** ByteBufferPool sized so its maximum pooled buffer capacity is at least
+     * {@link FusekiSystemConstants#jettyOutputBufferSize}. */
+    private static ByteBufferPool newByteBufferPool() {
+        int maxCapacity = FusekiSystemConstants.jettyOutputBufferSize;
+        return new ArrayByteBufferPool(0, 2048, maxCapacity, -1, -1, -1);
```
afs: Any particular reason for the choice of 2048 here? (the default 4096)

divyankshah: Hi @afs - Thank you for asking - I carried 2048 over from the suggested fix snippet in the original issue
report, without re-examining it against Fuseki's actual configured sizes. I just checked now:
all of Fuseki's buffer sizes (16KB header, 512KB request header, 5MB output) are exact
multiples of both 2048 and 4096.
I can switch  it to -1 to defer to Jetty's own default (4096) rather than hardcoding a
value. Let me know if i should push that change?

afs: The fix snippet was for an older Jetty version.

We haven't considered whether we should decrease the output buffer size (not to 64k obviously, but maybe 2MB is enough). All the result set and RDF writers buffer as well. In the new async Jetty12 architecture, the ByteBufferPool seems to play the role of transferring byte from application to the output thread.

Without information about whether this issue is having a visible impact, it is a bit guesswork. A SPARQL operation itself has significantly more overhead than, say, serving a static web page - the effect of a 64K buffer may there but insignificant in all the noise.

divyankshah: Hi @AniRamadoss - can we reproduce this and have logs for it for the current version?

AniRamadoss: Hi @divyankshah, I'll take a look and will have an update for you by the end of the week.

afs: For output usage, the byte buffers are direct (out of heap - which in modern Java is not limited and can expand. Just don't call it "malloc" :-)).

For now, I think the setting should be :

4096 or jetty default -1. It probabley aligns with VM page size.
FusekiSystemConstants.jettyOutputBufferSize = 2M

5MB was just a large number. Large also has longer latency to first output.

Because the result set and RDF writers are already buffered for short strings and single-character writing, the Jetty buffers are cross thread transfers. Using several, not one big one, should be just as good.

(And yes - we could change the writers to allow buffer allocation to injected but one thing at a time and see what the buffering costs are anyway.)

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/JettyServer.java
```diff
@@ -473,10 +475,18 @@ public static Server jettyServer(int minThreads, int maxThreads) {
         maxThreads = Math.max(minThreads, maxThreads);
         // Args reversed: Jetty uses (max,min)
         threadPool = new QueuedThreadPool(maxThreads, minThreads);
-        Server server = new Server(threadPool);
+        // Server(ThreadPool) alone installs a default 64KB ArrayByteBufferPool; pass ours explicitly.
+        Server server = new Server(threadPool, null, newByteBufferPool());
```
afs:
```suggestion
        Server server = new Server(threadPool, (Scheduler)null, newByteBufferPool());
```
to remind the reader what the null is for.

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/JettyServer.java
divyankshah: Thanks @afs, this makes sense - I'll make these changes and push shortly.

divyankshah: Hi @afs I pushed our default changes - I'm thinking of writing another unit test to validate whether our changes are actually effective, without needing a full HTTP load test. 

Jetty's ArrayByteBufferPool has a built-in diagnostic, getNoBucketDirectAcquires(), that tracks exactly this
failure mode internally, so I could acquire/release a buffer sized to jettyOutputBufferSize directly against the pool and assert that counter stays empty. I'd also verify it actually catches the bug by running the same test
against the old (broken) config to confirm it fails there.

Would this be useful, given the impact question is still open?

afs (approved):

jena-fuseki2/jena-fuseki-main/src/main/java/org/apache/jena/fuseki/main/JettyServer.java
afs: > Would this be useful, given the impact question is still open?

Yes, please.

I think the PR should go in whether there is a visible improvement or not. I doubt it is a negative and the only way to tell is to try. We are now at the start of the next release cycle so now is a good time. Having the PR changes in the code shows how Fuseki can take further control if it needs to.

divyankshah: Hi @afs, pushed the behavioural tests.

One builds a real FusekiServer, acquires/releases a buffer sized to
jettyOutputBufferSize, and asserts getNoBucketDirectAcquires() stays empty,
confirming the request gets pooled through our fixed path. The second does
the same against a bare Jetty default pool and asserts the opposite, proving
the check actually discriminates rather than always passing.

Verified manually too: reverted JettyServer.java to the old constructor first,
confirmed the first test fails, then restored the fix and it passed.

Let me know if it looks good or anything needs to be adjusted.

PR #4130, opened 2026-08-05
GH-2340: Migrate fuseki to execution unit tests with one server per test.

### LINKED ISSUE #2340: Convert Fuseki tests to use server-per-test, not server-per-testsuite.
### Change

This task is to complete the work of #2328 (Unstable build on MS Windows)

Yes

ThomasThelen:
GitHub issue resolved #2340 

Pull request Description:

Currently, test suites share a single server instance across them. These instances are started through the `@BeforeAll` and `@BeforeClass` methods and are torn down in the `@AfterAll`. #2340 is just one case of where the tight coupling can cause problems. Previous work was done on this in the past. For example, there are `OneServerPerTestSuite` and `CloseTestServers` flags to choose which tests run under coupled environments. Running them in a one server per test configuration added a neglible amount of runtime (0.1s) from a shoddy test on my laptop. I think the Jetty listener must be cheap enough.

This pull request makes tests use Jena on a server-per-test basis rather than the current server-per-testsuite. At a high level, this is effectively stripping a layer of abstraction away and looking at individual tests rather than the suite.

The GeoSPARQL tests were the last holdout of JUnit 4. Upgrading them to JUnit 5 allowed for using `@BeforeEach` and `@BeforeAll` which are heavily utilized throughout the changes. The upgrading also led to removing of the vintage JUnit engine dependency and adding junit-jupyter (6.x.x from the root BOM).

### Change Trends

The changes are spread across nearly 50 files and follow a similar pattern. These patterns include
1. `@BeforeClass` and `@BeforeAll` changing to `@BeforeEach` because we're now operating per test (function) not test suite (class). Example [here](https://github.com/apache/jena/compare/main...ThomasThelen:jena:issue_2340?expand=1#diff-072c86e69e3690e3dd2dbc20839a73142549ae26672994a9bae2815767ce9ae3R52)
2. The change from global state to test-local led to fewer static variables. Also related to dropping the abstraction layer to the test level. Example [here](https://github.com/apache/jena/compare/main...ThomasThelen:jena:issue_2340?expand=1#diff-5e5028b230c24dfbbe473a8c54823703c80a9ca025cb75cee9510a0c23abb8ddR59)
3. Because instances are no longer shared, there's no longer a need for clearing state. Example [here](https://github.com/apache/jena/compare/main...ThomasThelen:jena:issue_2340?expand=1#diff-d5382e5c13dde863b21a5be35bc5bcbad9ed23fae811e745996fbd1e8e762606L182)
4. Import changes from `org.junit.Test` to `org.junit.jupiter.*`. This was done to support the version change. Example [here](https://github.com/apache/jena/compare/main...ThomasThelen:jena:issue_2340?expand=1#diff-fe80d53479c6a74611aa814df6e6c3e3ce9ccd6b3eae7629fbb77d9b0c0847afR24)
5. Expecting cases to throw was changed to assertThrows, from JUnit 5. Example [here](https://github.com/apache/jena/compare/main...ThomasThelen:jena:issue_2340?expand=1#diff-fe80d53479c6a74611aa814df6e6c3e3ce9ccd6b3eae7629fbb77d9b0c0847afL235)

### Deleted files

It looks like several tests were created in isolation because of the shared state issues. These include

`TestFusekiShaclValidation2.java` and  `TestModelStore2.java`

### Misc

A few verbosity settings were hardcoded to true; these were updated to use the remaining test flag, [VerboseServer](https://github.com/apache/jena/pull/4130/changes#diff-273cae746ce900c7e6eafffd95632ae519a1e57158fd877b00dade7f18140e74R26).

The pull request was split into two commits so that the fuseki geosparql JUnit upgrade could easily be seen outside of the other changes. I can squash into a single one if wanted.

----

 - [x] Tests are included.

----


afs:
Hi @ThomasThelen,

yes, this would be good.
I will find time to review this overall some more but it will take awhile.

The current testing is the way it is for several reasons, some of which may well no longer be issues:

* Running out of TCP connections: under load, we have observed there being a delay in the kernel reclaiming network resources to the point where the test srun otu of TCP connections in ways that would never happen in normal use.

* MS Windows behaves differently - and not just to memory mapped files.

* Some tests are also testing the clean-do soemthing-reuse cycle whether expicit stated or implicitly.

Large PRs are difficult to review and the project is limited in it's capacity to handle PRs.

Can we start with `jena-fuseki-geosparql` as a separate PR? Getting that to be independent of test server management in Fuseki would be good all on its own.

> Upgrading them to JUnit 5 allowed for using @BeforeEach and @BeforeAll which are heavily utilized throughout the changes

Was that a blocker? Junit4 has `@Before` and `@BeforeClass`.

ThomasThelen:
> Large PRs are difficult to review and the project is limited in it's capacity to handle PRs.
> 
> Can we start with `jena-fuseki-geosparql` as a separate PR? Getting that to be independent of test server management in Fuseki would be good all on its own.

I completely understand it's a bit much of a horizontal surface in a single PR; I'll close this and as you suggest and start with the geosparql tests (not a bad idea with 1.2 around the corner, too). 


> > Upgrading them to JUnit 5 allowed for using @beforeeach and @BeforeAll which are heavily utilized throughout the changes
> 
> Was that a blocker? Junit4 has `@Before` and `@BeforeClass`.

It was a non-blocker and as you point out JUnit 4 has `@Before` which I think is functionally equivalent to the JUnit 5 `@BeforeEach`. `fuskei-geosparql` is the last JUnit 4 holdout in Fuseki so I figured I'd bump the version and use the new verbage. If this is a change that would typically require a new issue - I can play that way and refrain from including the change here.

afs:
>  If this is a change that would typically require a new issue

No need as it is changes to tests. 
I see the new PR has commits with log messages - for me, that's enough.

PR #4134, opened 2026-08-07
GH-4129: Fix rdf:reifies triple dropped when pre-bound variable is used inside an RDF 1.2 annotation

### LINKED ISSUE #4129: SPARQL CONSTRUCT: Missing rdf:reifies triple when pre-bound variable is in the reified triple
### Version

6.2.0

### What happened?

There appears to be a problem with the RDF 1.2 annotation syntax in CONSTRUCT queries when one of the variables is pre-bound. The rdf:reifies triple is missing from the output. It works fine for variables that are bound during the WHERE clause.

```
import java.util.Map;

import org.apache.jena.query.Query;
import org.apache.jena.query.QueryExecution;
import org.apache.jena.query.QueryExecutionFactory;
import org.apache.jena.query.QueryFactory;
import org.apache.jena.query.Syntax;
import org.apache.jena.rdf.model.Model;
import org.apache.jena.rdf.model.ModelFactory;
import org.apache.jena.rdf.model.ResourceFactory;
import org.apache.jena.riot.Lang;
import org.apache.jena.riot.RDFDataMgr;
import org.apache.jena.sparql.core.Var;
import org.apache.jena.sparql.syntax.syntaxtransform.QueryTransformOps;
import org.apache.jena.vocabulary.OWL;

/**
 * Standalone demo: runs a CONSTRUCT query with an RDF-star annotation
 * ({| ... |}) with $this substituted for owl:Thing before execution.
 */
public class RDFStarConstructDemo {

		String queryString =
				"PREFIX ex: <http://example.com/>\n" +
				"CONSTRUCT {\n" +
				"	$this ex:seeAlso ex:Nothing {| ex:tempTriple true |} .\n" +
				"}\n" +
				"WHERE {\n" +
				"}\n";

		Query query = QueryFactory.create(queryString, Syntax.syntaxARQ);

		query = QueryTransformOps.replaceVars(query, Map.of(
				Var.alloc("this"), ResourceFactory.createResource(OWL.Thing.getURI()).asNode()));

		try (QueryExecution qexec = QueryExecutionFactory.create(query, model)) {
			Model result = qexec.execConstruct();
			RDFDataMgr.write(System.out, result, Lang.TURTLE);
		}
	}
}
```

```shell
PREFIX ex: <http://example.com/>

[ ex:tempTriple  true ] .

<http://www.w3.org/2002/07/owl#Thing>
        ex:seeAlso  ex:Nothing .
```

None

adityaanikam:
GitHub issue resolved #4129

Pull request Description:

`QueryTransformOps.replaceVars()` (used to pre-bind a variable at the AST level,
before query execution) drops the `rdf:reifies` triple generated by an RDF 1.2
`{| ... |}` annotation when the variable being substituted is used inside the
reified triple.

Root cause: in `QueryTransformOps.mutateByQueryType()`'s `CONSTRUCT` case, each
quad's subject/predicate/object goes through `transformOrSame(Node, ExprTransform)`.
That method only checked `Var.isVar(node)` and returned any non-`Var` node
unchanged. The `rdf:reifies` triple's object is a quoted triple term
(`Node_Triple`) that itself contains the variable, not a plain `Var`, so it was
never substituted. The separate base triple *did* substitute correctly (its
subject is a plain `Var`), which is why the bug report shows the base triple
correct but the `rdf:reifies` triple missing entirely from the CONSTRUCT
output - it still held an unbound variable and was silently dropped when the
template was later instantiated.

The fix adds triple-term recursion to `transformOrSame`, mirroring the existing,
correct pattern already used at query-execution time in
`org.apache.jena.sparql.core.Substitute.substitute(Node, Binding)` (which is why
substituting via a WHERE-clause binding already worked, per the original report).

A new regression test, `transformTransformReplace_reifies_01` in
`TestQuerySyntaxTransform`, reproduces the exact query from the issue and
asserts both the base triple and the `rdf:reifies` triple's quoted triple have
the substitution applied.

----

 - [x] Tests are included.
 - [ ] Documentation change and updates are provided for the [Apache Jena website](https://github.com/apache/jena-site/) - not applicable, internal bug fix with no user-facing behavior change to document.

----

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/syntax/syntaxtransform/QueryTransformOps.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/syntax/syntaxtransform/QueryTransformOps.java
@@ -346,9 +346,8 @@ public class QueryTransformOps {
             return e2.getConstant().getNode();
         }
         // RDF 1.2 triple term (e.g. the object of an rdf:reifies triple from a {| ... |}
-        // annotation) may itself contain variables. Recurse into it so pre-binding a
-        // variable used inside a reified triple's annotation is applied consistently
-        // with the base triple, not silently skipped.
+        // annotation) may itself contain variables. Recurse into the triple term
+        // and replace any variables used inside a reified triple's annotation.
         if ( node.isTripleTerm() && ! node.isConcrete() ) {
             Triple triple = node.getTriple();
             Node s = transformOrSame(triple.getSubject(), exprTransform);
--- a/jena-arq/src/main/java/org/apache/jena/sparql/syntax/syntaxtransform/QueryTransformOps.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/syntax/syntaxtransform/QueryTransformOps.java
@@ -24,6 +24,7 @@ package org.apache.jena.sparql.syntax.syntaxtransform;
 import java.util.*;
 
 import org.apache.jena.graph.Node;
+import org.apache.jena.graph.NodeFactory;
 import org.apache.jena.graph.Triple;
 import org.apache.jena.query.Query;
 import org.apache.jena.query.QueryVisitor;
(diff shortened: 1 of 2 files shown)


jena-arq/src/test/java/org/apache/jena/sparql/syntax/syntaxtransform/TestQuerySyntaxTransform.java
```diff
@@ -158,6 +164,40 @@ public class TestQuerySyntaxTransform
                        "s", "<urn:ex:z>", "g", "<urn:ex:g>");
     }
 
+    @Test public void transformTransformReplace_reifies_01() {
+        String queryString =
+                "PREFIX ex: <http://example.com/>\n" +
+                "CONSTRUCT {\n" +
+                "  ?this ex:seeAlso ex:Nothing {| ex:tempTriple true |} .\n" +
+                "}\n" +
+                "WHERE {}\n";
```
afs:
```suggestion
        String queryString = """
                PREFIX ex: <http://example.com/>
                CONSTRUCT {
                    ?this ex:seeAlso ex:Nothing {| ex:tempTriple true |} .
                } WHERE {}
                """ ;
```
Triple-quoted string.

jena-arq/src/main/java/org/apache/jena/sparql/syntax/syntaxtransform/QueryTransformOps.java
```diff
@@ -336,14 +337,28 @@ private static Map<String, Node> transformJsonMapping(Map<String, Node> jsonMapp
...
+            if ( e2 == null )
+                return node;
+            if ( ! e2.isConstant() )
+                return node;
+            return e2.getConstant().getNode();
+        }
+        // RDF 1.2 triple term (e.g. the object of an rdf:reifies triple from a {| ... |}
+        // annotation) may itself contain variables. Recurse into it so pre-binding a
+        // variable used inside a reified triple's annotation is applied consistently
+        // with the base triple, not silently skipped.
```
afs:
```suggestion
        // RDF 1.2 triple term (e.g. the object of an rdf:reifies triple from a {| ... |}
        // annotation) may itself contain variables. Recurse into the triple term
        // and replace any variables used inside a reified triple's annotation.
```
The report was about (SHACL) pre-binding but this mechanism is general purpose - and is an approximation for pre-binding, not as it is defined in the SHACL 1.0 spec.

afs (approved):

afs (commented):
A couple of suggested tweaks but otherwise looks good.

PR #1459, opened 2022-07-28
JENA-2282: Fuseki2 Query Store

jamiefeiss:
Jira issue: https://issues.apache.org/jira/browse/JENA-2282

Pull request Description:

An initial implementation of a query library for the Fuseki2 Vue UI using Vuex. Adds two buttons to the dataset query page for saving & loading queries. From there, users can save queries to a query library that persists in local storage in the browser.

----

 - [ ] Tests are included.
 - [ ] Documentation change and updates are provided for the [Apache Jena website](https://github.com/apache/jena-site/)
 - [ ] Commits have been squashed to remove intermediate development commit messages.
 - [ ] Key commit messages start with the issue number (GH-xxxx or JENA-xxxx)

By submitting this pull request, I acknowledge that I am making a contribution to the Apache Software Foundation under the terms and conditions of the [Contributor's Agreement](https://www.apache.org/licenses/contributor-agreements.html).

----

See the [Apache Jena "Contributing" guide](https://github.com/apache/jena/blob/main/CONTRIBUTING.md).

PR #4140, opened 2026-08-10
GH-1680: Support create & delete operations on the prefix store through Fuseki UI

### LINKED ISSUE #1793: Fuseki UI: custom SPARQL query prefixes or examples

ThomasThelen:
GitHub issue resolved #1793

Pull request Description:

https://github.com/apache/jena/pull/1683 appears to be pretty stalled, so I thought I'd try to get it over the finish line and figured a new pull request would be fair game. The general idea in the linked issue/PR to include a way to have stateful prefix and example queries shown on the query page. Most of the feedback was centered around finding a place to put the prefix state (we need to save the prefixes _somewhere_). Some time after the pull request was made, https://github.com/apache/jena/pull/2538 was merged which includes a Prefix Store, an ideal place to store prefixes...

#2538 wasn't scoped to include Fuseki changes so the custom prefixes never surfaced to the query page and instead, the same default prefixes would be shown. This PR has two pieces; I would be happy to drop the second commit or split into a different PR if found out of scope (it contains the most LoC changes and can be functionally separated).

1. 22492af7b0d571c75facd5afb5068759cc62e47d: Adds the prefixes from the Prefix Store to the query page. The changes are fairly minimal here: there's a new computed value for the prefixes (the endpoint); a new method for fetching the values and setting them in the UI; and endpoint glue in the fuseki service definition. 
2. 4ac471a439136446cfa1502145b3762bf837b2ae: Adds CRUD operations for the prefixes. This inevitably came with some UI code, which is where most of the code changes are. There are two new form fields with validation, the submit button, and the confirm/cancel modal for deleting prefix maps.

## UI Changes

The main challenge was figuring out how to support CRUD operations on the prefixes in the UI. The final design I came up has a "shadow" prefix pill for adding new ones. When clicked, two forms appear beneath the prefix list.

Supporting the editing prefixes was a challenge. Clicking a prefix already has behaviour defined where the prefix is loaded into the yasgui editor. Furthermore, several can be selected at once which makes editing in a non-bulk way difficult (if it would even make sense). Rather than test my IQ on figuring out how to create a UX for editing prefixes... I punted. Creating a new prefix new and deleting are the only operations supported. If a user wants to edit a prefix they can delete and re-add it as a workaround - I'm happy to hear any other ideas.

## Demo

Here's what the user experience looks like 

https://github.com/user-attachments/assets/68d91aae-efe8-4dc3-8f37-4d4ac1d7277e

## Documentation

This _might_ need user facing documentation, but I'm not sure if it's minor enough where it's fine without. If anyone thinks this should be in the user docs, I'll happily add some.

## Tests

The tests were sort of brutal to get right but the gist is that correctness is checked in a lot of places by looking for the presence of HTML elements under different conditions.

## Misc

Issue #1793 technically says _or_ in the title... So I'm tempted to claim that on a technicality the issue can be closed along with #1683. The other reason being that #1459 takes care of the query portion. If necessary I can squash the two commits.

----

 - [x] Tests are included.

----

### DIFF
--- a/jena-fuseki2/jena-fuseki-ui/src/services/mock/json-server.js
+++ b/jena-fuseki2/jena-fuseki-ui/src/services/mock/json-server.js
@@ -16,6 +16,7 @@
  */
 
 import jsonServer from 'json-server'
+import { DEFAULT_PREFIXES } from '../../utils/prefixes.js'
 
 const PORT = process.env.FUSEKI_PORT || 3030
 
@@ -96,6 +97,16 @@ server.post('/\\$/datasets', (req, res) => {
         'srv.type': 'upload',
         'srv.description': 'File Upload',
         'srv.endpoints': ['upload']
+      },
+      {
+        'srv.type': 'prefixes-r',
+        'srv.description': 'Read prefixes',
+        'srv.endpoints': ['prefixes']
+      },
+      {
+        'srv.type': 'prefixes-rw',
+        'srv.description': 'Read-write prefixes',
+        'srv.endpoints': ['updatePrefixes']
       }
     ]
   }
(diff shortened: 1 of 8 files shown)


jena-fuseki2/jena-fuseki-ui/src/services/mock/json-server.js
```diff
@@ -96,6 +96,16 @@ server.post('/\\$/datasets', (req, res) => {
         'srv.type': 'upload',
         'srv.description': 'File Upload',
         'srv.endpoints': ['upload']
+      },
+      {
```
ThomasThelen, line 101: New mock elements for the prefix store

afs:
I changed the title to say "Fuseki UI" and added labels. 
Nowadays, Fuseki server has, independently, prefix management operations.

ThomasThelen:
I noticed there was a way to share a little duplicated functionality (keeping track of the default prefixes) while re-reviewing this and pushed the commits post-CI. The original failing run is[here](https://github.com/apache/jena/actions/runs/31349836390/job/93381962148).

It's suspicious that there's a failure in `jena-fuseki-main`'s `TestModShiro` when I was in jena-fuseki-ui`. I hit a failure on this test when I had a VPN enabled however, when turning it off the test passed.

PR #4161, opened 2026-08-21
GH-4160: Apply the filter-disjunction rewrite only when the disjuncts are mutually exclusive

### LINKED ISSUE #4160: TransformFilterDisjunction returns duplicate solutions for overlapping disjuncts
### Version

6.1.0

### What happened?

Hi there!

I ran some benchmarks against the [LDBC Semantic Publishing Benchmark](https://ldbcouncil.org/benchmarks/spb/) suite. One particular query (A14) turned out to be very slow so I had my Claude investigate the issue. It seems that when you write duplicate filter conditions it leads to ARQ evaluating the query twice. Example: FILTER(?f = X || ?f = X). 

I had it write a patch and test which I'd like to provide as an PR. What follows is a description of the bug generated by Claude:

## Summary

The default optimizer's `TransformFilterDisjunction` rewrites `filter(e1 || e2, P)` into
`(disjunction branch1 branch2)`, evaluating the pattern once per disjunct. When a solution
satisfies more than one disjunct it is returned once per satisfied disjunct, where the
filter returns it once. The rewrite is a union, not a partition, and nothing checks that
the disjuncts are mutually exclusive.

## Minimal repro

```java
Model m = ModelFactory.createDefaultModel();
RDFParser.fromString("""
    <http://e/s1> <http://e/fmt> <http://e/Interactive> .
    <http://e/s2> <http://e/fmt> <http://e/Interactive> .
    <http://e/s3> <http://e/fmt> <http://e/Other> .
    """, Lang.NTRIPLES).parse(m);
String q = """
    SELECT ?s WHERE {
      ?s <http://e/fmt> ?f
      FILTER(?f = <http://e/Interactive> || ?f = <http://e/Interactive>)
    }""";
try (QueryExecution qe = QueryExecutionFactory.create(q, m)) {
    System.out.println(ResultSetFormatter.consume(qe.execSelect()));
}
```

Prints **4**; the answer is **2** (s1 and s2, once each). With
`qe.getContext().set(ARQ.optFilterDisjunction, false)` or optimization off it prints 2.

The identical-disjunct form is not artificial: LDBC SPB's generated queries contain
`FILTER((?primaryFormat = cwork:InteractiveFormat) || (?primaryFormat = cwork:InteractiveFormat))`
(query template A14), so on that benchmark every ARQ-based engine evaluates the whole
nine-triple pattern twice and returns a doubled bag that only the query's `LIMIT` hides.

Overlap does not require identical disjuncts. Also affected, for example:

- `FILTER(?x = :c || ?x != :d)` — a solution with `?x = :c` satisfies both disjuncts and
  is returned twice (the current `TestTransformFilters.disjunction02` pins this expansion);
- `FILTER(?x = "1"^^xsd:integer || ?x = "01"^^xsd:integer)` — different terms, same value;
- `FILTER(?x = :c || ?y = :d)` — different variables, one solution can satisfy both.

## Cause

`TransformFilterDisjunction.expandDisjunction` creates one branch per disjunct with no
mutual-exclusion condition. `OpDisjunction` is executed as a plain union (bag semantics,
no dedup), which is only equivalent to the filter when at most one disjunct can be true
of any one solution.

## Proposed fix

Apply the transform only when every disjunct tests one and the same variable against a
constant (`=` or `sameTerm`) and the constants are pairwise known to be different values
(`NodeValue.notSameValueAs`, treating an indeterminate comparison as possibly equal) —
which makes the branches an exact partition. This keeps the motivating `?x IN (...)`
cases, including mixed IRI/literal lists, and leaves every other disjunction evaluated
as the filter it is.

```shell

```

Yes

faubulous:
GitHub issue resolved #4160

Pull request Description:

`TransformFilterDisjunction` rewrites `filter(e1 || e2, P)` into a disjunction that evaluates `P` once per disjunct, so a solution satisfying k disjuncts is returned k times where the filter returns it once. `FILTER(?x = :c || ?x = :c)` returns every matching row twice under the default optimizer, and `FILTER(?x = :c || ?x != :d)` returns rows with `?x = :c` twice; disabling `optFilterDisjunction` restores the correct multiset.

Two changes together restore filter semantics:

- **Repeated disjuncts are dropped first** — `(A || A)` is `A` — so the degenerate duplicate that generated queries really contain (LDBC SPB emits `FILTER(?pf = :c || ?pf = :c)`) collapses to a single equality that grounds the pattern, rather than being either doubled (before) or left unoptimized (declined).
- **The expansion is applied only when the disjuncts are mutually exclusive**: every remaining disjunct must test one and the same variable against a constant (`=` or `sameTerm`) and the constants must be pairwise known to be different values (`NodeValue.notSameValueAs`, with an indeterminate comparison treated as possibly equal). That makes the rewrite an exact partition. The motivating case — `?x IN (...)` over distinct constants, including mixed IRI/literal lists — keeps its expansion; every other disjunction is now evaluated as the filter it is.

Several existing algebra tests pinned the unsound expansions; their expectation is now that the filter is left alone.

----

 - [x] Tests are included.

### DIFF
--- a/jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformFilterDisjunction.java
+++ b/jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformFilterDisjunction.java
@@ -27,13 +27,20 @@ import java.util.List ;
 import java.util.Set ;
 
 import org.apache.jena.atlas.logging.Log ;
+import org.apache.jena.graph.Node ;
 import org.apache.jena.sparql.algebra.Op ;
 import org.apache.jena.sparql.algebra.TransformCopy ;
 import org.apache.jena.sparql.algebra.op.OpDisjunction ;
 import org.apache.jena.sparql.algebra.op.OpFilter ;
+import org.apache.jena.sparql.core.Var ;
+import org.apache.jena.sparql.expr.E_Equals ;
 import org.apache.jena.sparql.expr.E_LogicalOr ;
+import org.apache.jena.sparql.expr.E_SameTerm ;
 import org.apache.jena.sparql.expr.Expr ;
+import org.apache.jena.sparql.expr.ExprEvalException ;
+import org.apache.jena.sparql.expr.ExprFunction2 ;
 import org.apache.jena.sparql.expr.ExprList ;
+import org.apache.jena.sparql.expr.NodeValue ;
 
 /**
  * Filter disjunction. This covers the case of
@@ -42,6 +49,14 @@ import org.apache.jena.sparql.expr.ExprList ;
  * where either or both of {@code expr1} and {@code expr2} are equalities that help
  * ground the pattern. This includes {@code ?x IN (....)} so this optimization can a
  * significant improvement.
+ * <p>
+ * The rewrite evaluates the pattern once per disjunct, so it is only sound when at
+ * most one disjunct can be true of any one solution; otherwise a solution satisfying
+ * several disjuncts is returned once per satisfied disjunct where the filter returns
+ * it once. Repeated disjuncts are dropped first, and the rewrite is then applied only
+ * when every disjunct tests the same variable against a constant ({@code =} or
+ * {@code sameTerm}) and no one term can satisfy two of those tests. Any other
+ * disjunction is left as a filter.
  */
 
 public class TransformFilterDisjunction extends TransformCopy {
@@ -66,9 +81,6 @@ public class TransformFilterDisjunction extends TransformCopy {
 
         ExprList exprList2 = new ExprList();
         Op newOp = subOp;
-        // remember what's been seen so that FILTER(?x = <x> || ?x = <x> ) does not
-        // result in two transforms.
-        Set<Expr> doneSoFar = new HashSet<>();
 
         for ( Expr expr : exprList ) {
             if ( !isDisjunction(expr) ) {
@@ -77,25 +89,22 @@ public class TransformFilterDisjunction extends TransformCopy {
                 continue;
             }
 
-//            // Relies on expression equality.
-//            if ( doneSoFar.contains(expr) )
-//                continue ;
-//            // Must be canonical: ?x = <x> is the same as <x> = ?x
-//            doneSoFar.add(expr) ;
-
             Op op2 = expandDisjunction(expr, newOp);
-            if ( op2 != null )
-                newOp = op2;
+            if ( op2 == null ) {
+                // A disjunction this transform can not rewrite soundly.
+                // Leave it as a filter expression.
+                exprList2.add(expr);
+                continue;
+            }
+            newOp = op2;
         }
+
+        if ( newOp == subOp )
+            // No disjunction was expanded.
+            return super.transform(opFilter, subOp);
 
         if ( exprList2.isEmpty() )
             return newOp;
-
-        // There should have been at least on disjunction.
-        if ( newOp == subOp ) {
-            Log.warn(this, "FilterDisjunction assumption failure: didn't find a disjunction after all");
-            return super.transform(opFilter, subOp);
-        }
 
         // Put the non-disjunctions outside the disjunction and the pattern rewrite.
         Op opOther = OpFilter.filterBy(exprList2, newOp);
@@ -113,8 +122,26 @@ public class TransformFilterDisjunction extends TransformCopy {
         return (expr instanceof E_LogicalOr);
     }
 
+    /**
+     * Expand a disjunction into a union of the pattern grounded per disjunct, or null
+     * when that is not possible or not sound.
+     */
     public static Op expandDisjunction(Expr expr, Op subOp) {
         List<Expr> exprList = explodeDisjunction(new ArrayList<Expr>(), expr);
+
+        // (A || A) is A: drop repeated disjuncts rather than build identical branches.
+        // Generated queries really do contain the same disjunct twice - LDBC SPB writes
+        // FILTER(?pf = :c || ?pf = :c) - and a single disjunct then grounds the pattern.
+        List<Expr> distinct = new ArrayList<>(exprList.size());
+        Set<Object> seen = new HashSet<>();
+        for ( Expr e : exprList ) {
+            if ( seen.add(dedupKey(e)) )
+                distinct.add(e);
+        }
+        exprList = distinct;
+
+        if ( !isSafeDisjunction(exprList) )
+            return null;
 
         // All disjunctions - some can be done efficiently via assignments,
         // some can not (value tests).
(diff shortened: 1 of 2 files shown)


jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformFilterDisjunction.java
```diff
@@ -148,6 +171,81 @@ public static Op expandDisjunction(Expr expr, Op subOp) {
...
+        return constExpr.getConstant();
+    }
+
+    /** The variable of a disjunct {@link #constantTestedAgainst} accepted. */
+    private static Var singleVariable(Expr e) {
+        ExprFunction2 test = (ExprFunction2)e;
+        return test.getArg1().isVariable() ? test.getArg1().asVar() : test.getArg2().asVar();
+    }
+
+    private static boolean provablyDistinctValues(NodeValue nv1, NodeValue nv2) {
```
ThomasThelen: It could be quick enough to throw in a unit test for this function

jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformFilterDisjunction.java
```diff
@@ -148,6 +171,81 @@ public static Op expandDisjunction(Expr expr, Op subOp) {
...
+        for ( Expr e : exprList ) {
+            NodeValue constant = constantTestedAgainst(e, var);
+            if ( constant == null )
+                return false;
+            if ( var == null )
+                var = singleVariable(e);
+            constants.add(constant);
+        }
+
+        for ( int i = 0 ; i < constants.size() ; i++ ) {
```
ThomasThelen, line 213: This looks O(n^2) to me - is this a hot path that could have performance degraded?

jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformFilterDisjunction.java
```diff
@@ -148,6 +171,81 @@ public static Op expandDisjunction(Expr expr, Op subOp) {
...
+        }
+        return true;
+    }
+
+    /**
+     * The constant of a {@code variable = constant} or {@code sameTerm(variable, constant)}
+     * disjunct (either argument order), where the variable is {@code var} - or any
+     * variable when {@code var} is null. Null when the disjunct has another shape.
+     */
+    private static NodeValue constantTestedAgainst(Expr e, Var var) {
```
ThomasThelen, line 274: It looks like there are two places this sort of logic is being done; maybe the intent is to keep them separate. A third, similar implementation might benefit from sharing some code; I can't say for certain whether it's warranted here though. [Place 1](https://github.com/apache/jena/blob/790b3dc08fccb6be1ea2868b97bfcbae8f113062/jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformFilterInequality.java#L173), [Place 2](https://github.com/apache/jena/blob/790b3dc08fccb6be1ea2868b97bfcbae8f113062/jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformFilterEquality.java#L174)

ThomasThelen:
Testing the exclusive logic I ran into an issue where the same result is getting returned twice. This might be related to the dedup logic [here](https://github.com/apache/jena/pull/4161/changes#diff-976b008aaf18fe48bcf54a22c44c1f845cf0276ac53959d98b6a5146004f0881R129)

Save as data.ttl
```turtle
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
@prefix ex:  <http://example/> .

ex:obs1 ex:reading "NaN"^^xsd:double .
ex:obs2 ex:reading "1.5"^^xsd:double .
```


Save as query.rq
```sparql
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
SELECT * {
  ?s ?p ?x
  FILTER( sameTerm(?x, "NaN"^^xsd:double) || sameTerm("NaN"^^xsd:double, ?x) )
}
```

```
$ arq --datadata.ttl --query query.rq 
------------------------------------------------------------------------
| s                     | p                        | x                 |
========================================================================
| <http://example/obs1> | <http://example/reading> | "NaN"^^xsd:double |
| <http://example/obs1> | <http://example/reading> | "NaN"^^xsd:double |
------------------------------------------------------------------------

**Comment by faubulous** [comment]
@ThomasThelen Thanks for your comments! I'll have a look at it asap.

**Comment by ThomasThelen** [comment]
> @ThomasThelen Thanks for your comments! I'll have a look at it asap.

Thanks for your time! I'm not the most qualified to be looking at arq and the _semantics_ of the changes made here - but happy to look at structure as a once over

**Comment by faubulous** [comment]
@ThomasThelen Thanks — that is a real bug, and it is fixed and pushed.

You were right that the dedup is involved, but it turned out to be two defects combining.

**1. The exclusivity test was asking the wrong question.** `provablyDistinctValues` decides whether the constants are different *values*, via `NodeValue.notSameValueAs`. But `sameTerm` matches by *term*. For two identical `"NaN"^^xsd:double` constants, `NVCompare.sameValueAs` takes its same-term fast path into `sameExceptNaN`, which returns `false` for a NaN double or float — so `notSameValueAs(NaN, NaN)` is `true`. The two disjuncts were therefore declared mutually exclusive, even though every term equal to NaN satisfies both.

**2. The dedup was syntactic.** It used `List.contains`, i.e. `Expr.equals`, so `sameTerm(?x, "NaN"^^xsd:double)` and `sameTerm("NaN"^^xsd:double, ?x)` — one test written two ways — were not recognised as the same disjunct.

Both are addressed:

- disjuncts are now deduplicated on (operator, variable, constant) instead of on argument order, since `=` and `sameTerm` are both symmetric;
- two `sameTerm` disjuncts are exclusive exactly when the **terms** differ. Where at least one disjunct is `=`, a solution satisfying both makes the constants value-equal, so `notSameValueAs` remains the right test there.

Your query now returns one row, and it stays optimised — the two writings collapse into a single grounded branch:

```
(assign ((?x "NaN"^^xsd:double)) (bgp (triple ?s ?p "NaN"^^xsd:double)))
```

I checked each case against the same query with `optFilterDisjunction=false`, which is the semantics the optimizer has to preserve:

| query | optimized | reference |
|---|---|---|
| `sameTerm(?x,"NaN"^^xsd:double) \|\| sameTerm("NaN"^^xsd:double,?x)` | 1 | 1 |
| `?x = "1.5"^^xsd:double \|\| ?x = ex:nope` | 1 | 1 |
| `?x = ex:a \|\| ?x = ex:b \|\| ?x = ex:c` | 0 | 0 |
| `?x = "1.5"^^xsd:double \|\| ?x != ex:d` | 2 | 2 |

There is a regression test for the exact query you posted, plus direct unit tests for the comparison itself (see the reply on `provablyDistinctValues`). Full `jena-arq` suite: 16595 tests, 0 failures.

A side benefit of making the rule term-aware: `sameTerm(?x, "1"^^xsd:integer) || sameTerm(?x, "01"^^xsd:integer)` is genuinely exclusive — a term cannot be both — and is now expanded where the value-based rule declined it.

File: jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformFilterDisjunction.java (hunk 1 above)
Comment (faubulous) [reply jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformFilterDisjunction.java]: Done. 

Claude response:
`provablyDistinctValues` is now package-private and `TestTransformFilters` (same package) tests it directly:

- `NaN` / `NaN` → **true**. This one is worth pinning explicitly: it is value distinctness working as designed, and it is exactly why `sameTerm` disjuncts must not be judged by it — the bug you found on the PR.
- `"1"^^xsd:integer` / `"01"^^xsd:integer` → false (same value, different terms)
- `"1"^^xsd:integer` / `"1.0"^^xsd:double` → false (value-equal across the numeric tower)
- distinct / identical IRIs → true / false
- timezone-less `xsd:dateTime` against one with a timezone → false (indeterminate comparison, so nothing is proved)

The new pairwise predicate `provablyExclusive` is tested the same way, including the `sameTerm`/`NaN` case and the `sameTerm` pair whose terms differ but whose values are equal.

File: jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformFilterDisjunction.java (hunk 2 above)
Comment (faubulous) [reply jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformFilterDisjunction.java:213]: This one is rather tricky: We are looking at expressions like FILTER(A || B || A). In such scenarios we need to pairwise compare all possible variants to be on the safe side. However, this is done at query optimisation time and not for every solution. So it's only run once. Moreover, the amount of expressions to be evaluated is rather limited so the exponential O is not a huge problem here in practise.

But I thought of an alternative where we'd compute a hash value for every expression in the previous loop and then just check if we already have found a match. The issue is that this only works for non-literal values because of SPARQL literal comparison semantics (see explanation below). 

So we could add an optimised path here that checks for equality of non-literals quickly and only compare the remaining literals with O(n^2). Would that be OK with you or do you have other ideas?

### Literal Comparison Semantics
This is what Claude tells me.. but it sounds plausible:

The chain is `provablyDistinctValues` → `NodeValue.notSameValueAs` → `NVCompare.sameValueAs` → `XSDFuncOp.compareNumeric`, and that last one calls `classifyNumeric(fName, nv1, nv2)`, which picks the wider of the two operand types (integer → decimal → float → double) and compares in that type. The comparison type is a property of the pair, not of either element — so there is no per-element value to hash.

The concrete consequence is that numeric equality is not transitive:

```
"0.1"^^xsd:decimal vs "0.1"^^xsd:double → promoted to OP_DOUBLE → equal
"0.1"^^xsd:double vs "0.1000000000000000055511151231257827…"^^xsd:decimal → OP_DOUBLE → equal
"0.1"^^xsd:decimal vs that same decimal → OP_DECIMAL → BigDecimal.compareTo → not equal
```

File: jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformFilterDisjunction.java (hunk 3 above)
Comment (faubulous) [reply jena-arq/src/main/java/org/apache/jena/sparql/algebra/optimize/TransformFilterDisjunction.java:274]: Sorry for the spam. I did not anticipate that Claude would write responses in my name without asking. That's a no go. Never worked with it on public PRs so a newbie error. Sincere apologies.

So.. the way I see it is that these do provide similar functionality but differ in the details. There might be a way to consolidate them, but then we'd need to touch a lot more code in other classes. This widens the scope of this fix. Should this be handled in a different PR or do you want to address this here? To me the code duplication is another defect on its own already.

**Comment by faubulous** [comment]
Sure, no problem. I really appreciate Jena a lot and what you guys provide to the community. Thanks a lot in return! :+1: 

The case you spotted is pretty well observed. I had Claude write a fix for it and it seemed to be very trigger happy and pushed the results already. So now we can review it together.. :)

**Comment by afs** [comment]
A possibility: change the optimization to be for `IN`.

```
FILTER ( ?x IN (<uri1>, <uri2>,<uri3>) )
```
This would be more natural to write; the variable is in a single place, the values (which still need checking as to whether they are expressions etc) in another.

**Comment by faubulous** [comment]
@afs Are you suggesting to rewrite the original query? I completely agree with your suggestion. 

But as I mentioned in the [issue](https://github.com/apache/jena/issues/4160) this is part of a SPARQL benchmark suite which I can't change. So it seems people are writing this kind of queries and Jena falling back to evaluating the query twice or n-times is not a desirable behaviour in my opinion.

**Comment by afs** [comment]
No query rewrite.

Which forms actually show up for real in the benchmark?

```
FILTER((?primaryFormat = cwork:InteractiveFormat) || (?primaryFormat = cwork:InteractiveFormat))
```
but are there other forms?

It isn't applicable in many cases - the constants have to same-term matches so e.g. not numbers.

So restrict to a few safe cases (only one `||`  for example) if it makes a measurable difference.

Support the `IN` form as a second optimization.

PR #4186, opened 2026-09-01
GH-4185: Manage version of jspecify directly

### LINKED ISSUE #4185: `org.jspecify:jspecify` dependency convergence error
2026-09-01

We have dependency convergence problems in PRs #4177 #4180.

```
[ERROR]         +-com.github.ben-manes.caffeine:caffeine:jar:3.2.4:compile
[ERROR]           +-org.jspecify:jspecify:jar:1.0.0:compile

[ERROR]   +-com.google.guava:guava:jar:33.7.1-jre:compile
[ERROR]     +-org.jspecify:jspecify:jar:1.0.1:compile

[ERROR]     +-io.micrometer:micrometer-commons:jar:1.17.1:compile
[ERROR]       +-org.jspecify:jspecify:jar:1.0.1:compile
```
and several other micrometer dependencies.

`org.jspecify:jspecify:jar` is a collection of annotations for Java static analysis.

It is not required at runtime although that team recommend it is present.

`caffeine` will presumably catch-up.

We already exclude some runtime-optional exclusions for caffeine.

A fix is adding another exclusion on `caffeine` which might later be removed.

Jena has two uses of `jspecify` --

`jena-benchmarks/jena-benchmarks-jmh/src/test/java/org/apache/jena/sparql/engine/benchmark/QueryTaskResult.java`
`jena-core/src/main/java/org/apache/jena/mem/collection/JenaSet.java`

In `QueryTaskResult` it is on a call to `String.join` which does `return new String` and can't be null.
In `JenaSet`, use `Objects.requireNonNull`.

afs:
GitHub issue resolved #4185

Exclude jspecify: it can can cause maven version convergence problems.
It is not required at run time.

~Caveat: I am currently having problems with the build at the point of downloading corepack.~
That happens after version convergence has succeeded so I don't think it is related.
It may not affect this PR -- github builds see different caches.

----

### DIFF
--- a/jena-benchmarks/jena-benchmarks-jmh/src/test/java/org/apache/jena/sparql/engine/benchmark/QueryTaskResult.java
+++ b/jena-benchmarks/jena-benchmarks-jmh/src/test/java/org/apache/jena/sparql/engine/benchmark/QueryTaskResult.java
@@ -20,12 +20,9 @@
  */
 package org.apache.jena.sparql.engine.benchmark;
 
-import org.jspecify.annotations.NonNull;
-
 public record QueryTaskResult(String queryString, String originalOpString, String optimizedOpString,
                               long resultSetSize) {
 
-    @NonNull
     @Override
     public String toString() {
         return String.join("\n",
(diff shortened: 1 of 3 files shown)


afs:
This includes the version update from  #4177 and #4180.

PR #4244, opened 2026-09-20
GH-4239: xsd:boolean() of a NaN double is false

### LINKED ISSUE #4239: `xsd:boolean(...)` cast of `"NaN"^^xsd:double` returns `true`
### Version

6.2.0

### What happened?

**Reporter:** ting668

## Environment

- Apache Jena Version: `6.2.0`
- Java Version: `21.0.12+8-LTS`

## Description

While testing Apache Jena using an intuitive approach inspired by attribute constraints, I found that the numeric-to-boolean constructor returns `true` for a `NaN` double: `xsd:boolean("NaN"^^xsd:double)` evaluates to `true`, so a `FILTER` on the cast accepts a row it should reject.

Under the numeric-to-boolean cast, `NaN` and zero are `false` and non-zero values are `true`: `1.0` (as `xsd:double` or `xsd:float`) is `true`, while `"NaN"^^xsd:double` and `"0.0"^^xsd:double` are `false`.

## How to Reproduce and Expected Behavior

**Note:** The queries below are a **minimized, simplified example** reproducing the bug.

Buggy query (with the default graph loaded with the N-Triples data below):

```sparql
SELECT ?s WHERE { ?s <urn:pub:p> ?v . FILTER(<http://www.w3.org/2001/XMLSchema#boolean>(?v)) } ORDER BY ?s
```

Data (default graph, N-Triples):

```nt
<urn:pub:nanDouble> <urn:pub:p> "NaN"^^<http://www.w3.org/2001/XMLSchema#double> .
<urn:pub:zeroDouble> <urn:pub:p> "0.0"^^<http://www.w3.org/2001/XMLSchema#double> .
<urn:pub:oneDouble> <urn:pub:p> "1.0"^^<http://www.w3.org/2001/XMLSchema#double> .
<urn:pub:nanFloat> <urn:pub:p> "NaN"^^<http://www.w3.org/2001/XMLSchema#float> .
<urn:pub:oneFloat> <urn:pub:p> "1.0"^^<http://www.w3.org/2001/XMLSchema#float> .
<urn:pub:background> <urn:pub:other> <urn:pub:value> .
```

Expected behavior: only the `1.0` values cast to `true`:

```text
[{"s":"<urn:pub:oneDouble>"},{"s":"<urn:pub:oneFloat>"}]
```

Actual behavior: the row carrying `"NaN"^^xsd:double` is also returned:

```text
[{"s":"<urn:pub:nanDouble>"},{"s":"<urn:pub:oneDouble>"},{"s":"<urn:pub:oneFloat>"}]
```

In the same run the `"NaN"^^xsd:float` value was not accepted; the wrong `true` result was observed for the `xsd:double` `NaN`.

This suggests that the boolean cast of an `xsd:double` does not apply the specified mapping of `NaN` to `false`, so a value that should be rejected passes the filter.

```shell

```

None

anishmehta24:
GitHub issue resolved #4239

Pull request Description:

`CastXSD.castToBoolean` tested for NaN with `NodeValue.sameValueAs(nv, nvDoubleNaN)`. That is never true, since NaN is not the same value as anything (including itself, which `NVCompare.sameExceptNaN` enforces on the same-term fast path), so `xsd:boolean("NaN"^^xsd:double)` fell through to `true`. The float case only worked by accident: `"NaN"^^xsd:float` is a different term from `nvDoubleNaN`, so it skipped the fast path and hit `Double.compare(NaN, NaN) == 0` in `compareNumeric`.

The numeric branch now returns `XSDFuncOp.effectiveBooleanValue(nv)`, which already implements exactly the XPath casting rule the comment above the method quotes (zero, -0.0 and NaN are false, every other number true), so the three `sameValueAs` probes and their static imports go away.

Tests: `TestCastXSD.cast_to_boolean_06a` (double NaN, fails on `main`), plus `06b`/`06c` for ±INF. `TestCastXSD`, `TestExpressions2` and `TestNodeValue` pass.

----

 - [x] Tests are included.
 - [ ] Documentation change and updates are provided for the [Apache Jena website](https://github.com/apache/jena-site/) (not needed)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


afs:
The changes here are also in #4246.

anishmehta24:
Right, #4246 needs this commit (the strict NaN `sameValueAs` there exposes the float-NaN cast case), so I have folded both into #4246, one commit per issue, and closed this one to avoid the duplicate.

